PR #22, opened 2018-04-18
Basic Skeleton

blentz:


chambridge:
Looks like a conflict needs to be resolved and a couple minor linter fixes/ignores.
Otherwise looks fine to me. Will give the 👍 when build is green.

blentz:
Rebased the branch off of current master. All checks pass now.

requirements.txt
```diff
@@ -1 +1,8 @@
 pylint==1.8.4
+django>=2.0.0
```
adberglund: We can save this for a story around using pipenv, but this would be converted to a `Pipfile` then.

PR #42, opened 2018-04-18
add codecov.yml

blentz:


.codecov.yml
```diff
@@ -0,0 +1,27 @@
+codecov:
+  notify:
+    require_ci_to_pass: yes
+  strict_yaml_branch: master  # only use the latest copy on master branch
+
+coverage:
+  precision: 1
+  round: nearest
+  range: "90...100"
```
adberglund: Do we want a wider range for red-green?

chambridge: I'm open to suggestions.

adberglund: `80...100`? That way 95% coverage isn't yellow.

chambridge: Color doesn’t much matter to me. I’m colorblind so I just focus on the number. I like limiting at 90 for now. If people get hung up on the color we can address that later.

adberglund: Yes, this config option exclusively controls display color. This config will have no impact on the coverage percent/number and will not enforce a minimum coverage.  If we want to enforce 90% test coverage than we need to use [target](https://docs.codecov.io/docs/commit-status#section-target) It's admittedly a small annoyance, but since we're in the file now it seemed to me the time to address it. This may seem silly or like getting "hung up", but seeing yellow for 95% code coverage is an avoidable micro stressor that along with the countless others causes stress and anxiety throughout a day. If my brain sees green for 95% (which to me is an acceptable coverage) those chemical signals don't get triggered. Now, my assumption is that 95% is acceptable. If we feel that dropping to 95 is a real warning sign, then this is configured correctly and I should be concerned when I see that.

chargio: We have the opportunity to set up the standard right now. I am for 95% and don't allowing more than a change in 1% down, more than anything because if we have external collaborators that will force them to have full coverage.

blentz: Thanks for the clarification @adberglund . I agree with you - range should be 80, but target should be 90.

Issue #31, opened 2018-04-18
Setup coverage and threshold files to establish quality expectations and CodeCov integration


chambridge:
As a developer, I want collect the code coverage of unit tests and set a reasonable threshold so that code added to the project has associated test cases as new features are introduced; coverage should be checked for every pull request using CodeCov.

## Acceptance Criteria

- [x] Verify that coverage is gathered based on the settings of the coverage file
- [x] Verify third party integration with CodeCov
- [x] Verify that coverage must meet a threshold defined in ._codecov.yml_

chambridge:
I was expecting a `.coveragerc` file to be added as well.
Can we add a target into `tox.ini` and the `Makefile` to run test with coverage?

Expected to have CodeCov comments in the PRs after this is in master.

blentz:
Finished the rest of the expected work in #48

PR #48, opened 2018-04-19
add coveragerc (#31)

blentz:


.coveragerc
```diff
@@ -0,0 +1,29 @@
+# .coveragerc to control coverage.py
+[run]
+branch = True
+omit =
+  .tox/*
+  *virtualenv/*
```
adberglund: What do you think about omitting `database.py`, `settings.py`, `urls.py`, and `wsgi.py`? They are a mix of boilerplate and configs and likely won't have tests written against them.

blentz: I don't really have an opinion on it, but I don't think that should block this PR. 

We can expand the list of omissions in a future PR.

PR #49, opened 2018-04-19
Provide an example environment file for faster on-boarding.

chambridge:
* Setup usage of postgres via env file and docker compose
* Update make file to use postgres


docker-compose.yml
```diff
@@ -0,0 +1,9 @@
+version: '3'
+
+services:
+  db:
+    image: postgres:alpine
```
blentz: Please use CentOS as the base container.

https://github.com/sclorg/postgresql-container/

chambridge: Made the following updates to the **docker-compose.yml**.
```
    image: centos/postgresql-96-centos7
    environment:
    - POSTGRESQL_DATABASE=postgres
    - POSTGRESQL_USER=postgres
    - POSTGRESQL_PASSWORD=postgres
    - POSTGRESQL_ADMIN_PASSWORD=postgres
```
I haven't had much luck with this as I've not been able to remotely connect.

I dug into this a little and found the following:

http://suite.opengeo.org/docs/latest/dataadmin/pgGettingStarted/firstconnect.html#allowing-remote-connections

When I look at the */var/lib/pgsql/data/userdata/pg_hba.conf* it appears its only setup to consume local socket connections.
```
# TYPE  DATABASE        USER            ADDRESS                 METHOD

# "local" is for Unix domain socket connections only
local   all             all                                     trust
# IPv4 local connections:
host    all             all             127.0.0.1/32            trust
# IPv6 local connections:
host    all             all             ::1/128                 trust
# Allow replication connections from localhost, by a user with the
# replication privilege.
#local   replication     postgres                                trust
#host    replication     postgres        127.0.0.1/32            trust
#host    replication     postgres        ::1/128                 trust

#
# Custom OpenShift configuration starting at this point.
#

# Allow connections from all hosts.
host all all all md5

# Allow replication connections from all hosts.
host replication all all md5
```

I'm going to look for another Centos image to see if I'll have more luck. But if you know some other thing I've missed let me know.

blentz: If you look under examples/custom-config, there's docs on how to customize the pgsql config.

chambridge: Fixed, came down to an environment issue.

adberglund:
Closes #51

PR #52, opened 2018-04-20
add server status api endpoint (#36)

blentz:
Closes #36


saltcellar/api/status/model.py
```diff
@@ -0,0 +1,55 @@
...
+    global_identifier = models.CharField(
+        max_length=36,
+        null=False
+    )
+
+    def __str__(self):
+        """Convert to string."""
+        return '{' +\
+            'id:{}, global_identifier:{}'.format(self.id,
+                                                 self.global_identifier) + '}'
```
adberglund: Are the extra `{}` to mimic the appearance of a dict/JSON? 

Also, fun Python 3.6 feature--f-strings. You can type `f'id:{self.id}, global_idenfitfier:{self.global_idenfitfier}'` for less bulky variable substitution!

saltcellar/api/status/model.py
```diff
@@ -0,0 +1,55 @@
...
+
+    @staticmethod
+    @transaction.atomic
+    def create_or_retreive_server_id():
```
adberglund: It looks like we're using DRF below. In DRF world, CRUD type operations are the purview of a serializer http://www.django-rest-framework.org/tutorial/1-serialization/

saltcellar/api/status/view.py
```diff
@@ -0,0 +1,48 @@
...
+from api import API_VERSION
+from api.status.model import ServerInformation
+
+from saltcellar.environment import (commit,
+                                    modules,
+                                    platform_info,
+                                    python_version)
+
+
+@api_view(['GET'])
```
adberglund: Have you looked at class based views in DRF? There's a series of mixins that cover standard CRUD which makes writing APIs where the operations map 1:1 with your DB model ⚡️ fast. Since this does not do that we'd still be writing a custom retrieve/get method, so this is fine for now. I'll just note that we may want to use them in the future and for consistency we could refactor this to use class based views then.

saltcellar/saltcellar/environment.py
```diff
@@ -0,0 +1,98 @@
...
+import platform
+import subprocess
+import sys
+
+# pylint: disable=invalid-name
+# Get an instance of a logger
+logger = logging.getLogger(__name__)
+
+
+def commit():
```
adberglund: how about something like get_commit_info? `commit` makes me think it performs a commit.

saltcellar/saltcellar/environment.py
```diff
@@ -0,0 +1,98 @@
...
+    """Collect the build number for the server.
+
+    :returns: A build number
+    """
+    commit_info = os.environ.get('SALTCELLAR_COMMIT', None)
+    if commit_info is None:
+        try:
+            commit_info = subprocess.check_output(['git',
```
adberglund: Since Python 3.5 the preferred subprocess function is `run()`. You can put a stdout pipe in to capture output. You can also use a check flag if you want an error thrown on failure.

```
result = subprocess.run(["git", "describe", "--always"], stdout=subprocess.PIPE)
return result.stdout.strip().decode('utf-8')
```

saltcellar/saltcellar/environment.py
```diff
@@ -0,0 +1,98 @@
...
+                                                   'describe',
+                                                   '--always']).strip()
+            commit_info = commit_info.decode('utf-8')
+        except Exception:  # pylint: disable=broad-except
```
adberglund: If we're just passing, we can just not use the check flag so an error is not thrown, which would mean we can do away with the try/except.
If we do want to except, use something more specific than `Exception`.

saltcellar/saltcellar/environment.py
```diff
@@ -0,0 +1,98 @@
...
+def startup():
+    """Log environment at startup."""
+    logger.info('Platform:')
+    for name, value in platform_info().items():
+        logger.info('%s - %s ', name, value)
+
+    logger.info('Python: %s', python_version())
+    module_list = []
+    for name, value in modules().items():
+        mod = '{} - {}'.format(name, value)
```
adberglund: f-string opportunity. You could also do away with assigning the var and just use

```
module_list.append(f'{name} - {value})
```

saltcellar/saltcellar/urls.py
```diff
@@ -0,0 +1,41 @@
...
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+"""saltcellar URL Configuration
+
+The `urlpatterns` list routes URLs to views. For more information please see:
+    https://docs.djangoproject.com/en/2.0/topics/http/urls/
+Examples:
```
adberglund: Let's ✂️ the examples.

Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+RUN scl enable rh-python36 -- \
+    pip install pipenv
+
+RUN scl enable rh-python36 -- \
+    pipenv install --dev
+
+COPY saltcellar .
+
+ENTRYPOINT ["scl", "enable", "rh-postgresql96", "rh-python36", "--", "gunicorn"]
+CMD ["-c","config/gunicorn.py","config.wsgi"]
```
adberglund: Maybe I'm jumping the gun and this is in the works, but where's config/gunicorn.py?

saltcellar/api/admin.py
```diff
@@ -0,0 +1,19 @@
...
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+from django.contrib import admin
+
+# Register your models here.
```
adberglund: Since we have an admin file, let's register the new serverinfo model here.

PR #53, opened 2018-04-20
add requirements.txt for RTD

blentz:
Need a requirements.txt because RTD doesn't yet support pipenv. See: https://github.com/rtfd/readthedocs.org/issues/3181


docs/rtd_requirements.txt
```diff
@@ -0,0 +1,14 @@
+boto3==1.7.5
```
adberglund: What's the strategy for keeping this up to date?

chambridge: Diligence? We could have a simple `make doc-requirements` that redirects the output of `pipenv lock --requirements --dev` to `docs/rtd_requirements.txt`.

adberglund: Either/or if it becomes a problem we can automate.

PR #57, opened 2018-04-23
Create initial Open API swagger YAML. Closes #40.

### LINKED ISSUE #40: Create swagger.yml file to document our API (use OpenAPI v3)
As a developer, I want to be able to share our API via Swagger so that it can be consumed by other customer or they can generate client code using the defined API.

We plan on sharing/composing APIs across services using the Open API definition

## Acceptance Criteria

- [ ] Verify that a basic swagger.yml file exist describing the current API

chambridge:
* Document status endpoint.

### DIFF
--- a/docs/source/specs/openapi.yml
+++ b/docs/source/specs/openapi.yml
@@ -18,7 +18,7 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/Status"
-        default:
+        '500':
           description: unexpected error
           content:
             application/json:
--- a/Pipfile
+++ b/Pipfile
@@ -14,6 +14,7 @@ Django = ">=2.0.0"
 
 [dev-packages]
 pylint = "==1.8.4"
+sphinxcontrib-openapi = ">=0.3.2"
 sphinx-rtd-theme = ">=0.3.0"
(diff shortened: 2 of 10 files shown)


docs/source/specs/openapi.yml
```diff
@@ -0,0 +1,84 @@
+openapi: "3.0.0"
```
adberglund: I'm pretty swagger/openAPI ignorant. What's generating this YAML?

chambridge: Nothing is generating this YAML.

I'm actually in favor of this being a part of "documentation-driven development". You can have an issue to define the API contract for the UI to use and they are then able to use a mock of that API while the implementation is developed. QE can also use this doc to write test cases while development is in progress (ideally).

adberglund: Yep, totally into it. Creating stubs to unblock is 👍 . 

The root of my question was around that contract. If the swagger YAML always comes first, then there's way less risk of forgetting to update it with API changes.

chambridge:
Verified RTD build fix: 
http://saltcellar.readthedocs.io/en/40-swagger-api/
Will remove version after this is merged.

PR #61, opened 2018-04-24
56 openshift

adberglund:
Brett's OpenShift work, just cherry-picked out into its own PR.


init-app.sh
```diff
@@ -0,0 +1,16 @@
...
+    oc create -n openshift istag postgresql:9.6 --from-image=centos/postgresql-96-centos7
+fi
+OUT=$(oc get -n openshift -o yaml is python | grep -c 'tag: "3\.6"')
+if [ $OUT != 1 ]; then
+    oc create -n openshift istag python:3.6 --from-image=centos/python-36-centos7
+fi
+
+# TODO: add intelligence or user-prompt for git tag or somesuch
+oc new-app --file=openshift/saltcellar-template.yaml \
+           --code=https://github.com/SaltCellar/SaltCellar.git#36-status-endpoint
```
blentz: This is still pointing at my topic branch. We should point it at master.

In the future, we probably should make this more intelligent and have it auto-build from the current branch or some similar automation magic.

chambridge:
Do we want to add content to the README for developing with openshift (this can be a separate issue)?

blentz:
@chambridge I added a new doc under the `docs` dir for openshift-specific info. It's currently a stub, but we can expand on it.

PR #67, opened 2018-04-27
Create basic status endpoint. Closes #36.

chambridge:
Tried to make updates based on @blentz version in https://github.com/project-koku/koku/pull/52


koku/api/status/model.py
```diff
@@ -0,0 +1,104 @@
...
+
+    server_id = models.UUIDField(default=uuid.uuid4,
+                                 editable=False)
+
+    def get_commit(self):  # pylint: disable=R0201
+        """Collect the build number for the server.
+
+        :returns: A build number
+        """
+        commit_info = os.environ.get('KOKU_COMMIT', None)
```
chambridge: @blentz Are we able to "bake" the commit hash into the Openshift build in such a way that its accessible from the app?

blentz: Good question. There's at least two basic options, I think.

1. We can copy over the .git directory, install git, and let the code execute `git status` or similar.
2. As part of the s2i build, we could drop the commit info into a file, say, in the `.s2i` dir, and just have the code read it from there.

blentz: After some checking, it looks even easier. If `assemble` updates `.s2i/environment`, those environment variables should be pushed through into the built image's environment variables.

koku/api/status/model.py
```diff
@@ -0,0 +1,104 @@
...
+        if commit_info is None:
+            try:
+                commit_info = subprocess.check_output(['git',
+                                                       'describe',
+                                                       '--always'])
+                commit_info = commit_info.decode('utf-8').strip()
+            except subprocess.CalledProcessError:
```
adberglund: I still feel funny about this being in a try/except where the except is just a pass. `subprocess.run()` should usable with the check flag set to False so that an exception is not raised.

chambridge: updated.

PR #82, opened 2018-04-30
76 dev db openshift

adberglund:
Added ability to just run the DB out of OpenShift for local dev.


docs/source/openshift.rst
```diff
@@ -43,6 +43,28 @@ Alternatively, make commands are provided as a convenience. ::
...
 
+There are a few ways to use OpenShift while developing Koku. It is possible to spin up the entire application and its dependent services, or just the dependent services can be spun up while using the local Django dev server. ::
+
+  # Run everything through OpenShift
+  make oc-up-all
+
+  # Run *just* a database in Openshift, while running the server locally
+  make oc-up-db
+  # Run Django migrations to initialize the database
+  oc-run-migrations
```
blentz: Missing a `make` in there.

openshift/koku-template.yaml
```diff
@@ -98,7 +99,7 @@ objects:
         containers:
         - env:
           - name: DATABASE_SERVICE_NAME
-            value: POSTGRES_SQL
+            value: ${DATABASE_SERVICE_NAME}
```
blentz: Don't change this value. DATABASE_SERVICE_NAME is an env variable used by the Django `environ` module. 

So, every other place that is using the templated value by the same name is fine to change. Just not this one. :sweat_smile: 

We may want to rename the openshift template parameter to a different variable name just to reduce confusion. :stuck_out_tongue_winking_eye:

openshift/koku-template.yaml
```diff
@@ -371,19 +375,18 @@ parameters:
...
 - displayName: Database Name
   name: DATABASE_NAME
   required: true
-  value: default
+  value: koku
 - displayName: Database Username
-  from: '[a-zA-Z0-9]{16}'
-  generate: expression
   name: DATABASE_USER
+  value: kokuadmin
```
blentz: I'd prefer to leave this as generated by default.

You can pass in a set value from the cli (or `init-app.sh`), which overrides the generator.

adberglund: I can definitely do that. What do you see as the gain from a randomized user name? Let's at least append `user-` on the front so it gains some human readability. 

How do I pass in the database user through `init-app.sh`?

adberglund: Okay. The database directly in a pod in OpenShift is likely a stopgap while developing, switching to something hosted (like RDS if we were to stay with RDBMS/Postgres). I have the local db make command to give me consistent db access, which I view as essential for development. I am interested to hear your thoughts on tooling. 

Currently the `-p` flag is reserved for modifying the port in init-app.sh.

blentz: I think you're misunderstanding which `-p` I'm referring to.

Inside `init-app.sh` is the `oc new-app` command. Add a `--param KEY=VALUE` to the `oc new-app`. 

If you want to add a flag to `init-app.sh` too, just pick a different letter or use a positional argument, which doesn't require a flag. It should "just work". e.g. `init-app.sh -x foo positional=bar`

PR #84, opened 2018-04-30
fixes to get builds using gunicorn and clean up the Dockerfile

blentz:


Dockerfile
```diff
@@ -1,60 +1,44 @@
...
+FROM centos/python-36-centos7
 
 EXPOSE 8080
 
-ENV PYTHON_VERSION=3.6 \
-    NODEJS_VERSION=8 \
+ENV NODEJS_VERSION=8 \
+    NODEJS_SCL=rh-nodejs8 \
     NPM_RUN=start \
     NAME=nodejs \
```
chambridge: Wonder if this `NAME` value I brought over from the NODEJS container is unnecessary or could be _koku_?

PR #89, opened 2018-05-02
fix openshift template & friends - make environment variables work again!

blentz:


init-app.sh
```diff
@@ -104,20 +104,12 @@ eval set -- "$PARAMS"
...
-OUT=$(oc get -n ${OPENSHIFT_PROJECT} -o yaml is postgresql 2>/dev/null | grep -c 'tag: "9\.6"')
-if [ $OUT != 1 ]; then
-    oc create -n ${OPENSHIFT_PROJECT} istag postgresql:9.6 --from-image=centos/postgresql-96-centos7
-fi
-
-OUT=$(oc get -n ${OPENSHIFT_PROJECT} -o yaml is python 2>/dev/null | grep -c 'tag: "3\.6"')
-if [ $OUT != 1 ]; then
-    oc create -n ${OPENSHIFT_PROJECT} istag python:3.6 --from-image=centos/python-36-centos7
-fi
+oc apply -f openshift/koku-template.yaml
```
chambridge: Should we use the variable here instead of hard coding?

openshift/koku-template.yaml
```diff
@@ -190,14 +279,14 @@ objects:
...
               secretKeyRef:
                 key: database-user
-                name: koku
+                name: koku-secret
           - name: POSTGRESQL_PASSWORD
             valueFrom:
               secretKeyRef:
                 key: database-password
-                name: koku
+                name: koku-secret
```
chambridge: Should we be using `koku-secret` here or `${NAME}-secret`? Same question for the items above.

init-app.sh
blentz: good catch. yes, here.

openshift/koku-template.yaml
blentz: i've gone back and forth on that. my suggestion is to go for this as-is and give it some time to bake, then we can circle back later to further abstract it.

Issue #90, opened 2018-05-03
OpenShift app can't find DB credentials and test fails


adberglund:
- The env vars for the Django app need to be named `DATABASE_` and we are passing in `POSTGRES_`. 
- A test is failing now that we're getting the `OPENSHIFT_BUILD_COMMIT` environment variable

PR #91, opened 2018-05-03
Adb/fix build template

adberglund:
- Updated env vars for the Django app that need to be named DATABASE_ and we were passing in POSTGRES_.
- Fixed a test that was failing now that we're getting the OPENSHIFT_BUILD_COMMIT environment variable


adberglund:
Fixes #90

PR #109, opened 2018-05-07
73 openshift crashes slightly less often

adberglund:
In the end I think the most useful addition here is the new make command. It looks like OpenShift crashing the database is just a fact of life for now, and this will at least give the option of redeploying the whole project without tearing down the cluster.


Makefile
```diff
@@ -101,6 +103,14 @@ oc-up-all: oc-up oc-init
...
 oc-up-db: oc-up oc-create-db
 
+oc-reinit:
+	oc login -u developer
+	oc delete project myproject
+	sleep 60
+	oc new-project myproject
+	sleep 5
+	make oc-init
+
```
blentz: One alternative option is `oc delete all -l app=koku`. This is often faster than waiting for kubernetes to schedule the deletion of the entire project.

However, `delete all` doesn't delete the configMap, secrets, or pvc. So, if you want to delete those, it's two commands: `oc delete all -l app=koku &&  oc delete configmap/koku-env secret/koku-secret pvc/koku-pgsql`

adberglund: 👏 So much better than waiting for the project to get 💣 'd. Thank you!

PR #122, opened 2018-05-10
Create service admin on application startup. Closes #121.

### LINKED ISSUE #121: Create a Service Admin user for Customer On-boarding
Create a service admin user for Customer on-boarding.

## Implementation Details
On application startup we will check for the presence of the service admin user, if it doesn't exist we will create one.

The user information used to create the service admin user will come from the OSD environment settings

## Acceptance Criteria

- [ ] Verify that the Service Admin is created/exists at application startup

___
Associated with #93

chambridge:


koku/api/apps.py
```diff
@@ -14,8 +14,70 @@
...
 
 class ApiConfig(AppConfig):
+    """API application configuration"""
+
     name = 'api'
+
+    def ready(self):
+        """Action on application startup."""
+        # noqa: E402 pylint: disable=C0413
+        from django.db.utils import OperationalError
```
adberglund: Why not top level imports here (and in the below functions, too)?

koku/api/apps.py
```diff
@@ -14,8 +14,70 @@
...
+        """Log the status of the server at startup."""
+        # noqa: E402 pylint: disable=C0413
+        from api.status.model import Status
+        status_info = None
+        status_count = Status.objects.count()
+        if status_count == 0:
+            status_info = Status.objects.create()
+            status_info.save()
+        else:
+            status_info = Status.objects.get(pk=1)
```
adberglund: If we have more than instance of the API running ever this function may lie to us. It's a generated UUID, so I'm uncertain how we're tying it to the server anyway. 

Also, with Postgres the PK is likely a sequence which doesn't auto-reset to 1 if we end up wiping (or creating multiple Status objects--like might happen when testing). This can fail if the sequence increments and our status object has say pk=2. `Status.objects.all().first()` would just get the first entry regardless of pk. However, see the above comment about storing multiple servers.

koku/api/apps.py
```diff
@@ -14,8 +14,70 @@
...
+        from django.contrib.auth.models import User
+        service_user = ENVIRONMENT.get_value('SERVICE_ADMIN_USER',
+                                             default='admin')
+        service_pass = ENVIRONMENT.get_value('SERVICE_ADMIN_PASSWORD',
+                                             default='pass')
+
+        User.objects.create_superuser(service_user,
+                                      service_email,
+                                      service_pass)
+        logger.info('Created Service Admin: %s.', service_email)
```
adberglund: I assume this will change when we start using token auth, but do we have a plan for securely extracting the token out of this auto-generating superuser step?

koku/api/apps.py
blentz: Is there a reason these imports need to be local?

openshift/koku-template.yaml
```diff
@@ -53,12 +53,18 @@ objects:
...
+      template.openshift.io/expose-service_admin_password: "{.data['service-admin-password']}"
     name: ${DATABASE_SERVICE_NAME}-secret
   stringData:
     database-password: ${DATABASE_PASSWORD}
     database-user: ${DATABASE_USER}
     django-secret-key: ${DJANGO_SECRET_KEY}
     database-admin-password: ${DATABASE_ADMIN_PASSWORD}
+    service-admin-user: ${SERVICE_ADMIN_USER}
+    service-admin-email: ${SERVICE_ADMIN_EMAIL}
+    service-admin-password: ${SERVICE_ADMIN_PASSWORD}
```
blentz: Putting the django secrets in with the pgsql secrets seems like it's expanding the scope of the object. We should either rename it to something other than "${DATABASE_SERVICE_NAME}-secret" or we should create a new secret store for the app-specific secrets.

koku/api/apps.py
chambridge: Importing before the ready function causes the application to blow up as all server capabilities haven't been loaded and made ready. I had actually dig into that. #stackOverflow

koku/api/apps.py
chambridge: I'll use `first()`. We can evaluate whether the multi-instance issue has any impact. At this point the server uuid doesn't mean a lot to us. We could even consider dropping it and just never storing the object in the db.

koku/api/apps.py
chambridge: You can get access to the secrets created in OSD. That should give us the user/password to call the token endpoint. We could write a tool that does this.

openshift/koku-template.yaml
chambridge: 👍 I'll update this. Was a part of the template I didn't quite follow.

koku/api/apps.py
adberglund: If we don't plan on adding more columns to the Status model/table, or a way to tie it directly to hardware/a VM/a container??, I vote for dropping it. Not in this story, but in a follow up issue.

PR #123, opened 2018-05-10
make User.password field write-only, clean up tests

blentz:


koku/api/iam/test/tests_serializers.py
```diff
@@ -115,14 +112,21 @@ def test_update_user(self):
...
+                    'last_name': 'Wilson'}
         new_serializer = UserSerializer(current_user,
                                         new_data)
         if new_serializer.is_valid(raise_exception=True):
             new_serializer.save()
 
         # test the update
-        self.assertEqual(new_data['password'], User.objects.get(pk=1).password)
+        self.assertEqual(new_data['password'],
+                         User.objects.get(pk=1).password)
```
adberglund: We need to be really careful pulling records on default numeric primary keys if we switch to testing with Postgres, sequences are just going to bite us. It's also not the most reliable way of querying for a specific record in a table.

PR #148, opened 2018-05-17
Add ability to send an email on user creation. Closes #143.

### LINKED ISSUE #143: Send email on user creation
When users are created in the system they need to be alerted via email. Email should also contain a link allowing them to reset their password.

## Assumptions
User will have a valid email in the system.

#Implementation Details
We will be using the built in Django email functionality:
https://simpleisbetterthancomplex.com/tutorial/2016/06/13/how-to-send-email.html

____
Associated with #94.

chambridge:


koku/api/iam/serializers.py
```diff
@@ -18,22 +18,41 @@
...
 from django.core.validators import validate_email
 from django.db import transaction
 from rest_framework import serializers
 from rest_framework.validators import UniqueValidator
 
+from .email import new_user_reset_email
 from .models import Customer, ResetToken, User
 
 
+def generate_password():
```
adberglund: I think this is a temporary password (is it?) so it doesn't need-need to be as secure, but this is recommended for generating passwords https://docs.python.org/3.6/library/secrets.html instead of using random

openshift/koku-template.yaml
```diff
@@ -517,14 +519,25 @@ parameters:
...
-  from: 'cost-mgmt@redhat.com'
-  generate: expression
+  from: 'cost-mgmt@project-koku.com'
   name: SERVICE_ADMIN_EMAIL
   required: true
 - displayName: Service Admin Password
   from: '[a-zA-Z0-9]{16}'
   generate: expression
   name: SERVICE_ADMIN_PASSWORD
+- displayName: Email Service Host
```
adberglund: I think to pass these params into the container they need to go either in the secret, or configmap. HOST and PORT should probably be in the configmap, IDK if user should be secured in a secret or config map is fine?

koku/api/iam/serializers.py
```diff
@@ -18,22 +18,41 @@
...
+    """Generate a complex password."""
+    choices = '!@#$%^&*()_+' + string.digits \
+        + string.ascii_uppercase + string.ascii_lowercase
+
+    return ''.join(random.choices(choices, k=10))
```
blentz: NIST recently revised their password recommendations (see, for example: https://nakedsecurity.sophos.com/2016/08/18/nists-new-password-rules-what-you-need-to-know/ )

I'm pointing this out mostly to suggest that maybe we should make that `k` parameter a configurable setting. This isn't something worth blocking the PR on, though.

openshift/koku-template.yaml
blentz: IMO, usernames are fine in either the configmap or the secret.  

There's a small usability difference in co-locating all of the settings for a given feature into the same object versus splitting them between a secret and a configmap. It's not a huge deal, but it's worth noting. 

There's also the option of having a distinct configmap/secret per-feature, but I think that's only worth exploring if we need to expose pieces of the same data to different pods, like we do with the database creds.

koku/api/iam/serializers.py
chambridge: Yes, temporary. I'm gonna update the method name as well. But I'll switch over. Thanks for the info.

openshift/koku-template.yaml
chambridge: 🤦‍♂️ I want to blame meetings and task switching for this. Will update.

koku/api/iam/email.py
```diff
@@ -0,0 +1,40 @@
...
+
+"""Utility for emailing users on creation and password reset."""
+from django.core.mail import send_mail
+
+from koku.env import ENVIRONMENT
+
+SUBJECT = 'Welcome to Hybrid Cost Management'
+SENDER = ENVIRONMENT.get_value('EMAIL_SENDER',
+                               default='noreply@project-koku.com')
+BODY = 'Welcome to Hybrid Cost Management.' \
```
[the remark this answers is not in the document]
chambridge: Added templates based on django email template info. A plain text and html version.

PR #151, opened 2018-05-18
Add support for user password reset API. Closes #150.

### LINKED ISSUE #150: Support reset-password API
Support reset-password API.

## API Details

| API  | Detail|
 | --- | --- |
 | **Method** | PUT | 
 | **Path** | /api/v1/users/(user_id)/reset-password/ | 
 | **Body** | password, token | 
 | **Response** | 200 - uuid, name, email | 

## Acceptance Criteria

- [ ] Authenticated user can reset a password if they supply a valid user_id and token associated with that user_id
- [ ] Unauthenticated user cannot reset a password
- [ ] Authenticated user cannot reset password if the token is not valid
- [ ] Authenticated user cannot reset password if the token has already been used
- [ ] Authenticated user cannot reset password if the token has expired (1 day)

____
Associated with #94

chambridge:


koku/api/iam/view/user.py
```diff
@@ -190,3 +196,70 @@ def destroy(self, request, *args, **kwargs):
...
+                                                   used=False,
+                                                   expiration_date__gt=now)
+        if not token_qs.exists():
+            error = {
+                'token': [_('Token has expired or is invalid.')]
+            }
+            raise ValidationError(error)
+
+        reset_token = token_qs.first()
+        user.set_password(password)
```
adberglund: Might be nice to not allow their previous password to be provided as the current password.

chambridge: 👍

chambridge: @adberglund Done.

Issue #118, opened 2018-05-09
Deleteing a Customer, its users, and associated provider data


chambridge:
## User Story
As Service Admin, I want to be able to remove a customer and its associated data so that we aren't storing data we don't need.

## Impacts
API

## Role
Service Admin

## Questions
- Is a "cascade delete" of all associated items the right choice?
- Is there anything that needs to happen to remove our S3 privilege, etc.?

## API Details

| API  | Detail|
 | --- | --- |
 | **Method** | DELETE |
 | **Path** | /api/v1/customers/(customer_id)/ |
 | **Body** |  |
 | **Response** | 204 |

## Acceptance Criteria

- [ ] Verify the customer, users, provider and cost entry data has been removed.
- [ ] Verify that only a service admin can run this API.
- [ ] Verify that we appropriately handle bad/invalid IDs (404).

chambridge:
@LandoCalrizzian What is left to close this out from a test perspective?

LandoCalrizzian:
@chambridge I have verified that deleting a customer removes associated providers and users.

PR #167, opened 2018-05-22
Remove customers and associated regular users

dccurtis:
Primary logic to address #118 

**Problem to be solved:**
Koku needs the ability to remove a customer.  When removing a customer all of the regular users associated that customer must also be removed.

**What I did:**

- Added the DELETE /api/v1/customers/{id} operation
- Removing users associated with the customer authentication group
- Used Django transactions to rollback user deletions in the event that an exception is raised

**Functional Test Results:**
[customer_delete.txt](https://github.com/project-koku/koku/files/2026938/customer_delete.txt)

**Work Remaining:**
Remove provider and cost entry data.  I decided against adding stubs for these since there may be some refactoring anyway to integrate these removals into the customer removal flow.

I'm ok with either keeping this user story open until the provider and cost entry data removal is complete or we can open a new user story once integration is possible.


koku/api/iam/view/customer.py
```diff
@@ -165,3 +187,46 @@ def retrieve(self, request, *args, **kwargs):
...
+                {
+                    "Authorizaton": "Token 45138a913da44ab89532bab0352ef84b"
+                }
+
+            @apiParam {Number} id Customer unique ID.
+
+            @apiSuccessExample {json} Success-Response:
+                HTTP/1.1 204 NO CONTENT
+            """
+            customer = model.Customer.objects.filter(uuid=kwargs['uuid']).first()
```
chambridge: You could use `get_object_or_404()` here, as it may short circuit the logic if the uuid is not found and first has no data.  
https://github.com/project-koku/koku/blob/master/koku/api/iam/view/user.py#L241

koku/api/iam/view/customer.py
```diff
@@ -16,17 +16,39 @@
...
+from rest_framework.exceptions import APIException
 from rest_framework.permissions import IsAdminUser
 
 import api.iam.models as model
 import api.iam.serializers as serializers
 
 
+logging.basicConfig()
+LOG = logging.getLogger(__name__)
+LOG.setLevel(logging.INFO)
```
chambridge: You can create a logger but the level and initial setup should be handled by the Django settings.

koku/api/iam/view/customer.py
```diff
@@ -165,3 +187,46 @@ def retrieve(self, request, *args, **kwargs):
...
         """
         return super().retrieve(request=request, args=args, kwargs=kwargs)
+
+    @transaction.atomic
+    def destroy(self, request, *args, **kwargs):
+            """Delete a customer.
+
+            @api {delete} /api/v1/customers/:id/ Delete a customer
+            @apiName DeleteCustomers
+            @apiGroup UCustomers
```
chambridge: typo in the apiGroup
`@apiGroup UCustomers` --> `@apiGroup Customers`

koku/api/iam/view/customer.py
```diff
@@ -194,7 +193,7 @@ def destroy(self, request, *args, **kwargs):
 
             @api {delete} /api/v1/customers/:id/ Delete a customer
```
chambridge: `@api {delete} /api/v1/customers/:id/ Delete a customer` --> `@api {delete} /api/v1/customers/:uuid/ Delete a customer`

koku/api/iam/view/customer.py
```diff
@@ -209,22 +208,20 @@ def destroy(self, request, *args, **kwargs):
             @apiSuccessExample {json} Success-Response:
```
chambridge: `@apiParam {Number} id Customer unique ID.`  --> `@apiParam {String} uuid Customer unique ID.`

koku/api/iam/view/customer.py
```diff
@@ -209,22 +208,20 @@ def destroy(self, request, *args, **kwargs):
...
                 HTTP/1.1 204 NO CONTENT
             """
-            customer = model.Customer.objects.filter(uuid=kwargs['uuid']).first()
+            customer = get_object_or_404(model.Customer.objects.all(), uuid=kwargs['uuid'])
```
chambridge: `customer = get_object_or_404(model.Customer.objects.all(), uuid=kwargs['uuid'])` --> `customer = get_object_or_404(self.queryset, uuid=kwargs['uuid'])`

koku/api/iam/view/customer.py
dccurtis: Got it.  I was wondering if these ids should be uuid.  I see you just updated the other places where :id was being used.

Issue #97, opened 2018-05-04
Set User Preferences


chambridge:
## User Story
As a user, I want to be able to set my user preferences so that I can store things like the currency type I want to see data in.

## Impacts
API, UI, Docs

## Role
User

## Assumptions

- A user will have defaulted preferences (maybe in code instead of in the database).
- The US dollar will be the default currency.
- The time zone will default to UTC +0.
- The default language is EN-US.

## API Details

| API  | Detail|
 | --- | --- |
 | **Method** | PATCH | 
 | **Path** | /api/v1/customers/users/(user uuid)/preferences/| 
 | **Body** | currency, time zone, language | 
 | **Response** | currency, time zone, language | 

## UI Details
A view exists in order to set user preferences.

## Acceptance Criteria

- [ ] Verify that a user can store their preferences for service usage (like currency).

chambridge:
@blentz  Should we reopen until the API for PATCH is in?

PR #170, opened 2018-05-22
Add user preference view

blentz:
Implements the user preference view


koku/api/common/permissions/test/tests_owner.py
```diff
@@ -0,0 +1,82 @@
...
+        req = Mock(user=user, method='GET')
+        pref = Mock(spec=UserPreference, user=User(id=user.id, uuid=user.uuid))
+        o_perm = IsOwner()
+        result = o_perm.has_object_permission(request=req, view=None, obj=pref)
+        self.assertTrue(result)
+
+    def test_has_perm_auth_user_post_unowned(self):
+        """Test that an non-owner user cannot execute a POST."""
+        user = Mock(spec=User, is_authenticated=True)
+        req = Mock(user=user, method='POST')
```
chambridge: I don't see anything in the Permission that checks the method. Any reason for the specific test?
Also since we are creating preferences on user creation with defaults I'm not sure why we would check `POST` here. The Set Preference story calls for a `PATCH` https://github.com/project-koku/koku/issues/97

koku/api/iam/models.py
```diff
@@ -76,9 +76,14 @@ class UserPreference(models.Model):
                             unique=True, null=False)
     user = models.ForeignKey('User', null=False, on_delete=models.CASCADE)
     preference = JSONField(default=dict)
+    name = models.CharField(max_length=255, null=False, default=uuid4)
+    description = models.CharField(max_length=255, null=True)
+
+    class Meta:
+        ordering = ['preference']
```
chambridge: How does this choose to order items since its a dictionary? Does `name` make more sense?

koku/api/iam/serializers.py
```diff
@@ -95,7 +95,10 @@ def _create_default_preferences(user):
                     {'locale': settings.KOKU_DEFAULT_LOCALE}]
 
         for pref in defaults:
-            data = {'preference': pref, 'user': user}
+            data = {'preference': pref,
+                    'user': user,
+                    'name': list(pref.keys())[0],
+                    'description': 'default preference'}
```
chambridge: Maybe `'Default Preference'`, should we consider using `_()` so this is enabled for translation?

koku/api/iam/test/tests_user_preference_view.py
```diff
@@ -0,0 +1,310 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+"""Test the IAM views."""
```
chambridge: Test User Preference views.

koku/api/iam/test/tests_user_preference_view.py
```diff
@@ -0,0 +1,310 @@
...
+    """Tests the UserPreference view."""
+
+    def setUp(self):
+        """Set up the user view tests."""
+        super().setUp()
+        self.create_service_admin()
+
+        self.test_data = {'name': 'test-pref',
+                          'description': 'test-pref',
+                          'preference': {'test-pref': ['a',
```
chambridge: I think we should put controls around what we want to allow to be stored in preference. Don't want to allow storing random JSON data but the intended preferences called out in https://github.com/project-koku/koku/issues/97.

koku/api/iam/view/user_preference.py
```diff
@@ -0,0 +1,272 @@
...
+    serializer_class = serializers.UserPreferenceSerializer
+    authentication_classes = (TokenAuthentication,
+                              SessionAuthentication)
+    permission_classes = (IsOwner, IsAuthenticated)
+
+    def get_queryset(self):
+        """Get a queryset that only displays the owner's preferences."""
+        return self.queryset.filter(user=self.request.user)
+
+    def perform_create(self, serializer):
```
chambridge: You can remove this since you are doing anything different than the parent class.

koku/api/iam/view/user_preference.py
```diff
@@ -0,0 +1,272 @@
...
+from rest_framework import exceptions, mixins, viewsets
+from rest_framework.authentication import (SessionAuthentication,
+                                           TokenAuthentication)
+from rest_framework.permissions import IsAuthenticated
+
+import api.iam.models as models
+import api.iam.serializers as serializers
+from api.common.permissions.owner import IsOwner
+
+
```
chambridge: Not exactly as was described here: https://github.com/project-koku/koku/issues/97

But I am ok with it. May have misunderstood, but I assumed user -> user preference would be one-to-one. 

Was thinking a `PATCH` against `api/v1/users/(user_id)/preferences/` and a `GET` on `api/v1/users/(user_id)/preferences/`

koku/api/iam/view/user_preference.py
```diff
@@ -0,0 +1,272 @@
...
+                "Authorizaton": "Token 45138a913da44ab89532bab0352ef84b"
+            }
+
+        @apiParam {Number} id User unique ID.
+
+        @apiParam (Request Body) {String} preference a stringified JSON object
+        @apiParamExample {json} Request Body:
+            {
+                "name": "my-preference-name",
+                'preference': {'test-pref': ['a', ['b', 'c'], {'foo': 'bar'}]},
```
chambridge: Let's use an example for a preference we want as I don't think we want store any random JSON for preferences.

koku/api/iam/view/user_preference.py
```diff
@@ -0,0 +1,272 @@
...
+            }
+
+        @apiSuccess {Number} uuid The identifier of the preference.
```
chambridge: uuid is a String. I had to fix this in customer and user views:
```
@apiSuccess {String} uuid The identifier of the preference.
@apiSuccess {String} user_uuid The identifier of the user.
```

koku/api/iam/view/user_preference.py
```diff
@@ -0,0 +1,272 @@
...
+            raise exceptions.PermissionDenied()
+
+        user = models.User.objects.get(uuid=kwargs['user_uuid'])
+        request.data['user'] = user
+        return super().create(request=request, args=args, kwargs=kwargs)
+
+    def list(self, request, *args, **kwargs):
+        """Obtain the list of preferences for the user.
+
+        @api {get} /api/v1/user/:id/preferences/ Obtain a list of user preferences
```
chambridge: switch id --> uuid

koku/api/iam/view/user_preference.py
```diff
@@ -0,0 +1,272 @@
...
+        @apiVersion 1.0.0
+        @apiDescription Obtain the list of preferences for the user.
+
+        @apiHeader {String} token User authorizaton token.
+        @apiHeaderExample {json} Header-Example:
+            {
+                "Authorizaton": "Token 45138a913da44ab89532bab0352ef84b"
+            }
+
+        @apiParam {Number} id User unique ID.
```
chambridge: Same String comment and use uuid.

koku/api/iam/view/user_preference.py
```diff
@@ -0,0 +1,272 @@
...
+        @apiVersion 1.0.0
+        @apiDescription Get a user preference.
+
+        @apiHeader {String} token User authorizaton token.
+        @apiHeaderExample {json} Header-Example:
+            {
+                "Authorizaton": "Token 45138a913da44ab89532bab0352ef84b"
+            }
+
+        @apiParam {Number} user_id User unique ID.
```
chambridge: String instead of Number

koku/api/iam/view/user_preference.py
```diff
@@ -0,0 +1,272 @@
...
+        @apiVersion 1.0.0
+        @apiDescription Delete a user preference.
+
+        @apiHeader {String} token User authorizaton token.
+        @apiHeaderExample {json} Header-Example:
+            {
+                "Authorizaton": "Token 45138a913da44ab89532bab0352ef84b"
+            }
+
+        @apiParam {Number} user_id User unique ID.
```
chambridge: String instead of Number

koku/api/iam/view/user_preference.py
```diff
@@ -0,0 +1,272 @@
...
+        @apiVersion 1.0.0
+        @apiDescription Udpate a user preference.
+
+        @apiHeader {String} token User authorizaton token.
+        @apiHeaderExample {json} Header-Example:
+            {
+                "Authorizaton": "Token 45138a913da44ab89532bab0352ef84b"
+            }
+
+        @apiParam {Number} user_id User unique ID.
```
chambridge: String instead of Number

koku/api/iam/view/user_preference.py
```diff
@@ -0,0 +1,272 @@
...
+    lookup_field = 'uuid'
+    queryset = models.UserPreference.objects.all()
+    serializer_class = serializers.UserPreferenceSerializer
+    authentication_classes = (TokenAuthentication,
+                              SessionAuthentication)
+    permission_classes = (IsOwner, IsAuthenticated)
+
+    def get_queryset(self):
+        """Get a queryset that only displays the owner's preferences."""
+        return self.queryset.filter(user=self.request.user)
```
chambridge: With this queryset in place do we really need the permission `IsOwner`, since they can only see and act upon preferences that are associated with the calling user?

koku/api/iam/serializers.py
```diff
@@ -131,11 +134,14 @@ def create(self, validated_data):
...
 
+    user_uuid = serializers.StringRelatedField(source='user.uuid')
+
     class Meta:
         """Metadata for the serializer."""
 
         model = UserPreference
-        fields = ('__all__')
+        fields = ('uuid', 'name', 'description', 'preference', 'user', 'user_uuid')
+        extra_kwargs = {'user': {'write_only': True, 'required': False}}
```
chambridge: Have you looked at this in the Browesable API with `write_only: True` I'm betting it has a selection for user.

I'd guess you could accomplish what you are looking for  with `read_only: True` and without the use of the `user_uuid` field. You could look at the following for what was done with Providers to establish the `created_by`:
https://github.com/project-koku/koku/blob/master/koku/api/provider/serializers.py#L76

koku/api/iam/models.py
blentz: I added `name` late in development, long after I'd set the ordering. Now that `name` is in the schema, yes, that's a more sensible thing to order by.

I just needed to set this to something other than `id` or `uuid` because of how I'd written the tests for the defaults.

koku/api/iam/view/user_preference.py
blentz: Isn't User -> Preference a one-to-many relationship because we're expecting to set preference entries on a per-topic basis? e.g. one row for currency prefs, one row for timezone prefs.

I messaged you and that's what you said when we discussed it last.

chambridge: Like I said I'm ok with this. I think it was a lost in translation interpretation of what you were meaning vs. what I thought you meant. Definitely not asking you to do a big rewrite here.

koku/api/iam/serializers.py
blentz: The problem I was trying to solve was that `create()` needed to be able to set the `user_id`, but I wanted `get()` to display something more meaningful than the internal db `user_id`. I couldn't remember how we had solved that for `Customer`, so this was my lame kludge solution.

Issue #102, opened 2018-05-04
Configuring a Provider (AWS ARN & S3 Bucket)


chambridge:
## User Story
As a user, I want to be able to configure an AWS account so that the service can read the AWS Cost & Usage Reports from an S3 bucket within my account and obtain the AWS organization information to control cost visibility.

## Impacts
API, UI, Docs

## Role
User

## Assumptions

- We will provide docs (maybe embedded in the UI) to describe to the user the steps to configure S3 to capture AWS Cost CSV files.
- We will provide docs (maybe embedded in the UI) to describe to the user the steps to add a policy that provides the service access to the S3 bucket and AWS organization.
- The user will provide the ARN which we will verify access to the S3 bucket and AWS organization.

## API Details

| API  | Detail|
 | --- | --- |
 | **Method** | PATCH | 
 | **Path** | /api/v1/providers/(provider uuid)/ | 
 | **Body** | crn, bucket | 
 | **Response** | uuid, id, name, type (aws), crn, bucket | 

## UI Details
A view exist to configure an AWS account.

## Acceptance Criteria

- [ ] Verify that a user can configure an existing AWS account in their customer and the response provides the account (uuid, id, name, type, crn).
- [ ] Verify that an account cannot be configured if the provided CRN doesn’t allow access to an S3 bucket and AWS organization information.

chambridge:
Looks like this is done in a single POST in the Cloud Health API. Should we just collapse this into https://github.com/project-koku/koku/issues/99?

blentz:
We should probably aim to enable access via Role-based cross-account access - https://docs.aws.amazon.com/IAM/latest/UserGuide/tutorial_cross-account-with-roles.html

adberglund:
I agree with Brett. For a practical example I can point to how it's being done in cloudigrade. From what I remember seeing in the CloudHealth API from a screen share a week+ ago it looked like they provided two ways. 1) What Brett is suggesting. 2) actually creating a user in the customer account that has required access. Not sure why supporting both is important, but role based is nice because we don't have to keep any credentials. If the role exists, we can use `STS` to "assume role" and AWS handles the auth.

chambridge:
Based on https://github.com/project-koku/koku/issues/102#issuecomment-387482203 I have updated #99 so this can be collapsed into a single API.

PR #172, opened 2018-05-23
Enable Provider creation via API.

chambridge:
* AssumeRole produces Credentials
* Able to detect s3 bucket
* Able to describe AWS organization via client


koku/api/provider/serializers.py
```diff
@@ -15,6 +15,9 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 #
 """Provider Model Serializers."""
+import boto3
+import botocore
+import requests
```
adberglund: Same thing here `from requets.exceptions import ConnectionError`.

koku/api/provider/serializers.py
```diff
@@ -60,7 +63,61 @@ class Meta:
...
+    """Check for access to s3 bucket."""
+    s3_exists = True
+    s3_resource = boto3.resource(
+        's3',
+        aws_access_key_id=access_key_id,
+        aws_secret_access_key=secret_access_key,
+        aws_session_token=session_token,
+    )
+    try:
+        s3_resource.meta.client.head_bucket(Bucket=bucket)
```
adberglund: Since we're only interacting with the client, why not just make an s3 client, instead of the resource?

koku/api/provider/serializers.py
```diff
@@ -60,7 +63,61 @@ class Meta:
...
+    """Check for provider organization access."""
+    access_ok = True
+    org_client = boto3.client(
+        'organizations',
+        aws_access_key_id=access_key_id,
+        aws_secret_access_key=secret_access_key,
+        aws_session_token=session_token,
+    )
+    try:
+        org_client.describe_organization()
```
adberglund: This is future-thinking out loud, but once we've decided what our role policy is we might want to expand out this section of checks to make sure each permission in the policy is checked (e.g. instead of just checking for the bucket, see if we can add an event listener).

I still don't have a great view in my head of what we want to automate vs. what we're asking the customer to do as manual on-boarding in their AWS account. As an example, I know we need them to add an IAM role so that we can use assume_role, but will we ask them to setup the cost usage report, or are we trying to do that? I know this is outlined in #102 but maybe we can discuss this and produce the role JSON so we know explicitly what access will ask for.

chambridge: Understood. I have less AWS experience and am trying to learn this stuff on the fly. I created a separate issue for the API, because we need to document what we are asking the user to do (but maybe we can also make it better over time).

koku/api/provider/serializers.py
chambridge: will do.

koku/api/provider/serializers.py
```diff
@@ -60,7 +63,61 @@ class Meta:
...
+    )
+    credentials = assumed_role.get('Credentials')
+    access_key_id = credentials.get('AccessKeyId') if credentials else None
+    secret_access_key = credentials.get('SecretAccessKey') if credentials else None
+    session_token = credentials.get('SessionToken') if credentials else None
+
+    return (access_key_id, secret_access_key, session_token)
+
+
+def _check_s3_access(access_key_id, secret_access_key,
```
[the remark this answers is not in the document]
chambridge: What would the bucket for _our_ AWS account be? This may just be a disconnect I have from what we think we are doing.

koku/api/provider/serializers.py
adberglund: I think the con of using the assumed role is that we either need to grant broad access to all S3 buckets in our role JSON, or have the customer fill in their bucket name into the role JSON to restrict access down to that specific bucket. That becomes a little less copy-paste.

To me the pro of using our account is that once a customer is on-boarded, the bulk of our interaction with their AWS account is just pulling the reports from S3. S3 makes it really easy to provide bucket-level access to other accounts, so for data ingestion, if we're using our account, we wouldn't have to worry about the assume_role step, and the permissions are easily locked to a single bucket.


https://docs.aws.amazon.com/AmazonS3/latest/dev/s3-access-control.html talks about resource based and user based access policies. With resource based you can give another AWS account access to some or all actions on a single bucket.

PR #175, opened 2018-05-24
Bug fixes for issue found during Sprint 1 demo recording.

chambridge:
* OpenAPI doc had typo and was missing token endpoint
* Customer owners did not get default preferences
* Reset password endpoint incorrectly required authentication
* Reset password did not save the user after password reset
* Added filter by name support to user preferences

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -12,6 +12,7 @@ whitenoise = ">=3.3.0"
 Django = ">=2.0.0"
 psycopg2-binary = ">=2.7.0"
 drf-nested-routers = ">=0.90.0"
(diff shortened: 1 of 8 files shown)


Pipfile
```diff
@@ -12,6 +12,7 @@ whitenoise = ">=3.3.0"
 Django = ">=2.0.0"
 psycopg2-binary = ">=2.7.0"
 drf-nested-routers = ">=0.90.0"
+django-filter = ">=1.1.0"
```
blentz: I don't see where django-filter is added to `INSTALLED_APPS` in settings.

PR #177, opened 2018-05-24
update swagger spec; add preference

blentz:


docs/source/specs/openapi.yml
```diff
@@ -635,79 +829,124 @@ components:
...
-        - $ref: "#/components/schemas/User"
+        - $ref: '#/components/schemas/User'
+        - required:
+            - uuid
+      properties:
+        uuid:
+          type: string
+          format: uuid
+          example: 57e60f90-8c0c-4bd1-87a0-2143759aae1c
+    UserPreferenceIn:
```
chambridge: I don't think we need both `UserPreference` and `UserPreferenceIn`.

Your definition adds two `preference` fields when I look at it in the editor:
image

docs/source/specs/openapi.yml
```diff
@@ -635,79 +829,124 @@ components:
...
+        results:
+          type: array
+          items:
+            $ref: '#/components/schemas/UserPreferenceOut'
+    UserPreferenceOut:
+      allOf:
+        - $ref: '#/components/schemas/UserPreference'
         - required:
-          - uuid
+            - uuid
```
chambridge: You also need `- user` per current output:
```
{
    "uuid": "607fc4ec-0264-4c99-ab1b-50327a3693d8",
    "name": "currency",
    "description": "default preference",
    "preference": {
        "currency": "USD"
    },
    "user": {
        "uuid": "dc3cd5d1-5cfc-433e-b00a-cafe14beed4c",
        "username": "sergio",
        "email": "sergio@example.com"
    }
}
```

docs/source/specs/openapi.yml
```diff
@@ -635,79 +829,124 @@ components:
...
       properties:
         uuid:
           type: string
           format: uuid
-          example: "57e60f90-8c0c-4bd1-87a0-2143759aae1c"
+          example: 57e60f90-8c0c-4bd1-87a0-2143759aae1c
```
chambridge: Update:
```
    UserPreferenceOut:
      allOf:
        - $ref: '#/components/schemas/UserPreference'
        - required:
            - uuid
      properties:
        uuid:
          type: string
          format: uuid
          example: 57e60f90-8c0c-4bd1-87a0-2143759aae1c
        user:
          $ref: '#/components/schemas/UserOut'
```

docs/source/specs/openapi.yml
blentz: I see both a User and UserIn, Provider and ProviderIn, Customer and CustomerIn.  I'm not sure I understand why Preference would be different.

docs/source/specs/openapi.yml
```diff
@@ -309,14 +305,214 @@ paths:
...
+            application/json:
+              schema:
+                $ref: '#/components/schemas/UserPreferencePagination'
+        '500':
+          description: Unexpected Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+  '/api/v1/users/{user_uuid}/preferences/{pref_uuid}':
```
chambridge: Can you end this with a slash? Django can be particular so I want to have the docs always have the ending `/`.

`'/api/v1/users/{user_uuid}/preferences/{pref_uuid}/':`

docs/source/specs/openapi.yml
```diff
@@ -309,14 +305,214 @@ paths:
...
+            type: string
+            format: uuid
+        - name: pref_uuid
+          in: path
+          description: ID of preference to get
+          required: true
+          schema:
+            type: string
+            format: uuid
+      responses:
```
chambridge: Missing the request body:
```
      requestBody:
        description: Update to a Preference
        required: true
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/UserPreferenceIn'
```

PR #182, opened 2018-05-24
107 customer schema

adberglund:
I'm putting this up as a Work in Progress just to put forth the concept of multi-tenancy via django-tenant-schemas and also to get opinions on the models for the cost usage report.

I was initially thinking I'd just piggy-back the creation of a "tenant" on customer creation. It turns out we create a *lot* of customers in our unit tests so doing so basically creates another customer schema for every test we run. It pushed unit tests up to 200 seconds, so maybe mocking out that behavior in tests is good, or it should be a separate API call?


koku/api/apps.py
```diff
@@ -36,6 +36,7 @@ def ready(self):
         try:
             self.startup_status()
             self.check_and_create_service_admin()
+            self.get_or_create_public_schema()
```
blentz: I would like to find another way to do this (and the service_admin bits, too).

The problem with hooking into `ready()` is that it executes when you do *anything* with the app, including attempting to tab-complete `manage.py` commands on the CLI. (I'm regularly irritated by this because I get `status` output anytime I try to tab-complete `manage.py makemigrations`  :unamused: )

Anyhow, if this *needs* to be here, fine. But if it can live anywhere else, I'd really like to see it not get implemented here.

adberglund: I will look into using a migration (specifically a "data migration" to add our schema tenant, and, if I can, the service admin.  https://docs.djangoproject.com/en/1.10/topics/migrations/#data-migrations

PR #191, opened 2018-05-30
Enable provider removal. Closes #101

### LINKED ISSUE #101: Remove a Provider (AWS Account) for a Customer
## User Story
As a customer owner, I want to be able to remove an AWS account associated with my customer so that service will no longer try to gather cost visualization data.

## Impacts
API, UI, Docs

## Role
Customer Owner and/or User responsible for provider creation

## Assumptions
- An AWS account can only be associated with a single customer.

## API Details

| API  | Detail|
 | --- | --- |
 | **Method** | DELETE | 
 | **Path** | /api/v1/providers/(provider uuid)/ | 
 | **Body** |  | 
 | **Response** |  | 

## UI Details
A view exist to remove an existing AWS account.

## Acceptance Criteria

- [x] Verify that any customer owner can remove an existing AWS accounts for a customer.
- [x] Verify that the user (doesn't have to be owner) can remove the AWS account that it added.
- [x] Verify that a non-customer owner cannot remove an existing AWS account that they did not create.
- [x] Verify that upon customer delete that all associated provider data has been removed

dccurtis:
Enables provider removal to close #101.

Overview of change
* Introduced CustomerManager and ProviderManager to carry out common tasks that need to be done by in various parts of the code.  The idea was to create a new abstraction between the view and the database to consolidate all database marshaling and querying requests.  This will enable for more maintainable, re-usable, and testable code.

* Updated the customer delete view handling to also delete associated providers.

Functional Test Results
[provider_removal.txt](https://github.com/project-koku/koku/files/2054791/provider_removal.txt)

### DIFF
--- a/koku/api/iam/customer_manager.py
+++ b/koku/api/iam/customer_manager.py
@@ -65,6 +65,10 @@ class CustomerManager:
         except ValidationError as e:
             raise(CustomerManagerError(str(e)))
 
+    def get_model(self):
+        """Get the model object for the customer."""
+        return self.model
+
     def get_users_for_customer(self):
         """List of users that belong to the customer."""
         group = CustomerSerializer.get_authentication_group_for_customer(self.model)
--- a/koku/api/iam/customer_manager.py
+++ b/koku/api/iam/customer_manager.py
@@ -1,0 +1,84 @@
+#
+# Copyright 2018 Red Hat, Inc.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+"""Management capabilities for Customer functionality."""
+
+import logging
+
+from django.core.exceptions import ValidationError
+
+from api.iam.models import Customer
+from api.iam.serializers import CustomerSerializer
+
+LOG = logging.getLogger(__name__)
+
+
+class CustomerManagerError(Exception):
+    """General Exception class for CustomerManager errors."""
+
+    def __init__(self, message):
+        """Set custom error message for CustomerManager errors."""
+        self.message = message
+
+
+class CustomerManagerDoesNotExist(Exception):
+    """CustomerManager could not find customer error."""
+
+    def __init__(self, message):
+        """Set custom error message for CustomerManager errors."""
+        self.message = message
+
+
+class CustomerManagerPermissionError(Exception):
+    """CustomerManager permission error."""
+
+    def __init__(self, message):
+        """Set custom error message for CustomerManager errors."""
+        self.message = message
+
+
+class CustomerManager:
+    """Customer Manager to manage operations related to customer tasks."""
+
+    def __init__(self, uuid):
+        """Establish provider manager database objects."""
+        self.model = None
+        try:
+            query = Customer.objects.all().filter(uuid=uuid)
+            self.model = query.get()
+        except Customer.DoesNotExist as e:
+            msg = 'Could not find customer with uuid: {}'.format(uuid)
+            raise CustomerManagerDoesNotExist(msg)
+        except ValidationError as e:
+            raise(CustomerManagerError(str(e)))
+
+    def get_users_for_customer(self):
+        """List of users that belong to the customer."""
+        group = CustomerSerializer.get_authentication_group_for_customer(self.model)
+        users = CustomerSerializer.get_users_for_group(group)
+        return users
+
+    def remove_users(self, current_user):
+        """Remove users associated with this customer."""
+        if current_user.is_superuser:
+            users = self.get_users_for_customer()
+            if users:
+                for user in users:
+                    LOG.info('Removing User: {}'.format(user))
+                    user.delete()
+        else:
+            err_msg = '{} does not have pemrissions to remove users.'.format(current_user)
+            raise CustomerManagerPermissionError(err_msg)
--- a/koku/api/iam/customer_manager.py
+++ b/koku/api/iam/customer_manager.py
@@ -69,6 +69,10 @@ class CustomerManager:
         """Get the model object for the customer."""
         return self.model
 
+    def get_name(self):
+        """Get the name of the customer."""
+        return self.model.name
+
     def get_users_for_customer(self):
         """List of users that belong to the customer."""
         group = CustomerSerializer.get_authentication_group_for_customer(self.model)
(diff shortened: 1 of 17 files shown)


koku/api/iam/customer_manager.py
```diff
@@ -0,0 +1,92 @@
...
+from api.iam.models import Customer
+from api.iam.serializers import CustomerSerializer
+
+LOG = logging.getLogger(__name__)
+
+
+class CustomerManagerError(Exception):
+    """General Exception class for CustomerManager errors."""
+
+    def __init__(self, message):
```
adberglund: I like that this is explicit, but I think the init is unnecessary. Subclassing an Exception with a `pass` should also let you pass in your message just as you are down below.

koku/api/iam/customer_manager.py
```diff
@@ -0,0 +1,92 @@
...
+        """List of users that belong to the customer."""
+        group = CustomerSerializer.get_authentication_group_for_customer(self.model)
+        users = CustomerSerializer.get_users_for_group(group)
+        return users
+
+    def remove_users(self, current_user):
+        """Remove users associated with this customer."""
+        if current_user.is_superuser:
+            users = self.get_users_for_customer()
+            if users:
```
adberglund: I believe `get_users_for_group()` returns a `QuerySet` which is iterable, so the `if users` check is not needed. `for user in users` is all you need.

koku/api/iam/test/test_customer_manager.py
```diff
@@ -0,0 +1,141 @@
...
+
+    def setUp(self):
+        """Set up the Customer Manager tests."""
+        super().setUp()
+        self.create_service_admin()
+
+    def tearDown(self):
+        """Tear down the Customer Manager tests."""
+        super().tearDown()
+        User.objects.all().delete()
```
adberglund: I know you are just following suit with other tearDown methods, but it turns out the Django TestCase wraps each test in its own transaction, so deleting objects is superfluous. They just get rolled back after each test anyway.

koku/api/iam/customer_manager.py
```diff
@@ -0,0 +1,92 @@
...
+        return self.model.name
+
+    def get_users_for_customer(self):
+        """List of users that belong to the customer."""
+        group = CustomerSerializer.get_authentication_group_for_customer(self.model)
+        users = CustomerSerializer.get_users_for_group(group)
+        return users
+
+    def remove_users(self, current_user):
+        """Remove users associated with this customer."""
```
adberglund: In retrospect it seems obvious now, but it took me a running through the tests for a few minutes for it to really sink in that this method always deletes all available users for the customer, and that it performs differently based on the user getting passed in. It may benefit from a follow up description in the doc string. 

Tangentially, we (the team) haven't discussed docstring style, with the exception of api and doc formatting in some docstrings. We could maybe benefit from landing on a particular docstring style? Plug: I find the Google docstring style to be quite readable.

koku/api/iam/test/tests_customer_view.py
```diff
@@ -141,15 +174,44 @@ def test_delete_customer(self):
...
+        # Setup Providers
+        iam_arn = 'arn:aws:s3:::my_s3_bucket'
+        bucket_name = 'my_s3_bucket'
+        token = customer_1['owner']['token']
+        response = self._create_provider(bucket_name, iam_arn, token)
+        self.assertEqual(response.status_code, 201)
+
+        bucket2_name = 'my_other_s3_bucket'
+        response = self._create_provider(bucket2_name, iam_arn, token)
+        self.assertEqual(response.status_code, 201)
```
adberglund: I think we need to fail when the same ARN is used for different providers. ARN should be unique. I just looked at the model, and realized it does not have a unique constraint on it, and it probably should have it. I can make an issue for that to handle elsewhere.

koku/api/iam/test/tests_customer_view.py
```diff
@@ -141,15 +174,44 @@ def test_delete_customer(self):
...
+        client.credentials(HTTP_AUTHORIZATION=token)
+        response = client.get(provider_url)
+        self.assertEqual(response.status_code, 200)
+        json_result = response.json()
+        results = json_result.get('results')
+        self.assertIsNotNone(results)
+        self.assertEqual(len(results), 2)
+        provider_uuids = []
+        for provider in results:
+            provider_uuids.append(provider['uuid'])
```
adberglund: I know some people fine them intolerable, but I ❤️ comprehensions. No need to change this, but did you know you can do

```
provider_uuids = [provider['uuid'] for provider in results]
```

It initializes and populates the variable.

koku/api/provider/provider_manager.py
```diff
@@ -0,0 +1,81 @@
...
+    @transaction.atomic
+    def remove(self, current_user, customer_remove_context=False):
+        """Remove the provider with current_user."""
+        # Provider is removable by a service admin only if it's in the context of customer removal
+        force_removal_allowed = False
+        if current_user.is_superuser and customer_remove_context is True:
+            force_removal_allowed = True
+
+        if force_removal_allowed or self.is_removable_by_user(current_user):
+            LOG.info('Provider: {} removed by {}'.format(self.model.name, current_user.username))
```
adberglund: Nitpick: log after the actual delete in case the delete goes bad, the record is clean.

koku/api/provider/test/tests_views.py
```diff
@@ -204,3 +211,161 @@ def test_create_provider_as_service_admin(self):
...
+        other_user_data = {'username': 'sally', 'password': 'sallypassword', 'email': 'sally@me.com'}
+        other_user = self.create_user(owner_token, other_user_data)
+        self.assertEquals(other_user.status_code, 201)
+        other_user_token = self.get_token(other_user_data['username'], other_user_data['password'])
+
+        # Remove Provider as a regular user that belongs to the same company but did not create the Provider
+        url = reverse('provider-detail', args=[json_result.get('uuid')])
+        client = APIClient()
+        client.credentials(HTTP_AUTHORIZATION=other_user_token)
+        response = client.delete(url)
```
adberglund: Think this needs an assertion here?

koku/api/iam/test/test_customer_manager.py
dccurtis: good to know!

koku/api/iam/customer_manager.py
dccurtis: Yeah.  I definitely feel some strict style guidelines on docstrings is a great idea.  This is a good example to illustrate that we should look at a more uniform way to deal with permissions across koku.

koku/api/iam/test/tests_customer_view.py
dccurtis: Yes, we discussed offline that this should be addressed in a separate change.  I'll update the test case to be more realistic from the user perspective.

koku/api/iam/customer_manager.py
dccurtis: turns out it does return a NoneType.  Can't make this change after all.

koku/api/provider/provider_manager.py
dccurtis: good catch

koku/api/provider/test/tests_views.py
dccurtis: yup, before it was just passively verifying that nothing blew up.

PR #192, opened 2018-05-30
Fixes to get OpenShift app building and deploying again

adberglund:
There were a series of issues causing the template to not load, then the build to fail, and then the container/pod to crash on deploy.

### DIFF
--- a/openshift/init-app.sh
+++ b/openshift/init-app.sh
@@ -47,6 +47,7 @@ OPENSHIFT_USER='system:admin'
 OPENSHIFT_TEMPLATE_PATH='openshift/koku-template.yaml'
 CODE_REPO='https://github.com/project-koku/koku.git'
 REPO_BRANCH='master'
+EMAIL_SERVICE_PASSWORD=$EMAIL_SERVICE_PASSWORD
 
 # accept cli arguments for params
 PARAMS=""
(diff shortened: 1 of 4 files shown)


openshift/init-app.sh
```diff
@@ -47,6 +47,7 @@ OPENSHIFT_USER='system:admin'
 OPENSHIFT_TEMPLATE_PATH='openshift/koku-template.yaml'
 CODE_REPO='https://github.com/project-koku/koku.git'
 REPO_BRANCH='master'
+EMAIL_SERVICE_PASSWORD=$EMAIL_SERVICE_PASSWORD
```
adberglund: It might also be good to document that this is required in the docs PR that is out right now, that is if it hasn't already been @chambridge

openshift/s2i/bin/run
```diff
@@ -75,7 +75,7 @@ manage_file=$APP_HOME/manage.py
 if should_migrate; then
   if [[ -f "$manage_file" ]]; then
     echo "---> Migrating database ..."
-    python "$manage_file" migrate --noinput
+    python "$manage_file" migrate
```
blentz: I'm pretty sure we want the `--noinput` flag because we don't want `migrate` prompting us for things like missing default values. There's nobody at the terminal to respond to such prompting.

adberglund: Yep, adding that guy back in. No clue why this was failing on my local build, but it works again with the flag!

PR #198, opened 2018-05-31
Add basic reporting API for cost data, instance data, and storage data.

chambridge:
Initial API suggestion to support basic dashboards.

Assume we will continue to tweak for future enhancements.


docs/source/specs/openapi.yml
```diff
@@ -729,6 +771,135 @@ paths:
...
+        '500':
+          description: Unexpected Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+  /api/v1/reports/storage/:
+    get:
+      summary: Query to obtain storage reports
+      operationId: getCostReports
```
dccurtis: getStorageReports

docs/source/specs/openapi.yml
```diff
@@ -892,6 +1063,87 @@ components:
...
+      type: string
+      enum:
+        - account
+        - storage_type
+    ReportResolution:
+      type: string
+      enum:
+        - daily
+        - monthly
+    ReportTimeScope:
```
dccurtis: Maybe it's just me but I'm not sure how I feel about the negative numbers.  Is there any interesting background on this?

chambridge: I used a limited set from the Cloud Health API as the inspiration.
http://apidocs.cloudhealthtech.com/#reporting_understand-time-filters

docs/source/specs/openapi.yml
chambridge: Thanks. Copy & paste.

docs/source/specs/openapi.yml
```diff
@@ -729,6 +771,135 @@ paths:
...
+          name: resolution
+          schema:
+            $ref: '#/components/schemas/ReportResolution'
+          description: The resolution to apply to the cost report. Default resolution is daily.
+        - in: query
+          name: time_scope
+          schema:
+            $ref: '#/components/schemas/ReportTimeScope'
+          description: The time scope to apply to the cost report. Default resolution is '-10', meaning the last 10 days. Last 30 days would be '-30'. '-10' and '-30' are only valid when used with 'daily' resolution.'-1' is used for current month, '-2' is used for last month when resolution is 'monthly' otherwise invalid.
+      responses:
```
dccurtis: Since all of these end points are queries, shouldn't we also be returning a 404 or will they always get back a 200 with an empty report?

chambridge: My intent is we would give back 200 with an empty report.

dccurtis: ok.  I'll leave it to @dmiller9911 as to how that impacts him.

chambridge: Ideally we would still give back the bucketed time just with 'null' for the values to clarify no data for that time period

dccurtis: Should we go ahead and add `nullable: true` as explained [here](https://swagger.io/docs/specification/data-models/data-types/)

docs/source/specs/openapi.yml
```diff
@@ -729,6 +771,135 @@ paths:
...
+                $ref: '#/components/schemas/ReportCost'
+        '401':
+          description: Unauthorized
+        '500':
+          description: Unexpected Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+  /api/v1/reports/instances/:
```
blentz: How will this scale to include the full array of services?  I'm worried this is a bit too EC2-centric. How will this work for reports on e.g. Lambda or OpsWorks or SES?

What do you think about having the endpoint be `/api/v1/reports/inventory` with a `service` or `product` parameter (or kwarg passed as a dict to a `filter` param) to specify which pieces of the inventory to return?

chambridge: I'm ok with change to `inventory`. I suppose we have a default understanding that we will only return instances if no filter is provided for now (since that is the current requirement)?

chambridge:
@blentz Pushed an updated to address your comments.

docs/source/specs/openapi.yml
```diff
@@ -729,6 +771,105 @@ paths:
...
+          name: Authorization
+          description: e.g. Token 45138a913da44ab89532bab0352ef84b
+          schema:
+            type: string
+          required: true
+        - in: query
+          name: filter
+          schema:
+            $ref: '#/components/schemas/ReportFilter'
+          description: The filter to apply to the report.
```
blentz: Not a blocker, but beyond `filter`, we'll likely also need `group_by` and `order`.

docs/source/specs/openapi.yml
```diff
@@ -729,6 +771,105 @@ paths:
...
+                $ref: '#/components/schemas/ReportInventory'
+        '401':
+          description: Unauthorized
+        '500':
+          description: Unexpected Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+  /api/v1/reports/storage/:
```
blentz: I think now that we have the `/inventory/` endpoint, storage will be a portion of the inventory that can be filtered upon.

chambridge: ok. I will remove the storage endpoint and update the resource_scope to add storage types.

docs/source/specs/openapi.yml
chambridge: I'll pull `group_by` out of `filter`. What did you have in mind in terms of `order`? more than just `ascending` vs. `descending`?

chambridge:
@blentz Updated again.

docs/source/specs/openapi.yml
blentz: Yeah, ASC and DESC are fine.  

The short version is that in today's conversation about what the UI/UX will want to do, it'd be useful to be able to specify as part of the API query so that the response doesn't require additional re-sorting within the UI before it's displayed to the user.

So, the use case is, for example: `filter by ($service == "RDS"), group_by ($account && $region), order_by $cost DESC`

chambridge:
@blentz Updated API doc to be able to support your `order_by` and `group_by` case.

PR #205, opened 2018-06-01
Remove cost usage. Closes #173

### LINKED ISSUE #173: Remove cost usage data upon customer removal
## User Story
As a service admin, I want to ensure that cost usage data has been removed upon customer deletion.

## Impacts
Docs

## Role
Service Admin

## Acceptance Criteria

- [ ] Verify that upon customer delete that all associated cost entry data has been removed

dccurtis:
Removing customer tenant and cost usage data when a customer is removed.  This closes #173 .

Overview of the change:
* CustomerManager was updated to handle removing it's tenant.
* auto_drop_schema = True in our TenantMixin class will remove the associated reporting tables when the tenant is removed.  While working on this I verified that this is exactly what happens by setting this to False and peeking into the db.  I could see the tenant was removed but the reporting tables still existed.
* Updated customer view remove flow to remove the tenant after the users.

Since this is an internal and we don't have the report API ready, I'm not sure how QE can verify this, other then verifying the database contents directly. @LandoCalrizzian what do you think?
 
Functional Tests:
[issue_173_ut.txt](https://github.com/project-koku/koku/files/2063932/issue_173_ut.txt)


koku/api/iam/customer_manager.py
```diff
@@ -59,10 +60,26 @@ def __init__(self, uuid):
...
+            LOG.error(msg)
+
     def get_model(self):
         """Get the model object for the customer."""
         return self.model
 
+    def _get_tenant_query(self):
+        """Get the tenant model query for the customer."""
+        customer_schema_name = self.get_model().schema_name
+        query = Tenant.objects.all().filter(schema_name=customer_schema_name)
```
chambridge: I'm pretty sure you can simplify this from `Tenant.objects.all().filter(schema_name=customer_schema_name)` to `Tenant.objects.filter(schema_name=customer_schema_name)`

Similar statement for other filter calls here. No need to get all records first.

koku/api/iam/test/tests_customer_manager.py
```diff
@@ -135,3 +135,85 @@ def test_invalid_uuid(self):
...
+        tenant = Tenant(schema_name=customer_obj.schema_name)
+        tenant.save()
+
+        # Get manager
+        manager = CustomerManager(customer_uuid)
+        customer_obj = manager.get_model()
+        customer_schema_name = customer_obj.schema_name
+
+        # Verify tenant is returned from manager
+        tenant_obj = Tenant.objects.all().filter(schema_name=customer_schema_name).get()
```
chambridge: Same comment about not needing `all()`.

PR #208, opened 2018-06-06
Add CORS support through middleware

dmiller9911:
Added the django-cors-headers middleware support CORS. I have disabled CORS in production for now until I can figure out what the proper origin will be from insights.


koku/koku/settings.py
```diff
@@ -256,3 +258,10 @@
 KOKU_DEFAULT_CURRENCY = ENVIRONMENT.get_value('KOKU_DEFAULT_CURRENCY', default='USD')
 KOKU_DEFAULT_TIMEZONE = ENVIRONMENT.get_value('KOKU_DEFAULT_TIMEZONE', default='UTC')
 KOKU_DEFAULT_LOCALE = ENVIRONMENT.get_value('KOKU_DEFAULT_LOCALE', default='en_US.UTF-8')
+
+
+# Cors Setup
+# See https://github.com/ottoyiu/django-cors-headers
+
+# SECURITY WARNING: Replace this with proper origins once UI is deployed in insights
+CORS_ORIGIN_ALLOW_ALL = bool(DEBUG)
```
adberglund: There's a chance this doesn't do what you think it does. In Python `bool('any string')` will return `True` so `bool('False') == True`. If you could just make this `CORS_ORIGIN_ALLOW_ALL = DEBUG` and I'll file a bug against our debug assignment at the top of this file.

dmiller9911: ah thanks.  I was following what the linter was saying to do.

PR #209, opened 2018-06-06
apidoc: Add missing 'name' entry in create user preference request body description

LandoCalrizzian:
User create preference - 'name' & optional 'description' was missing from the request body description and example

Also, is there any requirement that the value of 'name' will match the value of the key in preference?

### DIFF
--- a/koku/api/iam/view/user_preference.py
+++ b/koku/api/iam/view/user_preference.py
@@ -76,11 +76,15 @@ class UserPreferenceViewSet(mixins.CreateModelMixin,
 
         @apiParam {String} user_uuid User unique ID.
 
-        @apiParam (Request Body) {String} preference a stringified JSON object
+        @apiParam (Request Body) {String} name User preference name.
+                                               No requirement to match the key in preference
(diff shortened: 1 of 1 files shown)


koku/api/iam/view/user_preference.py
```diff
@@ -76,11 +76,15 @@ def create(self, request, *args, **kwargs):
...
 
-        @apiParam (Request Body) {String} preference a stringified JSON object
+        @apiParam (Request Body) {String} name User preference name.
+                                               No requirement to match the key in preference
+        @apiParam (Request Body) {Object} preference Preference value
+        @apiParam (Request Body) {String} [description] The description of the user preference.
         @apiParamExample {json} Request Body:
             {
-                "name": "my-preference-name",
+                'name': 'my-preference-name',
```
chambridge: Should be double quotes to be valid JSON.

```
"name": "my-preference-name",
"preference": {"currency": "USD"},
"description": "description of my-preference-name"
```

LandoCalrizzian: Good catch. updated

PR #214, opened 2018-06-08
Convert DEBUG string to bool

adberglund:


koku/koku/settings.py
```diff
@@ -51,6 +51,11 @@
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = os.getenv('DJANGO_DEBUG', True)
 
+if DEBUG == 'True':
+    DEBUG = True
+elif DEBUG == 'False':
+    DEBUG = False
+
```
chambridge: Should we add a catch-all?
```
else:
    DEBUG = False
```

PR #229, opened 2018-06-15
apidoc: Only user can delete assigned provider

LandoCalrizzian:
api docs update to correctly state that the User authorization token is required to delete a provider


koku/api/provider/view.py
```diff
@@ -271,7 +271,7 @@ def destroy(self, request, *args, **kwargs):
             @apiVersion 1.0.0
             @apiDescription Delete a provider.
 
-            @apiHeader {String} token Service Admin authorization token.
+            @apiHeader {String} token User authorization token.
```
chambridge: If you want to be specific here its either the *Created By Use* or the *Customer Owner*.

LandoCalrizzian: I like that even better. Updating the doc

PR #233, opened 2018-06-19
Add report query support for total cost filtered by a time scope and sliced by resolution.

chambridge:
* Update OpenApi and Apidoc for time scope updates
* Update serializer parsing of time scope updates
* Utilize time_scope_units to determine interval of time (day or month)
* Utilize time_scope_value to determine the relative time frame
* Utilize resolution to determine data slicing (daily or monthly)


koku/api/report/queries.py
```diff
@@ -0,0 +1,331 @@
...
+        Returns:
+            (List[Dict]): List of data for all intervals sliced by resolution
+
+        """
+        current = {}
+        data = []
+        resolution = self.get_resolution()
+        if query_data:
+            current = query_data.pop()
+        while interval:
```
adberglund: A while loop is fine, but since a list is an iterable in Python, you will generally see a for loop instead `for interval in intervals`. That way you don't have to pop your intervals. Also if that interval is to be reused by another call to the function, it's been destroyed by pops.

```
>>> listy = [1,2,3]
>>> def go_pop(l):
...    while l:
...      l.pop()
...
>>> go_pop(listy)
>>> listy
[]
```

Likewise with the query data. If you don't want to alter the passed in list you could use a for loop with the builtin `enumerate`

```
for i, interval in enumerate(intervals):
    current = query_data[i]
```

koku/api/report/queries.py
```diff
@@ -0,0 +1,331 @@
...
+        return data
+
+    def _format_query_response(self):
+        """Format the query response with data.
+
+        Returns:
+            (Dict): Dictionary response of query params, data, and total
+
+        """
+        output = copy.deepcopy(self.query_parameters)
```
adberglund: 👍

koku/api/report/queries.py
```diff
@@ -0,0 +1,331 @@
...
+        if self.get_resolution() == 'daily':
+            trunc_func = TruncDay
+        else:
+            trunc_func = TruncMonth
+
+        with tenant_context(self.tenant):
+            start, end, days = self._get_time_frame_filter()
+            time_frame_query = AWSCostEntryLineItem.objects.filter(
+                usage_start__gte=start,
+                usage_end__lte=end)
```
adberglund: Not to be changed now, but just to keep in our back pocket if this query acts slowly with lots of data... The line item has a foreign key to the cost entry table, and an index on that foreign key. One thing to try later would be doing a query on the cost entry table to get the date ranges, then filter the line item table on that range of ids.

koku/api/report/queries.py
```diff
@@ -0,0 +1,331 @@
...
+
+        time_scope_units = self.get_filter_data('time_scope_units')
+        time_scope_value = self.get_filter_data('time_scope_value')
+        if not time_scope_value:
+            if not time_scope_units:
+                time_scope_value = -10
+            elif time_scope_units == 'month':
+                time_scope_value = -1
+            else:
+                time_scope_value = -10
```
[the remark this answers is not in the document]
chambridge: Uses the relative time index similar to  Cloud Health API:
http://apidocs.cloudhealthtech.com/#reporting_understand-time-filters

koku/api/report/queries.py
```diff
@@ -0,0 +1,331 @@
...
+    @staticmethod
+    def next_month(in_date):
+        """Return the first of the next month from the in_date.
+
+        Args:
+            in_date    (DateTime) input datetime
+        Returns:
+            (DateTime): First of the next month
+        """
+        dt_first = in_date.replace(day=1)
```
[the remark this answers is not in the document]
chambridge: done.

koku/api/report/queries.py
```diff
@@ -0,0 +1,331 @@
...
+        dt_up_month = dt_first + datetime.timedelta(days=32)
+        dt_next_month = dt_up_month.replace(day=1)
+        return dt_next_month
+
+    @staticmethod
+    def previous_month(in_date):
```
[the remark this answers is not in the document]
chambridge: done.

koku/api/report/queries.py
```diff
@@ -0,0 +1,331 @@
...
+            start_date    (DateTime) starting datetime
+            end_date      (DateTime) ending datetime
+        Returns:
+            (List[DateTime]): A list of days from the start date to end date
+        """
+        days = []
+        start_midnight = start_date.replace(hour=0)
+        end_midnight = end_date.replace(hour=0)
+        current = start_midnight
+        while current < end_midnight:
```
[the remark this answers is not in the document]
chambridge: done.

koku/api/report/queries.py
chambridge: I can swap to using a for loop for interval. Was using pop mostly because these lists were potentially of length.
```
interval = [<date=1>, <date=2>, <date=3>, <date=4>]
query_data = [<date=2>]
```
Destruction of the lists didn't matter here.

koku/api/report/queries.py
chambridge: ok.

koku/api/report/queries.py
adberglund: Nah, makes sense.

koku/api/report/queries.py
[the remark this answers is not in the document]
adberglund: Interesting. It's just my brain, but I find their approach non-intuitive. But at least it's explained. Thank you!

chambridge:
@adberglund Changes pushed.

PR #240, opened 2018-06-22
Fixing requirements for read the docs

adberglund:
We should maybe find a better way of automating the generation of the extra requirements file that read the docs needs.


chambridge:
Looks like they are getting close to a solution to just support pipenv:
https://github.com/rtfd/readthedocs.org/issues/3181

PR #265, opened 2018-07-12
Log check for Provider SNS notification. Closes #245

### LINKED ISSUE #245: Log check for Provider SNS notification
During provider initialization add a check for SNS setup to target endpoint.

If not found log an entry.

_____
Associated with: https://github.com/project-koku/masu/issues/7

dccurtis:
**Overview**
When a customer adds their provider information to Koku we would like to have a log to indicate whether or not the S3 bucket they are specifying is currently configured for SNS.

This is only a one-time check when the provider is added.  The additional check being added to the provider create workflow is for logging purposes only and will never prevent provider creation.

**Testing**
1. Create a provider with the SNS configuration set as expected. Verify logs
2. Create a provider with SNS configuration not setup (No events on S3 bucket). Verify logs
Results: [issue_245.txt](https://github.com/project-koku/koku/files/2190262/issue_245.txt)

**Later work**
There is an outstanding user story related to investigating how our SNS Topic Policy will be configured as well as what guidance we will provider for the customer.  As of now Masu is simply a basic SNS HTTP topic handler.  Similarly, on the Koku side, we are only logging what is configured (or not configured).  After the investigation user story is complete we should come back to this check and enhance it to provide more granular checks and logging.

Also, the work of providing regular configuration checks and alerting will be done in a subsequent user story/PR.

Closes #245

### DIFF
--- a/koku/providers/test/aws/tests_aws_provider.py
+++ b/koku/providers/test/aws/tests_aws_provider.py
@@ -22,7 +22,7 @@ import boto3
 from botocore.auth import SigV4Auth
 from botocore.exceptions import ClientError, NoCredentialsError
 from django.test import TestCase
-from moto import mock_s3, mock_sts
+from moto import mock_s3, mock_sns, mock_sts
 from providers.aws.aws_provider import (AWSProvider,
                                         _check_cost_report_access,
                                         _check_org_access,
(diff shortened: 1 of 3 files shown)


koku/providers/test/aws/tests_aws_provider.py
```diff
@@ -209,3 +209,79 @@ def test_check_cost_report_access_fail(self, mock_boto3):
...
+        config = {'TopicConfigurations': [{'TopicArn': topic, 'Events': ['s3:ObjectCreated:*']}]}
+        boto3.client('s3').put_bucket_notification_configuration(Bucket=bucket_name,
+                                                                 NotificationConfiguration=config)
+
+        provider_interface = AWSProvider()
+
+        try:
+            provider_interface.cost_usage_source_is_reachable(iam_arn, bucket_name)
+        except Exception:
+            self.fail('Unexpected Error')
```
blentz: I think this try/except should be an assertion, instead.

koku/providers/test/aws/tests_aws_provider.py
```diff
@@ -209,3 +209,79 @@ def test_check_cost_report_access_fail(self, mock_boto3):
...
+        # Create SNS Topic
+        conn = boto3.client('sns', region_name='us-east-1')
+        conn.create_topic(Name='CostUsageNotificationDemo')
+
+        provider_interface = AWSProvider()
+
+        try:
+            provider_interface.cost_usage_source_is_reachable(iam_arn, bucket_name)
+        except Exception:
+            self.fail('Unexpected Error')
```
blentz: Same as above, I think this should be an assertion.

koku/providers/test/aws/tests_aws_provider.py
dccurtis: I'm not expecting any exceptions to be thrown.  If one is thrown then it will fail the test.  I'm not sure how that's any different from an assertion.  Even if we went with an assertion what would you assert?

blentz: The test will fail when an exception is thrown, regardless. So, it's not necessary to explicitly call `self.fail()`. Simply not catching the exception will also be a failure - the try/catch is unnecessary.

I would expect a function that checks availability to return a boolean telling me whether or not the check(s) succeeded or not. So, then I'd expect the unit test to be an `self.assertTrue`

dccurtis: Yeah thats true it would fail regardless, the try/except is just there to have some failure message.  In this case 'Unexpected Error'

So the purpose of the tests is to establish the conditions that calling `cost_usage_source_is_reachable` can encounter.  Those are 1) S3 bucket has notifications turned on 2) S3 bucket has notifications turned off.  In both of these situations I want to make sure that no exceptions are thrown..  As far as verifying other behavior change all we are doing is logging a message, so there's no systemic change to verify other than the fact that no exceptions are thrown for S3 event notifications.

The `cost_usage_source_is_reachable` regrettably returns an exception when things are bad and does nothing when things are good.  This was a carry over from the original implementation in the serializer code, but anyway..

Do you see anything wrong with the actual implementation?

blentz: Hmmm...  I'd consider it a bug that the function doesn't return anything and relies on throwing exceptions when things are wrong. That's not a great experience. But, it's also not the code this PR is changing.  So, no, _technically_ there's nothing wrong with your specific PR.

dccurtis: Yup, 100% agree with that.  Before isolating the provider-specific code in koku, the provider serializer was essentially raising assertions as the filter checks.  Instead of changing the behavior I just carried that forward since I wanted it to work the same way after moving the logic.  It works, and it's usage is very isolated so I didn't see any real reason to spend a lot of time changing it..

PR #266, opened 2018-07-16
turn off APPEND_SLASH (closes #228)

### LINKED ISSUE #228: HTTP 500 Error when submitting data to a url that is missing a trailing slash
**Describe the bug**
POST requests w/o a trailing slash return a HTTP 500 Server Error message  

**To Reproduce**
1. Issue a Post request without the trailing slash

curl -X 'POST' -H 'Accept: */*' -H 'Accept-Encoding: gzip, deflate' -H 'Connection: keep-alive' -H 'Content-Length: 41' -H 'Content-Type: application/json' -H 'User-Agent: python-requests/2.19.1' -d '{"username": "admin", "password": "pass"}' 'http://127.0.0.1:8000/api/v1/token-auth'

**Expected behavior**
If enforcing the trailing slash requirement, return a 400 error with a message stating that a trailing slash is required --OR-- auto append the missing trailing slash to the url prior to processing the request

**Context (please complete the following information):**
 - Commit Hash from status endpoint: 3c9e278

**Additional Info**
GET requests are unaffected. Could also apply to PUT requests but haven't tested yet.

blentz:
This is a "technically correct" solution, in that we just don't issue redirects for requests that are missing the trailing slash.

I'm not sure if it's the way we want to resolve this particular bug, but it's _a_ solution. :stuck_out_tongue_winking_eye: 

Important Note: POST data is not resubmitted through a 301/302 redirect for good reasons. So, this is a case that if we want to be "user-friendly" and issue redirects for missing slashes, then the application itself needs to retain the POST data by saving it to the user's session somehow (issue a cookie or something similar).  So, the "better" solution is non-trivial.

I'm personally okay with our API just being more strict, which is something we can document as a requirement for any clients accessing our API.

Test results:
```
$ curl -s -o /dev/null -w "%{http_code}\n" -X 'POST' -H 'Accept: /' -H 'Accept-Encoding: gzip, deflate' -H 'Connection: keep-alive' -H 'Content-Length: 41' -H 'Content-Type: application/json' -H 'User-Agent: python-requests/2.19.1' -d '{"username": "admin", "password": "pass"}' 'http://koku-myproject.127.0.0.1.nip.io/api/v1/token-auth' 

404
```

### DIFF
--- a/openshift/koku-template.yaml
+++ b/openshift/koku-template.yaml
@@ -312,7 +312,7 @@ objects:
           livenessProbe:
             failureThreshold: 3
             httpGet:
-              path: /api/v1/status
+              path: /api/v1/status/
               port: 8080
               scheme: HTTP
             initialDelaySeconds: 30
@@ -324,7 +324,7 @@ objects:
               protocol: TCP
           readinessProbe:
             httpGet:
-              path: /api/v1/status
+              path: /api/v1/status/
               port: 8080
               scheme: HTTP
             initialDelaySeconds: 3


chambridge:
@project-koku/quality-engineers Is this the solution you were looking for related to issue: https://github.com/project-koku/koku/issues/228

I'm somewhat hesitant to change a "community" default and would tend to as Brett suggested just stress the documented APIs, but I get that the 500 error code is unappealing.

@blentz Do we get a 500 when `DEBUG=False`? I know sometimes there is a bit of difference.

blentz:
@chambridge - I didn't check, but I expect so. The basic behavior leads me to believe it'll be a 5xx error.

The basic flow is:

1. POST Request for `/foo` is submitted by user-agent.
2. Django issues a 301 redirect to `/foo/`
3. User-agent follows the redirect with a GET or an empty POST
4. The app throws a 500 because it got either the wrong request type or an empty params.

LandoCalrizzian:
@blentz @chambridge - I have no problem with the strict approach of requiring all slashes for every endpoint there was just some inconsistency when certain actions were forgiving of a missing trailing slash while others were not.

My main issue when filing the bug was with the return of a 500 error. In my opinion, if nothing was changed on how the trailing slash was enforced then the error should be caught and handled with a 4XX error.

PR #274, opened 2018-07-24
Enable "Local" provider type

dccurtis:
**Overview**
This change is a step towards creating a local, AWS-like provider to enable more flexibility for both development and testing.  AWS-like means the local directory structure is setup exactly like a cost usage report in an S3 bucket and the reports that it can process must conform to the AWS structure.

Project koku (koku, masu, nise and related services) have a significant dependency on AWS.  While it may be tempting to rely on AWS for all aspects of testing and development, there are some significant costs to doing so.  Some of those include:

* Having to manage AWS accounts and resources,
* Slower development cycles (if you are not working on anything related to direct interaction with the AWS API)
* End-to-end (masu and koku) presubmit/smoke testing on individual PR basis is not practical since a new S3 buckets would have to be programmatically created and populated with each test run to ensure maximum reliability.  Also, smoke tests should generally not have such external dependencies (should be able to be run in any environment, including offline).
* Doing anything at scale where hundreds of S3 buckets are required becomes extremely onerous using AWS.  With this change you can just make hundreds of local directories rather than provision and manage hundreds of S3 buckets.

Koku uses the environment variable `DJANGO_DEBUG` to put the services into a “DEBUG” mode.  The default setting for now is DEBUG=TRUE.  In this mode there is a new provider type called `Local`.  When using the `Local` provider type supply the `Billing Source: bucket` with a local directory path.  If the path exists then Koku will allow for the provider to be created.

When `DJANGO_DEBUG` is set to `False`, Koku is running in it’s production mode and the `Local` provider will not be available.

**What are the costs to using Local provider**
There are some costs to using `Local` provider.  While these may seem obvious it might be helpful to explicitly enumerate them here.  The Koku provider code is structured in such a way that any provider-specfic implementations are neatly isolated from the rest of the service.  These provider specific implementations populate a common interface that the rest of the koku service consumes.  This was done on purpose so that we could plug in other providers (include test providers like Local) and have it still be a meaningful tool for development and testing.

The following code is not exercised when using `Local` provider:
* `koku/providers/aws/aws_provider.py`: Validates that an S3 bucket exists, CUR access, STS access, and provides log info for SNS.

I am not pitching that any test automation using Local provider is a replacement for AWS integration testing, but rather a complement to it that can be ran in a more on-demand fashion.

**Related Work**
* Nise: Has the capability of generating cost usage reports locally by specifying a local directory as the `s3-bucket`
* Masu: PR coming soon to enable local report downloading.

**Testing**
Functional Test Steps:
1. Create a local S3 bucket and populate contents with nise
2. Start koku with DJANGO_DEBUG unset (defaults to True). Attempt to create a provider with 'Local' type
3. Attempt to create a Local provider for a directory that does not exist.  Verify provider creation fails
4. Start koku with DJANGO_DEBUG set to False and make sure 'Local' is not a valid provider type
5. Verify that an AWS provider still works

Results:
[local_provider_koku.txt](https://github.com/project-koku/koku/files/2223733/local_provider_koku.txt)


koku/koku/settings.py
```diff
@@ -189,7 +189,7 @@
     os.path.join(BASE_DIR, '..', 'apidoc'),
 ]
 
-STATICFILES_STORAGE = 'whitenoise.storage.CompressedManifestStaticFilesStorage'
+# STATICFILES_STORAGE = 'whitenoise.storage.CompressedManifestStaticFilesStorage'
```
dccurtis: This was causing problems with `DJANGO_DEBUG = False`.  This looks like its used to store .css files or some files needed to serve web page content.  Open to suggestions on how we should handle this.

adberglund: From http://whitenoise.evans.io/en/stable/django.html#make-sure-staticfiles-is-configured-correctly

> As part of deploying your application you’ll need to run ./manage.py collectstatic to put all your static files into STATIC_ROOT. 

Basically we aren't putting our static files where we're telling the thing to check for static files so it 💥 s. So I think if debug is false we just need a flow that is closer to deploying to production in which we run that first.

koku/providers/test/local/test_local_provider.py
```diff
@@ -0,0 +1,64 @@
...
+        """Verify that the cost usage source is authenticated and created."""
+        iam_arn = 'arn:aws:s3:::my_s3_bucket'
+        bucket_name = self.cur_source
+
+        provider_interface = LocalProvider()
+
+        try:
+            provider_interface.cost_usage_source_is_reachable(iam_arn, bucket_name)
+        except Exception:
+            self.fail('Unexpected Error')
```
adberglund: Pretty optional since the first thing I'd do if this failed is debug and find out what the exception is, but you could append the exception string to your fail string so we see which exception caused the failure.

koku/koku/settings.py
dccurtis: Thats it, thanks!  I'll revert this

PR #288, opened 2018-08-01
fix DEBUG boolean

blentz:
This resolves the error found in Travis runs:
`koku/koku/settings.py:55:8: W1508: os.getenv default type is builtins.bool. Expected str or None. (invalid-envvar-default)`

### DIFF
--- a/koku/koku/settings.py
+++ b/koku/koku/settings.py
@@ -52,8 +52,7 @@ SECRET_KEY = os.getenv(
 )
 
 # SECURITY WARNING: don't run with debug turned on in production!
-DEBUG = os.getenv('DJANGO_DEBUG', True)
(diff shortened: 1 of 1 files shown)


koku/koku/settings.py
```diff
@@ -52,8 +52,7 @@
 )
 
 # SECURITY WARNING: don't run with debug turned on in production!
-DEBUG = os.getenv('DJANGO_DEBUG', True)
-DEBUG = (DEBUG == 'True' or DEBUG is True)
+DEBUG = bool(os.getenv('DJANGO_DEBUG', 'True'))
```
adberglund: This is actually complaining about the default value given to the getenv call. This will work
```
DEBUG = os.getenv('DJANGO_DEBUG')
DEBUG = (DEBUG == 'True' or DEBUG is True)
```

PR #304, opened 2018-08-08
add region mapping table

blentz:
This adds the DB table for maintaining region mapping. The rest of this feature is implemented in Masu PR 132 .

It looks like Django picked up some of the local provider stuff from @dccurtis when creating the migrations, too.


koku/api/migrations/0017_auto_20180808_2134.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('api', '0016_auto_20180802_1911'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='provider',
+            name='type',
+            field=models.CharField(choices=[('AWS', 'AWS')], default='AWS', max_length=50),
```
adberglund: Why did we eliminate the local provider type?

chambridge: Sorry, I missed this change. I didn't focus very hard on the migrations, but we definitely don't want the local provider removed.

PR #305, opened 2018-08-09
Document that count is the distinct number of instances for the respective time period. Closes #275.

### LINKED ISSUE #275: Report inventory instance type: Total count different that sum of report line time count
**Describe the bug**
The inventory instance report total count is different than the sum of the count field from each returned line item.

**To Reproduce**
1. Generate and add nise data
2. call the instance trend endpoint /api/v1/reports/inventory/instance-type/

**Expected behavior**
The count property in the report total is the same number as the sum of the count from each line item

**Context (please complete the following information):**
 - OS: Fedora w/ local minishift instance
 - Commit Hash: e178aff

**Additional context**
I haven't tested this actual AWS data yet

chambridge:


koku/api/report/view.py
```diff
@@ -301,7 +301,7 @@ def instance_type(request):
 
     @apiSuccess {Object} group_by  The grouping to applied to the report.
     @apiSuccess {Object} filter  The filter to applied to the report.
-    @apiSuccess {Object} data  The report data.
+    @apiSuccess {Object} data  The data including types, usage, and distinct number of instances for a time period.
```
LandoCalrizzian: This makes it clearer that the line item count is distinct. To prevent any confusion can we add an additional date to the response for either of the instance types to clearly show that the count is distinct. 
`` 
 {
                "date": "2018-05-29",
                "accounts": [
                    {
                        "account": 111111111111 ,
                        "instance_types": [
                                {
                                    "instance_type": "t2.medium",
                                    "values": [
                                        {
                                            "date": "2018-05-29",
                                            "units": "Hrs",
                                            "instance_type": "t2.medium",
                                            "total":  4,
                                            "count": 1
                                        }
                                    ]
                                }
                            ]
                    }
                ]
            }
``
``
Then total would be updated include the additional total but the count is the same
"total": {
        "value": 38,
        "units": "Hrs",
        "count": 4
    }
``

PR #327, opened 2018-08-16
update Pipfile.lock

blentz:


chambridge:
Can you run `make requirements` so our read the docs requirements are updated too?

Issue #309, opened 2018-08-10
API allows duplicate entries for any user preference


LandoCalrizzian:
The API allows a user to create duplicates for any user preference

1. Create any user 
2. Create a user preference that is a duplicate any existing preference
Method: POST or PUT
Data:
    name: currency
    preference:
        name: currency
        value: USD
3. Update successful

**Expected behavior**
The creation/update should fail since the preference already exists

Context:
  - Commit Hash from status endpoint: 6ed68c

Additional Info:
Should the name and preference name key be unique?

LandoCalrizzian:
@chambridge 
This does allow duplicates (and name updates) when creating a preference name with the same spelling but different capitalization.
POST 
name: Currency
Preference: Currency: USD

PUT
uuid: <Same uuid as timezone>
preference: 
  'Timezone': 'UTC'

LandoCalrizzian:
While this still allows duplicates when capitalization differs from an existing preference, it no longer allows exact duplicates.

PR #334, opened 2018-08-24
validate preference values: locale, currency, timezone

### LINKED ISSUE #308: Invalid values are allowed when updating default user preferences via API
The API allows a user to set invalid values for any of the default user preferences: currency, locale, timezone

1. Create any user 
2. Update a user preference for any of the default preferences with an invalid value:
    currency -> Hugs
    locale -> en_FOO.LOL-EIGHT
    timezone -> LOL
3. Update successful

**Expected behavior**
The update should fail since the new value is not valid

Context:
  - Commit Hash from status endpoint: 6ed68c

Additional Info:
Should the default user preference be locked down to white listed values?  You can even rename the default user preference name, preference name and description

blentz:
This adds validation for the three main preferences: locale, timezone, and currency. 

This change also removes the incorrect uniqueness validation, which would prevent updating preferences. Instead, we'll just rely on the database to throw IntegrityErrors for now.

This should resolve #308.

Arbitrary preferences can still be set. I'm not fully convinced if this is a bug or a feature. The app isn't going to do anything with arbitrary preferences, so this behavior won't directly affect the app's functioning. There is still a theoretical potential for DOS if someone decides to fill up the DB with garbage, but that can be addressed in a future change.

### DIFF
--- a/koku/api/iam/test/tests_user_preference_view.py
+++ b/koku/api/iam/test/tests_user_preference_view.py
@@ -18,7 +18,6 @@
 
 from random import randint
 
-from django.core.exceptions import ValidationError
 from django.urls import reverse
 from faker import Faker
 from rest_framework import status
--- a/koku/api/iam/view/user_preference.py
+++ b/koku/api/iam/view/user_preference.py
@@ -18,7 +18,7 @@
 """View for User Preferences."""
 from django.forms.models import model_to_dict
 from django_filters.rest_framework import DjangoFilterBackend
-from rest_framework import exceptions, mixins, status, viewsets
+from rest_framework import exceptions, mixins, viewsets
 from rest_framework.authentication import (SessionAuthentication,
                                            TokenAuthentication)
 from rest_framework.permissions import IsAuthenticated
@@ -125,7 +125,6 @@ class UserPreferenceViewSet(mixins.CreateModelMixin,
                                                               user=user.username))
 
         return super().create(request=request, args=args, kwargs=kwargs)
-
 
     def list(self, request, *args, **kwargs):
         """Obtain the list of preferences for the user.


koku/api/iam/test/tests_serializers.py
```diff
@@ -219,11 +227,136 @@ def test_user_preference_duplicate(self):
...
         test_pref = {'foo': ['a', [1, 2, 3], {'b': 'c'}]}
-        data = {'name': 'foo', 'description': 'foo', 'preference': test_pref}
+        data = {'name': self.fake.word(),
+                'description': self.fake.text(),
+                'preference': test_pref}
         serializer = UserPreferenceSerializer(data=data, **kwargs)
         if serializer.is_valid(raise_exception=True):
             serializer.save()
 
+        with self.assertRaises(IntegrityError):
```
chambridge: The validation error was a fix for https://github.com/project-koku/koku/issues/309.

I'm pretty sure an Integrity Error will cause a 500 to display which isn't good.

blentz: I'm open to suggestions on how to solve that. 

I _think_ the way to avoid `IntegrityError` is to do a lookup for the preference before setting it, so that we can pass the existing `UserPreference` in to the serializer, so it does an `update()` instead of a `create()`.

chambridge: Just a thought, but you could keep the logic they way it was, but add in an additional check to get the request METHOD. If its a POST throw the error, if its a PUT/PATCH don't throw the error.

blentz: I've added a check in the view's `create()` method that will throw the proper error if the request is trying to create a preference that already exists.

PR #340, opened 2018-08-29
make ordering more generic; use OrderedDict to reduce data transform churn.

blentz:
This change makes the `order-by` parameter a more generic class property. This will facilitate ordering by arbitrary columns.

In addition, we stop mapping "cost" as a query parameter into meaning "total" internally. Instead, the query parameter for total cost will be `total` for both the API and the backend.

Third, there's some cleanups related to integrating with the summary table work. The new `ReportQueryHandler.cost` property is now used consistently throughout the class.

### DIFF
--- a/koku/api/utils.py
+++ b/koku/api/utils.py
@@ -28,7 +28,9 @@ from pint.errors import UndefinedUnitError
 class DateHelper():
     """Helper class with convenience functions."""
 
-    _now = timezone.now()
+    def __init__(self):
+        """Initialize when now is."""
+        self._now = timezone.now()
 
     @property
     def one_day(self):


koku/api/report/queries.py
```diff
@@ -85,36 +88,35 @@ def __init__(self, query_parameters, url_data,
...
+
+        self.operation = self.query_parameters.get('operation', OPERATION_SUM)
+        self._delta = self.query_parameters.get('delta')
+        self._limit = self.get_query_param_data('filter', 'limit')
         self._get_timeframe()
+        self.units_key = units_key
 
         if kwargs:
-            elements = ['accept_type', 'annotations', 'delta',
+            elements = ['accept_type', 'annotations', 'count', 'delta',
```
adberglund: Adding `count` here and using `setattr` overrides the property on the class. I had it as a property, and left the view using `resource_id` to allow for the path where the original `AWSCostEntryLineItem` table is used. In that case the referenced column is `resource_id` The property setter is supposed to override the value to `resource_count` when the summary table is used. If we are certain we never need to use the original resource_id column, we can do this, but then the property should be deleted. If we are uncertain let's not use setattr here and instead let the property handle it.

koku/api/report/queries.py
```diff
@@ -200,18 +202,40 @@ def get_query_param_data(self, dictkey, key, default=None):
...
-    def get_order(self):
-        """Extract the order parameters and apply the associated ordering.
+    @property
+    def order_field(self):
+        """Order-by field name.
+
+        The default is 'total'
+        """
+        order_by = self.query_parameters.get('order_by', self.default_ordering)
+        return list(order_by.keys()).pop()
```
adberglund: Can we take more than one order by value? What if I want to order by account and service?

koku/api/report/queries.py
```diff
@@ -744,16 +766,16 @@ def calculate_total(self, units_value):
             'time_scope_value',
             0
         )
-        total_filter['report_type'] = self._report_type
+        total_filter['report_type'] = getattr(self, '_report_type', 'costs')
```
adberglund: I hesitate to give this a default. The aggregates table is pretty specific to the current API types we have, so potentially bad data will be returned with that default if we forget to pass in the correct report type.

koku/api/report/queries.py
blentz: I'll look a bit more closely at this.

This is where leaving code comments about what's going on and why is super-helpful to call out this kind of subtle behavior. So much of the ReportQueryHandler is un-documented, which makes working on it a giant pain.

Though, I also think it definitely needs reworking to make these behaviors more well-defined and obvious.

PR #343, opened 2018-09-04
Refactor query filtering; clarify behavior, remove unnecessary functions.

blentz:
PR #340 is a pre-requisite for this change.

Changes:
* This reworks filter processing to ensure the translation from API query params to backend DB query filters is consistent.  There are two new classes, `QueryFilter` and `QueryFilterCollection` that handle arranging query filters into a unique list of filters ready for consumption by the Django ORM. This should eliminate the need to second-guess the contents of the query filters and make functions like `_strip_table_references()` unnecessary.

* I've also split `execute_query()` into two functions due to the large number of `if self.is_sum` checks indicating that use case is distinct enough to merit its own code path. I'm expecting this to allow future cleanups and refactoring now that it's clearer what each code branch is trying to accomplish.

Thoughts about the future:

Once this is merged, I'm hoping to apply a similar pattern for handling other elements of query processing, so that eventually the ordering and time_scope functions are also clearer.

### DIFF
--- a/koku/api/report/queries.py
+++ b/koku/api/report/queries.py
@@ -83,7 +83,14 @@ class QueryFilter(UserDict):
     parameter = None
 
     def __init__(self, table=None, field=None, operation=None, parameter=None):
-        """Constructor."""
+        """Constructor.
+
+        Args:
+            table (str) - The name of a DB table
+            field (str) - The name of a DB field
+            operation (str) - The name of a DB operation, e.g. 'in' or 'gte'
+            parameter (object) - A valid query target, e.g. a list or datetime
+        """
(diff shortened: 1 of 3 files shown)


blentz:
I've pushed changes that fix the issue you spotted, @adberglund .   Attached are my test results from queries where `time_scope_value` was set to -1, -2, and -10.

[filter_minus10_daily.txt](https://github.com/project-koku/koku/files/2358332/filter_minus10_daily.txt)
[filter_minus1_daily.txt](https://github.com/project-koku/koku/files/2358333/filter_minus1_daily.txt)
[filter_minus2_daily.txt](https://github.com/project-koku/koku/files/2358334/filter_minus2_daily.txt)

PR #347, opened 2018-09-07
AWSAccountAlias database schema changes

dccurtis:
Database schema changes to add a customer tenant table to hold all of their account ids and aliases.


chambridge:
A provider that is a consolidated account will have many child accounts that will have aliases. Masu gathers all of them, is my assumption @blentz, if that makes sense.

dccurtis:
Yeah, I considered making it a field in one of the report processing tables, to be next to `payer_account_id` or `usage_account_id`.

I decided against it because of a few reasons:
1.  With respect to aliases it really shouldn't matter if its the payer or usage account.  We want to have a generic mechanism to map ID -> alias for a given customer.
2. I didn't want to muddy the waters in the report processor.  Initially I thought maybe we should update the alias in the report processor itself but decided that wasn't the best idea since we would want to be able to refresh the alias table at anytime without firing up a report processor and a customer can update their alias before a new CUR is generated.
3.  Making the alias logic live independently will make it more re-usable with other providers, gives us more control for when alias updates are done, and also reduces the risk of regression by minimizing the changes to existing code.

PR #351, opened 2018-09-12
346 group by deltas

adberglund:
## Summary

The delta param for the reports APIs now returns data at the level of each grouped item. 
* We run the report query against the previous time frame, 
* Then get the grouped totals, package it in a dictionary keyed off the the group by values
* Calculate deltas against the query data for the current report timeframe


koku/api/report/queries.py
```diff
@@ -381,10 +383,12 @@ def get_resolution(self):
 
         if self.resolution == 'monthly':
             self.date_to_string = lambda dt: dt.strftime('%Y-%m')
+            self.string_to_date = lambda dt: datetime.datetime.strptime(dt, '%Y-%m').date()
```
blentz: This needs to be a function pointer, not an actual function call. Just like the `self.date_to_string` above it.

koku/api/report/queries.py
```diff
@@ -813,75 +818,127 @@ def execute_query(self):
...
-                                                        **self.query_filter)
-
         self.query_sum = query_sum
         self.query_data = data
         return self._format_query_response()
 
-    def calculate_delta(self, delta, query, query_sum, **kwargs):
-        """Calculate cost deltas.
+    def add_deltas(self, query_data, query_sum):
+        """Calculate and add cost deltas to a reasult set.
```
blentz: `result` not `reasult`

koku/api/report/queries.py
```diff
@@ -813,75 +818,127 @@ def execute_query(self):
...
         Returns:
-            (dict) with keys "value" and "percent"
+            (dict) A dictionary keyed off the grouped values for the report
 
         """
-        delta_filter = copy.deepcopy(kwargs)
+        if self.time_scope_value in [-1, -2]:
+            date_delta = relativedelta.relativedelta(months=1)
+        else:
+            date_delta = datetime.timedelta(days=10)
```
blentz: This doesn't cover the `time_scope_value=-30&time_scope_units=day` case.

koku/api/report/queries.py
```diff
@@ -813,75 +818,127 @@ def execute_query(self):
...
+        Returns:
+            (dict) A dictionary keyed off the grouped values for the report
 
-        query_delta = {'value': d_value, 'percent': d_percent}
-        return query_delta
+        """
+        if self.time_scope_value in [-1, -2]:
+            date_delta = relativedelta.relativedelta(months=1)
+        else:
+            date_delta = datetime.timedelta(days=10)
```
blentz: This doesn't cover the `time_scope_value=-30&time_scope_units=day` case.

koku/api/report/queries.py
adberglund: Can you give an example of what you mean?

blentz: Taking a second look, I'm wrong. This is fine as-is. Sorry.

PR #370, opened 2018-09-19
Creating OCP and OCP-local providers

### LINKED ISSUE #360: Allow API creation of an OCP provider
## User Story
As a user, I want to be able to add (configure) an OCP provider with Cost Management so I can obtain its usage information

## Impacts
API, UI, Docs

## Role
User

## Assumptions
- User has setup OCP installation with Insights Client configured

## API Details

| API  | Detail|
 | --- | --- |
 | **Method** | POST |
 | **Path** | /api/v1/providers/ |
 | **Response** | 201 |
**Body:**
```
{
"name": "RH-OCP-Test",
"type": "OCP",
            "authentication": {
                "provider_resource_name": "my-ocp-cluster-1"
            }
}
```

## UI Details

A button and wizard or modal to add an OCP provide

## Acceptance Criteria

- [ ] Verify that a user can add an OCP provider with a “OCP cluster id”

dccurtis:
Adding basic stubs for `OCP` and `OCP-local` providers.  

Testing:
1.  Attempt to add an AWS and AWS-local provider with the bucket parameter missing
2. Attempt to add an OSD and OSD-local provider with the bucket parameter provided.
3.  Add an AWS and AWS-local provider with the bucket parameter supplied
4.  Add OCP and OCP-local providers without bucket parameter

Results:
[masu_ocp_provider-ut.txt](https://github.com/project-koku/koku/files/2403555/masu_ocp_provider-ut.txt)

Closes #360


chambridge:
I'm not sure OCP would have a `billing_source` what is your expectation for what will be in that value? For us this will be internal as we always expect to get this from the Upload Service, but I don't think the customer would specify this.

dccurtis:
Yes thats true.  It shouldn't be specified by the API.  Since there are still details to be worked out around how we will be retrieving a specific report given a cluster-id from the upload service, I just left it for now so that it more easily fits into the existing provider infrastructure.

But what we do know is that the API shouldn't be required to specify this.  For the short term I can make the billing source optional in the serializer but we would still be creating the database entries for billing source.  Bucket would just be blank.  Does it make sense to do this or should we just not create the billing source database entries entirely and put it back if it turns out we need it as an internal detail? 

example:
```
POST /api/v1/providers/
HTTP 201 Created
Allow: GET, POST, HEAD, OPTIONS
Vary: Accept

{
    "uuid": "8e0da500-68e1-484a-b446-90488104701e",
    "name": "TestOCP",
    "type": "OCP",
    "authentication": {
        "uuid": "6f962de0-4541-435d-bff4-6835eb341b40",
        "provider_resource_name": "my-ocp-cluster-1"
    },
    "billing_source": {
        "uuid": "8df19d0d-b3d4-4dff-a126-e954ef488ab2",
        "bucket": ""
    },
    "customer": {
        "uuid": "7ea42f8a-8a66-40fb-9416-8057cffe98d9",
        "name": "TestCustomer",
        "owner": {
            "uuid": "f8bfd017-4249-4fcf-ac4f-10c37fbf686e",
            "username": "owner",
            "email": "owner@testcustomer.com"
        },
        "date_created": "2018-09-19T15:52:10.357230Z"
    },
    "created_by": {
        "uuid": "f8bfd017-4249-4fcf-ac4f-10c37fbf686e",
        "username": "owner",
        "email": "owner@testcustomer.com"
    }
}
```

chambridge:
So I would think that you should be able to make it optional, but in the serializer if its AWS throw an error if its not provided.

I think you could make the foreign key null for OCP, OCP-local provider.

dccurtis:
Yeah I agree, there needs to be some more work to ensure it's supplied for AWS

dccurtis:
I'll get the functional tests updated this evening, but the unit tests should cover most of it.  Things seemed to work as expected in my quick real world testing.

koku/api/provider/serializers.py
```diff
@@ -64,7 +64,7 @@ class ProviderBillingSourceSerializer(serializers.ModelSerializer):
 
     uuid = serializers.UUIDField(read_only=True)
     bucket = serializers.CharField(max_length=63, required=True,
-                                   allow_null=False, allow_blank=False)
+                                   allow_null=False, allow_blank=True)
```
chambridge: Do we need this change? Shouldn't we be covered by just having `billing_source` be null?

dccurtis: Needed for the Django API view since billing_source is being sent as a parameter

PR #372, opened 2018-09-19
Don't group by units, instead use an annotation

adberglund:
Fix a data loss issue in the API when multiple group bys are in play.


koku/api/report/queries.py
```diff
@@ -748,14 +749,14 @@ def execute_sum_query(self):
...
 
             query_order_by = ('-date', )
             if self.order_field != 'delta':
                 query_order_by += (self.order,)
 
-            query_data = query_data.values(*query_group_by_with_units)\
-                .annotate(total=Sum(self.aggregate_key))
+            query_data = query_data.values(*query_group_by)\
+                .annotate(total=Sum(self.aggregate_key))\
+                .annotate(units=Max(self.units_key))
```
blentz: How does `Max()` work with respect to strings?

In terms of having values like "bytes", "GiB", and "MiB", will it return the right thing?

adberglund: I believe in our current scope, we shouldn't hit the case where we're collecting different units in the same query. Prior to this change, returning the same units is already part of the contract, as we calculate totals and total deltas across the board under the assumption that all units are the same. True that with the group by we would then return each different unit type, but then we would be comparing apples and oranges. 

To answer the question directly. `Max()` is kind of funny with strings. In your case 'bytes' would return because for some reason lower case strings seem to be preferred in Max. The common problem in our data is either a unit (e.g. GiB) or null. In that case it will always select the non-null value over null, which is good for us.

PR #392, opened 2018-09-28
Implements order-by account-alias

blentz:
This implements ordering by account alias. The order-by parameter keyword is `account_alias`. The query parameter will look like this: `order_by[account_alias]=asc`


dccurtis:
Can you post some functional tests?  I think your missing account_alias in the OrderBySerializer class

chambridge:
Also did you check if order by service and region work? May just be updates to the serializer also. 
I would think some Swagger updates too.

blentz:
@dccurtis - I caught that myself when I was trying to generate some functional test output. :smile: 

@chambridge - I was going to check for those after alias is complete. I'll push some swagger updates, though.

blentz:
While it's not a great functional test, this shows that the basic query parameter is accepted and returns data:

```
curl -g -H "HTTP_X_RH_IDENTITY: ${IDENTITY}" \
    'http://'${HOST}'/api/v1/reports/inventory/instance-type/?filter[time_scope_value]=-1&filter[resolution]=monthly&filter[time_scope_units]=month&group_by[account]=*&order_by[account_alias]=asc' \
    | jq '.'
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   644  100   644    0     0   1190      0 --:--:-- --:--:-- --:--:--  1188
{
  "group_by": {
    "account": [
      "*"
    ]
  },
  "order_by": {
    "account_alias": "asc"
  },
  "filter": {
    "resolution": "monthly",
    "time_scope_value": "-1",
    "time_scope_units": "month"
  },
  "data": [
    {
      "date": "2018-09",
      "accounts": [
        {
          "account": "589173575009",
          "instance_types": [
            {
              "instance_type": "t2.micro",
              "values": [
                {
                  "instance_type": "t2.micro",
                  "date": "2018-09",
                  "units": "Hrs",
                  "account": "589173575009",
                  "total": 1306,
                  "account_alias": "redhat-hccm",
                  "count": 0
                }
              ]
            },
            {
              "instance_type": "t2.small",
              "values": [
                {
                  "instance_type": "t2.small",
                  "date": "2018-09",
                  "units": "Hrs",
                  "account": "589173575009",
                  "total": 654,
                  "account_alias": "redhat-hccm",
                  "count": 0
                }
              ]
            }
          ]
        }
      ]
    }
  ],
  "total": {
    "value": 1960,
    "count": 0,
    "units": null
  }
}

```

chambridge:
I'm not saying this is wrong, I just thought that `account_alias` was going to be populated with the `account_id` value if an alias was not set. Am I incorrect here @dccurtis @blentz?

dccurtis:
Yeah that was the intention. Need to look closer to see what’s going on with that.  Maybe the other account doesn’t have alias permissions setup and the init to account id isn’t happening. Admittedly I didn’t test that case with the quick account Id change. 

I can look at that outside of this change.

I haven’t done a lot of queries on the API but the count numbers look off to me.  The one with the alias is 0 and without is 17 sum at the bottom is 1?

dccurtis:
It looks like the exception path in masu (no alias access set on AWS) would not set the account_id as the default value.  I can change that if it's useful for the UI, just let me know.

chambridge:
@dccurtis I think it would be good to have the default in the exception path, but as you said it’s a different issue. I’m sure Victor setup the permissions without the alias permission.

docs/source/specs/openapi.yml
```diff
@@ -660,18 +660,24 @@ definitions:
...
+        type: string
+        enum:
+          - asc
+          - desc
     example: { 'total': 'asc' }
     description: The ordering to apply to the report. Default is ascending order for the data.
   ReportDelta:
     type: object
     properties:
       cost:
```
chambridge: shouldn't this be `delta`?

koku/api/report/queries.py
```diff
@@ -36,12 +36,12 @@
...
 
 from api.utils import DateHelper
-from reporting.models import (AWSAccountAlias,
-                              AWSCostEntryLineItem,
+from reporting.models import (AWSCostEntryLineItem,
                               AWSCostEntryLineItemAggregates,
                               AWSCostEntryLineItemDailySummary)
 
 
+logging.disable(0)
```
chambridge: Not clear why this was added. Maybe just for development?

docs/source/specs/openapi.yml
blentz: I'm not sure I understand what you're asking.

chambridge: https://github.com/project-koku/koku/blob/master/docs/source/specs/openapi.yml#L668

Report Delta, shouldn't the value be `delta` not `cost`

blentz: Good catch. You're right.

PR #393, opened 2018-09-28
First pass at OCP column mappings

dccurtis:
This will be a work in progress as the OCP report definition evolves but I wanted to get a first pass submitted to simplify the initial masu db access work.  Let me know if I missed any thing.


blentz:
Am I correct in thinking that we're mapping the OCP column names to the names being used in the AWS CUR?

PR #414, opened 2018-10-08
Add account alias to the aggregates table

adberglund:
Getting this in for a bugfix @chambridge is working on.


chambridge:
thank you @adberglund

PR #415, opened 2018-10-08
Intended to fix the break introduced by case insensitive and resolve the Patternfly filtering needs.

chambridge:
Items within a "category" are ORed together. Items in different "categories" are ANDed together.

Make appropriate updates for total & deltas to support account_alias updates as well.

FYI @dlabrecq 

**GET** `/api/v1/reports/costs/?filter[time_scope_value]=-1&filter[time_scope_units]=month&filter[resolution]=monthly&group_by[service]=eC2&group_by[service]=RDS&delta=true`

```
{
    "delta": {
        "value": 21.022314016,
        "percent": 0.0
    },
    "group_by": {
        "service": [
            "eC2",
            "RDS"
        ]
    },
    "filter": {
        "resolution": "monthly",
        "time_scope_value": "-1",
        "time_scope_units": "month"
    },
    "data": [
        {
            "date": "2018-10",
            "services": [
                {
                    "service": "AmazonEC2",
                    "values": [
                        {
                            "date": "2018-10",
                            "units": "USD",
                            "service": "AmazonEC2",
                            "total": 20.974,
                            "delta_value": 20.974,
                            "delta_percent": 0.0
                        }
                    ]
                },
                {
                    "service": "AmazonRDS",
                    "values": [
                        {
                            "date": "2018-10",
                            "units": "USD",
                            "service": "AmazonRDS",
                            "total": 0.048314016,
                            "delta_value": 0.048314016,
                            "delta_percent": 0.0
                        }
                    ]
                }
            ]
        }
    ],
    "total": {
        "value": 21.022314016,
        "units": "USD"
    }
}
```

### DIFF
--- a/koku/api/report/queries.py
+++ b/koku/api/report/queries.py
@@ -18,7 +18,7 @@
 import copy
 import datetime
 import logging
-from collections import OrderedDict, UserDict
+from collections import OrderedDict, UserDict, defaultdict
 from decimal import Decimal, DivisionByZero, InvalidOperation
 from itertools import groupby
 
@@ -183,15 +183,12 @@ class QueryFilterCollection(object):
     def compose(self):
         """Compose filters into a dict for submitting to Django's ORM."""
         composed_query = None
-        compose_dict = {}
+        compose_dict = defaultdict(list)
         for filt in self._filters:
             filt_key = filt.composed_query_string()
-            if compose_dict.get(filt_key) is None:
-                compose_dict[filt_key] = [filt]
-            else:
-                compose_dict[filt_key].append(filt)
+            compose_dict[filt_key].append(filt)
 
-        for _, filter_list in compose_dict.items():
+        for filter_list in compose_dict.values():
             or_filter = None
             for filter_item in filter_list:
                 if or_filter is None:
--- a/koku/api/report/queries.py
+++ b/koku/api/report/queries.py
@@ -18,7 +18,6 @@
 import copy
 import datetime
 import logging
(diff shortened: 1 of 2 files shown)


koku/api/report/queries.py
```diff
@@ -181,11 +182,28 @@ def add(self, query_filter=None, table=None, field=None, operation=None, paramet
...
         """Compose filters into a dict for submitting to Django's ORM."""
-        out = {}
+        composed_query = None
+        compose_dict = {}
         for filt in self._filters:
-            if filt.parameter is not None:
-                out.update(filt.composed_dict())
-        return out
+            filt_key = filt.composed_query_string()
+            if compose_dict.get(filt_key) is None:
```
adberglund: Definitely no reason to change this now, but I figured I'd insert some annoying Python standard library advice and say the collections package has a defaultdict class that lets you set an automatic default value for the value of a dict key. Set `compose_dict = defauldict(list)` and then you can just append away without checking if the key exists yet.

koku/api/report/queries.py
```diff
@@ -181,11 +182,28 @@ def add(self, query_filter=None, table=None, field=None, operation=None, paramet
...
+                compose_dict[filt_key] = [filt]
+            else:
+                compose_dict[filt_key].append(filt)
+
+        for _, filter_list in compose_dict.items():
```
blentz: If you're throwing away the key anyway, then just use `.values()` instead of `.items()`.

PR #436, opened 2018-10-24
Breaking up AWS API logic to prepare for code reuse with OCP

dccurtis:
This is a first pass at breaking out common logic that will be used for the OCP end points.  The main changes include:

1. Created a `aws` module to contain the AWS specific bits for queries, views, and serializers.
2. The `api.report.view/serializer/` module will contain common logic that can be used with both AWS and OCP.
3. The `aws.report.query_filter` contains the class definitions for `QueryFilter` and `QueryFilterCollection` since it can also be used between AWS and OCP.
4. `_generic_report` now contains parameters for provider-specific serializer and report query handlers.

Next Steps:
There will be further breakout of AWSReportQueryHandler to pull out common logic once OCP is in place.

*Testing*
1. Toggled between master and this branch and verified that APIs used by the UI are unchanged.


blentz:
"The existing query handler is tightly coupled to AWS behavior and the associated responses."

No. It isn't. The `ProviderMap` removes that tight coupling and most of the if/else logic you're talking about. There may be more work needed to line up the schemas so that we're querying the right fields across data sets. But, as of now, there is no provider-specific reference in the `ReportQueryHandler` class.

dccurtis:
Here's an example of the OCP API thus far:
```
GET /api/v1/reports/ocp/cpu/?filter[time_scope_units]=month&filter[time_scope_value]=-1&filter[resolution]=monthly&filter[limit]=5&group_by[project]=*
HTTP 200 OK
Allow: GET, OPTIONS
Vary: Accept

{
    "group_by": {
        "project": [
            "*"
        ]
    },
    "filter": {
        "resolution": "monthly",
        "time_scope_value": "-1",
        "time_scope_units": "month",
        "limit": 5
    },
    "data": [
        {
            "date": "2018-10",
            "projects": [
                {
                    "project": "openshift-web-console",
                    "values": [
                        {
                            "date": "2018-10",
                            "project": "openshift-web-console",
                            "cpu_limit": null,
                            "cpu_usage_core_hours": 0.862687,
                            "cpu_requests_core_hours": 4.753333
                        }
                    ]
                },
                {
                    "project": "monitoring",
                    "values": [
                        {
                            "date": "2018-10",
                            "project": "monitoring",
                            "cpu_limit": null,
                            "cpu_usage_core_hours": 7.861343,
                            "cpu_requests_core_hours": 17.920067
                        }
                    ]
                },
                {
                    "project": "metering",
                    "values": [
                        {
                            "date": "2018-10",
                            "project": "metering",
                            "cpu_limit": null,
                            "cpu_usage_core_hours": 4.464511,
                            "cpu_requests_core_hours": 53.985832
                        }
                    ]
                },
                {
                    "project": "default",
                    "values": [
                        {
                            "date": "2018-10",
                            "project": "default",
                            "cpu_limit": null,
                            "cpu_usage_core_hours": 0.119385,
                            "cpu_requests_core_hours": 9.506666
                        }
                    ]
                }
            ]
        }
    ],
    "total": {
        "cpu_usage_core_hours": 13.307928,
        "cpu_requests_core_hours": 86.165898
    }
}
```
One big difference between the AWS and OCP responses is that OCP has essentially 2 'totals'.  They are `usage` and `requested` values (both memory and cpu).  This goes against the gain of there being a single 'total' value that is summed for AWS.  On the `ProviderMap` front I wasn't able to fit the OCP into the existing map keys which all providers would have to share in order for it to be generically used in the guts of the handler.  There's other things as well, such as default_ordering, _get_group_by() options are hard-coded for AWS, etc.

I gave it an honest try to make it work but kept coming back to the conclusion that trying to fit all of this into a single monolithic handler just wasn't the best way forward.  Taking a step back, I envisioned all of this working in a single handler and could not see the value in it over having parent and sibling handlers.  If it was a monolith then we would be putting all other providers at risk whenever we add a new one and would also be unnecessarily constraining ourselves in what parameters and response we want for a particular API endpoint.

I would really like to understand what the monolithic (single class) handler will get us before spending any more time on that path trying to get it to work.  I'm completely open to the fact that I could be using the `ProviderMap` wrong or theres something I'm just generally missing.  But I think answering the question on the value of a single handler (keeping in mind that sibling handlers would be sharing probably 80-90% of the code from the parent) would be a good place to start.

blentz:
Well, considering that I continually get hammered about "make the smallest change necessary." So, now it's my turn to beat that same drum. Ripping all of this out and starting over isn't the "smallest change necessary."

I think we can look at each case you mention and re-work the individual functions to accommodate both over a series of PRs that focus on each piece.

I mean, really, it's not like we can't change the `default_ordering` variable to something else, or change how ordering works in light of both use cases. None of that requires throwing everything out.

chambridge:
Can we open an issue to add the updates to swagger?

dccurtis:
Will be sure to update swagger once the OCP changes land.

PR #438, opened 2018-10-24
Fix RTD build failure.

chambridge:
https://readthedocs.org/projects/koku/builds/7994907/


chambridge:
Fix here:
https://koku.readthedocs.io/en/fix-rtd-build/

PR #440, opened 2018-10-28
Preliminary OCP API work

dccurtis:
**Overview**
First pass at OCP API work.  Focus was on providing basic group_by functionality.  End points added are:

`/api/v1/inventory/ocp/cpu` and `/api/v1/inventory/ocp/memory`.  Group_by values are : `cluster`, `project`, and `node`.

**Testing**
The majority of this effort was spent on regression testing AWS since there was some refactoring work that had to happen.  The OCP testing included demonstrating the basic functionality of group_by and the other standard query parameters.

AWS Regression Testing:
1.  Issue all API calls that the UI performs to populate the dashboard with both master and the ocp_api branch.  Capture API response to a respective file and do a diff between files to ensure they are identical.

OCP Testing:
1. Verify that group_by project, node, and cluster works
2. Verify proper error message when group_by is invalid
3. Verify that resolution=daily works
4. Verify that group_by = specific value works
5. Verify that limit works
6. Verify CSV output

*Note* Additional unit testing is coming for the OCP query handler, but I wanted to get the review out since the AWS regression testing and basic functionality for OCP is complete.

**Test Results**
AWS Regression
[master.txt](https://github.com/project-koku/koku/files/2522848/master.txt)
[new_branch.txt](https://github.com/project-koku/koku/files/2522851/new_branch.txt)

Diff test
```
curtisd@docurtis-mac ~/D/aws_regression> diff master.txt new_branch.txt
curtisd@docurtis-mac ~/D/aws_regression>
```

OCP API Results:
[ocp_api_ut.txt](https://github.com/project-koku/koku/files/2522852/ocp_api_ut.txt)


dccurtis:
yes, theres still some duplication.  This was really just the minimum to get things in place.  I fully expect us to iterate on this as we go.  Unit tests are still underway and the query handler will take some time.  I'll continue to refine this.

docs/source/specs/openapi.yml
```diff
@@ -442,7 +442,7 @@ paths:
             $ref: '#/definitions/Error'
       security:
       - api_key: []
-  /reports/ocp/cpu/:
+  /reports/ocp/inventory/cpu/:
```
blentz: The Swagger paths don't match what's in `koku/api/urls.py`

Also, I'm not convinced we need the provider as part of the path. I don't know if we have plans to add collecting CPU & Memory stats out of AWS, but I think it's acceptable that the only handler for `/reports/inventory/cpu` right now is the OCP handler. It's not necessary to embed the provider into the URI.

dccurtis: Good catch on the openapi.yml.   I think the rest of the discussion regarding paths can continue when/if we do have provider overlap.  I vote for just keeping the provider in the path for now so it's clear when integrating with the UI what provider the end point is intended for.

PR #444, opened 2018-11-02
add Rate model

blentz:
This creates a `Rate` model as the first step for implementing a Price List.

Because we're coordinating changes between Masu and Koku, only the Rate model is being created here. The Serializer and Views will be handled in a future PR.


chambridge:
Maybe split this into two PRs?
- Model PR
- OCP/AWS reorganization

What you have for the model seems fine for now. Would it make any sense to store this as a TextField or JSON so we don't have update schemas a bunch while we work through it? Just a thought, some downside to it as the serializer can have greater controls, but I figure we're going to have to write quite a bit of validation anyways for these field.

koku/reporting/rate/models.py
```diff
@@ -0,0 +1,37 @@
...
+from django.db import models
+
+
+class Rate(models.Model):
+    """A rate for calculating costs.
+
+    A rate is (price * metric_usage / timeunit).
+    """
+
+    uuid = models.UUIDField(default=uuid4, editable=False,
```
adberglund: What is this uuid for? 

As it stands it's not the primary key, so there will also be an auto-incrementing id column here. Personally, I think we only need that auto-id field.

koku/reporting/rate/models.py
```diff
@@ -0,0 +1,37 @@
...
+#
+
+"""Models for identity and access management."""
+
+from uuid import uuid4
+
+from django.db import models
+
+
+class Rate(models.Model):
```
adberglund: This model could benefit from a unique constraint so that a rate can't be repeated in the db.

koku/reporting/rate/models.py
```diff
@@ -0,0 +1,37 @@
...
+    """A rate for calculating costs.
+
+    A rate is (price * metric_usage / timeunit).
```
adberglund: A rate should actually be just `price / unit`.

Some of the usage (like CPU) is currently reported in `cpu-core-hours`. So units are a little more complicated than just time. The rate is really price per thing we charge for.  So this rate would be `$ / cpu-core-hour` if using hours and USD. 

Some (like memory) is currently reported in GB. So here the unit would be `GB` and time would not factor in to the rate. This does raise the question of whether we want to or should be reporting memory usage in GB-hours as well.

koku/reporting/rate/models.py
chambridge: We leverage uuid for most models so that we aren't tied to an internal db id. I'm guessing this is done for consistency.

https://github.com/project-koku/koku/blob/master/koku/api/provider/models.py#L31

adberglund: Yeah, I get that. I think it makes less sense inside the tenant schema. No other models there use a uuid right now. Still, not holding back on this comment, just was looking for an answer of what it was really getting us.

blentz: I don't have a good reason for it. Mostly just following the pattern we're using in other models.

koku/reporting/rate/models.py
blentz: A rate always includes a time component. Some metrics have the time component built-in, like `cpu-core-hours`, which means we can set the `timeunit` field to `None` there. However, everything else is either a cost-over-time or a one-time charge.

As you say, GB-hours, not GB.

The model simply gives you the ability to specify whether it should be GB-hours or GB-days without needing to change the data collector. So, if all we're receiving is a raw count of units consumed, we can divide that count over the specified time interval.

koku/reporting/rate/models.py
blentz: That's tricky, because multiple rates can be applied to the same metric. 

So, I could do a unique_together on (metric, timeunit, price), but it doesn't really buy us much.

PR #454, opened 2018-11-09
Implement Rate CRUD API

blentz:
This implements a basic CRUD API for managing Rate objects. 

[functional_output.txt](https://github.com/project-koku/koku/files/2566812/functional_output.txt)


dccurtis:
Masu is currently expecting metric to be either `cpu` or `memory`.  Those choices were pretty arbitrary and I don't mind changing them.  But I do think it's important that there are checks on the rate API side to ensure masu and koku are in sync.  I'm fine with kicking this to another PR but just want to call it out since the user would have to know this to use the API successfully with masu

koku/api/report/rate/serializers.py
```diff
@@ -0,0 +1,31 @@
+#
```
chambridge: Why put these files under `koku/api/report`?
Would prefer `koku/api/rate` for this as its not really the report api.

koku/api/report/rate/serializers.py
```diff
@@ -0,0 +1,31 @@
...
+"""Rate serializer."""
+from rest_framework import serializers
+
+from reporting.rate.models import Rate
+
+
+class RateSerializer(serializers.ModelSerializer):
+    """Rate Serializer."""
+
```
chambridge: Can we open a follow up issue to add some input validation here?
Positive number rate,  maybe enums for metric and time unit?

PR #460, opened 2018-11-12
Adding OCP total charge to memory and CPU end points

dccurtis:
Adding `charge` to the `total` dictionary of the OCP API response.

Example:
```
{
    "group_by": {
        "node": [
            "*"
        ]
    },
    "filter": {
        "resolution": "monthly",
        "time_scope_value": "-1",
        "time_scope_units": "month"
    },
    "data": [
...
    ],
    "total": {
        "usage": 1079.191645,
        "request": 1451.458175,
        "charge": 217.718875
    }
}
```

**Testing**
1.  Issue a few OCP API report queries and verify that the total charge is updated correctly.

**Test Results**
[koku_charge_total_ut.txt](https://github.com/project-koku/koku/files/2573554/koku_charge_total_ut.txt)


dccurtis:
Yeah I was thinking the same thing.  I like both options really, but maybe for the near term its best to just keep it.  Once OCP is "done" we should remove it if its not used.

PR #476, opened 2018-11-19
Ocp deltas

adberglund:
## Summary
Deltas can now be calculated for OCP APIs. A previous fix for limits applied to AWS has also been applied to OCP. Deltas will likely change in the next sprint as we add in new delta comparisons, but hopefully that fits into this framework.
* Delta is now a string param that takes certain values by provider and API type
  * AWS/all APIs: `total`
  * OCP/memory and cpu APIs: `charge`, `usage`, `request`
  * OCP/charge API: `charge`

## Testing
* Tested against master for AWS delta values, since this change could impact AWS. 
* Added unit test that calculate deltas and make sure that the result coming back from the API are correct.


chambridge:
Can we update the example in swagger?
https://github.com/project-koku/koku/blob/master/docs/source/specs/openapi.yml#L1152
Looks like it was a string before (weird), but less to change.

docs/source/specs/openapi.yml
```diff
@@ -1148,9 +1148,9 @@ definitions:
     type: object
     properties:
       delta:
-        type: boolean
-    example: { 'delta': 'True' }
-    description: 'Toggle to include delta values in report.'
+        type: string
+    example: { 'delta': 'total' }
+    description: 'Field to calculate delta delta values for in report.'
```
adberglund:
```suggestion
    description: 'Field to calculate delta values for in report.'
```

PR #482, opened 2018-11-22
Updated OpenAPI for current server implementation

psav:
**NOTE:** This is perhaps more a change that needs to happen on the server side? ie not returning null/None for the next/previous. Regardless this PR is here to highlight an issue and hopefully facilitate a fix.

Generating the swagger client against the current specification doesn't
actually work. The previous and next fields are refined as required,
yet if there is only one record, deserialization fails as previous
and next are not allowed to be None.

The results attribute was also not present in the specification and so
the deserialization omitted it when using the python client.

* Removed the required marker for previous and next in the ListPagination
  object definition.
* Added results with type object to the ListPagination object.


chambridge:
@psav Just an FYI - we are currently just using the Django Rest Framework implementation for pagination: https://www.django-rest-framework.org/api-guide/pagination/#pagenumberpagination

Returning `null` is just part of its implementation.

PR #525, opened 2018-12-11
Add pod label key and value columns to the report daily summary table

adberglund:
## Summary
Add pod label key and value columns to the OCP daily summary table for use in reporting APIs


koku/reporting/provider/ocp/models.py
```diff
@@ -323,6 +331,9 @@ class Meta:
     usage_start = models.DateTimeField(null=False)
     usage_end = models.DateTimeField(null=False)
 
+    pod_label_key = models.CharField(max_length=253, null=True)
+    pod_label_value = models.CharField(max_length=253, null=True)
```
chambridge: What led you to this max length?

adberglund: Most kubernetes objects have that max length built in. I am just kind of assuming labels are the same, but we could lift this.

chambridge: Was just hoping it was doc related and wondering if you had dug up any links in case we needed it in the future.

PR #526, opened 2018-12-12
Fixed the ordering in the AWS total by annotation.

chambridge:
- Bug was because I was creating a set and this caused the order to vary.
- Added logging into middleware to capture Path+Querystring and Account and Username
- Added fallback units to AWS values

### DIFF
--- a/koku/api/report/aws/aws_query_handler.py
+++ b/koku/api/report/aws/aws_query_handler.py
@@ -64,9 +64,10 @@ class AWSReportQueryHandler(ReportQueryHandler):
             (Dict): query annotations dictionary
 
         """
+        units_fallback = self._mapper._report_type_map.get('units_fallback')
         annotations = {
             'date': self.date_trunc('usage_start'),
-            'units': Concat(self._mapper.units_key, Value(''))
+            'units': Coalesce(Value(units_fallback), Concat(self._mapper.units_key, Value('')))
         }
 
         # { query_param: database_field_name }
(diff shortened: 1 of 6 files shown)


koku/api/report/aws/aws_query_handler.py
```diff
@@ -64,9 +64,10 @@ def annotations(self):
             (Dict): query annotations dictionary
 
         """
+        units_fallback = self._mapper._report_type_map.get('units_fallback')
         annotations = {
             'date': self.date_trunc('usage_start'),
-            'units': Concat(self._mapper.units_key, Value(''))
+            'units': Coalesce(Value(units_fallback), Concat(self._mapper.units_key, Value('')))
```
adberglund: coalesce will return the first non-null arg, so units_fallback will always get returned.

chambridge: yep, I forgot to flip this one. Will fix. Nice catch.

koku/api/report/aws/aws_query_handler.py
```diff
@@ -67,7 +67,7 @@ def annotations(self):
         units_fallback = self._mapper._report_type_map.get('units_fallback')
         annotations = {
             'date': self.date_trunc('usage_start'),
-            'units': Coalesce(Value(units_fallback), Concat(self._mapper.units_key, Value('')))
+            'units': Coalesce(Concat(self._mapper.units_key, Value('')), Value(units_fallback))
```
adberglund: New problem here is that this will always return `''` as that's the fallback for self._mapper.units_key in the concat. Sorry. I think we need to get self._mapper.units_key to return a null when it's empty so we can fallback to the units_fallback properly.

chambridge: 🤦‍♂️  
I think the change below fixes it.

PR #566, opened 2019-01-10
add additional env vars to openshift template

blentz:
This adds additional environment variables to the openshift template that Koku supports. This will make it easier to access these variables.

### DIFF
--- a/openshift/koku-template.yaml
+++ b/openshift/koku-template.yaml
@@ -106,7 +106,6 @@ objects:
     django-log-handlers: "console"
     django-log-directory: ""
     django-logging-file: ""
-    api-path-prefix: ''
 - apiVersion: v1
   kind: Service
   metadata:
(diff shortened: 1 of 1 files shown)


openshift/koku-template.yaml
```diff
@@ -99,6 +99,14 @@ objects:
...
     django-debug: ${DJANGO_DEBUG}
     api-path-prefix: ${API_PATH_PREFIX}
+    development: "False"
+    koku-log-level: "INFO"
+    django-log-level: "INFO"
+    django-log-formatter: "simple"
+    django-log-handlers: "console"
+    django-log-directory: ""
+    django-logging-file: ""
+    api-path-prefix: ''
```
chambridge: You are duplicating the entry for `api-path-prefix` on line 101.

openshift/koku-template.yaml
```diff
@@ -323,6 +331,54 @@ objects:
...
+                  name: koku-env
+                  key: django-log-directory
+                  optional: true
+            - name: DJANGO_LOGGING_FILE
+              valueFrom:
+                configMapKeyRef:
+                  name: koku-env
+                  key: django-logging-file
+                  optional: true
+            - name: API_PATH_PREFIX
```
chambridge: You are duplicating the entry for `API_PATH_PREFIX` on line 304.

PR #591, opened 2019-01-25
Tag summary table

adberglund:
## Summary
This uses a new OCP tag summary table to get a quick list of available tag keys. Using the summary table and reducing down the set of tag keys passed to the serializer greatly reduces the time taken for a OCP report API call. 

## Notes
This is currently only implemented for OCP, but AWS will benefit from a similar approach. Also I think we can iterate on this slightly. I wanted to get something out to help QE, but we could potentially use the new summary table in parts (but not all of) the tags query handler.


koku/api/report/view.py
```diff
@@ -54,6 +55,21 @@ def process_query_parameters(url_data, provider_serializer, tag_keys=None):
...
+    """Reduce the set of tag keys based on those being queried."""
+    tag_key_set = set(tag_keys)
+    param_tag_keys = set()
+    for key, value in query_params.items():
+        if isinstance(value, dict) or isinstance(value, list):
+            for inner_param_key in value:
+                if inner_param_key in tag_key_set:
+                    param_tag_keys.add(inner_param_key)
+        if key in tag_key_set:
+            param_tag_keys.add(key)
```
adberglund:
```suggestion
            param_tag_keys.add(key)
        if value in tag_key_set:
            param_tag_keys.add(value)
```

PR #600, opened 2019-01-29
AWS report filter/query_by tags

dccurtis:
**Overview**
This feature works the same was as OCP where reports can filtered and group by with `tag:<tag_key>` 

For example:
```
http://localhost:8000/api/v1/reports/inventory/aws/instance-type/?filter[tag:environment]=dev
http://localhost:8000/api/v1/reports/inventory/aws/instance-type/?group_by[tag:environment]=*
```

**Testing:**
1.  Add static nise report data to include EC2 instances with labels:
2.  Add AWS-local provider to ingest the above nise data
3.  Ingest with masu
4.  Verify that AWS tags are returned from tag API:
5.  Issue /instance-type endpoint with no query parameters.  Verify that the response is correct:
6.  Filter the /instance-type with a filter on tag:environment for all values, correct value, and incorrect value
7.  Filter the /instance-type with a filter on tag:version for all values, correct value, and incorrect value
8.  group_by tag:version on the /instance-type endpoint
9.  group_by tag:environment on the /instance-type endpoint

**Testing Results**
[koku_aws_tag_api_ut.txt](https://github.com/project-koku/koku/files/2809150/koku_aws_tag_api_ut.txt)


koku/reporting/provider/aws/models.py
```diff
@@ -326,6 +326,7 @@ class Meta:
     unblended_cost = models.DecimalField(max_digits=17, decimal_places=9,
                                          null=True)
     resource_count = models.IntegerField(null=True)
+    tags = JSONField(null=True)
```
adberglund: I think now is also a good time to put an index on tags. If you look at `OCPUsageLineItemDailySummary` in the OCP models you can see that I used a GinIndex specific to Postgres JSON.

koku/api/report/test/tests_queries.py
```diff
@@ -1616,28 +1620,6 @@ def test_execute_query_orderby_alias(self):
 
         self.assertEqual(actual, expected)
 
-    def test_calculate_total(self):
```
adberglund: Why does this test go away?

dccurtis: whoops

PR #613, opened 2019-02-07
adding ocp storage column map entries

dccurtis:


dccurtis:
After further investigation to the impact of this change on the masu side it looks like some more model updates are needed to make this work.

PR #616, opened 2019-02-08
fix MVC breakage, prep for ocp costs

blentz:
This change cleans up the model objects that were leaking into the views, muddying up the separation of concerns between View and Controller objects.

I've also refactored the `_generic_report()` signature to help prevent similar problems in the future.

Second - I've tidied up a few AWS-specific references in the URLs to help ensure consistent naming when we add OCP cost reports.


koku/api/report/test/ocp/tests_views.py
```diff
@@ -220,9 +218,10 @@ def test_generic_report_ocp_cpu_success(self, mock_handler):
         request = Request(django_request)
         request.user = user
 
-        extras = {'report_type': 'cpu'}
-        response = _generic_report(request, QueryParamSerializer, OCPReportQueryHandler, **extras)
+        response = _generic_report(request, report='cpu', provider='ocp')
         self.assertIsInstance(response, Response)
+        # FIXME
+        # self.assertEqual(response.status_code, 200)
```
adberglund: Concerning that it's not returning a 200. What's up here?

koku/api/report/view.py
```diff
@@ -27,11 +27,85 @@
...
+        """Return the specified report dict.
+
+        Return default report dict if named report not found.
+
+        """
+        reports = self.reports(provider)
+        for rep in reports:
+            if report == rep.get('report'):
+                return rep
+        return self.get_report(provider, 'default')
```
adberglund: The default here could get get us into trouble if we forget to implement a new entry in the map or misspell something. It will happily supply things that we might not intend. Since we have to explicitly define each API endpoint, I don't think it's a bad idea to be explicit with the report types here.

koku/api/report/test/ocp/tests_views.py
blentz: Not my test. I don't know.

The test itself broke during my refactor, but was fixed by something else. So, I just left a note while I was in there that this test was missing a check of the return code. 

Currently, the test only checks that it gets a response, not whether the response is good.

koku/api/report/view.py
blentz: The problem I was solving was just that most reports end up using the same serializer and query handler. So, I opted for a defaulting mechanism instead of having a hugely redundant data structure.

I can reformulate it, but I agree with you that trade-off exists and I'm okay if you'd prefer to pick having an explicit, but highly redundant data structure instead. :slightly_smiling_face:

adberglund: Nah, I suppose it will fail in the query handler if the report type is not implemented there.

PR #631, opened 2019-02-14
A first pass at a cost API for OCP on AWS

blentz:
This implements costs API for OCP-on-AWS reports.


koku/api/report/queries.py
```diff
@@ -55,7 +55,7 @@ class ProviderMap(object):
 
     # main mapping data structure
     # this data should be considered static and read-only.
-    mapping = [
+    MAPPING = [
```
blentz: This is fixing a convention error. When I created the map, this should have been all-caps to denote a constant that shouldn't be modified.

koku/api/report/aws/aws_query_handler.py
```diff
@@ -92,6 +89,12 @@ def _format_query_response(self):
 
         return output
 
+    def _set_tag_filters(self, filters):
```
blentz: This is streamlining and simplifying the call-chain. 

We were passing `no_tag_query` up from this child class through the parent classes via kwargs purely to set this filter in the parent class to be consumed by this child class. 

Instead, we should do the right thing and override the method to insert the filter right here in this class. There's no need to involve the parent class at all.

koku/api/report/ocp_aws/ocp_aws_query_handler.py
```diff
@@ -38,18 +38,15 @@ def __init__(self, query_parameters, url_data,
             tenant    (String): the tenant to use to access CUR data
             kwargs    (Dict): A dictionary for internal query alteration based on path
         """
-        self._provider = 'OCP_AWS'
-        kwargs['provider'] = self._provider
-        super().__init__(query_parameters, url_data,
-                         tenant, **kwargs)
-        self.update_report_map()
-
-    def update_report_map(self):
```
blentz: This is a couple of simplifications. Realistically, we shouldn't be doing string-checking like this. We should be using `isinstance` instead. I'll make another pass in the future to clean this up further.

koku/api/report/queries.py
```diff
@@ -475,7 +490,8 @@ def __init__(self, provider, report_type):
         self._provider = provider
         self._report_type = report_type
 
-        self._map = ProviderMap.mapping
+        # FIXME: incorrect use of underscore convention.
+        # FIXME: these should be public.
```
blentz: I'll fix this in a future PR. I'm leaving the comment in case someone else wants to do it before I get a chance.

koku/api/report/queries.py
```diff
@@ -716,7 +737,7 @@ def annotations(self):
             (Dict): query annotations dictionary
 
         """
-        pass
+        raise NotImplementedError('Annotations must be defined by sub-classes.')
```
blentz: This improves behavior and testability. It also helps document the intended use.

chambridge:
I see most of the components I would expect here except I'd expect a serializer to handle the combination of group_bys and filters. Did I miss that?

Also I think there is an update to the openapi.yaml here too (wouldn't think it would be much more than a duplicate of the aws route with a name tweak 😉 ).

Do you have any example output for monthly/daily? Tried to pull and use the CI data set but since there are migrations I couldn't.

koku/api/report/queries.py
```diff
@@ -347,8 +347,23 @@ class ProviderMap(object):
             'group_by_options': ['account', 'service', 'region', 'cluster', 'project', 'node'],
             'tag_column': 'tags',
             'report_type': {
+                'costs': {
```
adberglund: Need to handle the group by project case.

koku/api/views.py
```diff
@@ -23,7 +23,10 @@
 from api.tags.ocp.view import ocp_tags
 from api.iam.view.user_preference import UserPreferenceViewSet
 from api.provider.view import ProviderViewSet
-from api.report.aws.view import costs as aws_costs, instance_type, storage
+from api.report.aws.view import (costs as aws_costs,
+                                 instance_type,
+                                 storage as aws_storage)
 from api.report.ocp.view import charges, cpu, memory
-from api.report.ocp_aws.view import (instance_type as ocp_aws_instance_type,
+from api.report.ocp_aws.view import (costs as ocp_costs,
```
adberglund: Maybe as ocp_aws_costs?

koku/reporting/migrations/0031_ocpawscostlineitemdailysummary_currency_code.py
```diff
@@ -1,4 +1,4 @@
-# Generated by Django 2.1.5 on 2019-02-13 19:41
```
adberglund: We shouldn't be altering already committed migrations like this. They are a record of changes made to the database and if they don't reflect what actually happened we can be out of sync, or our dev database build migrations might end up different from production.

koku/api/report/queries.py
```diff
@@ -371,7 +386,7 @@ class ProviderMap(object):
                     'default_ordering': {'total': 'desc'},
                 },
                 'storage_by_project': {
-                    'aggregates': {
+                    'aggregate': {
```
adberglund: Per the plural singluar argument this now leaves us with aggegate, and annotations. One singular, one plural. 

The instance type mapping wasn't updated, so I suspect that will break.

koku/api/report/view.py
```diff
@@ -129,6 +131,11 @@ def get_report(self, provider, report):
         for rep in reports:
             if report == rep.get('report'):
                 return rep
+
+        if report == 'default':
```
dccurtis: Trying to understand this part.  Where is `_generic_report` passing in `report='default'` to hit this case?

koku/api/report/view.py
```diff
@@ -129,6 +131,11 @@ def get_report(self, provider, report):
...
+            # avoid infinite recursion
+            return {}
+
         return self.get_report(provider, 'default')
```
dccurtis: ohh I see it's the recursive call...

koku/api/report/queries.py
blentz: Fair point, but if we need to change it, all instances of it needs to change not just new instances. And, well, your storage work only created inconsistency by using a different string than was being used elsewhere in the map. So, I'm fixing the inconsistency that was introduced.

Let's circle back in a future PR to migrate all cases of `aggregate` to `aggregates`.

koku/reporting/migrations/0031_ocpawscostlineitemdailysummary_currency_code.py
adberglund: Don't worry about the migrations, the end result is fine. so ignore this one.

koku/api/report/view.py
blentz: It's not. 

`_generic_report` passes in some string, say `foo`, and if the string falls through the for-loop, this function will recurse to look for the default. But we only want it to recurse once, not an infinite number of times. :slightly_smiling_face:

blentz:
@chambridge - I'll update the openapi spec. Regarding serializers - @adberglund was kind enough to implement the OCP-AWS Serializers with his OCP Storage work. They already exist, which is why you're not seeing them in this PR. :slightly_smiling_face:

docs/source/specs/openapi.yml
```diff
@@ -502,6 +502,45 @@ paths:
             $ref: '#/definitions/Error'
       security:
       - api_key: []
+  /reports/costs/ocp/:
+    get:
+      tags:
+        - Report
+      summary: Query to obtain cost reports
+      operationId: getCostReports
```
chambridge: image

Should update the operationId of this and the AWS report.

koku/api/report/queries.py
```diff
@@ -362,8 +361,22 @@ class ProviderMap(object):
...
                     'default_ordering': {'total': 'desc'},
                 },
+                'costs_by_project': {
+                    'aggregates': {'value': Sum('pod_cost')},
+                    'annotations': {
+                        'total': Sum('pod_cost'),
+                        'units': Coalesce(Max('currency_code'), Value('USD'))
+                    },
+                    'count': None,
+                    'delta_key': {'total': Sum('unblended_cost')},
```
adberglund: delta key should be pod_cost, too

Issue #652, opened 2019-02-19
TAGs API should include OCP storage volume labels


dccurtis:
Currently the `/api/v1/tags/ocp` API returns pod usage tags located in the `pod_labels` column of the `OCPUsageLineItemDailySummary` table.

OCP storage volume labels should also be included in the API response.  They are located in the `volume_labels` from the `OCPStorageLineItemDailySummary` table.

Details need to be thought through on this enhancement but at a minimum the developer should figure out whether or not the type of label should be included in the response and also if filter/group_by label type should be supported.

lcouzens:
Verified
Commit: 9e21cbb23de1645b26188c6c4079d04131ddf98e

PR #677, opened 2019-02-26
Adding OpenShift storage labels to tags endpoint

dccurtis:
This adds the storage labels to the endpoint /tags/openshift/.  In addition to that a new `type` field was added to the output response to indicate whether or not the label is for `pod` or `storage`.  Open to suggestions for better names on all.

**Testing**
1.  Issue /tags/openshift endpoint and verify both pod and storage labels are returned
2.  Use key_only boolean to show only keys are returned
3.  Filter all tags by pod type
4. Filter all tags by storage type
5.  Sanity check AWS tags

**Test Results**
[koku_issue_652_ut.txt](https://github.com/project-koku/koku/files/2907236/koku_issue_652_ut.txt)

Addresses #652


dccurtis:
Yeah that's correct.  Just to be consistent.  I'm good with removing it

dccurtis:
Updated the test doc as well with AWS examples

PR #679, opened 2019-02-26
refactor time_scope params

blentz:
This is a refactor of the time_scope_* and resolution query parameters. It's an API-breaking change, as the behaviors are now a bit different (and more consistent).

There's a couple of other bits of tidying up that snuck in, too.

### Changes to time_scope_units behavior 

*Old behavior:*
- When `time_scope_units = month`, the value returned was one month of data, with which month being determined by `time_scope_value`. 

e.g.  if today is February 26, then a query of `time_scope_value=-2&time_scope_units=month` would return data between January 1 and January 31.

*New behavior:*
- When `time_scope_units = month`, the value returned is a _range_ of data starting with the date `time_scope_value` months ago up to the current date. 

e.g. if today is February 26, then a query of `time_scope_value=-2&time_scope_units=month` will return the data between January 1 and February 26, inclusive.

*Unchanged behavior:*
- When `time_scope_units = day`, the value returned is data in the range between `time_scope_value` days ago and today.

e.g. if today is February 26, then a query of `time_scope_value=-10&time_scope_units=day` will return the data between February 16 and February 26, inclusive.

### Changes to defaults

*Old behavior:* 
The default for `time_scope_value` when `time_scope_units=day` was -10
The default for `time_scope_value` when  `time_scope_units=month` was -1

*New behavior:*
The defaults for `time_scope_value` is now -1 for all values of `time_scope_units`

*Old behavior:* 
The default for if no `time_scope_units` is given is the default is based on `time_scope_value` (None: day, [-1,-2]: month) 
The default for if no `time_scope_value` is given is the default is based on `time_scope_units` (day: -10, month: -1) 
The default for if no `resolution` is given varies based on `time_scope_value` (None: daily, [-1, -2]: monthly)

*New behavior:*
The default `time_scope_units` is day
The default `time_scope_value` is -1
The default `resolution` is daily


koku/api/report/test/ocp/test_ocp_query_handler.py
```diff
@@ -321,9 +310,12 @@ def test_add_current_month_deltas_no_previous_data_wo_query_data(self):
...
             result = handler.add_current_month_deltas(query_data, query_sum)
 
+            expected_delta = query_sum.get('usage', 0) - query_sum.get('request', 0)
+            expected_percent = query_sum.get('usage', 0) / query_sum.get('request', 0) * 100
+
             self.assertEqual(result, query_data)
-            self.assertEqual(handler.query_delta['value'], Decimal(0))
-            self.assertIsNone(handler.query_delta['percent'])
+            self.assertEqual(handler.query_delta['value'], expected_delta)
+            self.assertEqual(handler.query_delta['percent'], expected_percent)
```
blentz: This test change was required because the delta between `usage` and `request` was non-zero. I'm not sure what changed in the data generator to cause the values to differ, but it means the test wasn't testing the right thing - the expected values would never be zero, as far as I could tell.

koku/api/query_handler.py
```diff
@@ -66,13 +66,21 @@ def __init__(self, query_parameters, url_data,
...
+        if self.resolution == 'monthly':
+            self.date_to_string = lambda dt: dt.strftime('%Y-%m')
+            self.string_to_date = lambda dt: datetime.datetime.strptime(dt, '%Y-%m').date()
+            self.date_trunc = TruncMonthString
+            self.gen_time_interval = DateHelper().list_months
+        else:
+            self.date_to_string = lambda dt: dt.strftime('%Y-%m-%d')
+            self.string_to_date = lambda dt: datetime.datetime.strptime(dt, '%Y-%m-%d').date()
+            self.date_trunc = TruncDayString
+            self.gen_time_interval = DateHelper().list_days
```
blentz: This if-else block got moved in the conversion to using properties. I strongly suspect all of these things can be relocated closer to where they're being used, but I didn't want to expand the scope of this change quite that far. So, they can live in the constructor for now. :smile:

adberglund:
This will require a corresponding masu update to re-do how we populate `reporting_awscostentrylineitem_aggregates`. Currently -2 in that table refers to just the previous month's data.

In the following example you can see that the total does not match the data.

```
GET /local/v1/reports/aws/instance-types/?filter[time_scope_value]=-2&filter[time_scope_units]=month&filter[resolution]=monthly
HTTP 200 OK
Allow: OPTIONS, GET
Vary: Accept

{
    "filter": {
        "resolution": "monthly",
        "time_scope_value": "-2",
        "time_scope_units": "month"
    },
    "data": [
        {
            "date": "2019-01",
            "instance_types": [
                {
                    "instance_type": "m5.large",
                    "values": [
                        {
                            "instance_type": "m5.large",
                            "date": "2019-01",
                            "units": "Hrs",
                            "cost": 192.0,
                            "count": 8,
                            "total": 192.0
                        }
                    ]
                }
            ]
        },
        {
            "date": "2019-02",
            "instance_types": [
                {
                    "instance_type": "t2.micro",
                    "values": [
                        {
                            "instance_type": "t2.micro",
                            "date": "2019-02",
                            "units": "Hrs",
                            "cost": 3.1784,
                            "count": 45,
                            "total": 1024.0
                        }
                    ]
                },
                {
                    "instance_type": "t2.small",
                    "values": [
                        {
                            "instance_type": "t2.small",
                            "date": "2019-02",
                            "units": "Hrs",
                            "cost": 11.822,
                            "count": 22,
                            "total": 514.0
                        }
                    ]
                }
            ]
        }
    ],
    "total": {
        "units": "Hrs",
        "cost": 192.0,
        "count": 4,
        "value": 192.0
    }
}
```

chambridge:
Per the slack conversation from Luke:
```
Ok I had a quick look at the PR comments and I think I understand, my only question is in reference to the -1 for days or
```The default time_scope_units is day
The default time_scope_value is -1
The default resolution is daily```
So we return 1 day?
```
Would it make sense four our default to be a `month` instead of a `day` as cost management space users tend to care more about monthly data?

chambridge:
Ran the QE smoke test against this branch. Saw some expected failures but calling out some others for clarity for @project-koku/quality-engineers & @project-koku/ux-ui 
In a Slack conversation you said:
> The off-by-one errors are to be expected. There's several in our own unit tests that I fixed in my PR.

```
____________________________________________ test_api_ocp_volume_validate_items_past_days_daily[10-days-clusters] ____________________________________________

variable = 'cluster', units = ['day', '-10'], ocp_volume_report = <iqe_hccm.rest_api.koku_models.KokuOcpVolumeReport object at 0x112e2f828>

    @pytest.mark.smoke
    @pytest.mark.parametrize(
        "variable", ["node", "project", "cluster"], ids=("nodes", "projects", "clusters")
    )
    @pytest.mark.parametrize("units", [["day", "-10"], ["day", "-30"]], ids=("10-days", "30-days"))
    def test_api_ocp_volume_validate_items_past_days_daily(variable, units, ocp_volume_report):
        """Test to validate the days match the given param
    
        Polarion:
            assignee: lcouzens
        """
    
        if units[1] == "-10":
            days = 10
        else:
            days = 30
    
        ocp_volume_report.get(
            report_filter={
                "time_scope_units": units[0],
                "time_scope_value": units[1],
                "resolution": "daily",
            },
            group_by=[[variable, "*"]],
        )
    
        # Using ocp_volume_report.data to count all days regardless of blank values
        count = len(ocp_volume_report.data)
    
>       assert count == days
E       assert 11 == 10
E         -11
E         +10

iqe_hccm/tests/rest_api/v1/test_ocp_volume_reports.py:105: AssertionError
_____________________________________________ test_api_ocp_volume_validate_items_past_days_daily[30-days-nodes] ______________________________________________

variable = 'node', units = ['day', '-30'], ocp_volume_report = <iqe_hccm.rest_api.koku_models.KokuOcpVolumeReport object at 0x1145bf588>

    @pytest.mark.smoke
    @pytest.mark.parametrize(
        "variable", ["node", "project", "cluster"], ids=("nodes", "projects", "clusters")
    )
    @pytest.mark.parametrize("units", [["day", "-10"], ["day", "-30"]], ids=("10-days", "30-days"))
    def test_api_ocp_volume_validate_items_past_days_daily(variable, units, ocp_volume_report):
        """Test to validate the days match the given param
    
        Polarion:
            assignee: lcouzens
        """
    
        if units[1] == "-10":
            days = 10
        else:
            days = 30
    
        ocp_volume_report.get(
            report_filter={
                "time_scope_units": units[0],
                "time_scope_value": units[1],
                "resolution": "daily",
            },
            group_by=[[variable, "*"]],
        )
    
        # Using ocp_volume_report.data to count all days regardless of blank values
        count = len(ocp_volume_report.data)
    
>       assert count == days
E       assert 31 == 30
E         -31
E         +30

iqe_hccm/tests/rest_api/v1/test_ocp_volume_reports.py:105: AssertionError
```
Just want clarity on expectations so everyone is on the same page.

chambridge:
Another question around what `delta` means when we are looking at multiple months of data (just not clear what it means):

*GET* `/r/insights/platform/cost-management/v1/reports/aws/costs/?delta=cost&filter[time_scope_units]=month&filter[time_scope_value]=-2&filter[resolution]=monthly&group_by[account]=*&order_by[cost]=desc`
```
{
    "delta": {
        "value": 6141.005613538,
        "percent": 2232.37058143692
    },
    "group_by": {
        "account": [
            "*"
        ]
    },
    "order_by": {
        "cost": "desc"
    },
    "filter": {
        "resolution": "monthly",
        "time_scope_value": "-2",
        "time_scope_units": "month"
    },
    "data": [
        {
            "date": "2019-02",
            "accounts": [
                {
                    "account": "299748131127",
                    "values": [
                        {
                            "date": "2019-02",
                            "account": "299748131127",
                            "infrastructure_cost": {
                                "value": 6142.793729797,
                                "units": "USD"
                            },
                            "derived_cost": {
                                "value": 0.0,
                                "units": "USD"
                            },
                            "cost": {
                                "value": 6142.793729797,
                                "units": "USD"
                            },
                            "account_alias": "jonnyfive",
                            "delta_value": 6142.793729797,
                            "delta_percent": null
                        }
                    ]
                },
                {
                    "account": "256293412812",
                    "values": [
                        {
                            "date": "2019-02",
                            "account": "256293412812",
                            "infrastructure_cost": {
                                "value": 248.058719942,
                                "units": "USD"
                            },
                            "derived_cost": {
                                "value": 0.0,
                                "units": "USD"
                            },
                            "cost": {
                                "value": 248.058719942,
                                "units": "USD"
                            },
                            "account_alias": "256293412812",
                            "delta_value": -3.805070101,
                            "delta_percent": -1.5107650450072125
                        }
                    ]
                },
                {
                    "account": "589173575009",
                    "values": [
                        {
                            "date": "2019-02",
                            "account": "589173575009",
                            "infrastructure_cost": {
                                "value": 25.242150574,
                                "units": "USD"
                            },
                            "derived_cost": {
                                "value": 0.0,
                                "units": "USD"
                            },
                            "cost": {
                                "value": 25.242150574,
                                "units": "USD"
                            },
                            "account_alias": "hccm-alias",
                            "delta_value": 2.016953842,
                            "delta_percent": 8.684334799287246
                        }
                    ]
                }
            ]
        }
    ],
    "total": {
        "infrastructure_cost": {
            "value": 6416.094600313,
            "units": "USD"
        },
        "derived_cost": {
            "value": 0.0,
            "units": "USD"
        },
        "cost": {
            "value": 6416.094600313,
            "units": "USD"
        }
    }
}
```

blentz:
### RE: QE test failures

It's an "arrays start at 0" thing.

Instead of having "-1" mean "only yesterday" , -1 now means "today and yesterday".

So, when you say, "give me -1 days" you get 2 data points - today and yesterday, rather than 1.

That extends to the -10 case, where you now get 11 data points instead of 10.

### RE: Deltas

I'll need to investigate the delta code. It might need to change more because this is likely violating some assumptions it has about what the previous query is.

koku/api/report/test/ocp/helpers.py
```diff
@@ -262,7 +268,10 @@ def create_line_items(self, report_period, report):
                 'pod_labels': self._gen_pod_labels(report)
             }
             line_item = OCPUsageLineItem(**data)
-            line_item.save()
+            try:
+                line_item.save()
+            except IntegrityError as exc:
+                LOG.critical("Error in mock data generation: %s", exc)
```
blentz: This is a stop-gap to avoid tests failing on data generation, though it may still fail because the test is expecting N number of rows and it gets N-1 instead.

I didn't want to spend a lot of time crafting a more complete fix. But this should improve test reliability a bit in the meantime.

chambridge:
Just circling back to the deltas question above ^, have you investigated and can share output?

blentz:
@chambridge - I've pushed a fix to make sure the delta calculations are in line with the new behavior.

To answer your question about what the deltas mean - It will be the delta between the requested time interval and an interval of the same size immediately previous to the requested interval. e.g. If the request is "-7 days", it'll be a delta between this week and last week. If the request is "-2 months", it'll be a delta between the current two months and the two months prior.

This does introduce a flaw in the calculation that we'll need to address. This math works fine going by days. But the monthly version makes no guarantees that you're comparing the same number of days between the intervals. But this flaw is present in the current behavior as well - the delta between February (28 days) and either January or March (both 31 days) is similarly skewed.

I'll open a tech debt issue so we can circle back to it in a future PR after we discuss how we want to handle that edge case.

Issue #699, opened 2019-03-02
Wrong error message being shown when not specifying bucket


lcouzens:
**Describe the bug**
Wrong error message being shown when not specifying bucket name when adding an AWS provider through the API

**To Reproduce**
1. Go to '/v1/providers/'
2. Fill in correct AWS details but leave bucket name blank
3. See error
```
errors": [
        {
            "detail": "Unable to obtain credentials with using arn:aws:iam::111111111111:role/LocalAWSProvider.",
            "source": "provider_resource_name",
            "status": 400
        }
```
**Expected behavior**
Message should say 'Bucket name must not be blank'

Additional info:
If you try adding AWS provider but switch the type to AWS-local you then get the correct error message.
```
Bucket is a required parameter for AWS.
```

lcouzens:
Verified
Commit: 0daaf8012966f31e9fa15a20282716743197e915
Automation: test_api_aws_provider_create_no_billing_source

Issue #698, opened 2019-03-02
Currently able to add OCP provider without ARN/Cluster ID


lcouzens:
**Describe the bug**
Currently able to add OCP provider without ARN/Cluster ID

**To Reproduce**
1. Go to '/v1/providers/'
2. Set Name
3. Set type to OCP
4. Hit submit
5. Provider added
```
"uuid": "caa03676-8b5a-4a04-b5b5-cc0b132b2433",
            "name": "test",
            "type": "OCP",
            "authentication": {
                "uuid": "38a78a7e-2250-4e19-b4c2-f44dbef8599b",
                "provider_resource_name": ""
            },
```
**Expected behavior**
Error message should appear saying ARN/Cluster ID is required

lcouzens:
Verified
Commit: 0daaf8012966f31e9fa15a20282716743197e915
Automation: test_api_ocp_provider_create_no_resource_name

Issue #697, opened 2019-03-02
Typo/extra space in error message


lcouzens:
**Describe the bug**
Typo/extra space in error message

**To Reproduce**
1. Go to v1/providers/
2. Try to add an AWS provider without specifying 'Provider Resource Name' 

```
            "detail": "Unable to obtain credentials with using .",
```
Error message should actually say something like 'Provider resource name must not be blank'

Also I'm not sure the wording makes sense in general when you provider an ARN:
```
"detail": "Unable to obtain credentials with using test.",
```
'with using' 
I think we should drop one word or reword this.
'Unable to obtain credentials using test' I think makes more sense.

lcouzens:
Verified
Commit: 0daaf8012966f31e9fa15a20282716743197e915
Automation: test_api_aws_provider_create_no_resource_name

PR #707, opened 2019-03-05
Add small provider test case fixes.

chambridge:
Associated with: #697 #698 #699

### DIFF
--- a/koku/providers/aws/aws_provider.py
+++ b/koku/providers/aws/aws_provider.py
@@ -135,23 +135,29 @@ class AWSProvider(ProviderInterface):
 
     def cost_usage_source_is_reachable(self, credential_name, storage_resource_name):
         """Verify that the S3 bucket exists and is reachable."""
+        if not credential_name or len(credential_name.strip()) == 0:
+            key = 'authentication.provider_resource_name'
+            message = 'Provider resource name is a required parameter for AWS' \
+                ' and must not be blank.'
+            raise serializers.ValidationError(error_obj(key, message))
+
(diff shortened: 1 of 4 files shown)


koku/providers/aws/aws_provider.py
```diff
@@ -135,31 +135,37 @@ def name(self):
 
     def cost_usage_source_is_reachable(self, credential_name, storage_resource_name):
         """Verify that the S3 bucket exists and is reachable."""
+        if not credential_name or len(credential_name.strip()) == 0:
```
blentz: You could use `.isspace()` here.

(See: https://docs.python.org/3.6/library/stdtypes.html#str.isspace )

koku/providers/aws/aws_provider.py
```diff
@@ -135,31 +135,37 @@ def name(self):
...
             key = 'provider_resource_name'
-            message = 'Unable to obtain credentials with using {}.'.format(
+            message = 'Unable to access account resources with ARN {}.'.format(
                 credential_name)
             raise serializers.ValidationError(error_obj(key, message))
 
-        if not storage_resource_name:
-            key = 'bucket'
-            message = 'Bucket is a required parameter for AWS.'
+        if not storage_resource_name or len(storage_resource_name.strip()) == 0:
```
blentz: `isspace()` opportunity here, too.

koku/providers/aws/aws_provider.py
chambridge: Good to know 👍

PR #718, opened 2019-03-08
Remove duplicate tag keys

adberglund:
## Summary
Filter out duplicate keys when multiple sources contribute to the key list

## Testing

Before: 

```
GET /local/v1/tags/openshift/?filter[time_scope_value]=-2&key_only=true
HTTP 200 OK
Allow: OPTIONS, GET
Vary: Accept

{
    "meta": {
        "count": 9,
        "filter": {
            "time_scope_value": "-2"
        },
        "key_only": true
    },
    "links": {
        "first": "/local/v1/tags/openshift/?filter%5Btime_scope_value%5D=-2&key_only=true&limit=100&offset=0",
        "next": null,
        "previous": null,
        "last": "/local/v1/tags/openshift/?filter%5Btime_scope_value%5D=-2&key_only=true&limit=100&offset=0"
    },
    "data": [
        "key1",
        "key2",
        "key3",
        "key3",
        "key4",
        "key4",
        "storageclass",
        "vc_key2",
        "vc_key3"
    ]
}
```

After:

```
GET /local/v1/tags/openshift/?filter[time_scope_value]=-2&key_only=true
HTTP 200 OK
Allow: OPTIONS, GET
Vary: Accept

{
    "filter": {
        "time_scope_value": "-2"
    },
    "key_only": true,
    "data": [
        "key1",
        "key2",
        "key3",
        "key4",
        "storageclass",
        "vc_key2",
        "vc_key3"
    ]
}
```

### DIFF
--- a/koku/api/report/test/ocp/tests_views.py
+++ b/koku/api/report/test/ocp/tests_views.py
@@ -778,13 +778,17 @@ class OCPReportViewTest(IamTestCase):
 
     def test_execute_query_with_tag_filter(self):
         """Test that data is filtered by tag key."""
+        handler = OCPTagQueryHandler('?type=pod', {'type': 'pod'}, self.tenant)
+        tag_keys = handler.get_tag_keys()
+        filter_key = tag_keys[0]
+
         with tenant_context(self.tenant):
             labels = OCPUsageLineItemDailySummary.objects\
-                .filter(usage_start__date__gte=self.ten_days_ago)\
+                .filter(usage_start__gte=self.ten_days_ago)\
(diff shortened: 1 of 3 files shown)


chambridge:
Do we need a set here too:
https://github.com/project-koku/koku/pull/718/files#diff-15229760e25a5831ac9dbc834694165cR153

adberglund:
@chambridge I don't believe so no. That result comes from a `.values()` call which is deduped by a group by. Then the data gets processed into a shared merged dictionary which is okay. Do need to update more tests that are periodically failing now. I keep getting a perfect run locally and failing in travis.

Issue #607, opened 2019-02-05
API for determining whether OCP clusters are running on AWS


chambridge:
## User Story
As a user, if I have OCP clusters running on AWS, I want to see the cost of my clusters that are running on AWS so that I can understand cost of running these OCP clusters broken down by AWS cost line items; however, if I do not have any clusters running on AWS I should not have an option to view this information.

## Impacts
API, UI, Docs

## Role
User

## Assumptions
- This is special case provider information for OCP

## API Details

| API  | Detail|
 | --- | --- |
 | **Method** | GET |
 | **Path** | /api/v1/providers/ |
 **Response:**
```
{
	"count": 1,
	"next": null,
	"previous": null,
	"results": [{
		"uuid": "e5206320-9286-4b0b-ac3c-7150102a21d2",
		"name": "ocp-hccm",
		"type": "OCP",
		"authentication": {
			"uuid": "fc15ebd5-8c1e-4306-b8ff-8fcc1347e0f2",
			"provider_resource_name": "9b6fc2fe-427b-4457-9f85-3211dd564158"
		},,
		"customer": {
			"uuid": "5065621e-5a39-495f-98b9-341b5d6b782f",
			"account_id": "6193296",
			"date_created": "2019-02-04T19:41:50.386195Z"
		},
		"created_by": {
			"uuid": "212a19fe-77c2-4395-aebb-5251f9d26fe6",
			"username": "cost-demo",
			"email": "chambrid@redhat.com"
		},
	       "infrastructure": "AWS",
		"stats": {
			"2019-01-01": [{
				"assembly_id": "f23124d1-cdb0-468d-a84a-a0adaa783280",
				"billing_period_start": "2019-01-01",
				"files_processed": "0/1",
				"last_process_start_date": "2019-02-04 19:55:15",
				"last_process_complete_date": null,
				"summary_data_creation_datetime": "2019-02-04 19:55:22",
				"summary_data_updated_datetime": "2019-02-04 20:06:02"
			}]
		}
	}]
}
```

## UI Details
UI can determine whether an OCP cluster/provider is running on AWS by looking at the information found in the providers endpoint. OCP clusters/providers will have a read-only `infrastructure` field that could be `AWS` or `Unknown`.

## Acceptance Criteria

- [x] OCP providers have "infrastructure" field that provides either "AWS" or "Unknown" correctly

chambridge:
Maybe filter by infrastructure type too?

dchorvat1:
Verified
[...]

PR #754, opened 2019-03-20
Adding infrastructure field to provider response

dccurtis:
Addresses #607 

The provider endpoint now has a new field called `infrastructure` which will designate what infrastructure a provider is running on.

The primary use case for the short term is OCP on AWS.  OCP clusters running on AWS (resource_id matches and usage period aligns between the two providers) will now have `infrastructure=AWS` when we detect that an OCP cluster is running on AWS.

All providers not detected to be running on AWS have `infrastructure=Unknown`.  If we encounter an error while trying to determine the infrastructure the API response has `infrastructure=Unknown-Error`.

There is still room for improvement to cleanup the test data generation.  First candidate would be to decouple the `FakeAWSCostData` from the `ReportQueryTest` test suite and then update the new `OCPAWSReportDataGenerator().add_aws_data_to_tenant()` to use it instead.  I started doing this work but I was hitting the point of diminishing returns for focusing on new sprint content.

**Testing**
1.  Add AWS provider with data and ingest
2. Add OCP provider with matching date range and resource_id
3. Check the provider list and see that both AWS and OCP are infrastructure="Unknown"
4.  Ingest the OCP data and recheck the infrastruture field.  Verify that it is "AWS"
5.  Add another OCP Provider and verify that the Insfrastructure is "Unknown" and the original OCP-on-AWS provider is "AWS"

**Test Results**
[koku_607_ut.txt](https://github.com/project-koku/koku/files/2988471/koku_607_ut.txt)


koku/providers/aws/aws_provider.py
```diff
@@ -179,3 +179,7 @@ def cost_usage_source_is_reachable(self, credential_name, storage_resource_name)
                      topics_string, storage_resource_name)
         else:
             LOG.info('SNS is not configured for S3 bucket %s', storage_resource_name)
+
+    def infrastructure_type_meat(self, provider_uuid, tenant):
```
blentz: I the `_meat` bit intentional?  I don't understand what it's supposed to mean.

dccurtis: ha, I forgot to come back and search/replace for a better name ;)

koku/providers/ocp/ocp_provider.py
```diff
@@ -17,9 +17,13 @@
 """OCP service provider implementation to be used by Koku."""
 import logging
 
+from django.core.exceptions import ObjectDoesNotExist, ValidationError
 from django.utils.translation import ugettext as _
 from rest_framework import serializers
+from tenant_schemas.utils import tenant_context
 
+from api.provider.models import Provider
+from reporting.provider.ocp_aws.models import OCPAWSUsageLineItemDaily
```
adberglund:
```suggestion
from reporting.provider.ocp_aws.models import OCPAWSCostLineItemDailySummary
```

koku/providers/ocp/ocp_provider.py
```diff
@@ -57,3 +69,28 @@ def cost_usage_source_is_reachable(self, cluster_id, storage_resource_name):
         message = 'Stub to verify that OCP report for cluster {} is accessible.'.format(
                   cluster_id)
         LOG.info(message)
+
+    def _is_on_aws(self, tenant, resource_name):
+        """Determine if provider is running on AWS."""
+        with tenant_context(tenant):
+            objects = OCPAWSUsageLineItemDaily.objects.all()
```
adberglund:
```suggestion
            objects = OCPAWSCostLineItemDailySummary.objects.all()
```

koku/providers/provider_interface.py
```diff
@@ -62,3 +62,20 @@ def cost_usage_source_is_reachable(self, credential_name, storage_resource_name)
 
         """
         pass
+
+    @abstractmethod
+    def infrastructure_type_meat(self, provider_uuid, schema_name):
```
adberglund: 🍖

koku/providers/ocp/ocp_provider.py
dccurtis: If you don't mind, I'd like to open a tech debt story to take care of this in a separate PR.  This change basically messes up the testing as it stands now.  I started making the test data updates but decided it might be better to get this in so the UI can integrate and come back for the performance improvements.

adberglund: Sounds good to me.

PR #766, opened 2019-03-22
Validate AWS delta value support based on path.

chambridge:
Now provides expected output:

*GET* `/api/cost-management/v1/reports/aws/storage/?filter[time_scope_units]=month&filter[time_scope_value]=-1&filter[resolution]=monthly&delta=cost`

```
{
    "delta": {
        "delta": "\"cost\" is not a valid choice."
    }
}
```


chambridge:
When I looked at that across all the endpoints it got quite a bit more complex. We can certainly go in that direction when needed but chose to just go with the above for now. Can certainly see us doing that for GA though.

PR #782, opened 2019-04-01
Set provider name filter lookup expression to contain

### LINKED ISSUE #781: Filtering provider by its name should show providers that contains the name
**Describe the bug**
related to https://github.com/project-koku/koku-ui/issues/649

When filtering providers by their names, I would expect to see matches that contain the name that passed as argument. However, I see only providers that match exactly to the given name.

For example, passing `/providers/?name=provider-1` will show only provider-1 while `/providers/?name=provider` will return no match.

**To Reproduce**
1. Create two providers. For example OCP providers with the names: provider-1 and provider-2
2. Send a GET request to providers with the name parameter set to provider
3. You will see no match

**Expected behavior**
See provider-1 and provider-2 as the provider string is contained in both of them.

**Screenshots**
Partial string:
Screen Shot 2019-04-01 at 9 47 59 AM

Full String:
Screen Shot 2019-04-01 at 9 48 14 AM

boaz0:
fixes #781 

This is going to filter providers that their names contain the `name` parameter passed in the request.

### DIFF
--- a/koku/api/provider/test/tests_views.py
+++ b/koku/api/provider/test/tests_views.py
@@ -189,6 +189,40 @@ class ProviderViewTest(IamTestCase):
         self.assertEqual(json_result.get('stats'), {})
         self.assertEqual(json_result.get('infrastructure'), 'Unknown')
 
+    def test_filter_providers_by_name_contains(self):
+        """Test that providers that contain name appear."""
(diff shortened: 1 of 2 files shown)


boaz0:
I believe the failed tests are not related to this change. Although I might be wrong.

Issue #783, opened 2019-04-02
Configure autoscaling for all DeploymentConfigs


blentz:
## User Story
As service operator, I want my pods to autoscale so that I don't need to scale my pods manually.

## Impacts
API

## Role
Service Operator

## Assumptions
- The DeploymentConfig supports autoscaling (StatefulSets do not support autoscale)

## Implementation Details
- Because we have limited operational data, just do a "best guess" when setting thresholds for when we should autoscale.
-

## Acceptance Criteria

- [ ] Autoscaling is defined in each deploymentconfig
- [ ]

PR #800, opened 2019-04-09
add autoscaling

blentz:
Configuring autoscaling for #783


chambridge:
So I think the next part of the effort is to update the existing deployments with the updated config. But great stuff!

Issue #799, opened 2019-04-09
Apply pricing rates to multiple clusters


blentz:
## User Story
As a koku user, 
I want to specify which OpenShift clusters my price list applies to,
so that I can manage one price list for multiple OpenShift clusters.

See also: https://docs.google.com/document/d/1N06BRnzdAM5gTI62I43t3b8o3DgV-lBw9R7rVV4MZXQ/edit?usp=sharing

## Impacts
- API
- UI
- Docs

## Role
Koku User

## Assumptions
- API stores a list of objects that can be associated with a set of Rates
-

## API Details

| API  | Detail|
 | --- | --- |
 | **Method** | GET/POST/PUT/PATCH/DELETE |
 | **Path** | /api/v1/?/ |
 | **Body** |  |
 | **Response** | |

## UI Details
The UI should allow the user to associate a Price List with a set of objects (probably tags).

## Acceptance Criteria

- [ ] Koku stores an association between Rates and other objects (e.g. tags, clusters, etc.).
- [ ] The relationship between Rates and other objects is 1-to-M. (One price list can apply to many other things.)
- [ ] The prices defined by the Rates are applied to all objects associated with that price list

chambridge:
remove the provider_uuid from rates model:
https://github.com/project-koku/koku/blob/master/koku/rates/models.py#L47

Add foreign key from rates to provider (e.g.):
https://github.com/project-koku/koku/blob/master/koku/api/provider/models.py#L80

Need to be able to see if we can foreign key to another schema since providers are in public schema and rates is in tenant schema.

_Nice to have:_
Migrate data if possible (time box to 1 day; otherwise we will blow away rates and recreate).

Need to support PUT (update) on provider.

investigate cost/benefit of moving provider to tenant schema or rates to public (preferred in provider in tenant). What are the impacts.
^prototype the above.

Rates API should take/show a list of providers (the list can be empty).

lcouzens:
Verified
Commit: d4de1a9035d4005a307d7116578f40b26f4d34f7

PR #829, opened 2019-04-23
Support OpenShift one-to-many rates to clusters

dccurtis:
Addresses #799 

See also masu change: https://github.com/project-koku/masu/pull/420

Rate API and database model changes to support multiple OpenShift providers being associated with a single rate instance.

**API changes:**
* Provider UUID parameter is now a list of uuids: `provider_uuid: uuid` -> `provider_uuids: [uuid1, uuid2]`
* `provider_uuids` is optional
* `PATCH` is not supported; I think we should support it in a future change since updating provider uuids is cumbersome with `PUT`

**System behavior changes**
* Removing a provider now removes the association with that provider and the rate.  The rate instance itself remains.
* Removing a rate instance will remove the association with the provider. *Be Careful*
* A single OpenShift provider can only have a one metric instance associated with it at a time (i.e. can not have multiple `cpu_core_request_per_hour` metrics).  An error is thrown for `POST`

**Testing**
* QE OCP derived-cost tests pass after necessary updates.

1.  Add rates to two providers on master
2.  Add both providers to a new rate
3.  Remove one of the providers with the PUT operation from step 2
4.  Remove the other provider on the memory rate and verify the rate remains with no providers
5.  Delete rate with no providers (2131a827-17cb-441e-ac5e-ea02f90f247e)
6.  Delete rate with providers (2b4f38d0-f063-44f9-bff8-b07604d297c8) and verify ratemap entry is also removed
7. Attempt to add a rate to a provider where the rate already exists.  Verify error Response
8. Create a metric with no provider_uuids supplied
9.  Add a provider to the rate with no providers (69812bc7-9fa0-4d93-b103-8738a3387442)
10.  Remove provider "6604ba3a-6a43-4366-98dc-988143da6652" and verify the rate map entry is removed as well as the other rate it was associated with (24b51457-5aaf-4ce3-bd39-b591c573232c)
11.  Remove rate with provider associated.  Verify mapping is removed as well as the rate instance
12.  Add two rates for a provider and veirfy the query filter works
13.  Ingest data for the provider with rates and verify that charges are calclated in masu
14.  Add Rate with provider, create another rate of same type (no providers), update no-provider rate with the same provider.  Verify error is thrown

**Test Results**
[koku_rate_1_to_many_ut.txt](https://github.com/project-koku/koku/files/3108514/koku_rate_1_to_many_ut.txt)


dccurtis:
@boaz0 Just want to bring your attention to this since it involves some rate API changes.  Let me know if you have any questions.

PR #853, opened 2019-05-06
Match rate API json data for PUT/POST to response

dccurtis:
PUT/POST operations now accept json data that matches the rate API response.  Having the two json formations was causing unnecessary confusion.  Now if the user wants to make a change via PUT, they can simply copy/paste the GET response, update the fields as needed and issue PUT.

**Testing**
1. Create a new rate with data that matches the rates API response
2.  Get the current rate, copy/paste response to json data and PUT with a modification to it
3.  Change the user-friendly API elements and issue PUT, verify that the change is ignored
4. Run all API tests and verify no regressions from status-quo

**Test Results**
[koku_rate_api_ut.txt](https://github.com/project-koku/koku/files/3149724/koku_rate_api_ut.txt)


dccurtis:
@boaz0 FYI, this is a change to the json data POST/PUT for creating and modifying rates.  You're not doing that currently, but I wanted to give you the heads up that this is merging incase you see anything change on your end.

PR #875, opened 2019-05-24
add OCP+AWS tags endpoint, fixes #840

### LINKED ISSUE #840: OCP on AWS: group_by tag not working
### Proposed Solution
Create a filtered API endpoint for AWS tags that returns just the AWS tags for OCP on AWS. 
A possible implementation could be to use the `/infrastructures/aws/` path akin to our reporting APIs, and this could give the the appropriate AWS tags, but would have to live on the `/tags/openshift/` path in order to match. That might be confusing, but it could set us up for other providers on AWS. Alternatively we could just have a filter on the `/tags/aws/` endpoint that just pulls openshift matched tags. 

**Describe the bug**
Cannot group_by tag keys with the `/api/cost-management/v1/reports/openshift/infrastructures/aws/costs` API. No value is returned, which results in an empty table.

The group_by tag key still works for OCP details and AWS details.

**To Reproduce**
1. In the Koku UI, go to 'OpenShift on AWS'
2. Click on 'Group cost by'
3. Select 'TagKey: 5dc_key' or any tag key

**Expected behavior**
Not sure when this stopped working, but we're still able to obtain a list of tag keys. The user should be able to group_by the list of tags obtained.

**Screenshots**
Screen Shot 2019-04-29 at 2 11 40 PM

Full API used:
/api/cost-management/v1/reports/openshift/infrastructures/aws/costs/?delta=cost&filter%5Blimit%5D=10&filter%5Boffset%5D=0&filter%5Bresolution%5D=monthly&filter%5Btime_scope_units%5D=month&filter%5Btime_scope_value%5D=-1&group_by%5Btag%3A5dc_key%5D=%2A&order_by%5Bcost%5D=desc

blentz:
This adds an API endpoint for querying OCP+AWS tags.

New endpoint: `<prefix>/tags/openshift/infrastructures/aws/`

Functional test:
```
COMMAND: curl -v -g -H 'HTTP_X_RH_IDENTITY: ImlkZW50aXR5Ijp7ImFjY291bnRfbnVtYmVyIjoiMTAwMDEiLCJ0eXBlIjoiVXNlciIsInVzZXIiOnsidXNlcm5hbWUiOiJ0ZXN0X2N1c3RvbWVyIiwiZW1haWwiOiJibGVudHoraGNjbS10ZXN0QHJlZGhhdC5jb20ifX0K' http://koku-myproject.192.168.42.57.nip.io/api/v1/tags/openshift/infrastructures/aws/
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0* Could not resolve host: ImlkZW50aXR5Ijp7ImFjY291bnRfbnVtYmVyIjoiMTAwMDEiLCJ0eXBlIjoiVXNlciIsInVzZXIiOnsidXNlcm5hbWUiOiJ0ZXN0X2N1c3RvbWVyIiwiZW1haWwiOiJibGVudHoraGNjbS10ZXN0QHJlZGhhdC5jb20ifX0K'
* Closing connection 0
curl: (6) Could not resolve host: ImlkZW50aXR5Ijp7ImFjY291bnRfbnVtYmVyIjoiMTAwMDEiLCJ0eXBlIjoiVXNlciIsInVzZXIiOnsidXNlcm5hbWUiOiJ0ZXN0X2N1c3RvbWVyIiwiZW1haWwiOiJibGVudHoraGNjbS10ZXN0QHJlZGhhdC5jb20ifX0K'
*   Trying 192.168.42.57...
* TCP_NODELAY set
* Connected to koku-myproject.192.168.42.57.nip.io (192.168.42.57) port 80 (#1)
> GET /api/v1/tags/openshift/infrastructures/aws/ HTTP/1.1
> Host: koku-myproject.192.168.42.57.nip.io
> User-Agent: curl/7.64.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Server: gunicorn/19.9.0
< Date: Wed, 29 May 2019 18:40:22 GMT
< Vary: Accept, Origin
< Allow: GET, OPTIONS
< X-Frame-Options: SAMEORIGIN
< Content-Length: 795
< Set-Cookie: d02bf7bff81a220d14ef9d513c8d6521=cca6d168bebfd9d0691fea70b880d1a1; path=/; HttpOnly
< Cache-control: private
< 
{ [795 bytes data]
100   795  100   795    0     0   5129      0 --:--:-- --:--:-- --:--:--  5129
* Connection #1 to host koku-myproject.192.168.42.57.nip.io left intact


{
  "meta": {
    "count": 9,
    "key_only": false,
    "filter": {
      "time_scope_value": "-10",
      "time_scope_units": "day",
      "resolution": "daily"
    }
  },
  "links": {
    "first": "/api/v1/tags/openshift/infrastructures/aws/?limit=100&offset=0",
    "next": null,
    "previous": null,
    "last": "/api/v1/tags/openshift/infrastructures/aws/?limit=100&offset=0"
  },
  "data": [
    {
      "key": "c25_key",
      "values": [
        "c25_value"
      ]
    },
    {
      "key": "insights_project",
      "values": [
        "cost management"
      ]
    },
    {
      "key": "key1",
      "values": [
        "label_value1"
      ],
      "type": "pod"
    },
    {
      "key": "key2",
      "values": [
        "label_value2"
      ],
      "type": "pod"
    },
    {
      "key": "key3",
      "values": [
        "label_value3"
      ],
      "type": "storage"
    },
    {
      "key": "key4",
      "values": [
        "label_value4"
      ],
      "type": "storage"
    },
    {
      "key": "storageclass",
      "values": [
        "bravo"
      ],
      "type": "storage"
    },
    {
      "key": "vc_key2",
      "values": [
        "label_value2"
      ],
      "type": "storage"
    },
    {
      "key": "workload-type",
      "values": [
        "other"
      ]
    }
  ]
}

```


adberglund:
It's possible I am misunderstanding something, but it looks like this PR creates an API endpoint that would return all OCP and all AWS tags in one call. What we need is an endpoint that returns the subset of AWS tags that have been tag matched to an OpenShift cluster. The problem today is that we've been using the AWS tags endpoint, and some of those tags don't work for OCP on AWS data.

blentz:
@adberglund Okay. I'll make it do that.

koku/api/report/view.py
```diff
@@ -116,7 +119,13 @@ class ClassMapper(object):
                     'serializer': OCPAWSQueryParamSerializer,
                     'query_handler': OCPAWSReportQueryHandler,
                     'tag_handler': [AWSTagsSummary]
-                }
+                },
+                {
+                    'report': 'tags',
+                    'serializer': OCPAWSTagsQueryParamSerializer,
+                    'query_handler': OCPAWSTagQueryHandler,
+                    'tag_handler': [AWSTagsSummary, OCPUsagePodLabelSummary]
```
adberglund: Right now we only look at the filtered subset of AWSTags that matched, so the handler field can drop OCP Usage labels. 

```suggestion
                    'tag_handler': [AWSTagsSummary]
```

koku/api/tags/ocp_aws/queries.py
```diff
@@ -0,0 +1,39 @@
...
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+
+"""OCP-on-AWS Tag Query Handling."""
+from api.tags.aws.queries import AWSTagQueryHandler
+from api.tags.ocp.queries import OCPTagQueryHandler
+from reporting.models import OCPAWSCostLineItemDailySummary
+
+
+class OCPAWSTagQueryHandler(AWSTagQueryHandler, OCPTagQueryHandler):
```
adberglund: Same thing here just need to subclass the 1 tag handler.

```suggestion
class OCPAWSTagQueryHandler(AWSTagQueryHandler):
```

koku/api/tags/serializers.py
```diff
@@ -175,3 +175,8 @@ def validate_filter(self, value):
         """
         validate_field(self, 'filter', AWSFilterSerializer, value)
         return value
+
+
+class OCPAWSTagsQueryParamSerializer(AWSTagsQueryParamSerializer,
```
adberglund: Should only need to subclass the AWS Serializer.

```suggestion
class OCPAWSTagsQueryParamSerializer(AWSTagsQueryParamSerializer)
```

koku/api/tags/serializers.py
```diff
@@ -175,3 +175,8 @@ def validate_filter(self, value):
...
+                                     OCPTagsQueryParamSerializer):
```
adberglund:
```suggestion
```

koku/api/tags/serializers.py
blentz: That is incorrect. We need to subclass from both to get the union of serializer fields. Otherwise, we'd have no way to validate the fields across both AWS and OCP datasets.

koku/api/tags/ocp_aws/queries.py
blentz: Ultimately, the code calls all the way down to TagHandler for most things. So, today, inheriting from one of two classes that are subclasses of TagHandler is technically sufficient. You're correct.

But, if there are any future changes to either class, the OCP+AWS class will need that behavior, too. So, if we remove the inheritance we'll either A) need to remember the inheritance chain to know when there's a feature missing (likely to result in bugs) or B) need to add the multiple inheritance back in at that point in time.

I don't see that we gain anything other than an increased risk of bugs.

koku/api/tags/queries.py
```diff
@@ -41,8 +41,10 @@
 class TagQueryHandler(QueryHandler):
     """Handles tag queries and responses."""
 
+    _DEFAULT_ORDERING = {'tags': 'asc'}
+
     def __init__(self, query_parameters, url_data,
-                 tenant, data_sources, **kwargs):
+                 tenant, default_ordering=None, **kwargs):
```
dccurtis: Can we add back `data_sources` as a required parameter?  It's required for this class to do its main job; query the database for tags.

This code would require that the subclass author knows that it needs to set this list after instantiating super.  Sort of a lousy developer experience when they will inevitably forget to set it and need to dig into the parent's source code to understand why it's not working.

koku/api/tags/ocp_aws/views.py
```diff
@@ -0,0 +1,66 @@
...
+from api.report.view import _generic_report
+
+
+@api_view(http_method_names=['GET'])
+@permission_classes([AllowAny])
+@renderer_classes(tuple(api_settings.DEFAULT_RENDERER_CLASSES))
+def ocp_aws_tags(request):
+    """Get OpenShift-on-AWS tags.
+
+    @api {get} /cost-management/v1/tags/openshift/infrastructures/aws/
```
dccurtis: Can you also update `openapi.json`?

koku/api/tags/serializers.py
adberglund: After discussion, I agree that it's a good thing to have the ability to filter on AWS account or OpenShift project for this API, but the multiple inheritance situation is causing 

```
GET /local/v1/tags/openshift/infrastructures/aws/?filter[project]=1
HTTP 400 Bad Request
Allow: GET, OPTIONS
Vary: Accept

{
    "filter": {
        "project": [
        ]
    }
}
```

koku/api/tags/ocp_aws/queries.py
adberglund: Here too, I'm okay with multiple inheritance.

adberglund:
Unrelated to multiple inheritance, but while testing the two filter options, I noticed that account filtering doesn't work. It isn't registering my filter param as a filter.

```
GET /local/v1/tags/openshift/infrastructures/aws/?filter[account]=abc
HTTP 200 OK
Allow: GET, OPTIONS
Vary: Accept

{
    "meta": {
        "count": 4,
        "filter": {
            "time_scope_value": "-10",
            "time_scope_units": "day",
            "resolution": "daily"
        },
        "key_only": false
    },
    "links": {
        "first": "/local/v1/tags/openshift/infrastructures/aws/?filter%5Baccount%5D=abc&limit=100&offset=0",
        "next": null,
        "previous": null,
        "last": "/local/v1/tags/openshift/infrastructures/aws/?filter%5Baccount%5D=abc&limit=100&offset=0"
    },
    "data": [
        {
            "key": "app",
            "values": [
                "analytics",
                "catalog",
                "cost"
            ]
        },
        {
            "key": "environment",
            "values": [
                "dev"
            ]
        },
        {
            "key": "storageclass",
            "values": [
                "bravo",
                "charlie",
                "delta",
                "epsilon"
            ]
        },
        {
            "key": "version",
            "values": [
                "beta",
                "gamma",
                "master",
                "prod"
            ]
        }
    ]
}
```

koku/api/tags/queries.py
blentz: It's the same thing as DRF's serializers requiring you to set fields attributes on the class. Data sources is really a class attribute, not parameters.

But, there is definitely some tidying up I can do, to make it clearer how sub-classes need to be implemented.

adberglund: DRF's serializers are intentionally designed to be flexible and sub-classed to meet a wide array of use cases. Here we have a very specific class that we maintain and are the sole users of. If the param is required for the class to work, we should have the param. 

Please defer tidying up. Let's fix broken functionality and get this to a place where we can merge it. We have a full sprint on deck and need to move on.

blentz: The param is not required for the class to work. Thus, the change.

dccurtis: without setting `self.data_sorurces` then `get_tags()` will not do anything since it will be walking over an empty list.  To me that seems like it's required. 🤷‍♂

koku/api/tags/serializers.py
blentz: I've pushed a fix for this. It passes validation now:

```
COMMAND: curl -v -g -H 'HTTP_X_RH_IDENTITY: ImlkZW50aXR5Ijp7ImFjY291bnRfbnVtYmVyIjoiMTAwMDEiLCJ0eXBlIjoiVXNlciIsInVzZXIiOnsidXNlcm5hbWUiOiJ0ZXN0X2N1c3RvbWVyIiwiZW1haWwiOiJibGVudHoraGNjbS10ZXN0QHJlZGhhdC5jb20ifX0K' http://localhost:8000/api/v1/tags/openshift/infrastructures/aws/?filter[project]=abc
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0* Could not resolve host: ImlkZW50aXR5Ijp7ImFjY291bnRfbnVtYmVyIjoiMTAwMDEiLCJ0eXBlIjoiVXNlciIsInVzZXIiOnsidXNlcm5hbWUiOiJ0ZXN0X2N1c3RvbWVyIiwiZW1haWwiOiJibGVudHoraGNjbS10ZXN0QHJlZGhhdC5jb20ifX0K'
* Closing connection 0
curl: (6) Could not resolve host: ImlkZW50aXR5Ijp7ImFjY291bnRfbnVtYmVyIjoiMTAwMDEiLCJ0eXBlIjoiVXNlciIsInVzZXIiOnsidXNlcm5hbWUiOiJ0ZXN0X2N1c3RvbWVyIiwiZW1haWwiOiJibGVudHoraGNjbS10ZXN0QHJlZGhhdC5jb20ifX0K'
*   Trying ::1...
* TCP_NODELAY set
* connect to ::1 port 8000 failed: Connection refused
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 8000 (#1)
> GET /api/v1/tags/openshift/infrastructures/aws/?filter[project]=abc HTTP/1.1
> Host: localhost:8000
> User-Agent: curl/7.64.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Date: Thu, 30 May 2019 17:55:19 GMT
< Server: WSGIServer/0.2 CPython/3.7.3
< Vary: Accept, Origin
< Allow: GET, OPTIONS
< X-Frame-Options: SAMEORIGIN
< Content-Length: 378
< 
{ [378 bytes data]
100   378  100   378    0     0    818      0 --:--:-- --:--:-- --:--:--   818
* Connection #1 to host localhost left intact


{
  "meta": {
    "count": 0,
    "filter": {
      "project": [
        "abc"
      ],
      "time_scope_value": "-10",
      "time_scope_units": "day",
      "resolution": "daily"
    },
    "key_only": false
  },
  "links": {
    "first": "/api/v1/tags/openshift/infrastructures/aws/?filter%5Bproject%5D=abc&limit=100&offset=0",
    "next": null,
    "previous": null,
    "last": "/api/v1/tags/openshift/infrastructures/aws/?filter%5Bproject%5D=abc&limit=100&offset=0"
  },
  "data": []
}

```

koku/api/report/view.py
blentz: I've pushed this fix.

koku/api/tags/queries.py
blentz: It's not required *to be a parameter*. It's a required class attribute.

There is a difference.

The class needs the value, but external consumers of the class are setting that value by passing it in. It's model data defined by sub-classes. You don't need to pass it around from child to parent via __init__().

PR #877, opened 2019-05-31
872 metric map

adberglund:
## Summary
To support the price list (AKA cost model) flow we need to provide an API that lists the available metrics for a source type. This is a very simple GET/List API that shows available metrics. It has an optional `source_type` param that will filter the result set based on the source. This also shifts the ownership of allowed metrics to this new model and away from the Rate model, making this the "source of truth" for metrics.

## Example

```
GET /local/v1/metrics/?source_type=OCP
HTTP 200 OK
Allow: GET, HEAD, OPTIONS
Vary: Accept

{
    "meta": {
        "count": 6
    },
    "links": {
        "first": "/local/v1/metrics/?limit=10&offset=0&source_type=OCP",
        "next": null,
        "previous": null,
        "last": "/local/v1/metrics/?limit=10&offset=0&source_type=OCP"
    },
    "data": [
        {
            "source_type": "OpenShift Container Platform",
            "metric": "cpu_core_request_per_hour",
            "label_metric": "CPU",
            "label_measurement": "Request",
            "label_measurement_unit": "core-hours"
        },
        {
            "source_type": "OpenShift Container Platform",
            "metric": "cpu_core_usage_per_hour",
            "label_metric": "CPU",
            "label_measurement": "Usage",
            "label_measurement_unit": "core-hours"
        },
        {
            "source_type": "OpenShift Container Platform",
            "metric": "memory_gb_request_per_hour",
            "label_metric": "Memory",
            "label_measurement": "Request",
            "label_measurement_unit": "GB-hours"
        },
        {
            "source_type": "OpenShift Container Platform",
            "metric": "memory_gb_usage_per_hour",
            "label_metric": "Memory",
            "label_measurement": "Usage",
            "label_measurement_unit": "GB-hours"
        },
        {
            "source_type": "OpenShift Container Platform",
            "metric": "storage_gb_request_per_month",
            "label_metric": "Storage",
            "label_measurement": "Request",
            "label_measurement_unit": "GB-month"
        },
        {
            "source_type": "OpenShift Container Platform",
            "metric": "storage_gb_usage_per_month",
            "label_metric": "Storage",
            "label_measurement": "Usage",
            "label_measurement_unit": "GB-month"
        }
    ]
}
```


chambridge:
Providing an API here allows the UI to decouple its self as we add new metrics and capabilities. They just need to call the API  and populate the price list with the various metrics and doesn't need to be a two step process where the API add the feature and then the UI has to add the feature; its all just programatic.

I'd suggest an update to the naming though `ui_*` could be change to something like `label_*` and as we support translation these labels would return the appropriate translation.

dccurtis:
I think it makes sense for the backend to be in charge of what the UI gets with respect to database objects.  Think of it in terms of the MVC design pattern.  These metric names and types need to be 100% aligned to what our price list object will be (Model). And the API needs to be able to always present a consistent view of the world with respect to the price list objects and the names of things (Controller).  Allowing the View (UI) to go off on their own and do whatever they want with these model objects could get very difficult to manage.  For this case in particular the API will be supplying data to a simple drop down list, so I'm having a hard time understanding why we would want the UI to be involved any more then a simple GET operation

adberglund:
@blentz The backend should absolutely control the available metrics for price lists (cost models). The fact that we've added a few columns to give the UI a convenient name is a nicety. The database is a fine place to store things like that. If some sort of re-categorization were to happen, we'd need to know about it too. When do you see CPU/Memory/Storage and Request/Usage changing?

blentz:
@adberglund - 

The list of available metrics isn't the issue - gating which metrics are available is a consequence of other design decisions we've made. That's not the problem I'm trying to solve here.

I don't agree that the DB is the right place to store a set of friendly strings where the UI is driving the definition and organization. As said above, it creates a dependency that pushes the maintenance burden onto the API and away from the real stakeholder.

I've been in this situation before on other projects; it really isn't a good idea to create this dependency. The API should really stick to delivering data based on the metric names alone and leave any translation up to the UI.

I'm happy to discuss it further. If we need schedule a meeting to talk through it, let's do that.

blentz:
After some offline discussion, we've clarified some information I was missing. Now this PR makes more sense to me. I have no other objections. :smiley:

PR #903, opened 2019-06-11
Moving database code from masu

dccurtis:
This change moves the database access code from masu to koku.  The purpose of this is to get the git history established on these files before they are converted to DjangoORM and land in their final directory structure.

I put these into a `masu` directory simply to make it easy to integrate masu logic into koku and also could be a directory structure that could have the entire report ingest/processing logic be a standalone Django app.

I left out the `__init__.py` in the koku.masu.test.database package so that the tests are skipped.  Theres also a flake8 `masu` line to skip this package until integration is complete.


dccurtis:
@werwty awesome, thanks for confirming 👍 So I'm not going to squash commits when I merge, just in case it would plow over all the hard work of keeping the history 😂

Issue #885, opened 2019-06-06
Add Celery to Koku


adberglund:
## Objective
We should integrate celery into our Koku Django app so that the eventual addition of masu tasks are supported.

## Assumptions
- We can follow RBAC as a template (https://github.com/RedHatInsights/insights-rbac) 

## Acceptance Criteria

- [ ] A celery worker can be run from koku

lcouzens:
Verified

PR #904, opened 2019-06-11
Add celery to Koku

werwty:
https://github.com/project-koku/koku/issues/885

Running docker-compose up will now start a rabbitmq,
koku_worker, and koku_beat container.


koku/koku/settings.py
```diff
@@ -382,3 +382,9 @@
 MASU_BASE_URL = 'http://{}:{}/'.format(MASU_SERVICE_HOST, MASU_SERVICE_PORT)
 
 MASU_API_REPORT_DATA = 'api/v1/report_data/'
+
+# AMQP Message Broker
+RABBITMQ_HOST = os.getenv('RABBITMQ_HOST', 'rabbit')
```
dccurtis: could you make this default to`localhost`?  It should default to `rabbit` in the docker-compose

docker-compose.yml
```diff
@@ -57,6 +58,69 @@ services:
...
+      working_dir: /koku/koku
+      entrypoint: ['celery', '-A', 'koku', 'worker', '-l', 'info']
+      environment:
+        - DATABASE_ENGINE=postgresql
+        - DATABASE_NAME=${DATABASE_NAME-postgres}
+        - DATABASE_HOST=db
+        - DATABASE_PORT=5432
+        - DATABASE_USER=${DATABASE_USER-postgres}
+        - DATABASE_PASSWORD=${DATABASE_PASSWORD-postgres}
+        - RABBITMQ_HOST=rabbit
```
dccurtis: Just looked over at masu and this is how it is.  Ideally it would pick up the environment variable `RABBITMQ_HOST` and default to rabbit `${RABBITMQ_HOST-rabbit}`.  In fact all of the docker-compose environment variables should probably do this.  Not a big deal if you don't want to change it here.  Main thing is we shouldn't be using the docker-compose host names in the `koku/settings.py`

werwty: Should I also be adding the rabbitmq environment variables to .env.example?

dccurtis: My only only hesitation is that the default values of setting it in the .env example can get sort of confusing.  For example.  When using docker-compose the `RABBITMQ_HOST` needs to be `rabbit` if you are using the rabbit service from docker-compose (which I can imagine would be the majority when using our compose file).  But under any other context (local, OpenShift) it really doesn't make sense.  

I wanted to make the docker-compose deployment be as standalone as possible where it basically defaults to the right values for other services in the compose yaml.  Only thing people would have to add is the AWS secrets.

That being said we could force the user to set all environment variables and put the defaults in different .env example files.  Like `.env.example.docker`, `.env.example.openshift', `.env.example.local`.  So if you're going to make that change I think pulling all defaults into separate example files would make the most sense.  I can see pro/cons on both sides of this. :)

dccurtis:
I'm noticing some error messages saying `invalid port number: "None"` for the beat and worker service.  Any idea where this is coming from?  Guessing its something with `DJANGO_SETTINGS_MODULE`?

```
(bihan_fork) ➜  bihan_fork git:(885-celery-koku) ✗ docker-compose up koku-worker
koku_rabbit is up-to-date
Creating koku_worker ... done
Attaching to koku_worker
koku_worker    | [2019-06-12 00:57:01,786] INFO: Django setup.
koku_worker    | [2019-06-12 00:57:01,790] INFO: Celery autodiscover tasks.
koku_worker    | [2019-06-12 00:57:01,799] WARNING: invalid port number: "None"
koku_worker    |
koku_worker    | /opt/app-root/lib/python3.6/site-packages/celery/platforms.py:801: RuntimeWarning: You're running the worker with superuser privileges: this is
koku_worker    | absolutely not recommended!
koku_worker    |
koku_worker    | Please specify a different user using the --uid option.
koku_worker    |
koku_worker    | User information: uid=0 euid=0 gid=0 egid=0
koku_worker    |
koku_worker    |   uid=uid, euid=euid, gid=gid, egid=egid,
koku_worker    |
koku_worker    |  -------------- celery@koku_worker v4.3.0 (rhubarb)
koku_worker    | ---- **** -----
koku_worker    | --- * ***  * -- Linux-4.9.125-linuxkit-x86_64-with-centos-7.6.1810-Core 2019-06-12 00:57:02
koku_worker    | -- * - **** ---
koku_worker    | - ** ---------- [config]
koku_worker    | - ** ---------- .> app:         koku:0x7fa691ba7668
koku_worker    | - ** ---------- .> transport:   amqp://guest:**@rabbit:5672//
koku_worker    | - ** ---------- .> results:     disabled://
koku_worker    | - *** --- * --- .> concurrency: 4 (prefork)
koku_worker    | -- ******* ---- .> task events: OFF (enable -E to monitor tasks in this worker)
koku_worker    | --- ***** -----
koku_worker    |  -------------- [queues]
koku_worker    |                 .> celery           exchange=celery(direct) key=celery
koku_worker    |
koku_worker    |
koku_worker    | [tasks]
koku_worker    |
koku_worker    |
koku_worker    | [2019-06-12 00:57:02,751: INFO/MainProcess] Connected to amqp://guest:**@rabbit:5672//
koku_worker    | Connected to amqp://guest:**@rabbit:5672//
koku_worker    | [2019-06-12 00:57:02,770: INFO/MainProcess] mingle: searching for neighbors
koku_worker    | mingle: searching for neighbors
koku_worker    | [2019-06-12 00:57:03,803: INFO/MainProcess] mingle: all alone
koku_worker    | mingle: all alone
koku_worker    | [2019-06-12 00:57:03,804] WARNING: invalid port number: "None"
koku_worker    |
koku_worker    | [2019-06-12 00:57:03,804: WARNING/MainProcess] invalid port number: "None"
koku_worker    |
koku_worker    | invalid port number: "None"
koku_worker    |
koku_worker    | [2019-06-12 00:57:03,831: INFO/MainProcess] celery@koku_worker ready.
koku_worker    | celery@koku_worker ready.
koku_worker    | [2019-06-12 00:57:05,809] WARNING: invalid port number: "None"
koku_worker    |
koku_worker    | [2019-06-12 00:57:05,809: WARNING/MainProcess] invalid port number: "None"
koku_worker    |
koku_worker    | invalid port number: "None"
koku_worker    |
koku_worker    | [2019-06-12 00:57:07,813] ERROR: Query failed to return results.
koku_worker    | [2019-06-12 00:57:07,813: ERROR/MainProcess] Query failed to return results.
koku_worker    | Query failed to return results.
koku_worker    | [2019-06-12 00:59:07,773] WARNING: invalid port number: "None"
koku_worker    |
koku_worker    | [2019-06-12 00:59:07,773: WARNING/MainProcess] invalid port number: "None"
koku_worker    |
koku_worker    | invalid port number: "None"
koku_worker    |
koku_worker    | [2019-06-12 00:59:09,778] WARNING: invalid port number: "None"
koku_worker    |
koku_worker    | [2019-06-12 00:59:09,778: WARNING/MainProcess] invalid port number: "None"
koku_worker    |
koku_worker    | invalid port number: "None"
koku_worker    |
koku_worker    | [2019-06-12 00:59:11,784] WARNING: invalid port number: "None"
koku_worker    |
koku_worker    | [2019-06-12 00:59:11,784: WARNING/MainProcess] invalid port number: "None"
koku_worker    |
koku_worker    | invalid port number: "None"
koku_worker    |
koku_worker    | [2019-06-12 00:59:13,787] ERROR: Query failed to return results.
koku_worker    | [2019-06-12 00:59:13,787: ERROR/MainProcess] Query failed to return results.
koku_worker    | Query failed to return results.
```

werwty:
Good catch. I had `DATABASE_NAME` and `DATABASE_PORT` in the docker-compose env instead of `POSTGRES_SQL_SERVICE_PORT` and `POSTGRES_SQL_SERVICE_HOST`. The tasks still work in that case, but Django is definitely not happy. 

Fixed in latest change.

docker-compose.yml
werwty: I'll not touch the `.env` file in that case. The defaults in the docker-compose environment are reasonable. And if there is a need to change the defaults settings in it often, we can pull out the env variables into a separate file then.

dccurtis:
Awesome thanks. Looks good!

PR #906, opened 2019-06-13
Openshift 4 provider documentation update

adberglund:
## Summary 
This is a documentation update to reflect changes for Openshift 4 and the most recent Operator Metering version. We now access the operator metering API through an OpenShift route. 

I apologize for the whitespace changes caused automatically by my editor, there are several lines in the diff that are not actual changes.

### DIFF
--- a/docs/source/providers/ocp.rst
+++ b/docs/source/providers/ocp.rst
@@ -1,12 +1,12 @@
 Adding an OCP Provider
 #######################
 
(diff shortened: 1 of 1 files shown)


docs/source/providers/ocp.rst
```diff
@@ -1,15 +1,16 @@
...
 
 Dependencies
 ************
 
-We require the following dependencies in order to obtain the OCP operator metering usage data. 
+We require the following dependencies in order to obtain the OCP operator metering usage data.
 
 - You must be running OCP version 3.11 or newer.
 - You must `install Operator Metering <https://github.com/operator-framework/operator-metering/blob/master/Documentation/install-metering.md>`_ on your OCP cluster.
+- The Operator Metering installation must `expose a route <https://github.com/operator-framework/operator-metering/blob/master/Documentation/configuring-reporting-operator.md#openshift-route>`_ for the reporting operator.
```
adberglund: This step is part of the operator metering install process, so I don't know if it actually belongs in the same bullet as installing Operator Metering?

dayleparker: I think it can go either way -- having it a separate bullet highlights it a bit more as a prerequisite, so I vote to keep it the way you wrote it.

Issue #887, opened 2019-06-06
Add Masu Database Accessors to Koku and Integrate Django ORM


adberglund:
## Objective
We want to move the database accessor layer over to Koku in preparation for moving the rest of the masu code base.

## Assumptions
- The functionality of the accessors will remain unchanged
- The database accessors will use Django ORM instead of SQLAlchemy

## Acceptance Criteria

- [ ] Database accessor unit tests pass in koku

blentz:
Note: The openshift templates are still pointing at the masu git repo. The git URL in the templates needs to be updated in tandem with this work.

adberglund:
@lcouzens This is just part of the general "make sure everything passes" testing for koku/masu merge.

lcouzens:
Verified

PR #917, opened 2019-06-28
Djangoify Masu DB accessors.

werwty:
https://github.com/project-koku/koku/issues/887

Things left to do:

- [x]  Fix ~20 failing tests
- [x] Update docstrings
- [x] Clean up lint errors
- [x] Document changes to public methods

DEPRECATED METHODS:
- `koku.masu.database.engine.create_engine()`  (DB Engine creation is no longer necessary)
- `koku.masu.database.koku_database_access.close_session()` Sessions to connect to the db is no longer necessary, since everything lives in a single django session.
- `koku.masu.database.koku_database_access.get_base()` 
- `koku.masu.database.koku_database_access.get_engine()` 
- `koku.masu.database.koku_database_access.get_meta()` 
- `koku.masu.database.report_stats_db_accessor.get_cursor_position()` 
- `koku.masu.database.report_stats_db_accessor.set_cursor_position()` 
- `koku.masu.database.koku_database_access.add()` In the previous implementation this created an object but didn't save it. In the new implementation the object is saved to the database. It doesn't make sense to have `use_savepoint` if we don't save the object. I can change this to only create a python instance of the object, but that requires more code to save it, ex:
  ```
  obj = accessor.add(**kwargs)
  obj.save()
  ```
  The old pattern of 
  ```
  accessor.add(**kwargs)
  accessor.commit()
  ```
  would not work since we're creating a python instance of the model. An alternative would be to use Django's transactions, disable autocommit, and make it so commit has to be called manually. Let me know if that pattern is preferable.


blentz:
For context, the `use_savepoint` stuff was added here: https://github.com/project-koku/masu/pull/412

It was implemented to handle a specific issue we were seeing during bulk updates. The goal was to allow bulk updates to insert/update as many rows as possible, catch and log any rows that failed to insert/update without causing a fatal error that resulted in the entire bulk update not updating any rows.

I'm pretty sure Django's ORM has a better way to handle that. So, that behavior does not need to be preserved so long as we're handling that case some other way.

Pipfile
```diff
@@ -37,6 +38,8 @@ coverage = ">=4.5.0"
 codecov = ">=2.0.0"
 faker = ">=0.8.14"
 pylint-django = "==2.0.9"
+pydocstyle = "~=3.0.0"
```
werwty: Open issue for this: https://gitlab.com/pycqa/flake8-docstrings/issues/36
I'll revert this change if the issue is resolved before this PR gets merged.

koku/masu/database/provider_billing_source_db_accessor.py
```diff
@@ -54,11 +54,11 @@ def get_uuid(self):
         Args:
             None
         Returns:
-            (String): "UUID v4",
+            (UUID): "UUID v4",
```
dccurtis: I believe these public functions returning UUID are still returning a String, correct?

koku/masu/test/__init__.py
```diff
@@ -1 +1,73 @@
...
+                                                   schema_name=cls.schema)
+            cls.tenant = Tenant(schema_name=cls.schema)
+            cls.tenant.save()
+
+        cls.ocp_test_provider_uuid = '3c6e687e-1a09-4a05-970c-2ccf44b0952e'
+        cls.aws_test_provider_uuid = '6e212746-484a-40cd-bba0-09a19d132d64'
+        cls.aws_provider_resource_name = 'arn:aws:iam::111111111111:role/CostManagement'
+        cls.ocp_provider_resource_name = 'my-ocp-cluster-1'
+        cls.aws_test_billing_source = 'test-bucket'
+        cls.ocp_test_billing_source = 'test-bucket'
```
dccurtis: This should be `None` since OCP doesn't use billing source

koku/masu/test/database/test_ocp_rate_db_accessor.py
```diff
@@ -48,12 +54,12 @@ def setUpClass(cls):
...
 
     def setUp(self):
-        """"Set up a test with database objects."""
+        """Set up a test with database objects."""
         super().setUp()
-        if self.accessor._conn.closed:
-            self.accessor._conn = self.accessor._db.connect()
-        if self.accessor._pg2_conn.closed:
-            self.accessor._pg2_conn = self.accessor._get_psycopg2_connection()
+        # if self.accessor._conn.closed:
```
dccurtis: Remove these?

koku/masu/database/provider_billing_source_db_accessor.py
werwty: Oops, no these are returning uuids now. But you're right, these should continue to return Strings since this is a public function. Fixed~

PR #934, opened 2019-07-21
Filter cost models by name or source type

### LINKED ISSUE #932: Filter cost models by name or source type
## User Story
As a user I'd like to filter cost models by name or source type so that I can quickly and efficiently find what I'm looking for.

## Assumptions
- This will add some filter params to the GET List APIs to return a subset of data

## API Details

| API  | Detail|
 | --- | --- |
 | **Method** | GET |
 | **Path** | /api/v1/costmodels/ |

## Acceptance Criteria

- [ ] Filtering on source type returns cost models for the specific type.
- [ ] Filtering on name does case insensitive partial string matching to return matching cost models

boaz0:
closes #932


boaz0:
Thanks @adberglund

PR #978, opened 2019-08-09
Model changes required for Azure

blentz:
After some discussion with @adberglund and the rest of the team, we've decided on the following implementation plan for Azure.

1. Add new JSON fields for storing authentication and billing-bucket information. Azure requires more data than AWS for just about any API query. Future clouds are also likely to do their own thing. So, a JSONField will enable us to remain flexible to support any cloud's requirements.
2. Implement Azure provider using the new fields.
3. At a future date, we'll migrate the AWS, OCP, and OCP+AWS providers to using the new fields.
4. We'll remove the old fields once all providers are using the new JSONFields.

This PR is for the model changes described above. There will be future PRs for the rest of the Azure provider implementation.


koku/api/provider/serializers.py
```diff
@@ -45,28 +44,64 @@ class ProviderAuthenticationSerializer(serializers.ModelSerializer):
...
+    def to_internal_value(self, data):
+        """Convert incoming data to a native value.
+
+        This method is being overridden to handle optional/deprecated fields.
+        """
+        if 'credentials' not in data:
+            data['credentials'] = {}
+
+        if 'provider_resource_name' not in data:
+            data['provider_resource_name'] = 'None'
```
adberglund: Instead of storing the string "None" in the database, I think we should use a check constraint that requires either provider_resource_name or credentials to be null. This would require making both fields nullable, but the check constraint would ensure 1 of them is not null. See: https://docs.djangoproject.com/en/2.2/ref/models/constraints/#checkconstraint

koku/api/provider/serializers.py
```diff
@@ -45,28 +44,64 @@ class ProviderAuthenticationSerializer(serializers.ModelSerializer):
...
+    def to_internal_value(self, data):
+        """Convert incoming data to a native value.
+
+        This method is being overridden to handle optional/deprecated fields.
+        """
+        if 'data_source' not in data:
+            data['data_source'] = {}
+
+        if 'bucket' not in data:
+            data['bucket'] = 'None'
```
adberglund: Same check constraint suggestion here.

koku/api/provider/serializers.py
blentz: Sure thing. That sounds like a much better way to achieve the same goal. Thanks for the suggestion. ^_^

Issue #889, opened 2019-06-06
Update Docker Compose workflow for Masu in Koku


adberglund:
## Objective
We want to update docker-compose files so that development environments that use Docker are synced with the koku/masu merge.

## Acceptance Criteria

- [ ] Docker Compose spins up what it needs to

lcouzens:
Verified

Issue #888, opened 2019-06-06
Move masu code to Django app within koku


adberglund:
## Objective
Once the database access layer is in place in koku, we should move the rest of the masu code base over to a Django "app" inside the koku application. 

## Assumptions
- With a functioning database access layer, this should "just work"

## Acceptance Criteria

- [ ] All masu unit tests work
- [ ] All celery tasks function 
- [ ] The normal flow for data processing proceeds without error

adberglund:
We want to keep our masu git history as well if possible. Reference https://gbayer.com/development/moving-files-from-one-git-repository-to-another-preserving-history/ if this is possbile?

blentz:
Note: The openshift templates are still pointing at the masu git repo. The git URL in the templates needs to be updated in tandem with this work.

werwty:
Branch here with most of the processor tests fixed: https://github.com/werwty/koku/tree/888-masu-mv

I'm putting this back onto the backlog since I will be PTO until 8/2.

lcouzens:
Verified

PR #989, opened 2019-08-12
Enable masu ingest/processing in koku

dccurtis:
Addresses #888 #889

Enabling masu ingest and processing logic in Koku repo.

**Testing**
1. Generate Nise OCP payload
2. Start Koku and connect koku-listener to upload service network
3.  Add OCP Provider
4. Upload OCP payload to upload service.
5. Verify OCP processing is complete.
6.  Add AWS provider.
7. Ingest data with masu-server /download
8. Verify AWS processing is complete.


koku/masu/api/views.py
```diff
@@ -1,5 +1,5 @@
 #
-# Copyright 2018 Red Hat, Inc.
+# Copyright 2019 Red Hat, Inc.
```
blentz: Please don't modify the Copyright stanza. It is not something that gets updated after it has been written.

It needs to reflect the *first* year the file was written, not the current year.

koku/masu/database/report_db_accessor_base.py
```diff
@@ -115,11 +113,13 @@ def create_new_temp_table(self, table_name, columns):
...
         table_sql = base_sql + column_sql
-        self._cursor.execute(table_sql)
+        with connection.cursor() as cursor:
+            cursor.db.set_schema(self.schema)
+            cursor.execute(table_sql)
 
         return temp_table_name
 
-    # pylint: disable=too-many-arguments
+    # pylint: disable=too-many-arguments,too-many-locals
```
blentz: Instead of disabling the `too-many-locals` check, let's see if we can reduce the number of local variables in this function.

I can already spot a couple of places where a local variable is being assigned when it's not needed.

koku/masu/api/views.py
dccurtis: This is a new file, so I think 2019 is ok here. Right?  Looks like GitHub saw this as a file name change from worker.py to views.py?

dccurtis:
@blentz As for OpenShift templates.  @adberglund is close on those.   Planning on a separate PR for that one 👍

koku/masu/api/views.py
blentz: If it's a new file, why is there a diff of changed lines?

dccurtis: Yeah I really can't explain why it's showing up like this, but bottom line is that this file is new.  All related to getting views setup in Django

PR #992, opened 2019-08-13
Pip pipenv fixes

adberglund:
## Summary
This removes the hard requirement for pip and pipenv versions. This also removes the psycopg2-binary requirement. The non-binary requirement is in our Pipfile.lock regardless, I believe because Django uses it. Switched to using Django's cursor instead of a raw psycopg2 cursor to keep things in Django. 

This also removes the `--sequential` flag on pipenv installs with instructions for Mac OSX users on how to get a working ssl C library for building the psycopg2 wheel. 

## Testing
A koku build passed in minishift with the pip and pipenv requirements stripped. 
Tox ran without the sequential flag with the readme update applied.


Makefile
```diff
@@ -252,7 +252,7 @@ oc-create-imagestream:
 	$(OC_PARAMS) $(MAKE) __oc-create-object
 
 oc-create-koku-api: OC_OBJECT := 'bc/$(NAME) dc/$(NAME)'
-oc-create-koku-api: OC_PARAMETER_FILE := $(NAME).env
+oc-create-koku-api: OC_PARAMETER_FILE := $(NAME)-api.env
```
adberglund:
```suggestion
oc-create-koku-api: OC_PARAMETER_FILE := $(NAME).env
```

PR #995, opened 2019-08-13
Migrationless test runner

adberglund:
## Summary
This subclasses the Django DiscoveryRunner and creates our tenant at database creation time, so it's already available for each test class. This means we don't have to run migrations for any tests that don't explicitly create a new tenant as part of the test. Should speed tests up some.

### DIFF
--- a/koku/api/iam/test/iam_test_case.py
+++ b/koku/api/iam/test/iam_test_case.py
@@ -23,9 +23,9 @@ from django.db import connection
 from django.test import TestCase
 from faker import Faker
 
-from ..models import Customer, Tenant
-from ..serializers import create_schema_name
-from ...common import RH_IDENTITY_HEADER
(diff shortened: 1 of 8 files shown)


koku/api/iam/test/iam_test_case.py
```diff
@@ -45,22 +46,22 @@ def setUpClass(cls):
...
+        cls.tenant = Tenant.objects.get_or_create(schema_name=cls.schema_name)[0]
         cls.tenant.save()
         cls.headers = cls.request_context['request'].META
 
     @classmethod
     def tearDownClass(cls):
         """Tear down the class."""
         connection.set_schema_to_public()
-        cls.tenant.delete()
+        # cls.tenant.delete()
```
adberglund:
```suggestion
```

Issue #981, opened 2019-08-09
API to create and check status of data dump requests


infinitewarp:
## User Story
As a Koku API user, I want an API where I can request a data dump and check on the status of my previous data dump requests.

## Assumptions and Questions
- User can create and view requests but not delete or update them.
- API is a standard DRF create-list-read endpoint.
- User can specify start and end dates for limiting the data.
  - Is start date inclusive and end date exclusive? TBD in a later story.
- Gathering and sending of the data to S3 will be implemented in a future story. For now, this is just the initial infrastructure for accepting and tracking the request.
- Check with other Koku devs and APIs to ensure consistency with all attribute names and behaviors.
- Should a superuser be able to get all dump requests? What happens if a superuser creates a dump request?
  - Not today. For now, only limit the API to the logged-in requesting user.

## Acceptance Criteria

- [ ] API is served at `/v1/dataexportrequests/`
  - [ ] post to create, get to list, get with uuid to fetch individual request
  - [ ] list uses standard pagination
  - [ ] user limited
- [ ] required post params are:
  - `start_date` with ISO-8601 formatted date string
  - `end_date` with ISO-8601 formatted date string
    - must be greater than `start_date` (if `end_date` is exclusive)
- [ ] returned responses include attributes:
  - `uuid`  (automatically set upon creation)
  - `start_date`
  - `end_date`
  - `status`
  - `created_timestamp` (automatically set upon creation)
  - `updated_timestamp` (automatically set upon creation or update)
- [ ] The requesting user ID is automatically stored internally with the request so the future data collection can filter appropriately based on that user's access (RBAC).
- [ ] The datetime at which the request was created is also stored in the model.
- [ ] The `status` attribute can be one of the following strings, although only `pending` is normally visible right now:
  - `pending` (the initial state upon creation)
  - `complete`
- [ ] API returns 400 error upon post/creation:
  - [ ] if `start_date` and `end_date` are not both defined.
  - [ ] if `start_date` and `end_date` are not both ISO-8601 formatted date strings.
  - [ ] if `end_date` is not greater than `start_date`.
  - [ ] with details in response explaining _all_ problematic inputs (e.g. if both `start_date` and `end_date` are malformed, the response indicates both of them).

infinitewarp:
Future possible enhancement not in this issue: add the customer's target S3 bucket name to the model and API.

dchorvat1:
verified

PR #998, opened 2019-08-14
create model, serializer, and view for user-initiated data dump requests

infinitewarp:
for https://github.com/project-koku/koku/issues/981


koku/api/dump/views.py
```diff
@@ -0,0 +1,29 @@
...
+    mixins.CreateModelMixin,
+    mixins.RetrieveModelMixin,
+    mixins.ListModelMixin,
+    viewsets.GenericViewSet,
+):
+    """Create, retrieve, or list dump requests."""
+
+    queryset = DumpRequest.objects.all()
+    serializer_class = DumpRequestSerializer
+    permission_classes = (AllowAny,)
```
adberglund: This is probably not something we'll want `AllowAny` on since it is customer data. Since we have RBAC around who can view things in the API, we should have RBAC for this.

koku/api/dump/models.py
```diff
@@ -0,0 +1,48 @@
+"""Models for user-initiated data dump requests."""
+
+from uuid import uuid4
+
+from django.db import models
+
+
+class DumpRequest(models.Model):
```
adberglund: It's rare that I comment on naming, but maybe we can call all everything related to pulling raw data out of the database a DataExport instead of Dump. I think the user experience there is a little better.

koku/api/dump/models.py
```diff
@@ -0,0 +1,48 @@
...
+    """A user's request for a data dump."""
+
+    PENDING = 'pending'
+    COMPLETE = 'complete'
+    STATUS_CHOICES = ((PENDING, 'Pending'), (COMPLETE, 'Complete'))
+
+    uuid = models.UUIDField(default=uuid4, editable=False, unique=True, null=False)
+    created_by = models.ForeignKey('User', null=False, on_delete=models.CASCADE)
+    created_at = models.DateTimeField(auto_now_add=True)
+    updated_at = models.DateTimeField(auto_now=True)
```
adberglund: Just for consistency with other database models let's use `created_timestamp` and `updated_timestamp`.

koku/api/dump/models.py
```diff
@@ -0,0 +1,48 @@
...
+
+from django.db import models
+
+
+class DumpRequest(models.Model):
+    """A user's request for a data dump."""
+
+    PENDING = 'pending'
+    COMPLETE = 'complete'
+    STATUS_CHOICES = ((PENDING, 'Pending'), (COMPLETE, 'Complete'))
```
blentz: Is there some reason three local variables are being assigned here? 

I don't see any use of `PENDING` or `COMPLETE` beyond its use in `STATUS_CHOICES`. So, I think that can be collapsed into just a single variable.

infinitewarp: This is Django's recommended practice for defining choices. See: https://docs.djangoproject.com/en/2.2/ref/models/fields/#choices

> Generally, it’s best to define choices inside a model class, and to define a suitably-named constant for each value:

Also, they will likely be used in future commits as this feature grows out, and when they are, referencing a class variable will better than proliferating strings.

koku/api/dump/models.py
infinitewarp: "dump" has been the term of choice since we were originally talking about just getting data via `pg_dump`, but I understand the hesitation on language…

💩

koku/api/dump/views.py
infinitewarp: This is an area of the system that I've only started trying to figure out. What specific `permissions_classes` do we currently have that you think would make sense here? I saw `OpenShiftAccessPermission` and `AWSAccessPermission`, but they don't look like the right fit. Should I make a new permission class?

(I may be totally wrong, though, because I don't yet understand those classes and their use cases fully.)

FWIW, `DumpRequestSerializer.create` sets the currently authenticated user on the model and `DumpRequestViewSet.get_serializer` limits the results to the currently authenticated user making the request. Keep in mind also that there is no actual customer cost management data accessible through _this_ model+serializer+view. The `DumpRequest` (rename pending) only acts like a request ticket for an export to happen, and the user can check back later to see the state of the export process. I would expect the async export process itself (not part of this PR) to do appropriate RBAC filtering for its various data collection and filtering operations.

koku/api/dump/models.py
infinitewarp: The classes have been renamed in the latest push to `DataExportRequest`*.

koku/api/dump/models.py
infinitewarp: Updated to match in the latest push.

infinitewarp:
The latest push just rebases over the latest `master` and adds the new API definitions to the `openapi.json` file.

koku/api/dataexport/models.py
```diff
@@ -0,0 +1,59 @@
...
+
+class DataExportRequest(models.Model):
+    """A user's request for a data export."""
+
+    PENDING = 'pending'
+    COMPLETE = 'complete'
+    STATUS_CHOICES = ((PENDING, 'Pending'), (COMPLETE, 'Complete'))
+
+    uuid = models.UUIDField(default=uuid4, editable=False, unique=True, null=False)
+    created_by = models.ForeignKey('User', null=False, on_delete=models.CASCADE)
```
adberglund: Definitely something that can (if needed) be iterated on in future PRs, but we typically associate things to a customer/provider too so we a) know which tenant to pull the data from and b) which data in the tenant needs pulling. User is obviously useful because we get an email address out of it, but we may end up needing more as this is implemented.

infinitewarp: Sounds good! I'll hold onto that suggestion for future iterations. 👍

PR #1013, opened 2019-08-16
remove apidoc & dependencies

### LINKED ISSUE #1003: Remove api-doc
## Objective
We have landed on openapi as our official API documentation scheme. We can remove api-doc generation to remove a dependency on node and reduce build time. 

## Acceptance Criteria

- [ ] Koku build no longer makes api-doc
- [ ] Documentation is updated

blentz:
This fixes #1003 .

Removing apidoc and its dependencies.

This PR includes these main changes:
- remove apidoc references
- change openshift buildstrategy to sourceStrategy
- remove Dockerfile and s2i scripts
- refactored handling of staticfiles to be a bit simpler


Dockerfile-env
```diff
@@ -1,10 +0,0 @@
-FROM centos/python-36-centos7
```
adberglund: We need to keep Dockerfile-env for the docker-compose dev flow.

blentz: Okay. 

Just be aware that this means the docker-compose workflow users are now responsible for tracking any changes to the Dockerfile made upstream now that the OpenShift workflow no longer uses a custom Dockerfile.

blentz: Fixed in a730244

Pipfile.lock
```diff
@@ -985,6 +985,40 @@
             ],
             "version": "==2.8.3"
         },
+        "psycopg2-binary": {
```
adberglund: This should have recently been removed as a dependency. The binary is nice, but other things already require non binary psycopg2, so I removed it from the Pipfile.

blentz: Okay. I'll pull it back out.

I did a `pipenv clean` recently and saw it get removed, but thought it was an error. Glad to know it's not. :smiley:

Issue #957, opened 2019-08-01
Download Azure Cost Export (cost and usage report)


adberglund:
## Objective
We want to successfully download an Azure report.

## Impacts
Ex. API, UI, Docs

## Assumptions
-We will need to generate a concept of a manifest for Azure.
-

## API Details

| API  | Detail|
 | --- | --- |
 | **Method** | GET/POST/PUT/PATCH/DELETE |
 | **Path** | /api/v1/?/ |
 | **Body** |  |
 | **Response** | |

## Acceptance Criteria

- [ ]
- [ ]

lcouzens:
Verified

PR #1014, opened 2019-08-16
Azure Report Downloader

dccurtis:
Addresses #957 

Major additions
* Azure report downloader
* Masu changes needed to support https://github.com/project-koku/koku/pull/978
* Azure report processing stubs.  Added to make sure the downloader was returning values that conforms to the rest of the processing pipeline.

**Testing**
1.  Add Azure Provider:
2.  Trigger download and verify files were downloaded

**Test Results**
[azure_downloader_ut.txt](https://github.com/project-koku/koku/files/3511208/azure_downloader_ut.txt)


dccurtis:
Hat tip to @chambridge for the jump start on the `AzureSevice()` part of this. 👍

Pipfile
```diff
@@ -32,6 +32,9 @@ python-dateutil = "*"
 bs4 = "*"
 azure = "==4.0.0"
 azure-mgmt-costmanagement = "==0.1.0"
+isodate = "==0.6.0"
+azure-storage-blob = "*"
+azure-storage = "*"
```
blentz: How do you feel about unlocked versus version-locked dependencies?

I could go either way, but seeing as most everything else is version-locked, we may just want all new deps to be locked as well.

koku/masu/external/accounts/db/cur_accounts_db.py
```diff
@@ -22,7 +22,26 @@
 
 # pylint: disable=too-few-public-methods
 class CURAccountsDB(CURAccountsInterface):
-    """Provider interface defnition."""
+    """Provider interface definition."""
+
+    @staticmethod
+    def get_authentication(provider):
+        """Return either provider_resource_name or credentials."""
+        if provider.authentication.provider_resource_name:
```
blentz: For tracking purposes, it might be good to include a comment or extra info in the docstring. This is only conditional because the `provider_resource_name` is deprecated. That will make it easier to find these sections of code when we go to remove the deprecated bits.

koku/masu/external/accounts/db/cur_accounts_db.py
```diff
@@ -22,7 +22,26 @@
...
+            return provider.authentication.provider_resource_name
+        elif provider.authentication.credentials:
+            return provider.authentication.credentials
+        return None
+
+    @staticmethod
+    def get_billing_source(provider):
+        """Return either bucket or data_source."""
+        if provider.billing_source:
+            if provider.billing_source.bucket:
```
blentz: Same as my comment about `get_authentication()` - might be good to have a comment about the deprecation.

koku/masu/external/downloader/azure/azure_service.py
```diff
@@ -0,0 +1,96 @@
...
+
+        if not self._factory.credentials:
+            raise ValueError('Azure Service credentials are not configured.')
+
+    def get_cost_export_for_key(self, key, container_name):
+        """Get the latest cost export file from given storage account container."""
+        report = None
+        cloud_storage_account = self._cloud_storage_account
+        blockblob_service = cloud_storage_account.create_block_blob_service()
+        blob_list = blockblob_service.list_blobs(container_name)
```
blentz: Should the calls to `create_block_blob_service()` or `list_blobs` be wrapped in a try/except block?

They're remote API calls that could raise an `AzureException`. Is that going to be caught anywhere?

koku/masu/external/downloader/azure/azure_service.py
```diff
@@ -0,0 +1,96 @@
...
+        for blob in blob_list:
+            if key == blob.name:
+                report = blob
+                break
```
blentz: This for/if loop could be simplified using `map()` and/or `filter()`

Pipfile
adberglund: If we lock, pyup should just keep advancing the version number anyway, so they're effectively equal.

blentz: @adberglund  - The difference, I think, is that pyup won't do major-version upgrades. It only does incremental upgrades where `*` will accept any version, including a major update.

dccurtis: I meant to come back later and lock these.  Thanks for the reminder!

koku/masu/external/downloader/azure/azure_report_downloader.py
```diff
@@ -0,0 +1,241 @@
...
+    @staticmethod
+    def _get_azure_client(credentials, billing_source):
+        subscription_id = credentials.get('subscription_id')
+        tenant_id = credentials.get('tenant_id')
+        client_id = credentials.get('client_id')
+        client_secret = credentials.get('client_secret')
+        resource_group_name = billing_source.get('resource_group')
+        storage_account_name = billing_source.get('storage_account')
+
+        service = AzureService(subscription_id, tenant_id, client_id, client_secret,
```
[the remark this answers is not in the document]
dccurtis: Yes, I was planning on adding more error checking here so we can get useful errors whenever any of these fields is missing.  I'll come back to this when I do the Azure-local downloader.

koku/masu/external/downloader/azure/azure_service.py
dccurtis: yeah thats a good point.  I'll add that

Issue #974, opened 2019-08-06
Openapi spec file missing some of the endpoints


dchorvat1:
**Describe the bug**
openapi.json is missing `/metrics/` and `/costmodels/` (there's still `/rates/`) endpoints, so we can't reach them using the auto generated api client

**To Reproduce**
1. Go to /api/cost-management/v1/openapi.json
2. See error

**Expected behavior**
All endpoints specification

**Additional context**
commit afc618f654ae481af10e77e88f70164c0e9e042e

adberglund:
Might be a good idea to rename the cost models metrics endpoint name to not be confused with Prometheus metrics.

dchorvat1:
@werwty In the current spec, there is no `/metrics/` endpoint, could we also get that one?

werwty:
@dchorvat1 +1 on it

werwty:
https://github.com/project-koku/koku/pull/1107 has been merged.

dchorvat1:
Thanks @werwty, the only thing that the generated client was complaining about was a missing slash in `/metrics/`. I put out a PR for that (https://github.com/project-koku/koku/pull/1123), can you review so it can be merged?

PR #1019, opened 2019-08-20
Add cost models to openapi.json.

werwty:
- add /costmodels endpoint and remove /rates endpoints

https://github.com/project-koku/koku/issues/974


werwty:
Demo that the CLI generated from this spec works: https://asciinema.org/a/4OEwPb4R2UzE6gUdQmighTzfX

docs/source/specs/openapi.json
```diff
@@ -705,31 +720,41 @@
...
+                                    "$ref": "#/components/schemas/Error"
+                                }
+                            }
+                        }
+                    },
                     "404": {
                         "description": "Not Found",
                         "content": {
-                            "application/json": {
+                            "*/*": {
```
infinitewarp: Do we ever (under normal circumstances) return a response without json?

werwty: I don't _think_ so. Koku's exception handler encodes errors as dicts: https://github.com/project-koku/koku/blob/master/koku/api/common/exception_handler.py#L67

And by default uses the json renderer: https://github.com/project-koku/koku/blob/master/koku/koku/settings.py#L260

The content type for errors in the openapi.json is kinda inconsistent. It's mostly `application/json` but there's a couple of `*/*`. I'm not sure if that's an oversight, or if I'm missing something...

adberglund: I think it would be safe to go with `application/json`

werwty: It's `application/json`
```
> GET /api/cost-management/v1/costmodels/34b2b968-6181-4104-b183-6d47d8243e95/ HTTP/1.1
> Host: 0.0.0.0:8000
> User-Agent: curl/7.61.1
> Accept: */*
> 
< HTTP/1.1 404 Not Found
< Date: Tue, 20 Aug 2019 18:33:49 GMT
< Server: WSGIServer/0.2 CPython/3.6.3
< Vary: Accept, Origin
< Allow: GET, PUT, PATCH, DELETE, HEAD, OPTIONS
< X-Frame-Options: SAMEORIGIN
< Content-Length: 67
< 
{ [67 bytes data]
100    67  100    67    0     0   3722      0 --:--:-- --:--:-- --:--:--  3722
* Connection #0 to host 0.0.0.0 left intact
{
  "errors": [
    {
      "detail": "Not found.",
      "source": "detail",
      "status": 404
    }
  ]
}
```

PR #1008, opened 2019-08-15
Let's run tests in parallel

adberglund:
## Summary
This will hopefully speed up our test runs with parallel execution. 

### Local Run

```
----------------------------------------------------------------------
Ran 1157 tests in 432.099s

OK
Destroying test database for alias 'default' ('test_koku_1')...
Destroying test database for alias 'default' ('test_koku_2')...
Destroying test database for alias 'default' ('test_koku_3')...
Destroying test database for alias 'default' ('test_koku_4')...
Destroying test database for alias 'default' ('test_koku_5')...
Destroying test database for alias 'default' ('test_koku_6')...
Destroying test database for alias 'default' ('test_koku_7')...
Destroying test database for alias 'default' ('test_koku_8')...
Destroying test database for alias 'default' ('test_koku')...
```

PR #1021, opened 2019-08-20
Revert "Let's run tests in parallel (#1008)"

infinitewarp:
This reverts commit 3005eebcdbd286e05ce8fdb65e9c961578f0d8fe.

After discussing with @blentz, I'm reverting this feature because when (some? all? only sometimes?) tests fail and parallel is enabled, the test runner crashes out without giving useful information. For example, a recent run:

```
TypeError: can't pickle traceback objects

ERROR: InvocationError for command /home/travis/build/project-koku/koku/.tox/py36/bin/coverage run --parallel-mode --concurrency=multiprocessing koku/manage.py test --noinput --parallel -v 2 koku/ (exited with code 1)
```

It might also be related to a seemingly random/intermittent error I saw that results in databases being dropped too early:

```
setUpClass (api.provider.test.tests_serializers.AdminProviderSerializerTest) failed:

    OperationalError('FATAL:  database "test_koku_test_1" does not

    exist\nDETAIL:  It seems to have just been dropped or renamed.\n',)
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -404,16 +404,13 @@ oc-up-db: oc-up oc-create-db
 docker-down:
 	docker-compose down
 
-docker-down-db:
-	docker-compose stop db
-
 docker-logs:
 	docker-compose logs -f
 
 docker-rabbit:
 	docker-compose up -d rabbit
 
-docker-reinitdb: docker-down-db remove-db docker-up-db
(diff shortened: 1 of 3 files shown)


adberglund:
I would prefer https://github.com/project-koku/koku/pull/1022 over a full revert as the ability to run in parallel is still useful.

PR #1027, opened 2019-08-21
1004 move openshift cost summary

maskarb:
Resolves issue 1004

update_cost_summary_table is now stand alone method in ocp_cloud_summary_updater. Method is called regardless of provider and only after derived costs are calculated.

Testing:
1. Add aws provider.
2. Trigger download and verify that cost_summary_table is updated.


tox.ini
```diff
@@ -65,7 +65,7 @@ commands =
   /bin/sh {toxinidir}/scripts/check_postgres_running.sh
   /bin/sh {toxinidir}/scripts/create_test_db_user.sh
   pipenv install --dev --ignore-pipfile
-  coverage run --parallel-mode --concurrency=multiprocessing {toxinidir}/koku/manage.py test --noinput --parallel -v 2 {posargs: koku/}
+  coverage run --parallel-mode --concurrency=multiprocessing {toxinidir}/koku/manage.py test --noinput -v 2 {posargs: koku/}
```
adberglund:
```suggestion
  coverage run --parallel-mode --concurrency=multiprocessing {toxinidir}/koku/manage.py test --noinput --parallel -v 2 {posargs: koku/}
```

dccurtis:
We should be able to remove `populate_cost_summary_table` from here too with this change. 👍 

https://github.com/project-koku/koku/blob/4261acbe914efd4a84b48949a918ecf836449c55/koku/masu/processor/ocp/ocp_report_charge_updater.py#L280

Up to you if you want to remove it now or we can always do it in a follow on PR.

Issue #982, opened 2019-08-09
Configure destination S3 bucket for customer data export


infinitewarp:
## User Story
As a Koku user, I want to provide my writable S3 bucket name so Koku can send my data exports to it.

When a customer user requests a new data export, the user must include with that request the name of the bucket to which we will write the data export files.

## Assumptions and Questions
- This does not mean creating a new API. This intends to add a `bucket_name` field to the `dataexportrequest` API that was recently added.
- For now, assume only AWS S3 bucket name is required.
- The user has already configured the bucket to allow our AWS account write access.
- A future iteration may expand this to support Azure Storage.

## Acceptance Criteria
- [ ] user's writable S3 bucket information is retained with the data export request

infinitewarp:
Last week I discussed this issue with Andrew and Chris. The current advised path is simply to require the bucket name as an input with the initial data export request. In a future iteration we may ask for this information earlier in the customer/user on-boarding process, but that is beyond the scope of this immediate work.

I am updating the issue description to match that solution.

lcouzens:
Verified.
Commit: 96a12fc0e55ee08d444f5919b549ec575683a988

PR #1036, opened 2019-08-22
Add bucket_name to DataExportRequest

ilya-w:
https://github.com/project-koku/koku/issues/982


koku/api/dataexport/models.py
```diff
@@ -21,6 +21,7 @@ class DataExportRequest(models.Model):
     status = models.CharField(max_length=32, choices=STATUS_CHOICES, default=PENDING)
     start_date = models.DateField(null=False)
     end_date = models.DateField(null=False)
+    bucket_name = models.CharField(max_length=63)
```
infinitewarp: I think this needs a False null/blank setting because it should be a required value.

koku/api/dataexport/serializers.py
```diff
@@ -18,6 +18,7 @@ class Meta:
             'start_date',
             'end_date',
             'status',
+            'bucket_name',
```
infinitewarp: We should also add bucket_name to create_only_fields because we don't plan to let the user change the request.

ilya-w: I was wondering about that, but there was nothing in the story about it being immutable, so I left it as it was. Will make the change now.

koku/api/dataexport/models.py
ilya-w: Null and blank default to False.
https://docs.djangoproject.com/en/2.2/ref/models/fields/#null

```
Field.null¶
If True, Django will store empty values as NULL in the database. Default is False.
```
```
Field.blank¶
If True, the field is allowed to be blank. Default is False.
```

PR #1044, opened 2019-08-27
Add provider names to the cost model response

### LINKED ISSUE #1043: Cost Model should show provider names and uuids
## User Story
As UI developer, I want to see names of all the providers that are using a specific cost model so that I can display it in my UI without doing over-fetching.

## Impacts
API

## Role
user

## Assumptions
None

## API Details

| API  | Detail|
 | --- | --- |
 | **Method** | GET |
 | **Path** | /api/v1/costmodel/ |
 | **Body** |  |
 | **Response** | [{"uuid": "32ad1f8961addb", "providers": [{"uuid": "45db", "name": "p1"}, {"uuid": "51cef", "name": "p2"}] }] |

## UI Details
None

## Acceptance Criteria

- [ ] Get provider uuid and name from cost model

boaz0:
closes #1043 

This should resolve the problem mentioned in #1043.

I thought about using on-to-many relationship between cost-model and provider but I guess it's not possible as "cost model" is a tenant app and "api.provider" is shared app.

Anyway, I went with displaying the required data through the serializer. Let me know what you think.


adberglund:
Provider list just needs to look like this in the openapi

```
"providers": {
    "type": "array",
    "items": {
        "type": "object",
        "properties": {
            "uuid": {
                "type": "string"
            },
            "name": {
                "type": "string"
            }
        }
    }
}

```

boaz0:
Thanks @adberglund - updated openapi.json again.

PR #1069, opened 2019-08-29
Sync S3 lifecycle policy

ilya-w:
This should probably be merged after the preceding stories are done and merged.

### DIFF
--- a/scripts/sync_s3_lifecycle.py
+++ b/scripts/sync_s3_lifecycle.py
@@ -1,0 +1,42 @@
+#!/usr/bin/env python3
+""" sync s3 lifecycle policy """
+
+import os
+from distutils.util import strtobool
+
+import boto3
+
+
+def sync_s3_lifecycle():
+    s3 = boto3.resource('s3')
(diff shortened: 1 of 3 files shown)


scripts/sync_s3_lifecycle.py
```diff
@@ -0,0 +1,43 @@
...
+def sync_s3_lifecycle():
+    s3 = boto3.resource('s3')
+    bucket_lifecycle_configuration = \
+        s3.BucketLifecycleConfiguration(os.environ.get('S3_BUCKET_NAME'))
+    bucket_lifecycle_configuration.put(
+        LifecycleConfiguration={
+            'Rules': [
+                {
+                    'Expiration': {
+                        'Days': 1465
```
infinitewarp: Does `1465` have a special meaning? Should it be configurable?

ilya-w: My apologized, I went over the story requirements and realized there was no AC for object expiration. This has been removed.

Issue #1018, opened 2019-08-20
Cost models endpoint accepts invalid queries 


lcouzens:
**Describe the bug**
Cost models endpoint accepts invalid queries 

**To Reproduce**
1. Go to '/api/cost-management/v1/costmodels/'
2. add invalid query `/api/cost-management/v1/costmodels/?wrong=query`
3. See returned data

**Expected behavior**
Expected to see something like:
```
{
    "wrong": [
        "Unsupported parameter or invalid value"
    ]
}
```

**Context (please complete the following information):**
 - OS: Fedora
 - Browser: Firefox
 - Commit Hash from status endpoint: "18c1cc664d035147e7eac0659e6f470c59380c99"

**Additional context**
Reproducible locally and on CI

Automation: test_api_cost_model_negative_endpoint

adberglund:
I think to fix this we need to have a white list of params in the view itself and check to make sure that th incoming params are in the whitelist. If they are not then we throw an error similar to the error provided by the reports api serializer.

lcouzens:
Verified
Commit: `8ebad1f91f71f6dc97d3e5dd91ff341dd6d8bd45`

PR #1072, opened 2019-08-29
1018 reject invalid queries

maskarb:
reference #1018 

Testing:

1. Go to '/api/cost-management/v1/costmodels/'
2. add invalid query /api/cost-management/v1/costmodels/?wrong=query
3. See returned data
```python
HTTP 400 Bad Request
Allow: GET, POST, HEAD, OPTIONS
Vary: Accept

{
    "errors": [
        {
            "detail": "Cannot resolve keyword 'wrong' into field. Choices are: costmodelmap, created_timestamp, description, name, rates, source_type, updated_timestamp, uuid",
            "source": "detail",
            "status": 400
        }
    ]
}
```


maskarb:
@blentz 
I understood those docstrings to be for apidocs. Since we moved away from apidocs, I removed them.

I have no issues putting them back.

blentz:
The formatting is apidoc, sure. But it's still documentation. 

Feel free to rewrite it if you don't want it to be in apidoc format, but please don't remove documentation entirely.

adberglund:
@blentz Those docstrings weren't up-to-date and without apidoc, I don't think maintaining them is good. The OpenAPI spec is our API documentation "source of truth" and I think that's good enough for detailed API documentation. I would actually go a step further in this PR and just eliminate the `create`, `list`, and `retrieve` methods in this file as all they're doing is calling super(), and they were only ever there to generate apidoc documentation.

PR #1078, opened 2019-09-04
Upload normalized data to koku's s3 bucket.

werwty:
- Add script to create S3 bucket and set its policies.


koku/masu/processor/report_processor_base.py
```diff
@@ -108,13 +114,27 @@ def _write_processed_rows_to_csv(self):
...
 
-    def _save_to_db(self, temp_table, report_db_accessor):
+    def _save_to_db(self, temp_table, report_db_accessor, date):
         """Save current batch of records to the database."""
         columns = tuple(self.processed_report.line_items[0].keys())
         csv_file = self._write_processed_rows_to_csv()
-
         report_db_accessor.commit()
 
+        upload_path = "{storage_prefix}/{account_name}/{year}/{month}/{table_name}.csv".format(
```
infinitewarp: I think this needs single quotes to make the linter happy.

Same on line 130.

koku/api/dataexport/uploader/__init__.py
```diff
@@ -2,7 +2,7 @@
 from abc import ABC, abstractmethod
 
 import boto3
-
+from django.conf import settings
```
infinitewarp: The linter wants an extra line after this so that there are two blank lines between imports and the first class.

openshift/celery-worker.yaml
```diff
@@ -305,6 +305,24 @@ objects:
...
+              valueFrom:
+                configMapKeyRef:
+                  name: ${NAME}-app
+                  key: s3-bucket-name
+                  optional: false
+            - name: S3_STORAGE_PREFIX
+              valueFrom:
+                configMapKeyRef:
+                  name: ${NAME}-app
+                  key: s3-bucket-path
```
infinitewarp: Why the rename? We should be consistent between the name and key to make future searching/refactoring efforts easier. I have no strong preference for `S3_STORAGE_PREFIX` versus `S3_BUCKET_PATH`, and if the latter looks better, let's make them all like that.

werwty: An oversight. It is now `S3_BUCKET_PATH` everywhere.

PR #1090, opened 2019-09-10
1033 cost model markup

maskarb:
~work in progress for cost model markup.~

Ready for review. User can add markup to cost model and markup is successfully calculated during data summarization.

Will add reporting api changes in a follow up PR.


tox.ini
```diff
@@ -64,7 +64,7 @@ deps =
 commands =
   /bin/sh {toxinidir}/scripts/check_postgres_running.sh
   /bin/sh {toxinidir}/scripts/create_test_db_user.sh
-  pipenv install --dev --ignore-pipfile
+  ; pipenv install --dev --ignore-pipfile
```
dccurtis: Don't forget to uncomment ;)

PR #1093, opened 2019-09-11
Fix cost models check fields to exclude pagination params

### LINKED ISSUE #1091: Api returns 400 when using limit/offset query parameters in cost model api
**Describe the bug**
Using `limit` or `offset` query parameters in `/api/cost-management/v1/costmodels/` results in 400 response and 

**To Reproduce**
1. Go to `https://ci.cloud.redhat.com/api/cost-management/v1/costmodels/?offset=0&limit=10`
2. See error

**Expected behavior**
200 response with a list of cost models

**Response**
```
{
    "errors": [
        {
            "detail": "Cannot resolve keyword 'limit' into field. Choices are: costmodelmap, created_timestamp, description, name, rates, source_type, updated_timestamp, uuid",
            "source": "detail",
            "status": 400
        }
    ]
}
```

**Additional context**
commit afbc1ac47dc809a704e082d3415b8e4eb8811500
tested on CI

boaz0:
closes #1091

### DIFF
--- a/koku/cost_models/view.py
+++ b/koku/cost_models/view.py
@@ -119,8 +119,10 @@ class CostModelViewSet(mixins.CreateModelMixin,
     @staticmethod
     def check_fields(dict_, model, exception):
         """Check if GET fields are valid."""
+        valid_query_params = ['limit', 'offset', 'provider_uuid']
+        cost_models_params = {k: dict_.get(k) for k in dict_.keys() if k not in valid_query_params}
         try:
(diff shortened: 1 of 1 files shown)


boaz0:
Nooice!
I will update now.
Thanks @maskarb

boaz0:
That's so kwel :dark_sunglasses:  removed 4 lines to get the same results! :call_me_hand:

Issue #891, opened 2019-06-06
Create Koku Sources Client service


adberglund:
## Objective
We need to listen to the sources kafka topic `platform.sources.event-stream` so that we can act on newly registered sources.

Listening to the sources kafka topic should be done in an independent service from the Koku API application.  This service will be responsible for handling sources kafka message handling, Sources API response, and creation of roleARN to construct a koku provider object.  The service would then make appropriate koku `/provider` REST API calls to ensure that the koku application and Insight-Sources are always in sync.

## Assumptions
- So that this can be done independent of the database model setup in our system, for this issue, we can just log the message we receive. 
- koku sources client will live in it's own repository

## Acceptance Criteria

- [ ] We listen and log a message from Insights Sources service
- [ ] Service is deployable via docker-compose and OpenShift

lcouzens:
Verified
Commit: "6bac6962d08f693740f8147c966bfdb5ce39a7c3"

PR #1111, opened 2019-09-12
Backend integration with Platform Sources

dccurtis:
Addresses #891 

Backend integration with Platform-Sources.  This change includes integration for OCP and AWS.  For the time being the AWS roleARN is stored in the password field of the Platform-Services Authentication secure data store.  Since the Platform-Sources backend does not handle application-specific Source setup a `/billing_source` endpoint was added to the Sources Integration service to allow adding of the S3 bucket for AWS providers.

Items that still need to be addressed are:
1. Add Azure provider creation.
2. Add OpenShift template
3. Get Platform-Sourceses backend running with docker-compose to finish off the docker-compose based deployment.
4. Potentially move from HTTP API based provider creation and deletion to a direct database implementation if necessary.  The reason why I didn't go down this road initially is because we have a lot of logic tied to the /providers endpoint that would need to be handled going the direct database route.  The churn (both in terms of code changes and risk) didn't seem justified at this point in time.

**Testing**

1. Add OCP Source and verify Koku Provider is created.
2. Add AWS Soruce and verify Koku Provider is created.
3. Delete OCP Source verify Koku Provider is removed.
4. Delete AWS Source verify Koku Provider is removed.
5. Stop Koku API server.  Add source, verify that provider is created once koku API is running again.
6. Stop Sources integration service.  Remove source from step 5.  Verify that provider is deleted once service is running again.
7. Add Source when both koku API and Sources integraiton services are down.  Start API and intgration service one by one, and verify provider is created when both are running.
8. With Sources integration service down.  Add and remove the same source.  Start integraiton service and verify that the provider was not created in koku.
9.  Add OCP source and attempt to add a billing_source via the /billing_source API.  Verify that it fails.
10. Add billing_source to a no existent source.  Verify that it fails.
11. Attempt to start Sources integraiton service with Sources backend down.  Verify warning message is displayed.

**Test Results**
[sources_backend_ut.txt](https://github.com/project-koku/koku/files/3611032/sources_backend_ut.txt)


koku/api/provider/models.py
```diff
@@ -144,6 +144,28 @@ class Meta:
...
+        """Meta for Sources."""
+
+        db_table = 'api_sources'
+
+    source_id = models.IntegerField(primary_key=True)
+    name = models.CharField(max_length=256, null=True)
+    source_type = models.CharField(max_length=50, null=False)
+    authentication = models.CharField(max_length=128, null=False)
+    billing_source = models.CharField(max_length=128, null=True)
+    koku_uuid = models.CharField(max_length=512, null=True)
```
adberglund: I would consider making this a real foreign key to our provider table. That would also make it clearer that this is a provider uuid.

koku/api/provider/models.py
```diff
@@ -144,6 +144,28 @@ class Meta:
...
+    auth_header = models.CharField(max_length=512, null=True)
+    pending_delete = models.BooleanField(default=False)
+    offset = models.IntegerField(null=False)
```
adberglund: Can you throw a comment on auth_header, pending_delete, and offset just to document what they represent.

Something like 

```
# offset is the ...
offest = 
```

koku/api/provider/models.py
dccurtis: Yes, using uuid for now since all provider sync communication is still HTTP based.  I agree to update this if we ever change that to be direct database updates.

PR #1122, opened 2019-09-17
Fix internal error when adding sources (MultipleObjectsReturned)

boaz0:
Because in the current database there might be duplicates in
billing source the "get_or_create" function raises an exception
saying it assumes there is no more than 1 item of the billing source.

To fix that MultipleObjectsReturned is handled in the following way:
 - print in the log all the duplicates for that billing source
 - get the first item from the duplicates and use that

Using the logs to find all the duplicates will allow later to alter
current database or somehow update old records to avoid such case

//cc @chargio

### DIFF
--- a/koku/api/provider/serializers.py
+++ b/koku/api/provider/serializers.py
@@ -17,6 +17,7 @@
 """Provider Model Serializers."""
 import logging
 
+from django.core.exceptions import MultipleObjectsReturned
 from django.db import transaction
 from django.utils.translation import ugettext as _
 from providers.provider_access import ProviderAccessor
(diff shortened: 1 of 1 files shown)


adberglund:
@boaz0 Thanks for looking into this. The issue we were seeing in production is actually already fixed in CI and just needs to make it to production. In this case we don't want to handle the exception, we need it to not happen, as we don't know which record is true if multiple are returned, so just picking `.first()` could give an incorrect result.

boaz0:
@adberglund I see :cat2: . If this is fixed, I will close this PR.
Thanks for checking this out.

PR #1127, opened 2019-09-17
Platform-Sources Integration OpenShift template

dccurtis:
Adding Sources integration service OpenShift templates as well as status endpoint for liveness probes.


openshift/sources.yaml
```diff
@@ -0,0 +1,430 @@
...
+                  name: ${NAME}-sources
+                  key: sources-port
+                  optional: false
+            - name: RETRY_SECONDS
+              valueFrom:
+                configMapKeyRef:
+                  name: ${NAME}-messaging
+                  key: kafka-retry-interval
+                  optional: false
+            - name: LOG_LEVELKOKU_API_PATH_PREFIX
```
adberglund:
```suggestion
            - name: LOG_LEVEL
```

PR #1145, opened 2019-09-20
configure cache-control headers for all views.

blentz:
This configures cache-control headers for each view to ensure we are not caching things that shouldn't be cached and that we are maintaining privacy of critical user data.

### DIFF
--- a/koku/api/iam/view/user_preference.py
+++ b/koku/api/iam/view/user_preference.py
@@ -24,8 +24,8 @@ from rest_framework.permissions import AllowAny
 
 import api.iam.models as models
 import api.iam.serializers as serializers
-from api.common.permissions.object_owner import IsObjectOwner
 from api.common import RH_IDENTITY_HEADER
+from api.common.permissions.object_owner import IsObjectOwner
 
 
 class UserPreferenceViewSet(mixins.CreateModelMixin,
--- a/koku/api/iam/view/user_preference.py
+++ b/koku/api/iam/view/user_preference.py
@@ -16,15 +16,13 @@
 #
 
 """View for User Preferences."""
-from django.views.decorators.cache import cache_control, never_cache
-from django.views.decorators.vary import vary_on_headers
+from django.views.decorators.cache import never_cache
 from django_filters.rest_framework import DjangoFilterBackend
 from rest_framework import exceptions, mixins, viewsets
 from rest_framework.permissions import AllowAny
 
 import api.iam.models as models
 import api.iam.serializers as serializers
-from api.common import RH_IDENTITY_HEADER
 from api.common.permissions.object_owner import IsObjectOwner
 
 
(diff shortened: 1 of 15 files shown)


koku/api/iam/view/user_preference.py
```diff
@@ -115,6 +118,8 @@ def create(self, request, *args, **kwargs):
 
         return super().create(request=request, args=args, kwargs=kwargs)
 
+    @cache_control(private=True)
+    @vary_on_headers('User-Agent', 'Cookie')
```
chambridge: I think the header we need to vary on is the `x-rh-identity` header.
https://github.com/project-koku/koku/blob/8ab793cccce66efad3a97d3dae085ec334eeba7c/koku/api/common/__init__.py#L4

koku/api/provider/view.py
```diff
@@ -115,10 +118,13 @@ def get_queryset(self):
...
         return queryset
 
+    @never_cache
     def create(self, request, *args, **kwargs):
         """Create a Provider."""
         return super().create(request=request, args=args, kwargs=kwargs)
 
+    @cache_control(private=True)
+    @vary_on_headers(RH_IDENTITY_HEADER)
     def list(self, request, *args, **kwargs):
```
dccurtis: If we are going to be caching this, could we invalidate the cache whenever a provider is created or removed?

blentz: We can, but as the saying goes, there are two Very Hard Problems in computing - naming things and cache invalidation.

To invalidate an object, we need to know it's key in the cache. The cache key is generated based on attributes and headers from the request and hashed values (i.e. the url is hashed and used as part of the key). The Vary header in our response specifies other headers to use as part of the cache key hash.

(See: https://github.com/django/django/blob/master/django/utils/cache.py#L304 )

If there's a concern about returning cached responses and there isn't a reasonable TTL we can set (e.g. would a 5-minute cache TTL work where a 60-minute TTL won't?), we probably shouldn't cache the view at all.

chambridge: Sure. I guess the thing that makes the most sense here is the caching around the reports view as that is where our real time is spent. Perhaps without expending a bunch of time on cache invalidation we just focus on caching just for reports and go with `@never_cache` for the rest? We could open an issue around the invalidation approach for other APIs.

koku/cost_models/view.py
```diff
@@ -142,10 +145,12 @@ def get_queryset(self):  # noqa: C901
                     LOG.error(queryset_error)
         return queryset
 
+    @never_cache
     def create(self, request, *args, **kwargs):
         """Create a rate."""
         return super().create(request=request, args=args, kwargs=kwargs)
 
+    @vary_on_headers(RH_IDENTITY_HEADER)
```
adberglund: Can we invalidate the list/retreieve cache if someone does a create, update, or delete operation?

blentz: Doug asked the same question. Short answer: Not easily enough to be worthwhile.

I had a side-chat with Doug about it and our thinking is just to `@never_cache` views where we need to show state changes like this.

dccurtis: I had the same question for `/providers`.  I suggest we do the same thing here and not cache `/costmodels`

PR #1146, opened 2019-09-20
Adding Azure to Sources Client integration

dccurtis:
Adds Azure to the Koku-Sources Client integration service.  Platform-Sources stores `client_id`, `client_secret`, and `tenant_id`.  In Koku's current provider on boarding we require these values in addition to `subscription_id`, `storage_group`, and `storage_account`.

We must supply the `storage_group` and `storage_account` values to the existing `/billing_source` Sources Client HTTP endpoint.  A new endpoint `/authentications` was added to supply the `subscription_id`.

**Testing**
1. Create Azure Source in Platform-Sources backend
2.  Associate Storage Information with Azure Source.
3.  Associate Subscription ID with Azure Source.  Verify provider is created
4.  Add Azure Provider to Koku directly, then add source.  Verify that koku_uuid is updated in source.

**Test Results**
[azure_sources_ut.txt](https://github.com/project-koku/koku/files/3636818/azure_sources_ut.txt)


koku/api/provider/models.py
```diff
@@ -178,10 +178,10 @@ class Meta:
     source_type = models.CharField(max_length=50, null=False)
 
     # Provider authentication (AWS roleARN, OCP Sources UID, etc.)
-    authentication = models.CharField(max_length=128, null=False)
+    authentication = JSONField(null=False, default=dict)
```
blentz: How are we going to handle migrating existing rows? The migration doesn't appear to cover it, unless there's some sort of default.

koku/sources/api/urls.py
```diff
@@ -16,11 +16,12 @@
 """Describes the urls and patterns for the API application."""
 from django.conf.urls import url
 from sources.api.status import get_status
-from sources.api.views import billing_source
+from sources.api.views import authentications, billing_source
```
blentz: Is there some reason authentication is plural while billing_source is singular?

koku/sources/kafka_listener.py
```diff
@@ -149,39 +167,20 @@ def _sources_network_info_sync(source_id, auth_header):
 
     if source_type_id == 1:
         source_type = 'OCP'
-        authentication = source_details.get('uid')
+        authentication = {'resource_name': source_details.get('uid')}
     elif source_type_id == 2:
         source_type = 'AWS'
-        authentication = sources_network.get_aws_role_arn()
+        authentication = {'resource_name': sources_network.get_aws_role_arn()}
+    elif source_type_id == 3:
```
blentz: It might be good to move these magic ID numbers into a set of constants (or a list/dict of constants).

koku/sources/koku_http_client.py
```diff
@@ -41,17 +43,91 @@ def __init__(self, auth_header):
         header = {'x-rh-identity': auth_header, 'sources-client': 'True'}
         self._identity_header = header
 
-    def create_provider(self, name, provider_type, authentication, billing_source):
-        """Koku HTTP call to create provider."""
-        url = '{}/{}/'.format(self._base_url, 'providers')
-        json_data = {'name': name, 'type': provider_type}
+    @staticmethod
+    def _get_dict_from_text_field(value):
```
blentz: Why would private class methods need to be static?  

Should these be functions (outside of a class) instead? Or should we make them public class methods (remove the preceding `_`)?

I think either of those options is closer to best practice.

koku/sources/koku_http_client.py
```diff
@@ -41,17 +43,91 @@ def __init__(self, auth_header):
...
+            auth = {'credentials': authentication.get('credentials')}
+        else:
+            raise KokuHTTPClientError('Missing credentials')
+        return auth
+
+    def _authentication_for_aws(self, authentication):
+        return self._build_provider_resource_name_auth(authentication)
+
+    def _authentication_for_ocp(self, authentication):
+        return self._build_provider_resource_name_auth(authentication)
```
blentz: You're calling a staticmethod non-statically here. Perhaps the methods they're calling shouldn't have the `@staticmethod` decorator?

koku/sources/storage.py
```diff
@@ -26,17 +26,46 @@ class SourcesStorageError(Exception):
     """Sources Storage error."""
```
blentz: It looks like this `storage.py` file could be getting a bit long. You might want to consider breaking it up into per-provider files.

koku/api/provider/models.py
dccurtis: Since we don't have any data in CI or production I didn't bother making the migration change.

koku/sources/api/urls.py
dccurtis: No reason.  Other then I had `/authentications` on the mind from what the Platform-Sources calls it.  I re-named this to `/authentication`

koku/sources/kafka_listener.py
dccurtis: Good idea.  Added some constants

koku/sources/koku_http_client.py
dccurtis: Well these are private utility functions that's only intended to be used from this class.  They don't need any information about any particular object so there's no reference to self needed.

The purpose is to build the `/provider` HTTP json data so I don't have a use case for them outside of this class.

koku/sources/koku_http_client.py
dccurtis: Yeah, it does look odd I guess.  It needs a reference to itself to get the static method.

koku/api/provider/models.py
blentz: I'm okay with that.

koku/sources/koku_http_client.py
blentz: If they're private functions, let's drop the `@staticmethod` decorator and call them directly from inside the class.

koku/sources/koku_http_client.py
blentz: To call it statically, you'd use this:

```
KokuHTTPClient._authentication_for_ocp()
```

But then lint would complain about you calling a class' private methods. :grin: 

So, as above, let's keep this code as-is, but drop the `@staticmethod` decorator from the method.

dccurtis: Ok Ill remove `@staticmethod`.  My editor isn't happy with it but it seems to be getting past our tox/lint setup so lets do it.

koku/sources/koku_http_client.py
```diff
@@ -41,17 +43,88 @@ def __init__(self, auth_header):
...
+    def get_authentication_for_provider(self, provider_type, authentication):
+        """Build authentication json data for provider type."""
+        provider_map = {'AWS': self._authentication_for_aws,
+                        'OCP': self._authentication_for_ocp,
+                        'AZURE': self._authentication_for_azure}
+        provider_fn = provider_map.get(provider_type)
+        if provider_fn:
+            return provider_fn(authentication)
 
+    @staticmethod
```
blentz:
```suggestion
```

koku/sources/koku_http_client.py
```diff
@@ -41,17 +43,88 @@ def __init__(self, auth_header):
...
+    def _build_provider_bucket(billing_source):
+        if billing_source.get('bucket') is not None:
+            billing = {'bucket': billing_source.get('bucket')}
+        else:
+            raise KokuHTTPClientError('Missing bucket')
+        return billing
+
+    @staticmethod
```
blentz:
```suggestion
```

Issue #1149, opened 2019-09-23
OpenShift builds failing on install of googleapis-common-protos==1.6.0


adberglund:
**Describe the bug**
OpenShift builds are failing trying to install googleapis-common-protos==1.6.0

**To Reproduce**
1. Trigger koku build in OpenShift

**Traceback**

```
An error occurred while installing googleapis-common-protos==1.6.0 --hash=sha256:e61b8ed5e36b976b487c6e7b15f31bb10c7a0ca7bd5c0e837f4afab64b53a0c6! Will try again.
--
  | Installing initially failed dependencies…
  | [pipenv.exceptions.InstallError]:   File "/opt/app-root/src/.local/venvs/pipenv/lib/python3.6/site-packages/pipenv/core.py", line 1874, in do_install
  | [pipenv.exceptions.InstallError]:       keep_outdated=keep_outdated
  | [pipenv.exceptions.InstallError]:   File "/opt/app-root/src/.local/venvs/pipenv/lib/python3.6/site-packages/pipenv/core.py", line 1253, in do_init
  | [pipenv.exceptions.InstallError]:       pypi_mirror=pypi_mirror,
  | [pipenv.exceptions.InstallError]:   File "/opt/app-root/src/.local/venvs/pipenv/lib/python3.6/site-packages/pipenv/core.py", line 859, in do_install_dependencies
  | [pipenv.exceptions.InstallError]:       retry_list, procs, failed_deps_queue, requirements_dir, **install_kwargs
  | [pipenv.exceptions.InstallError]:   File "/opt/app-root/src/.local/venvs/pipenv/lib/python3.6/site-packages/pipenv/core.py", line 763, in batch_install
  | [pipenv.exceptions.InstallError]:       _cleanup_procs(procs, not blocking, failed_deps_queue, retry=retry)
  | [pipenv.exceptions.InstallError]:   File "/opt/app-root/src/.local/venvs/pipenv/lib/python3.6/site-packages/pipenv/core.py", line 681, in _cleanup_procs
  | [pipenv.exceptions.InstallError]:       raise exceptions.InstallError(c.dep.name, extra=err_lines)
[...]

PR #1152, opened 2019-09-23
upgrade pip to latest during build

infinitewarp:
Since we recently added `google-cloud-python` as a dependency, the build step has started failing as in #1149. This is because we have old versions of `pip` and `setuptools` in the base build image.

This change enables a feature in our build image to upgrade to the latest `pip`, `setuptools`, and `wheel` _before_ `pipenv` starts its own environment installation process.

See also: https://github.com/googleapis/google-cloud-python/issues/2990 and https://github.com/pypa/setuptools/issues/885

### DIFF
--- a/openshift/koku.yaml
+++ b/openshift/koku.yaml
@@ -83,6 +83,8 @@ objects:
             value: ${PIPENV_PYPI_MIRROR}
           - name: ENABLE_PIPENV
             value: "true"
+          - name: UPGRADE_PIP_TO_LATEST
+            value: "true"
           - name: APP_CONFIG
             valueFrom:
               configMapKeyRef:


infinitewarp:
FWIW, I tested with my local minishift deployment to verify that setting this environment variable resolves the build error.

infinitewarp:
For reference, this is the feature that gets enabled:

https://github.com/sclorg/s2i-python-container/blob/master/3.6/s2i/bin/assemble#L41
```bash
if [[ ! -z "$UPGRADE_PIP_TO_LATEST" ]]; then
  echo "---> Upgrading pip to latest version ..."
  pip install -U pip setuptools wheel
fi
```

PR #1165, opened 2019-09-25
Bugfix ocp aws markup cost

adberglund:
## Summary
Because the markup_cost fields are new, most of the existing data in the database has a `null` for that field. When you add a value + a null in SQL, you get a null, which was causing our cost and markup_cost fields to come back as `null` and in turn, our code to format the values into a dictionary with value and unit was not working. 

I added lots of `Coalesce()` functions to cost and markup_cost fields to default to 0 when a null is in either field. See https://docs.djangoproject.com/en/2.2/ref/models/database-functions/#coalesce if unfamiliar.

I also switched all fields related to markup cost to `markup_cost` both in aggregates and annotations, regardless of whether it is calculated from the markup_cost or project_markup_cost. The code to dictionary-ize the fields was not accounting for the different names, so having everything called `markup_cost` will get picked up by that now. 

While testing all of this, I found a bug in the create test customer script for the skip API route, and threw in a refactor to make sure each provider created has the correct billing and auth id associated with them.

## Testing
Ran OCP on AWS nise datasets locally and checked OCP on AWS, AWS, an OCP APIs to see if all fields returned in the correct format. I had to simulate the `null` field case, because running fresh data caused the "correct" `0` to be placed in the markup_cost column when there was no markup. Intentionally set that column to null in tables to check that the null value does not cause the API field to return null, but defaults to 0 now.


koku/api/report/azure/provider_map.py
```diff
@@ -70,15 +70,15 @@ def __init__(self, provider, report_type):
                         'annotations': {
                             'infrastructure_cost': Sum('pretax_cost'),
                             'derived_cost': Value(0, output_field=DecimalField()),
-                            'markup_costs': Sum('markup_cost'),
+                            'markup_cost': Sum('markup_cost'),
```
maskarb: Why not use Coalesce and move cost above markup_cost?

Issue #1079, opened 2019-09-05
Use safe templating for templated raw SQL


adberglund:
## Objective
We don't want SQL Injection, and despite our current variables being controlled by us, it would still be safer to rely on paramaterization for the strings getting inserted into our SQL. 

Let's look at using https://github.com/hashedin/jinjasql for our SQL which will swap out or variables and return a list of parameters for the SQL. 

## Additional Information

* We need to modify our variables in our SQL to be Jinja variables (effectively switching from single brackets `{variable}` to double `{{variable}}`
  * SQL files are located here: https://github.com/project-koku/koku/tree/master/koku/masu/database/sql

* We need to switch our db_accessors that run this SQL from using string `.format()` calls to running JinjaSQL and getting the modified SQL string and bind_params.

* Update https://github.com/project-koku/koku/blob/master/koku/masu/database/report_db_accessor_base.py#L415 commit_and_vacuum to pass the sql and bind params to the `cursor.execute()` call

## Acceptance Criteria

- [ ] SQL produces identical results

nbon12:
@adberglund where is the unsafe SQL?

adberglund:
@nbon12 just a reminded to put a comment on the issue to help with QA. 

QA: There should be no functional difference here. This was just tackling technical debt and the goal is for all processed and summarized data to be exactly the same as it was before.

lcouzens:
Verified.
Commit: 96a12fc0e55ee08d444f5919b549ec575683a988

PR #1192, opened 2019-10-01
Safer sql awscostentrylineitem daily

nbon12:
Story: https://github.com/project-koku/koku/issues/1079
Iteratively improving the safety of SQL statements, this PR introduces the usage of SQL parameters separately passed into the database, rather than SQL and variables in the same string. Using [JinjaSQL](https://github.com/hashedin/jinjasql) to generate the variables list separately from the SQL


koku/masu/database/aws_report_db_accessor.py
```diff
@@ -197,23 +197,27 @@ def populate_line_item_daily_table(self, start_date, end_date, bill_ids):
 
         """
         table_name = AWS_CUR_TABLE_MAP['line_item_daily']
+        from jinjasql import JinjaSql
+        j = JinjaSql()
```
nbon12: I'm planning to move this out of the function into the heading of the file

koku/masu/database/report_db_accessor_base.py
```diff
@@ -424,7 +424,10 @@ def _commit_and_vacuum(self, table, sql, start=None, end=None):
             transaction.savepoint_commit(KokuDBAccess._savepoints.pop())
         with connection.cursor() as cursor:
             cursor.db.set_schema(self.schema)
-            cursor.execute(sql)
+            if bind_params is not None:
+                cursor.execute(sql, bind_params)
+            else:
+                cursor.execute(sql)
```
blentz: I don't think a conditional is needed here. This should be functionally equivalent and using a keyword arg means not needing to care about whether `bind_params` is None or not:

```suggestion
                cursor.execute(sql, bind_params=bind_params)
```

koku/masu/database/aws_report_db_accessor.py
```diff
@@ -197,23 +198,26 @@ def populate_line_item_daily_table(self, start_date, end_date, bill_ids):
 
         """
         table_name = AWS_CUR_TABLE_MAP['line_item_daily']
+
+        jinjaSql = JinjaSql()
```
adberglund: This is a nitpick, but consistent with our code. Use lowercase where it makes sense and separate with underscores. 
```suggestion
        jinja_sql = JinjaSql()
```

koku/masu/database/report_db_accessor_base.py
nbon12: change added

nbon12:
FYI, table names are not made SQL-safe in this PR, see https://github.com/hashedin/jinjasql#sql-safe-strings

Issue #1195, opened 2019-10-02
Handle Authentication.create events in Sources Client


dccurtis:
Platform Sources Wizard can create authentication objects either before or after Cost Management is attached to a particular source.  We need to handle Authentication messages. i.e. Storing roleARN or Azure client and tenant info either before or after `Application.create`.

Currently we are unable to create AWS or Azure sources in CI  because of this timing discrepancy.

lcouzens:
Verified
Commit: "6bac6962d08f693740f8147c966bfdb5ce39a7c3"

PR #1196, opened 2019-10-02
Handle Authentication.create messages from Platform Sources

dccurtis:
Addresses #1195 

AWS and Azure store credential information in the Platform Sources authentication store.  The Authentication information can be stored before or after the Cost Management application is attached to the source.  This change handles this event and updates the Source Authentication information regardless of this order.

**Testing**
1. Add Source, endpoint, attach cost management application, authentication, add S3 bucket.  Verify Koku Provider is created.
2. Add Source, endpoint, authentication, attach cost management application, add S3 bucket.  Verify Koku Provider is created.
3.  Add OCP and Azure Sources make sure they are created with helper script.

**Test Results**
[issue_1195_ut.txt](https://github.com/project-koku/koku/files/3679254/issue_1195_ut.txt)


koku/sources/kafka_listener.py
```diff
@@ -221,11 +292,16 @@ def sources_network_info(source_id, auth_header):
                 await EVENT_LOOP.run_in_executor(pool, sources_network_info,
                                                  msg_data.get('source_id'),
                                                  msg_data.get('auth_header'))
+        elif msg_data.get('event_type') == KAFKA_AUTHENTICATION_CREATE:
+            with concurrent.futures.ThreadPoolExecutor() as pool:
+                await EVENT_LOOP.run_in_executor(pool, sources_network_auth_info,
```
adberglund: 🔑

Issue #1180, opened 2019-09-30
Azure costs reporting 10x higher than actual


adberglund:
**Describe the bug**
We are reporting 10x the cost for Azure than our actual bill right now. The problem appears to be in report processing. I believe we are duplicating data as the `_daily` table we load data into should match what Azure claims is our current cost.

```
select sum(pretax_cost) from reporting_azurecostentrylineitem_daily where usage_date_time >= '2019-09-01';
      sum
---------------
 155.882472574
```

Looking a little further, I believe we want to take either the AWS approach or the OCP approach for Azure processing.

**AWS Approach)** Wipe the table for the billing period prior to processing a file. 
**OCP approach)** Write the entire file to a temporary table, and then merge that temp table into the main table with an `ON CONFLICT... DO UPDATE` clause.

**Expected behavior**
We need a proper check to avoid duplicating data when running the Azure processor.

lcouzens:
Verified
Commit: "1dc0c990bdde6af21a1c4930c21d7216da6dde1b"
Automation: Raw calc tests coming soon

PR #1202, opened 2019-10-02
Add conflict columns to Azure report processing to avoid double processing

dccurtis:
Address #1180 

Using conflict columns and temp tables to do updates on line items that we have already seen before.

**Testing**
1. On master ingest a report and get API azure values
2. Ingest same report and resample API with new branch.  Verify that it matches #1
3. Re-generate a new Azure Export and re-ingest with new branch.  Verify values do not change since they were exported only a few minutes apart
4.  Re-load master and ingest data.  Then recreate an export and see that the data doubles

**Test Results**
[issue_1180_ut.txt](https://github.com/project-koku/koku/files/3682803/issue_1180_ut.txt)


koku/masu/database/azure_report_db_accessor.py
```diff
@@ -153,3 +154,36 @@ def populate_markup_cost(self, markup, bill_ids=None):
             else:
                 AzureCostEntryLineItemDailySummary.objects.\
                     update(markup_cost=(F('pretax_cost') * markup))
+
+    # pylint: disable=too-many-arguments,arguments-differ
```
blentz: Let's try avoid these disables, if it's reasonably feasible to do so. 

I'd like to at least avoid disabling the `arguments-differ` check. I suspect that might be as easy as adding the `conflict_columns` parameter to the base class, then updating the method signatures of the other implementing classes.

PR #1212, opened 2019-10-03
Add support for aws glacier storage_class in the dataexport syncer.

werwty:
- When the file being synced is in glacier storage, restore that file and try
  to sync again later.

- When the file being synced is in IA-Standard, it can be copied with the usual `copy_from`, no additional check is needed.

### DIFF
--- a/koku/masu/celery/tasks.py
+++ b/koku/masu/celery/tasks.py
@@ -24,12 +24,13 @@ import collections
 from datetime import date
 
 from botocore.exceptions import ClientError
+from celery.exceptions import MaxRetriesExceededError
 from celery.utils.log import get_task_logger
 from dateutil.relativedelta import relativedelta
 from django.conf import settings
 
 from api.dataexport.models import DataExportRequest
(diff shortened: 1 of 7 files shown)


koku/masu/celery/tasks.py
```diff
@@ -259,6 +270,12 @@ def sync_data_to_customer(dump_request_uuid):
         dump_request.status = DataExportRequest.ERROR
         dump_request.save()
         return
-
+    except SyncedFileInColdStorageError as e:
+        LOG.info(
+            f'One of the request files is currently in cold storage, DataExportRequest'
+            f'{dump_request.uuid}. This task will automatically retry.')
+        dump_request.status = DataExportRequest.ERROR
```
infinitewarp: I don't think we should set the `status` to `ERROR` since the process is actually working normally despite the interruption. I typically think of error states as terminal states. Maybe we should add a new status here like `RETRYING` or `WAITING` to indicate that things are still going to be processed?

Issue #960, opened 2019-08-01
Refactor QueryHandler Initialization


blentz:
This issue was brought into focus while working on #938. 

The QueryHandler object tree has a really brittle, complex initialization sequence. This makes working on changes in this area more complex and difficult than it probably needs to be.

Of particular note are the steps that depend upon looking up the group_by parameter to make decisions.

This issue is to redesign and refactor QueryHandler initialization.

Proposed change:

1. Create a new object to hold data values passed into the QueryHandler from the Views. Let's call it "QueryData" for now. This object will be responsible for taking values from the `_generic_report()` view function, processing those values into the form required by the QueryHandler, and providing a set of getters or properties that the QueryHandler can use to access data coming in from the views.

This "QueryData" object is expected to replace the `url_data`, `query_parameters`, and `kwargs` parameters currently required by the QueryHandler. Instead of passing multiple individual data points, we'll pass a single QueryData object into the QueryHandler. 

The information contained within the QueryData should be "ready to go" and require no additional processing by the QueryHandler. (i.e. QueryHandler treats QueryData as read-only.)

2. Untangle the initialization flow. The basic flow for the whole QueryHandler should look something like this:

a. Receive parsed view data from the constructor parameters.
b. Look up model information required to compile the report out of the ProviderMap
c. Compile the requested query (i.e. `execute_query()`)

So, as part of refactoring the initialization portion of that workflow, we need to distinguish between View Data, Model References, and any metadata (e.g. class attributes) that are inherent to the QueryHandler objects themselves.

adberglund:
Since there is refactor involved here let's discuss as a team before starting work on this and agree on the proposal before beginning work. 

I put this card in the Design column so it would not get lost in the backlog or new issues (both columns are bloated). I don't anticipate getting to this until after Azure, and wanted to keep it visible on the board so we didn't lose track of it. Just trying something new.

blentz:
We discussed this and agreed to do a small/quick (1-2 hours) POC of the idea to help show the code impact. Once a POC is ready, we'll meet again to discuss sizing and timing on a full implementation.

blentz:
I've pushed a branch to show off some code around the QueryHandler Initialization discussion:
https://github.com/project-koku/koku/compare/960_queryhandler

This isn't fully functional quite yet, but it should at least show off the idea enough for us to discuss a go/no-go decision and stuff like timing/priority.

The main highlights are:
[...]

PR #1234, opened 2019-10-09
QueryHandler initialization refactor

blentz:
This PR relates to #960 . This refactors the QueryHandler initialization process to streamline information-passing from the Views.

Change Highlights:
- Add a new `QueryParameters` object whose responsibility is performing any transforms on data received from an `HTTPRequest` required by the `QueryHandler`. (i.e. The `QueryHandler` should receive all of the data it needs, in the format it needs to be in.)
- Moved several utility functions into the `QueryParameters` object. (e.g. the access-related functions were moved into `QueryParameters`)
- Moved and Removed class and instance attributes from QueryHandler to DRY things up and improve the separation of concerns. (e.g. `QueryHandler._accept_type` was moved to `QueryParameters.accept_type`)
- Restructured the calls to `super().__init__()` from sub-classes for a clearer initialization ordering. (There's more that could be done here, but this is a definite improvement.)
- A net reduction of about 600 lines of code due to the constructor simplification. This is mostly due to simplifying the object instantiation and removing the duplication of parameters and URI declarations in the relevant unit tests.


koku/api/report/queries.py
```diff
@@ -359,7 +325,7 @@ def _get_group_by(self):
         # For that ranking to work we can't also group by instance_type.
         inherent_group_by = self._mapper._report_type_map.get('group_by')
         if (inherent_group_by and not (group_by and self._limit)):
-            group_by += inherent_group_by
+            group_by = group_by + list(set(inherent_group_by) - set(group_by))
```
blentz: This is a bugfix. 

The refactor exposed an issue where `list += other_list` was adding duplicate entries to the merged result. This was causing problems with the report data nesting. So, this ensures we de-duplicate the lists.

koku/api/report/test/__init__.py
```diff
@@ -1 +1,455 @@
...
+
+from faker import Faker
+
+from api.query_params import QueryParameters
+from api.utils import DateHelper
+
+LOG = logging.getLogger(__name__)
+
+
+class FakeAWSCostData:
```
blentz: Moving this test class isn't required for the refactor. It was a "while I was there" decision to break apart the massive `test_queries` test file, which was approaching 2000 lines.

koku/api/report/test/azure/tests_azure_query_handler.py
```diff
@@ -51,13 +50,13 @@ def setUp(self):
         self.generator = AzureReportDataGenerator(self.tenant)
         self.generator.add_data_to_tenant()
 
-    def get_totals_by_time_scope(self, aggregates, filter=None):
+    def get_totals_by_time_scope(self, aggregates, filters=None):
```
blentz: This argument rename is to make the linter happy, so that we're not overriding the built-in `filter` keyword.

koku/api/report/aws/view.py
```diff
@@ -27,10 +27,10 @@ class AWSView(ReportView):
...
 
     permission_classes = [AwsAccessPermission]
-    provider = 'aws'
-    _serializer = QueryParamSerializer
-    _query_handler = AWSReportQueryHandler
-    _tag_handler = [AWSTagsSummary]
+    provider = 'AWS'
+    serializer = QueryParamSerializer
+    query_handler = AWSReportQueryHandler
+    tag_handler = [AWSTagsSummary]
```
blentz: This bit is me fixing an incorrect choice I made in a previous PR. These attributes are not private, particularly with how `QueryParameters` needs to access them. So, I'm removing the leading `_` and making these public attributes.

dccurtis:
I just ran through all of the API tests and this change didn't introduce any new regressions.  Same results vs master.

blentz:
Functional test results:

[960_queryhandler.txt](https://github.com/project-koku/koku/files/3735678/960_queryhandler.txt)
[master.txt](https://github.com/project-koku/koku/files/3735679/master.txt)

```
blentz▌(e) koku│~▌130▌diff -u master.txt 960_queryhandler.txt 
blentz▌(e) koku│~▌
```

PR #1251, opened 2019-10-11
Allow long AWS product names

adberglund:
## Summary 
We are failing to load AWS data on a character limit.


dccurtis:
Might have been too quick on the approval, but would a TextField be ok here?

PR #1256, opened 2019-10-11
Use textfield instead.

adberglund:


adberglund:
Can't cherry pick migrations. Oops.

PR #1258, opened 2019-10-11
Alter column memory on product to be text

adberglund:
## Summary
We can leave this open in case we hit others, but didn't want to forget it. Turns out this column can be... whatever it wants to be depending on the service it comes from.


adberglund:
Believe this will actually just require a change to the processor to parse the value out

PR #1272, opened 2019-10-16
Fix syncer paths

infinitewarp:


koku/api/dataexport/syncer/__init__.py
```diff
@@ -10,6 +11,7 @@
 from django.utils.translation import gettext as _
 
 LOG = get_task_logger(__name__)
+_PROVIDER_TYPES = ['aws', 'ocp', 'azure']
```
adberglund: A note for the future: 

We're tracking a list of provider types in a few places right now. Currently it's not super convenient to get at the list in the Providers model as it's a list of tuples (see https://github.com/project-koku/koku/blob/master/koku/api/provider/models.py#L110-L121) 

I also just did this last week: https://github.com/project-koku/koku/blob/master/koku/api/provider/serializers.py#L33

I think this is prime for future refactor where we move the tracked list of providers to the provider model file (or a similar common place) and always importing from there.

Issue #1269, opened 2019-10-16
Azure Conflict Columns are not sufficient


dccurtis:
The Azure conflict columns are not sufficient in processing Azure line item data.

Report in Azure portal to reproduce (10/16/19) `cost/costreport/20191001-20191031/costreport_82b55d01-5b53-449c-93d8-35f4b8b1ad90.csv`

```
koku_db           | 2019-10-16 19:48:15.579 UTC [77] ERROR:  ON CONFLICT DO UPDATE command cannot affect row a second time
koku_db           | 2019-10-16 19:48:15.579 UTC [77] HINT:  Ensure that no rows proposed for insertion within the same command have duplicate constrained values.
koku_db           | 2019-10-16 19:48:15.579 UTC [77] STATEMENT:
koku_db           | 	            INSERT INTO reporting_azurecostentrylineitem_daily (subscription_guid,usage_date_time,usage_quantity,pretax_cost,offer_id,tags,cost_entry_bill_id,cost_entry_product_id,meter_id)
koku_db           | 	                SELECT subscription_guid,usage_date_time,usage_quantity,pretax_cost,offer_id,tags,cost_entry_bill_id,cost_entry_product_id,meter_id
koku_db           | 	                FROM reporting_azurecostentrylineitem_daily_9c4b132a_b6be_4f5c_825c_fa7e0eb4816f
koku_db           | 	                ON CONFLICT (cost_entry_bill_id,cost_entry_product_id,meter_id,subscription_guid,usage_date_time) DO UPDATE
koku_db           | 	                SET subscription_guid = excluded.subscription_guid,usage_date_time = excluded.usage_date_time,usage_quantity = excluded.usage_quantity,pretax_cost = excluded.pretax_cost,offer_id = excluded.offer_id,tags = excluded.tags,cost_entry_bill_id = excluded.cost_entry_bill_id,cost_entry_product_id = excluded.cost_entry_product_id,meter_id = excluded.meter_id
```

lcouzens:
Verified.
Commit: 96a12fc0e55ee08d444f5919b549ec575683a988

PR #1280, opened 2019-10-18
No longer use conflict columns to upsert Azure data

dccurtis:
The conflict columns approach to processing Azure data has proven to be insufficient for Azure data.  This change adopts the mechanism that AWS uses for processing line item data.

**Testing**
1.  Ran all API QE tests
2. Ingested multiple exports of Azure data and verified no duplication was occurring.

Addresses #1269


koku/masu/processor/report_processor_base.py
```diff
@@ -38,7 +41,7 @@ class ReportProcessorBase():
     Base object class for downloading cost reports from a cloud provider.
     """
 
-    def __init__(self, schema_name, report_path, compression, provider_id, processed_report):
+    def __init__(self, schema_name, report_path, compression, provider_id, manifest_id, processed_report):
```
blentz: This constructor is getting a bit big. It might be good to consider using kwargs or passing in an object.

koku/masu/database/azure_report_db_accessor.py
```diff
@@ -164,3 +165,11 @@ def populate_markup_cost(self, markup, bill_ids=None):
             else:
                 AzureCostEntryLineItemDailySummary.objects.\
                     update(markup_cost=(F('pretax_cost') * markup))
+
+    def get_lineitem_query_for_billid(self, bill_id):
+        """Get the Azure cost entry line item for a given bill query."""
+        table_name = AzureCostEntryLineItemDaily
```
maskarb: can we make this `get_daily_query_for_billid`?
In AWS, we have:
```
get_lineitem_query_for_billid:
    table_name = AWSCostEntryLineItem
get_daily_query_for_billid:
    table_name = AWSCostEntryLineItemDaily
```

blentz: Agreed. :+1:

dccurtis: So this is where Azure departs from AWS and OCP.  For Azure the line item is already daily (We don't have a separate Daily table for Azure).  I made this line item to be more explicit that it's the line item data (and it just happens to have a Daily the table name to make it clear that it's daily data.  Obviously the naming conventions of the tables was not successful :) But hopefully that makes sense.  Let me know if it doesn't

PR #1281, opened 2019-10-18
Add report processing for GCP

werwty:


koku/reporting/provider/gcp/models.py
```diff
@@ -0,0 +1,54 @@
...
+
+    """
+
+    class Meta:
+        """Meta for GCPCostEntryBill."""
+
+        unique_together = ('billing_period_start', 'provider')
+
+    billing_period_start = models.DateTimeField(null=False)
+    billing_period_end = models.DateTimeField(null=False)
```
katherine-black: Minor style issue, null by default is false.

koku/reporting/provider/gcp/models.py
```diff
@@ -0,0 +1,54 @@
...
+    summary_data_creation_datetime = models.DateTimeField(null=True)
+    summary_data_updated_datetime = models.DateTimeField(null=True)
+    finalized_datetime = models.DateTimeField(null=True)
+    derived_cost_datetime = models.DateTimeField(null=True)
```
katherine-black: We should consider adding `blank=True` here just in case we ever want to rely on form validation in Django.

koku/reporting/provider/gcp/models.py
```diff
@@ -0,0 +1,54 @@
...
+
+    provider = models.ForeignKey(Provider, null=True, on_delete=models.CASCADE)
+
+
+class GCPProject(models.Model):
+    """The per Project information for GCP."""
+
+    account_id = models.CharField(max_length=20)
+    project_number = models.BigIntegerField()
+    project_id = models.CharField(null=False, unique=True, max_length=256)
```
katherine-black: As above `null=False` does nothing here

koku/reporting/provider/gcp/models.py
```diff
@@ -0,0 +1,54 @@
...
+    project = models.ForeignKey(GCPProject, on_delete=models.PROTECT)
+    measurement_type = models.CharField(max_length=512, null=False)
+    consumption = models.BigIntegerField()
+    unit = models.CharField(max_length=63, null=True)
+    cost = models.DecimalField(max_digits=17, decimal_places=9,
+                               null=True)
+    currency = models.CharField(max_length=10, null=False)
+    description = models.CharField(max_length=256,)
+    start_time = models.DateTimeField(null=False)
+    end_time = models.DateTimeField(null=False)
```
katherine-black: Same as all the other `null`/`blank` suggestions above.

koku/masu/processor/gcp/gcp_report_processor.py
```diff
@@ -0,0 +1,184 @@
...
+            project (reporting.provider.gcp.models.GCPProject): A GCP Project
+
+        """
+        data = self._get_data_for_table(row, GCPCostEntryLineItemDaily._meta.db_table)
+        if not GCPCostEntryLineItemDaily.objects.filter(
+                project=project,
+                line_item_type=data['line_item_type'],
+                start_time=data['start_time']
+        ).exists():
+            line_item = GCPCostEntryLineItemDaily(
```
adberglund: Processing line items using the ORM is going to kill us on performance. Right now for AWS and Azure we are using a bulk csv read method which is about as fast as you can load bulk data into the DB. See https://github.com/project-koku/koku/blob/master/koku/masu/database/report_db_accessor_base.py#L166

We are already hitting some processing time concerns with AWS. I did some testing with this in the AWS processor early this week with a large file, trying the ORM just for a benchmark. 

With the bulk upload I was able to process the file in `224 seconds`
With the ORM and the exact same bulk create setup you have here I went up to `729 seconds`. 

So I think we really, really need to stick with the csv upload method.

werwty: @adberglund AWS has monthly data right? For GCP data comes in daily csvs.
Since each batch of data to process is going to be a lot smaller would this still be a concern for GCP?

adberglund: @werwty Really depends on the account and the data source. So far in general we have under-estimated account sizes and that is currently biting us. 

It also looks like we might need to do the BigQuery export to get true line item data and all of the columns we will likely need. That will bump up the file size, too.

werwty: I suppose that's fair enough. I'll rewrite this to use `bulk_insert_rows`

werwty:
Demo: https://asciinema.org/a/raTfrDtPdwc4THzhg3BnIHbJG

@katherine-black @adberglund  Can I get another review on this?

Some implementation notes:

In a recent PR the Azure processor stopped using conflict columns to insert line items:
https://github.com/project-koku/koku/pull/1280
Unfortunately this does not work for GCP. That implementation relies on 
`self._delete_line_items(AzureReportDBAccessor, self.column_map)` getting called when processing begins, which deletes all lines items in a manifest report period for a provider.

For this to work with GCP the line items that needs to get deleted is any matching the `Start Time` and `Project ID`. The `Start Time` is not the manifest start time, the manifest is per month, where the `Start Time` for a report is by day. We do not know these values until we start processing the report. And these values could differ in the report (we have reports with 2 days worth of data, and reports with multiple projects in it). If we remodeled the manifest to be daily, this implementation is viable. 

Also during my work on this I discovered that GCP has duplicate line items in some cases:
```
01C2AB-2F30E0-1EF054 | com.google.cloud/services/compute-engine/VmimageG1Small | 2019-09-30T00:00:00-07:00 | 2019-10-01T00:00:00-07:00 | 244082564888 | com.google.cloud/services/compute-engine/VmimageG1Small | 15276 | seconds | FreeTrial:Credit-01C2AB-2F30E0-1EF054 | -0.109053 | USD | 0.109053 | USD | 244082564888 | fine-way-252818 | My First Project |   | Small Instance with 1 VCPU running in Americas

01C2AB-2F30E0-1EF054 | com.google.cloud/services/compute-engine/VmimageG1Small | 2019-09-30T00:00:00-07:00 | 2019-10-01T00:00:00-07:00 | 244082564888 | com.google.cloud/services/compute-engine/VmimageG1Small | 1 | seconds | FreeTrial:Credit-01C2AB-2F30E0-1EF054 | -7E-06 | USD | 7E-06 | USD | 244082564888 | fine-way-252818 | My First Project |   | Small Instance with 1 VCPU running in Americas
```

The merge temp table method would only work with all non-duplicate rows. So some work is done in process to consolidate these.

koku/masu/processor/gcp/gcp_report_processor.py
```diff
@@ -0,0 +1,250 @@
...
+        report_date_range = utils.month_date_range(parser.parse(start_time))
+        start_date, end_date = report_date_range.split('-')
+
+        start_date_utc = parser.parse(start_date).replace(hour=0, minute=0, tzinfo=pytz.UTC)
+        end_date_utc = parser.parse(end_date).replace(hour=0, minute=0, tzinfo=pytz.UTC)
+
+        data = {
+            'billing_period_start': datetime.strftime(start_date_utc, '%Y-%m-%d %H:%M%z'),
+            'billing_period_end': datetime.strftime(end_date_utc, '%Y-%m-%d %H:%M%z'),
+            'provider_id': self._provider_id,
```
adberglund:
```suggestion
            'provider_id': self._provider_uuid,
```

koku/masu/processor/gcp/gcp_report_processor.py
```diff
@@ -0,0 +1,250 @@
...
+            report_path (str): Where the report file lives in the file system
+            compression (CONST): How the report file is compressed.
+                Accepted values: UNCOMPRESSED, GZIP_COMPRESSED
+
+        """
+        super().__init__(
+            schema_name=schema_name,
+            report_path=report_path,
+            compression=compression,
+            provider_id=provider_id,
```
adberglund:
```suggestion
            provider_uuid=provider_uuid,
```

koku/masu/processor/gcp/gcp_report_processor.py
```diff
@@ -0,0 +1,250 @@
...
+        """Clear a batch of rows after they've been saved."""
+        self.line_items = []
+        self.unique_line_items = {}
+
+
+class GCPReportProcessor(ReportProcessorBase):
+    """Cost Usage Report processor."""
+
+    def __init__(self, schema_name, report_path, compression,
+                 provider_id, manifest_id=None):
```
adberglund:
```suggestion
                 provider_uuid, manifest_id=None):
```

koku/masu/database/koku_database_access.py
```diff
@@ -184,10 +183,10 @@ def savepoint(self, func, *args, **kwargs):
         """
         with schema_context(self.schema):
             try:
-                sid = django_savepoint()
+                sid = transaction.savepoint()
```
infinitewarp: Is this a holdover from your earlier implementation iteration that attempted to use proper Django model objects? Are these transaction-related changes still necessary?

werwty: I'm cleaning up some of the imports since they were originally:
```
from django.db import transaction
from django.db.transaction import get_connection, savepoint_commit, savepoint_rollback
from django.db.transaction import savepoint as django_savepoint
```
This is still functionally the same, just cleaner.

koku/reporting/provider/gcp/models.py
```diff
@@ -0,0 +1,56 @@
...
+
+    class Meta:
+        """Meta for GCPCostEntryLineItemDaily."""
+
+        unique_together = ('start_time', 'line_item_type', 'project')
+
+    line_item_type = models.CharField(max_length=256)
+    cost_entry_bill = models.ForeignKey(
+        GCPCostEntryBill, on_delete=models.PROTECT)
+    project = models.ForeignKey(GCPProject, on_delete=models.PROTECT)
```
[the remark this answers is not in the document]
werwty: You are correct- this has now been set to `CASCADE`. Thanks!

Issue #1284, opened 2019-10-21
API to return our cloud account ID for use in sources


adberglund:
## User Story
As a sources UI engineer I would like to have an API to call to get the AWS account ID for cost management so that I do not have to hard code an account ID I do not own.

## Assumptions
- Our sources flow has a user setting up an IAM role in *their* AWS account and allowing *our* AWS access to that role via the boto3 assume_role command.
- We currently just hard code our AWS account ID in the cost management sources wizard, but with that workflow now moving to platform sources, we should really be giving that account ID via an API. 

## API Details

| API  | Detail|
 | --- | --- |
 | **Method** | GET |
 | **Path** | /api/v1/cloudaccounts/?type=AWS |
 | **Body** | N/A |
 | **Response** | `{"account_id": "9999999999"}` |

## Acceptance Criteria

- [ ] The API returns the correct account ID for our cost management AWS account
- [ ] The API is filterable by source type (although we only need AWS today)

nbon12:
@adberglund a new table will be created called Attributes, with the following columns

- Name
- Value
- Description

The first row entry would be the tuple <"aws_account_id", "XXXXXXXXX", "Cost Management's AWS Account ID">

A user can make a GET request to /api/v1/attributes/?name=aws_account_id and receive the following response in the payload: {"aws_account_id", "XXXXXXXXX", "Cost Management's AWS Account ID"}

lcouzens:
@nbon12 I am goner move this back for a minor tweak. Me being really picky, but all our current endpoints use `-` over `_` so we should probably keep consistent here too.

For example:
`cost-management`
`instance-types`

nbon12:
@lcouzens I merged the PR to rename the endpoint and moved the issue to QA.

lcouzens:
Verified
commit: 9418759f4f87a6e7b87ff2b356cb32ca583a9478

PR #1295, opened 2019-10-22
Feature/#1284

nbon12:
Add API for getting AWS Account ID
Issue #1284 

- New table for "CloudAccounts" is created. The purpose of this table is to store cloud account IDs, and any other name-value pair.

- The columns are "Name", "Value", "Description"

- The first row in this table is the tuple: <"aws_account_id", "yourawsaccountidhere", "Cost Management's AWS Account ID">
- The only allowed operation is GET. To modify this value, Cost Management must log in to the postgres database directly.

Screen Shot 2019-10-25 at 3 18 49 PM


nbon12:
Screen Shot 2019-10-22 at 9 44 47 AM

blentz:
I'm not sure we need a whole new table for this. Or, perhaps we need to rework our existing Provider table to address this particular issue.

Within the role ARN that we already have is the account number.

The ARN format is this:
`arn:partition:service:region:account:resource`

So, when we get an ARN, like this:  `arn:aws:iam::123456789012:role/S3Access`, the account number is already embedded in the ARN. All we have to do is parse the ARN and extract the account number. 

We already have an ARN parsing class in our unit testing code (see: koku/masu/test/util/aws/test_common.py ). 

We can move that and use it for this purpose: to extract the Account number from the ARN we already have. Whether we store the account number separately in its own field, or just reparse ARNs whenever we need the account number is an open question we could talk about further.

koku/attribute/apps.py
```diff
@@ -0,0 +1,23 @@
+#
+# Copyright 2018 Red Hat, Inc.
```
dccurtis:
```suggestion
# Copyright 2019 Red Hat, Inc.
```

koku/attribute/migrations/0001_initial.py
```diff
@@ -0,0 +1,25 @@
...
+    operations = [
+        migrations.CreateModel(
+            name='Attribute',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('name', models.CharField(help_text='The name of the attribute', max_length=255)),
+                ('value', models.TextField(null=True)),
+                ('description', models.TextField(null=True)),
+                ('updated_timestamp', models.DateTimeField(auto_now=True, null=True)),
+                ('created_timestamp', models.DateTimeField(auto_now_add=True, null=True)),
```
dccurtis: What's the reason for keeping track of two dates?  I can see maybe keeping one timestamp when the field is updated.

koku/attribute/models.py
```diff
@@ -0,0 +1,28 @@
+#
+# Copyright 2018 Red Hat, Inc.
```
dccurtis:
```suggestion
# Copyright 2019 Red Hat, Inc.
```

koku/attribute/tests/test_model.py
```diff
@@ -0,0 +1,33 @@
+#
+# Copyright 2018 Red Hat, Inc.
```
dccurtis:
```suggestion
# Copyright 2019 Red Hat, Inc.
```

koku/attribute/urls.py
```diff
@@ -0,0 +1,28 @@
+#
+# Copyright 2018 Red Hat, Inc.
```
dccurtis:
```suggestion
# Copyright 2019 Red Hat, Inc.
```

koku/attribute/views.py
```diff
@@ -0,0 +1,28 @@
+#
+# Copyright 2018 Red Hat, Inc.
```
dccurtis:
```suggestion
# Copyright 2019 Red Hat, Inc.
```

koku/koku/settings.py
```diff
@@ -86,6 +86,7 @@
     'reporting_common',
     'cost_models',
     'sources',
+    'attribute'
```
dccurtis: I think this should probably live in the api app instead of a new Django app.

koku/attribute/urls.py
```diff
@@ -0,0 +1,28 @@
...
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+
+"""URLs for Attribute."""
+from rest_framework.routers import DefaultRouter
+from .views import AttributeViewSet
+from django.conf.urls import include, url
+
+ROUTER = DefaultRouter()
+ROUTER.register(r'attributes', AttributeViewSet, base_name='attributes')
```
dccurtis: the URL name `/attributes` isn't very descriptive as to what this API is for.

nbon12:
Sure, name it whatever you want. I didn’t want to name it myself so I picked a generic name that can mean anything so that people can name things later, but if you want to name it in the codebase itself instead of giving ourselves the flexibility of naming things in the database, that’s fine too.

nbon12:
> A few things.
> 
> The issue in 1284 is trying to address a specific "customer" requirement (even if the sources frontend team is our customer in this case) and I think we should try to focus specifically on that need. Unless you have good evidence for what other attributes we would be storing, I think we're over-complicating this issue by generalizing it.
> 
> Also, please move the API to the api app in django, not its own app.

@adberglund  What's the reason for including this in the api app instead of its own app?

dccurtis:
@nbon12 I think you have to add the new endpoint here https://github.com/project-koku/koku/blob/09acee78d401a4656f555bab528eaf74bdabf449/koku/koku/middleware.py#L48

This shouldn't have any RBAC authorization since it's going to be called as an implementation detail to the frontend.

Hopefully this will fix the failing tests.

adberglund:
Let's add the account ID we want to be returned by the API. Check out https://docs.djangoproject.com/en/2.2/topics/migrations/#data-migrations for how to use a Django data migration to do this.

dlabrecq:
We would probably call this API lazily, when needed for the sources wizard. However, since the sources creation is being ported by the Catalog team, I don't know where else the koku-ui would use this? Regardless, the koku-i should cache all API calls for 30 min.

PR #1305, opened 2019-10-24
release_2019.10.24.0

adberglund:
## Summary
Primarily merging into stable to get OCP on AWS report type fix released.


adberglund:
Will re-open a PR at a specific commit so this doesn't keep growing larger with master

Issue #1310, opened 2019-10-25
Modify Sources Client integration service to use a provided cluster-id instead of sources uuid


dccurtis:
## User Story
As a user, I want my OCP cluster to be added when on boarding a new OCP cluster through Platform Sources.

The Sources Client integration service is currently coded to use the Sources UUID as the cluster ID.  After discussions with members from UX and Platform Sources, we decided to keep the OCP on boarding flow the same as it is today where a cluster ID from Korekuta is provided to Platform Sources UI to onboard the cluster into Cost Management.

## Impacts
API, UI, Docs

## Assumptions and Questions
- Same Korekuta flow as today (however, a new issue will be opened to replace the UUID generated today with a cluster uuid returned from the OCP cluster resource object)
- The Koku-Sources client flow will match that of AWS and Azure, where the frontend will have to supply the cluster id directly to the koku micro service instead of going to the platform sources service.
- Sources Client service API endpoints need to be enhanced to allow for an API call to supply the cluster_id for OCP source types.  Most obvious candidate for this is the existing endpoint `<koku-sources-client>/authentication` with the JSON data`{"source_id": 3, "credentials": {"cluster_id": "2639de71-ca37-4a17-a104-17665a50e7fc"}}`

## API Details

| API  | Detail|
 | --- | --- |
 | **Method** POST |
 | **Path** | <koku-sources-client>/api/v1/authentication/ |
 | **Body** |  {"source_id": 3, "credentials": {"cluster_id": "2639de71-ca37-4a17-a104-17665a50e7fc"}} |
 | **Response** | {"source_id": 3, "credentials": {"cluster_id": "2639de71-ca37-4a17-a104-17665a50e7fc"}} |

## UI Details
Platform Sources UI will enhance the Cost Management specific wizard pages for OCP type to send the cluster-id value to the Koku micro service.  Same as what is being done for AWS

## Acceptance Criteria

- [ ] An OCP cluster can be onboard to cost management via Platform Sources + Koku Sources Client /authentication endpoint
- [ ] All standard cost management Source CRUD works.

dccurtis:
After thinking this over some and discussing with Platform Sources this morning, I think we will be storing the cluster ID on the Platform Sources side since Platform Sources will ultimately integrate with OCM.

If we decide to go this direction, there will be no `/authentication` work but instead be only Sources Client updates to fetch the "OCM" (korekuta provided) cluster_id from platform sources.

lcouzens:
Verified
[...]

PR #1319, opened 2019-10-28
Fetching Sources imported field for OCP cluster ID if available.

dccurtis:
Current implementation is to use the Source guid as the OCP cluster ID in cost management.  This change allows for a user supplied cluster-id to be associated with an OCP Platform Source.

This ID would be returned from [Korekuta](https://github.com/project-koku/korekuta) and entered by the user on the Platform Sources UI when associated Cost Management to the OpenShift source.

I kept the Source guid fallback in place so that any attempts to create an OCP Cost Source in CI would be "successful".  Once the front end implements the OCP Cost Management pages we could remove the fallback.

@agrare we discussed this briefly but I wanted to get your sign off before proceeding.  Here's an example of what the sources table would look like for an OCP source with and without a provided cluster id.  Do you want this `imported` field to be a more standard json structure to be more descriptive or is the simple UUID ok? 

```
sources_api_development=# select * from sources;
 id |            name            |                 uid                  |         created_at         |         updated_at         | tenant_id | source_type_id | version | availability_status |               imported
----+----------------------------+--------------------------------------+----------------------------+----------------------------+-----------+----------------+---------+---------------------+--------------------------------------
  1 | Test OCP Source cluster-id | b6255061-4330-4a5e-9214-62ebffc5dde1 | 2019-10-28 20:53:48.908224 | 2019-10-28 20:53:48.908224 |         1 |              1 |         |                     | 8770d031-9b46-40d9-be12-52929af682f5
  2 | Test OCP Source            | 3adbef49-e9f3-4cb9-bee3-3d834c837bd3 | 2019-10-28 20:54:58.993587 | 2019-10-28 20:54:58.993587 |         1 |              1 |         |                     |
(2 rows)
```

**Testing**
1. Create OCP Source with cluster-id provided.
2. Create OCP source without cluster-id provided.  Verify it falls back to Source uid.
3. Verify that the Sources DB has entries as expected

**Test Results**
[ocp_sources_cluster_id_ut.txt](https://github.com/project-koku/koku/files/3795775/ocp_sources_cluster_id_ut.txt)

Addresses #1310


agrare:
The value of `imported` is intended to be what other service it was imported from for display purposes.  If we need a "native" uuid we could add another column for that.

agrare:
@dccurtis If it is the main identifying uuid of an OCP cluster then it should probably go in the sources table.  If you want to put a PR in to sources that'd be great, I think something like `source_uid` (we have `source_ref` for the "native" reference in other places) or `native_uid` would work.

dccurtis:
Sources PR https://github.com/ManageIQ/sources-api/pull/117

PR #1332, opened 2019-10-30
Fetch batches

infinitewarp:
This is necessary because the previous `fetchall` gathered _all_ of the results _immediately_ to build a complete list in memory, but I had (incorrectly) expected that call to act as a generator for large result sets. This means that Celery workers die (or are `SIGKILL`ed) as they exceed their memory limitations.

Ideally, we would fetch large sets of records using named cursors, and that should _internally_ work in batches "for free", but I couldn't reasonably figure out how to make named cursors play nicely with transactions plus our django-tenant-schemas implementation.

For now, I'm just batching results inside a transaction in a somewhat manual way. If someone else here has a better or cleaner implementation suggestion, I'm happy to hear it.

### DIFF
--- a/koku/masu/util/upload.py
+++ b/koku/masu/util/upload.py
@@ -10,6 +10,7 @@ from api.dataexport.uploader import AwsS3Uploader
 from masu.util.common import NamedTemporaryGZip
 
 LOG = logging.getLogger(__name__)
+_DB_FETCH_BATCH_SIZE = 2000
 
 
 def get_upload_path(account_name, provider_type, date, table_name, daily=False):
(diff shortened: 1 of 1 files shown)


koku/masu/util/upload.py
```diff
@@ -81,20 +82,25 @@ def query_and_upload_to_s3(schema, table_export_setting, date_range):
...
+                        'end_date': the_date if iterate_daily else end_date,
+                    },
+                )
+                # Don't upload if result set is empty
+                if cursor.rowcount == 0:
+                    return
+                with NamedTemporaryGZip() as temp_file:
+                    writer = csv.writer(temp_file, quotechar='"', quoting=csv.QUOTE_MINIMAL)
+                    writer.writerow([field.name for field in cursor.description])
+                    while True:
```
blentz: You could make this a touch more compact, if you want.

```
            while cursor.rowcount > 0:
                with NamedTemporaryGZip() as temp_file:
                    writer = csv.writer(temp_file, quotechar='"', quoting=csv.QUOTE_MINIMAL)
                    writer.writerow([field.name for field in cursor.description])
```

infinitewarp:
We closed this branch and PR and created new ones because we wanted to preserve the failed Travis job logs for analysis.

Issue #1315, opened 2019-10-28
split archive data on provider id, not just type


infinitewarp:
## User Story
As a Koku customer, I want my exported data archive to be split on provider id so I can analyze data at the provider level.

We currently save all of the data at the provider _type_ level (e.g. `aws`). We will need to change the way we query the data, the path where the data is stored, and the path where the data is fetched+copied to the customer's S3 bucket.

S3 file paths should look like `acct10001/aws/{provider_id}/2019/10/12/foo.csv`.

## Assumptions and Questions
- Maybe a future enhancement: Allow the customer to ask for only a specific provider ID, not "give me everything within this time".

## Acceptance Criteria

- [ ] When the scheduled job runs to archive data daily, the files we write to Koku's S3 bucket include the relevant provider ID in the path.
- [ ] When the customer requests their data export and we copy files to the requested bucket, we include the appropriate provider IDs in the paths of the copied files.
  - This means we probably need to iterate in code the provider IDs so we can define appropriate S3 path prefixes.

lcouzens:
Verified

PR #1352, opened 2019-11-04
split data export archive paths on provider uuid

infinitewarp:
for #1315

### DIFF
--- a/koku/api/dataexport/syncer/__init__.py
+++ b/koku/api/dataexport/syncer/__init__.py
@@ -1,4 +1,5 @@
 """Data export syncer."""
+import uuid
 from abc import ABC, abstractmethod
 from datetime import timedelta
 from itertools import product
@@ -107,6 +108,15 @@ class AwsS3Syncer(SyncerInterface):
 
         """
         if settings.ENABLE_S3_ARCHIVING:
+            log_uuid = str(uuid.uuid4())
+            LOG.info(
+                '%s Beginning sync_bucket to %s for %s from %s to %s',
+                log_uuid,
+                s3_destination_bucket_name,
+                schema_name,
+                date_range[0],
+                date_range[1],
+            )
             start_date, end_date = date_range
             # rrule is inclusive for both dates, so we need to make end_date exclusive
             end_date = end_date - timedelta(days=1)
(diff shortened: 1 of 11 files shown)


koku/masu/util/upload.py
```diff
@@ -13,13 +13,16 @@
...
 
-def get_upload_path(account_name, provider_type, date, table_name, daily=False):
+def get_upload_path(
+    account_name, provider_type, provider_uuid, date, table_name, daily=False
+):
     """
     Get the s3 upload_path for a file.
 
     Args:
         account_name (str): Koku user account (schema) name.
```
werwty: Small suggestion: s/account_name/schema_name/

infinitewarp:
demo of current functionality: https://asciinema.org/a/279862

koku/api/dataexport/syncer/__init__.py
```diff
@@ -94,40 +97,70 @@ def _copy_object(self, s3_destination_bucket, source_object):
...
                 for source_object in self.s3_source_bucket.objects.filter(
-                    Prefix=(
-                        f'{settings.S3_BUCKET_PATH}/{account}/{provider_type}/'
-                        f'{day.year:04d}/{day.month:02d}/{day.day:02d}/'
-                    )
+                    Prefix=prefix
                 ):
                     self._copy_object(s3_destination_bucket, source_object)
+
+            log_uuid = str(uuid.uuid4())
```
werwty: Is the point of this uuid to keep track of a single sync task?
If so shouldn't we be using the uuid declared above?

koku/masu/celery/tasks.py
```diff
@@ -228,15 +75,29 @@ def upload_normalized_data():
     accounts, _ = Orchestrator.get_accounts()
 
     # Deduplicate schema_name since accounts may have the same schema_name but different providers
```
werwty: This comment is no longer valid. Accounts is built from the list of providers: https://github.com/project-koku/koku/blob/master/koku/masu/external/accounts/db/cur_accounts_db.py#L76 so we don't need deduplication. 

You could remove
```
    schema_providers = set(
        (account['schema_name'], account['provider_uuid']) for account in accounts
    )
```
and iterate through the accounts. This save only a marginal amount of time so I completely understand if you want to leave as is.

koku/api/dataexport/syncer/__init__.py
infinitewarp: Good catch! Fixing this in the next push.

koku/masu/celery/tasks.py
infinitewarp: TIL! I'll prune this out with the next push.

Issue #1354, opened 2019-11-05
Push source status to Platform Sources on create and edit


dccurtis:
## User Story
As user, I want my source status to be available to the platform when I create and edit a cost management source.

Platform Sources has recently provided a mechanism for applications to supply source status.  The two fields to update are availability_status and availability_status_error (https://github.com/ManageIQ/sources-api/blob/92e3fa6c9afe1bbae93b1b067ee12a45078e65f5/public/doc/openapi-3-v1.0.0.json#L1193) on an Application object.

## Impacts
API, UI, Docs

## Assumptions
- For the first pass we can supply the detail string for /provider 400 responses as the detailed error message.  A docs/ux should weigh in on this to make sure the language is consistent across other apps and we are not inadvertently revealing any sensitive information.
- I will bring up with the platform team whether or not the error message string will change in the future to support localization.  In that case it could be that we will return an error-id with relevant arguments and the platform could retrieve the string appropriate for the user's locale.

## UI Details
UX mocks are underway for the Platform to determine how the error status messages should be displayed for the user.

## Acceptance Criteria

- [ ] Creating a source with invalid cost management details should fail Koku-provider creation and the relevant status should be saved in platform sources
- [ ] Editing a successfully created cost management source with bad information should result in the Koku-provider being marked as inactive as well as the appropriate error status being saved in platform sources.  The source should be edited again with correct information to make the provider to be active with the "available" status in platform sources.

lcouzens:
Verified
Commit: 5f65d324af3f801b7f2ee6f20f7a6da5ba6162d7

PR #1355, opened 2019-11-05
Push cost management application status to platform on create/edit

dccurtis:
Platform Sources recently enhanced their API to allow for applications to set source status (https://github.com/ManageIQ/sources-api/pull/116)

This change make the Sources Client push the source status to the platform on source create and edit.  The error messages being saved on the platform is currently the detailed string from the `/provider` 400 error response.  I will be opening a docs/UX issue to review these messages to ensure they are consistent with the platform and we are not inadvertently revealing any sensitive information.

**Testing**
1. Add an AWS Source with invalid bucket name and veirfy that the platform sources /application response has status strings set correctly and the koku provider has not been created yet.
2.  Edit the source with the correct bucket name and verify the koku provider is created (and is active).  Verify that the avaiable status is set on platform sources
3.  Edit the AWS source with the wrong bucket.  Verify status is updated in platform sources and that the koku provider is marked not active.
4.  Edit bucket again with correct bucket and verify provider is active and platform sources status is avaiable.

**Test Results**
[sources_push_status_ut.txt](https://github.com/project-koku/koku/files/3809378/sources_push_status_ut.txt)

Addresses #1354


koku/sources/koku_http_client.py
```diff
@@ -115,6 +115,17 @@ def get_billing_source_for_provider(self, provider_type, billing_source):
         if provider_fn:
             return provider_fn(billing_source)
 
+    def _handle_bad_requests(self, response):
+        """Helper method to raise exceptions with error message string for Platform Sources."""
+        if response.status_code == 401 or response.status_code == 403:
+            raise KokuHTTPClientNonRecoverableError('Insufficient Permissions')
+        if response.status_code == 400:
```
blentz: It might be good to have this catch any 4xx code:

```suggestion
        elif response.status_code >= 400 and response.status_code <= 499:
```

dccurtis: So the 401 and 403 is returning an `Insufficient Permissions` string.
The actual error message string that we want to return is from a 400 (message from the koku provider method `cost_usage_source_is_reachable -> serializers.ValidationError(error_obj(key, message))`)

For the `create_provider` flow if it gets passed this check it is looking for any status code that's not the expected path
```
        self._handle_bad_requests(r)

        if r.status_code != 201:
            raise KokuHTTPClientNonRecoverableError(str(r.status_code))
```

Similar behavior on the `update_provider` method only there is also a `404` check

By the way, I'm working with docs/ux to review all of these strings and was planning on coming back and updating accordingly.

blentz: That works for me! ^_^

Issue #1241, opened 2019-10-10
OpenShift and OpenShift on AWS Returning Slightly Different Costs


adberglund:
**Describe the bug**
After merging the change to use a single reporting daily summary table for OpenShift (bypassing the need for the costs summary table) our OpenShift and OpenShift on AWS APIs are returning slightly different costs. 

**To Reproduce**
1. Hit api/cost-management/v1/reports/openshift/costs/?filter[time_scope_value]=-1
2. Hit api/cost-management/v1/reports/openshift/infrastructures/aws/costs/?filter[time_scope_value]=-1
  - this one is probably the accurate view of the data
3. Observe infrastructure cost!

**Expected behavior**
We should have the same infrastructure cost on both APIs.

@adberglund has volunteered to walk through and explain the SQL behind these results if you have any questions.

In the even that this is an API issue, here are a few links to places to look. 

- OCP execute query method: https://github.com/project-koku/koku/blob/master/koku/api/report/ocp/query_handler.py#L115
- OCP provider map where aspects of the query are defined: https://github.com/project-koku/koku/blob/master/koku/api/report/ocp/provider_map.py
- OCP AWS execute query: https://github.com/project-koku/koku/blob/master/koku/api/report/ocp_aws/query_handler.py#L57
- OCP AWS provider map: https://github.com/project-koku/koku/blob/master/koku/api/report/ocp_aws/provider_map.py

In the even that this is a SQL bug

- The data in this table to be the source of truth for cost since it is, per the group by aws_id, it should not have any duplicates of AWS line item costs: https://github.com/project-koku/koku/blob/master/koku/masu/database/sql/reporting_ocpawscostlineitem_daily_summary.sql#L1129
- Check the data in this table to make sure the costs are the same: https://github.com/project-koku/koku/blob/master/koku/masu/database/sql/reporting_ocpawscostlineitem_daily_summary.sql#L1197
- Finally check this block to see if the cost discrepancy is occurring at this step: https://github.com/project-koku/koku/blob/master/koku/masu/database/sql/reporting_ocpcosts_summary.sql#L1-L55

lcouzens:
Verified

PR #1366, opened 2019-11-06
Increase precision of decimals, standardize stored decimals to have 9 digits of precision, calculated ones to have 15.

werwty:
- Standardize decimals to have 24 max digits, and 9 decimal places where possible.
    In the case of the decimal already having 24 max digits and 6 decimal we change it
    to 27 max_digits and 9 decimal, on the off chance that 15 digit numbers are already in the db.

- For calculated fields have at 15 decimal places. Not having enough precision causes rounding errors when we sum calculated fields. 

- Fix numbers generated for tests.
    Our tests generates numbers with too many max_digits. With the new standard of 24 max_digits,
    9 of which are decimals, we can have integrals of 15 digits.
    Some numbers generated in our tests are 3600 * 1e9 * 32 which is more than 15 digits,
    change these to use 1e7 instead.

- Update tests to use Decimal math instead of floating point math.

When this change is deployed please re-run the summarization for OCP with the super secret admin endpoint. Which would regenerate the ocpcostusagedailysummary tables with the values from ocp on aws which already has 9 decimal places. This should  hopefully fix the reporting issue https://github.com/project-koku/koku/issues/1241


werwty:
I just tested this again, and there's still a slight mismatch in the ocp vs ocp on aws numbers. I'm slapping WIP on this and going to continue looking into it.

werwty:
> I just tested this again, and there's still a slight mismatch in the ocp vs ocp on aws numbers. I'm slapping WIP on this and going to continue looking into it.

This was caused by us not keeping enough precision on calculated fields. 

For example the unblended_cost in reporting_ocpawscostlineitem_project_daily_summary is calculated from daily.unblended_cost/daily.shared_pods. When this is stored in the temporary table reporting_ocpawscostlineitem_project_daily_summary_{{uuid}} the field type in postgres is numeric, which does not coerce input values to any precision[0]. But when this is copied over to reporting_ocpawscostlineitem_project_daily_summary the type is coerced to numeric(24,9). Which results in some loss of precision, especially noticeable when a lot of them are summed together.

Since Django does not allow us to specify an unlimited Decimal Field, an alternative solution is to just store the calculated decimals at a higher precision than non-calculated decimals. This way we should be able to minimize rounding errors. I've updated the calculated fields to have at least 15 digits of precision. This corrects the values generated for the nise test data:

```
postgres=# select SUM(unblended_cost) from reporting_ocpawscostlineitem_daily_summary WHERE usage_start>='2019-11-01' AND usage_start<='2019-11-10';
         sum         
---------------------
 359.373150110000000
(1 row)

postgres=# select SUM(unblended_cost) from reporting_ocpawscostlineitem_project_daily_summary WHERE usage_start>='2019-11-01' AND usage_start<='2019-11-10';
         sum         
---------------------
 359.373150110000000
(1 row)

postgres=# select SUM(infra_cost) from reporting_ocpusagelineitem_daily_summary WHERE usage_start>='2019-11-01' AND usage_start<='2019-11-10';
         sum         
---------------------
 359.373150110000000
(1 row)

postgres=# select SUM(infra_cost) from reporting_ocpcosts_summary WHERE usage_start>='2019-11-01' AND usage_start<='2019-11-10';
         sum         
---------------------
 359.373150110000000
(1 row)
```

PR #1375, opened 2019-11-07
Updating source_status API to conform with Platform Sources application status API

dccurtis:
The Sources Client pushes status to the platform on source create/edits by updating the `availability_status` and `availability_status_error` on the appropriate application object in the platform backend.

This change returns the same values in the form of an API handle for the platform to call.  Further updates to the calling convention or response may be needed once platform sources designs are complete (@agrare).

**Testing**
1.  Add AWS Source.  Verify that the /source_status returns an error about missing a bucket.  Add bucket and verify the availability_status is available.
2.  Edit the AWS source from step 1 with an incorrect bucket and verify the status message and that the provider is not active. Edit again with correct bucket and verify provider is active.
3.  Add OCP source and verify that the status is available.
4.  Add an Azure Source without the storage and subscription_id and verify the status is unavailable.
5. Verify that 404 is returned for when a query for a non-existent source

**Test Results**
[sources_status_endpoint_ut.txt](https://github.com/project-koku/koku/files/3821844/sources_status_endpoint_ut.txt)


nbon12:
Associated with #1356

PR #1377, opened 2019-11-08
Use the correct column name for Azure filtering

adberglund:
## Summary 
Azure processing was failing because of a bad column/filter reference. We select the correct column for Azure filtering.

## Testing

1. Downloaded and processed an initial Azure report successfully.
2. Generated a new report and successfully downloaded and processed. (This step is what was failing)


koku/masu/test/processor/aws/test_aws_report_processor.py
```diff
@@ -178,11 +178,13 @@ def test_process_default(self):
...
+
         expected = (
-            f'INFO:masu.processor.report_processor_base:Deleting data for:\n'
-            f' schema_name: acct10001\n'
-            f' provider_uuid: {self.aws_provider_uuid}\n'
-            f' bill date: {bill_date}'
+            f'INFO:masu.processor.report_processor_base:Processing bill starting on {bill_date}.\n'
+            f' Processing entire month.\n'
+            f' schema_name: {self.schema},\n'
+            f'provider_uuid: {self.aws_provider_uuid},\n'
```
maskarb:
```suggestion
            f' provider_uuid: {self.aws_provider_uuid},\n'
```

Issue #1390, opened 2019-11-11
OpenAPI spec for sources_status


nbon12:
## User Story
As a QA person, I want the sources_status to have an openapi spec to make it easier to generate the testing client. 
Related to Issue #1309
Related PRs: #1356 #1375 
Luke: "We have internal endpoint on koku-masu which we create a openapi spec for to make it easier to generate our client so we may want to do the same for this"
Doug: "yeah we probably should have one, even though it’s called internally
we could do it now or once we finalize the request/response conventions with the platform"

## Impacts
Sources client

## Assumptions
- This spec may change when request/response conventions are finalized with the platform

## API Details

http://localhost:4000/api/cost-management/v1/source_status/?source_id=1

| API  | Detail|
 | --- | --- |
 | **Method** | GET/POST/PUT/PATCH/DELETE |
 | **Path** | /api/cost-management/v1/source_status/?source_id=X |

![Screen Shot 2019-11-11 at 5 00 52 PM](https://user-images.githubusercontent.com/17408265/68624463-e3219d00-04a4-11ea-927f-727c1a8355aa.png)

## UI Details
- None

## Acceptance Criteria

- [ ] An openapi spec is created for sources_status
- [ ]

lcouzens:
Verified
Commit: "6bac6962d08f693740f8147c966bfdb5ce39a7c3"

Issue #1309, opened 2019-10-25
Create API to return provider status to Platform Sources


dccurtis:
## User Story
As cost management user, I need my cost management account's status to be available to the insights platform so I can be informed when my account is not healthy.

## Impacts
API, UI, Docs

## Assumptions and Questions
- Platform Sources will be calling this API periodically.  Exact frequency is not yet known but it will be most likely hourly or daily.
- The exact calling convention and expected response is not yet known.  Platform Sources design work is forthcoming.  I know that it will be requesting status for Source ID: X and expecting a simple boolean type response.
- In order to start this work I propose that we make an API such as `/source-status/X` to return the source status for ID X.  The response can be a simple json such as `{'source_id: X', 'status': true}`. Before beginning this work we should reach out to Platform Sources and see if they have an API definition.  If they do use that, if they don't maybe propose something or try to get a heads up on what it will be so we can minimize changes for integration.
- `ProviderAccessor().cost_usage_source_ready()` can be used as the business logic behind the source status.  The `sources` database table has the `source_id`, `authentication`, and `billing_source` arguments needed for `ProviderAccessor`.
- TODO for whoever implements this: Figure out how `ProviderStatus` fits into this, and consider the flows @blentz describes in which we may set an "error" status if things fail during processing.
- Q: Are there any specific RBAC concerns for this? Should this be a special internal API that only the platform can access? Should this be available to customers (thus RBAC)?
  - A: We probably want this to be an internal API. Sources client is currently internal... there may be something we can reuse there if we want this to be internal.
  - Also please have a discussion with the platform sources team about how this may affect the standard interface for other services that implement this.

## API Details

Anything that takes a source ID and returns a boolean status for the source.  Ultimately this will need to conform with what Platform Sources defines.  We will be their first customer so we should reach out to them before starting this work to get the latest status.

| API  | Detail|
 | --- | --- |
 | **Method** | GET/ |
 | **Path** | /api/v1/source-status/?source_id=X |
 | **Body** |  |
 | **Response** | `{True}` |
|**Response Code ** | 200_OK |

## UI Details
TBD.  UX mocks are underway.

## Acceptance Criteria

- [ ] API returns successful for an AWS, and Azure source that is configured correctly.
- [ ] API returns failure for an AWS, and Azure source that becomes misconfigured on the cloud backend.  This would mean setting it up correctly initially than changing something on the AWS or Azure side such that our checks fail.
[...]

PR #1392, opened 2019-11-12
add openapi specification for source_status

nbon12:
This PR adds an openapi spec for the internal sources_status API.
This PR is for issue #1390 and it's parent issue #1309 
add openapi.json to sources folder


koku/sources/openapi.json
```diff
@@ -0,0 +1,108 @@
...
+                  "Available": {
+                    "value": {
+                      "availability_status": "available",
+                      "availability_status_error": ""
+                    }
+                  },
+                  "NotAvailable": {
+                    "value": {
+                      "availability_status": "unavailabile",
+                      "availability_status_error": "Bucket cost-usage-buckett could not be found with arn:aws:iam::589173575009:role/CostManagement."
```
dccurtis:
```suggestion
                      "availability_status_error": "Bucket cost-usage-bucket could not be found with arn:aws:iam::999999999999:role/CostManagement."
```

koku/sources/openapi.json
```diff
@@ -0,0 +1,108 @@
+{
+  "openapi": "3.0.0",
+  "info": {
+    "description": "The Sources Status API.  You can find out more about     Cost Management at [https://github.com/project-koku/](https://github.com/project-koku/).",
```
dccurtis:
```suggestion
    "description": "The Sources Status API.  You can find out more about Cost Management at [https://github.com/project-koku/](https://github.com/project-koku/).",
```

koku/sources/openapi.json
```diff
@@ -0,0 +1,108 @@
...
+    "title": "Sources Status API",
+    "license": {
+      "name": "AGPL-3.0",
+      "url": "https://opensource.org/licenses/AGPL-3.0"
+    }
+  },
+  "tags": [
+    {
+      "name": "Source Status",
+      "description": "Operations about report downloads"
```
dccurtis:
```suggestion
      "description": "Platform sources cost management availability status"
```

koku/sources/openapi.json
nbon12: in this example, the error was a real error that I intentionally spelled bucket with two 't's in order to cause the error to occur, otherwise the error would not occur. 
I can rename it to something like "Bucket nonexistantbucketexamplename could not(...)" etc

nbon12: resolved as per changing the name to nonexistantbucketexample

PR #1416, opened 2019-11-17
WIP: Report query tests refactor

maskarb:
This PR builds on #1415.

Creates an AWSReportDataGenerator and moves AWS specific query tests into the AWS folder.


maskarb:
Will open separate PR to separate out the query tests refactor from the views tests refactor.

Issue #1051, opened 2019-08-27
Add cost model id/name to providers API


adberglund:
## Note

We should take this as an opportunity to look at the providers API and make sure this isn't overloading the API with additional info. We also need to get more info from the the UI as to why the provider needs the cost model. 

Why we need this: https://marvelapp.com/73ii7a7/screen/57707337 

## User Story
As a user I'd like to see the cost models associated with my provider/source when I view my providers. 

At a minimum, we want to include the name (but ideally also include the id?) with the response from the providers API.

From Boaz: 
>When creating a new cost model the user can select sources that will use the new cost model. In that step the cost model name of a given source should be displayed next to it. However, there is no easy way to get this information in one API request.

## Impacts
Ex. API, UI, Docs

## Assumptions
- This should be an easy addition to the providers API right now
- The CostModelMap ties a cost model to a provider using it.
- API developers will work closely with UI developers (@boaz0?) to ensure the UI will display this new value as expected, and the back-end and front-end work is done together in this one story.
- Attention API developer! Please engage in conversations early!

## Acceptance Criteria

- [ ] Associated cost models are displayed when hitting the `/providers/` API 
- [ ] The name of the associated cost models are displayed in the UI, per https://marvelapp.com/73ii7a7/screen/57707337
- [ ] OpenAPI JSON includes the new value.

adberglund:
Let's make sure this is an actual need for the UI before implementing anything.

maskarb:
Not needed at this time.

boaz0:
Hi @maskarb and @adberglund 

During the fit & finish meetings I had with Natalie and Dan, it came to me that this is required for assigning sources to a cost model UI.

In addition, there is another related requirement which is filtering providers by cost model name.

lcouzens:
Verified
Commit: 184a84ca5cec9057ffc7f64ad3fc324ea08d9bd8
Tests: test_api_ocp_provider_returns_cost_model, test_api_aws_provider_returns_cost_model, test_api_azure_provider_returns_cost_model

PR #1418, opened 2019-11-18
include cost models info on get/list provider response

infinitewarp:
for #1051


infinitewarp:
Demo of API changes: https://asciinema.org/a/282875

Issue #1407, opened 2019-11-14
Create product manifest for product security consumption.


chambridge:
## User Story
As a service developer, I want to be able to provide a manifest of the run-time dependencies in a machine readable fashion so that product security can review dependencies for any needed vulnerability remedeation.

## Impacts
Backend

## Implementation Details
Another application (python 3.6) has created a manifest  in the following fashion:
[Drift app - manifest creation](https://github.com/RedHatInsights/drift-backend/search?q=manifest&unscoped_q=manifest)

We can review and iterate as necessary. We would like perform this update as part of our `make requirements` target.

## Acceptance Criteria

- [ ] We produce a manifest file of dependencies that is consumable by product security.
- [ ] We fail travis builds if dependencies are changed but the manifest is not updated.

lcouzens:
Verified.

PR #1431, opened 2019-11-20
Add manifest file creation and update flow for product security.

chambridge:
Issue: https://github.com/project-koku/koku/issues/1407


scripts/create_manifest.py
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env python
+
```
blentz:
```suggestion
#
# Copyright 2019 Red Hat, Inc.
#
# This program is free software: you can redistribute it and/or modify
# it under the terms of the GNU Affero General Public License as
# published by the Free Software Foundation, either version 3 of the
# License, or (at your option) any later version.
#
# This program is distributed in the hope that it will be useful,
# but WITHOUT ANY WARRANTY; without even the implied warranty of
# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
# GNU Affero General Public License for more details.
#
# You should have received a copy of the GNU Affero General Public License
# along with this program.  If not, see <https://www.gnu.org/licenses/>.
#

```

scripts/create_manifest.py
```diff
@@ -0,0 +1,36 @@
...
+lockfile = {}
+
+with open("Pipfile.lock") as json_file:
+    lockfile = json.load(json_file)
+
+with open("koku-manifest", "w") as manifest:
+    for name, value in sorted(lockfile["default"].items()):
+        if "version" in value:
+            version = value["version"].replace("=", "")
+            manifest.write("koku/python:3.6=%s:%s\n" % (name, version))
```
adberglund: In case we do upgrade the Python version what do you think about grabbing the Python version this way 

```
>>> import sys
>>> sys.version
'3.7.4 (default, Sep  7 2019, 18:27:02) \n[Clang 10.0.1 (clang-1001.0.46.4)]'
>>> sys.version.split(' ')[0]
'3.7.4'
```

chambridge: Yeah, I thought about doing it programmatically, but since this is run by the developer (who right now could be running 3.7), I thought It should be a more intentional choice. But I'm flexible; I can change it if you prefer. Or I could look into some mechanism where this is generated from within our docker image; but that may be slow and heavy handed.

blentz: +1 to using `sys.version`.

adberglund: @chambridge That's a good point. Since my example _clearly_ had 3.7. Turns out you can't brew install anything these days without it auto-updating your Python. I'll have to be careful here...

chambridge: @blentz @adberglund Updated to programmatically obtain python version

chambridge: This actually causes a problem in travis when the 3.7 version runs as it detects a difference in the manifest since it runs the create manifest and compares it:
```
$ .travis/check_manifest.sh
Pipfile.lock changed without updating koku-manifest. Run 'make manifest' to update.
```
I'm going to move back to us having to update this when we change the running python.

PR #1434, opened 2019-11-21
Release 2019.11.21.0

adberglund:
## Summary
* Release config change for celery to use a single worker process.
* Release bugfix for `no such savepoint` error in processors.


adberglund:
We ended up needing another solution than what was to be released here. Closing.

PR #1437, opened 2019-11-21
Remove savepoint, switch to autocommit

adberglund:
## Summary
This switches masu to running in autocommit mode (Django's default mode) and removes save-pointing. Save-pointing is causing our report processor to crash and we aren't really making use of savepoints, in fact we were manually committing after every raw SQL execution. For atomic transactions we can use Django's `transaction.atomic()` where we need it.


koku/masu/test/providers/test_report_stats_db_accessor.py
```diff
@@ -134,29 +120,21 @@ def test_update(self):
...
-        saver.close_session()
 
     def test_log_last_completed_datetime(self):
         """Test convience function for last completed processing time."""
         saver = ReportStatsDBAccessor('myreport', self.manifest_id)
         saver.log_last_completed_datetime()
-        saver.commit()
 
         # FIXME: missing asserts
         saver.delete()
```
maskarb: Are these tests doing anything?

adberglund: Discussed offline, and updated these tests to have asserts.

Issue #1238, opened 2019-10-10
Configure OCP cost as $X/month/node


infinitewarp:
## Update January 20, 2020

This has come back to dev after initial QA. There's two bits we need to add in to pass QA. 

-  When we create the summary line item for the node cost here: https://github.com/project-koku/koku/pull/1438/files#diff-5132f3be415e3032124d987a0aff4939R770 we are missing some important information. 
   -  We need to alter that method to create a new row for each node. 
   - We need each new row to have `cluster_id`, `cluster_alias`, and `report_period_id` as included data
   - Adding these fields will cause this row to be deleted when we delete a provider
- Once we have a row per node in this table, then we need to update the APIs to return sum('monthly_cost') for the cluster when we group by cluster. Having the cluster_id and cluster_alias added to the row will allow us to do this. 

## User Story
As a Koku customer, I want to see my billing cost for OCP usage as my configured $X/month/node.

## Assumptions
- This needs UX discussion around how we gather inputs.
  - We accept that this initial implementation is API-only and may be revised as UI/UX discussions unfold.
- See: https://cloud.redhat.com/beta/cost-management/cost-models
- Input definition is to perhaps add "node" alongside CPU/Memory/etc in the "add rate" UI, etc. We accept that we are going to make some assumptions about this interaction and may need to revisit the implementation once UX provides feedback.
- If defined, this applies to _all_ OCP nodes for the customer's provider/source.
- In terms of unit testing, we do not intend to exhaustively verify the SQL beyond the pattern used to currently test/verify the affected reports. We will likely spot-check some particular use cases manually or have very specific unit tests for specific known use cases.

## Questions and Answers
- Q: how is this output value presented in the UI?
- A: This affects derived cost. Cost model has 2 components: price list and markup. Anything from price list we call a derived cost. Anything from markup we call markup cost. We also return infrastructure cost which is just the cost directly from the provider.
- Q: do we calculate fractional rates, round, or ceiling? (e.g. does 1 node running only for 10 days of a 30-day month mean it accrues 1/3 the rate?) If fractional, does this really mean that we internally multiply and treat the rates as "per second"?
- A: Regardless of how much you use it, you get the whole month charge. So, this effectively acts like a ceiling. This is intended to mimic the customer contract with openshift. This means that if you look at the cost for a node on Day 1 vs Day 30 of the month, you see this contribute to the full cost either way. We are not going to consider amortizing the cost across the month's days.
[...]

PR #1438, opened 2019-11-21
Configure OCP cost as $X/month/node 

werwty:
Monthly ocp node cost is calculated and stores in the 'monthly_cost' column in
OCPUsageLineItemDailySummary. and is calculated by the total number of unique nodes * node cost.

Note that this PR does not include monthly_cost in the groupby[project] view, as that needs a separate discussion. 
https://github.com/project-koku/koku/issues/1238


werwty:
Demo: https://asciinema.org/a/y8rCV80zmY5DETv1mS6MOT1eT

PR #1456, opened 2019-11-27
add docs and update smoke testing scripts

blentz:
This PR adds some initial documentation discussing how to run the unit and smoke tests for koku.

I've also made two updates:
1. Updated the smoke testing scripts to work on Linux using an OpenShift-based environment. 
2. Updated the `settings.local.yaml` to use the non-deprecated auth format.


docs/source/testing.rst
```diff
@@ -0,0 +1,98 @@
...
+    $ make stop-forwarding-ports
+
+Smoke testing
+=============
+
+Running the QE functional and smoke tests is achieved through the scripts provided in ``koku.git/testing``
+
+Requirements
+------------
+In order to run the test suite, there need to be clones of the following git repositories:
```
dccurtis: Are these requirements needed for local OpenShift IQE testing?  When running with compose at least you don't have to have any of these installed or configured on your system since they are pre-built in the IQE container.

docs/source/testing.rst
```diff
@@ -0,0 +1,98 @@
...
+
+  - koku
+  - iqe
+  - hccm-plugin
+
+You will need to edit the settings file to match your environment: ``koku.git/testing/conf/settings.local.yaml``
```
dccurtis: Similar question as above.  Are changes to this yaml needed for OpenShift local testing?  The way docker-compose is configured currently the user should never have to touch these files.

blentz: Yes. You need to point it to the endpoint, which isn't going to be on localhost.

docs/source/testing.rst
blentz: The `e2e-deploy.sh` script leverages the `iqe` command to deploy a portion of the environment.

Issue #1439, opened 2019-11-21
Consolidate Sources API endpoints and enable PATCH


dccurtis:
## User Story
As user I want to have a single API where I can and do limited editing on my cost management sources.  I want to be able to edit the cost management source `authentication` and `billing_source` values.  These values are not stored within the Platform Sources backend so they need to be editable via this API so they can be set by the UI.  The `sources-status` API should also be consolidated to the new API with an action method on a particular source instance.

## Impacts
API, UI, Docs

## Assumptions
- Existing `billing_source`, `authentication`, and `source-status` APIs continue to work until the frontend and test automation can be moved to the new API.  Once they are transitioned we can remove the old API endpoints.

## API Details
`GET` Example
```
GET /api/cost-management/v1/sources/
HTTP 200 OK
Allow: GET, HEAD, OPTIONS
Cache-Control: max-age=0, no-cache, no-store, must-revalidate
Content-Type: application/json
Expires: Thu, 21 Nov 2019 21:28:08 GMT
Vary: Accept

{
    "meta": {
        "count": 1
    },
    "links": {
        "first": "/api/cost-management/v1/sources/?limit=10&offset=0",
        "next": null,
        "previous": null,
        "last": "/api/cost-management/v1/sources/?limit=10&offset=0"
    },
    "data": [
        {
            "source_id": 1,
            "name": "Test Azure Source",
            "source_type": "AZURE",
            "authentication": {
                "credentials": {
                    "client_id": "0f10dbdf-be25-452d-871b-4e32d19a4249",
                    "tenant_id": "3be09d48-9d77-468c-99ba-36be84238834",
                    "client_secret": "b6f61b57-bc7f-46be-a019-66dccd8b7dfe",
                    "subscription_id": "3b63027f-af79-4ddd-b6f2-e7b0ac6e7546"
                }
            },
            "billing_source": {
                "data_source": {
                    "resource_group": "TESTRG",
                    "storage_account": "mysa1"
                }
            },
            "koku_uuid": "fcb45f97-7d60-43e1-8574-596716570dcb",
            "source_uuid": "fcb45f97-7d60-43e1-8574-596716570dcb"
        }
    ]
}
```
`GET` specific source id
```
GET /api/cost-management/v1/sources/1/
HTTP 200 OK
Allow: GET, PATCH, HEAD, OPTIONS
Cache-Control: max-age=0, no-cache, no-store, must-revalidate
Content-Type: application/json
[...]

PR #1481, opened 2019-12-03
Consolidate Sources APIs and enable PATCH

dccurtis:
The sources `/billing_source` and `/authentication` were very basic.  Designed to provide just enough functionality to enable basic backend sources integration.

Now that we have integration well underway and I've gotten some feedback from the API consumers (UI and QE), I'd like to make some changes to simplify the Koku-Sources API interactions.

With this change there is now a single `/sources` endpoint (also a Koku-api proxy) which takes the `source_id` as a path parameter.  The GET response provides information into the source which was previously only accessible via direct db calls.  On the Sources Client there is an internal `/sources/<source_id>/status/` endpoint which returns the same thing as `/source-status`.  This internal API hook to be used by the platform to provide source status information.  Work here is still underway so this calling convention is subject to change.

Now API consumers will be able to use `PATCH` to set the authentication and billing_source information on a source. (yes, for Azure you can set both with a single API call!)

I'm keeping the old `/billing_source` and `/authentication` endpoints around until we move QE tests and UI code to the new `/sources` API.

The API also respects the user's account ID and only returns sources that belong to the users account.

**Testing**
1. Add AWS Source to platform with auth_header account id 10001.
2.  Attempt to add bucket with auth_header account id 10002.  Verify that the /sources API returns 404.
3.  Access internal status api for source id 1 and verify source is unavaiable.
4.  Add bucket with auth_header account id 10001.  Verify provider is created successfully

**Test Results**
[sources_api_enhance_ut.txt](https://github.com/project-koku/koku/files/3918929/sources_api_enhance_ut.txt)

Addresses #1439 

cc: @rvsia @boaz0


koku/sources/koku_http_client.py
```diff
@@ -155,6 +155,7 @@ def update_provider(self, provider_uuid, name, provider_type, authentication, bi
                      'billing_source': self.get_billing_source_for_provider(provider_type,
                                                                             billing_source)}
         try:
+            print(f'UPDATE JSON_DATA: {str(json_data)}')
```
maskarb: Should this be a log and not a `print`?

dccurtis: ahh forgot to remove that 👍

Issue #848, opened 2019-05-02
filter_by with specific set of values should override group_by with wildcard


adberglund:
**Describe the bug**
When using the koku reporting APIs, group by with wildcard overrides a filter with specific values. 
E.g. filter[service]=AmazonEC2&group_by[service]=* returns data for all services, despite a specific filter being set. 

**Expected behavior**
The filter should be respected

adberglund:
See https://github.com/project-koku/koku-ui/pull/847 for some examples

blentz:
Essentially, this means that `filter[service]=S3&group_by[service]=*` should be equivalent to `group_by[service]=S3`. 

So, those two queries should return the same report or one of those should raise an error that instructs the user to use the equivalent form.

dccurtis:
@nbon12 Here's the issue we were discussing

nbon12:
The endpoint is 
api/cost-management/v1/reports/aws/storage/

nbon12:
Regarding: "So, those two queries should return the same report or one of those should raise an error that instructs the user to use the equivalent form."
should it return the same report, or raise an error? @blentz  @adberglund @dccurtis 

Option1: filter[service]=AmazonS3&group_by[service]=*  is equivalent to group_by[service]=AmazonS3
Option2: filter[service]=AmazonS3&group_by[service]=* is an error with a message "Please use group_by[service]=AmazonS3"

nbon12:
@adberglund which APIs are affected by this bug? 
I notice it affects /api/cost-management/v1/reports/aws/storage/

nbon12:
blocked due to IQE pipeline failure: https://gitlab.cee.redhat.com/insights-qe/hccm-plugin/-/jobs/1163812

lcouzens:
Verified
Commit: 712a742a78b0c0593169c8a8266f4392603477be
Automation: test_api_ocp/azure/aws_cost_filter_respecting

PR #1484, opened 2019-12-04
OLD: bug#848 fix group_by[NAME]=*&filter[NAME]=X

nbon12:
This PR addresses issue #848 
To test this:

Given a request, for example:
```
        GET
        api/reports/aws/storage/?group_by[region]=*
            &group_by[service]=AmazonS3, AmazonRDS
            &filter[region]=us-west-1
            &filter[region]=us-west-2
```
        
        The request should receive the SAME response as this:

```
        GET
        api/reports/aws/storage/?group_by[region]=us-west-1, us-west-2
            &group_by[service]=AmazonS3, AmazonRDS
```

If the two responses are not equal, this test fails.


koku/api/report/aws/query_handler.py
```diff
@@ -234,3 +236,40 @@ def calculate_total(self, **units):
...
+            parameters (ReturnDict): The parameters object
+        """
+        #import pudb
+        # pu.db
+
+        # For every group_by[X]=*
+        # If there is a filter[X]=Y
+        # Set group_by[X]=Y
+        # For example parameters._parameters['group_by']['region'] = ['eu-west-3']
+
```
nbon12: TODO: this is where the transformation will happen

Issue #1398, opened 2019-11-13
Missing instance-type data for some dates in Azure api report


msager27:
**Describe the bug**
I'm querying the Azure instance-type report for 10 days of data with daily resolution and grouping by instance-type.  I'm using Nise to generate data.  The instance types generated are Standard_A0 and Standard_B2s.  In my report, I'm seeing both instance types for every date except one.  I've run this multiple times and have consistently seen this behavior, although not always for the same date.  The CSV files seem to have the proper data.

I'll provide what I have.  I've also discussed this with Doug and I think he's been able to recreate this behavior in his environment.

Note to devs: try loading the data here and check the database to ensure that each day actually has the expected data. If the files have that data, then it's likely a processing bug.

**To Reproduce**
1. Create 4 Azure-local providers
2. Generate Nise data
3. Ingest the data
4. Query the report, e.g. http://localhost:8000/api/cost-management/v1/reports/azure/instance-types/?filter[resolution]=daily&filter[time_scope_value]=-10&filter[time_scope_units]=day&group_by[instance_type]=*

**Expected behavior**
I'm expecting to see both instance types for every date based on what I'm seeing in the CSV files.  However, I'm not seeing one for Standard_A0 on 2019-11-06 (in a prior run, it was missing Standard_B2s for 2019-11-03).  I'll attach the CSV files.

**Screenshots**
![11-06-entry](https://user-images.githubusercontent.com/53917070/68720361-f2741980-057c-11ea-8aed-e6239b954be7.png)
![11-07-entry](https://user-images.githubusercontent.com/53917070/68720332-d96b6880-057c-11ea-8f6c-4d3cb99fe347.png)
![CSV-Standard_A0-grep-11-06](https://user-images.githubusercontent.com/53917070/68720341-e12b0d00-057c-11ea-827c-7ee5c28bd6e0.png)

**Context (please complete the following information):**
 - OS: Fedora 30

**Additional context**
[csv_files.tar.gz](https://github.com/project-koku/koku/files/3838753/csv_files.tar.gz)

lcouzens:
Verified
Commit: e92a2ed6473fc26e13b932a0c0ab4c59a82215a1

PR #1526, opened 2019-12-10
add instance_type key from AdditionalInfo to AzureCostEntryProductService

maskarb:
#1398 
This PR adds an `instance_type` key to the AzureCostEntryProductService table.

Testing:
[csv_files.tar.gz](https://github.com/project-koku/koku/files/3945147/csv_files.tar.gz)
1. Using attached data, add 4 azure-local providers and ingest data. This json can be used in Postman to create a provider (update the export_name and name fields accordingly):
```
{
    "name": "Test Azure Source - 2",
    "type": "AZURE-local",
    "authentication": {
        "credentials": {
            "subscription_id": "12345678-1234-5678-1234-567812345678",
            "tenant_id": "12345678-1234-5678-1234-567812345678",
            "client_id": "12345678-1234-5678-1234-567812345678",
            "client_secret": "12345"
        }
    }, "billing_source": {
        "data_source": {
            "resource_group": {
                "directory": "",
                "export_name": "hccmnise02"
            },
            "storage_account": {
                "local_dir": "/tmp/local_container",
                "container": ""
            }
        }
    }
}
```
2. Query the Azure data: i.e. `?filter[resolution]=daily&filter[time_scope_value]=-2&group_by[instance_type]=*`
3. See that `no-instance-type`, `Standard_A0`, and `Standard_B2` show for each day and correspond to the appropriate csv.

Results:
[1398_db.txt](https://github.com/project-koku/koku/files/3945273/1398_db.txt)

Additional Testing:
1. Run iqe smoke tests.


koku/reporting/provider/azure/models.py
```diff
@@ -61,6 +61,7 @@ class Meta:
     service_name = models.CharField(max_length=50, null=False)
     service_info1 = models.TextField(null=True)
     service_info2 = models.TextField(null=True)
+    instance_type = models.CharField(max_length=50, null=True)
```
dccurtis: With this model change we can probably update the summary SQL here https://github.com/project-koku/koku/blob/8aaee11d1b0e33711660f0202f333440079a3acc/koku/masu/database/sql/reporting_azurecostentrylineitem_daily_summary.sql#L9

Also will need an initial migration change for the existing database to populate this field.

koku/masu/processor/azure/azure_report_processor.py
```diff
@@ -167,10 +168,12 @@ def _create_cost_entry_product(self, row, report_db_accessor):
         """
         table_name = AzureCostEntryProductService
         instance_id = row.get('InstanceId')
+        additional_info = json.loads(row.get('AdditionalInfo'))
```
dccurtis: We might want to consider using https://pypi.org/project/simplejson/ here since it's more performance oriented with the C extension.  I've never personally used it but would be happy to help investigate if you need.  Only reason for mentioning it here is that that we have seen fairly long line item processing times for AWS.

All this said... Azure has considerably fewer lines since it's already daily so we could also open an issue to switch to simplejson if necessary in the future.

koku/masu/test/processor/azure/test_azure_report_processor.py
```diff
@@ -211,6 +211,7 @@ def test_azure_process_duplicates(self):
         # Wipe stale data
         with schema_context(self.schema):
             self.accessor._get_db_obj_query(AZURE_REPORT_TABLE_MAP['line_item']).delete()
+            # self.accessor._get_db_obj_query(AZURE_REPORT_TABLE_MAP['product']).delete()
```
maskarb:
```suggestion
```

koku/masu/processor/azure/azure_report_processor.py
maskarb: Implemented `ujson`

PR #1552, opened 2019-12-12
Bug #848 fix

nbon12:
This PR addresses issue #848 
To test this:

Given a request, for example:
```
        GET
        api/reports/aws/storage/?group_by[region]=*
            &group_by[service]=AmazonS3, AmazonRDS
            &filter[region]=us-west-1
            &filter[region]=us-west-2
```
        
        The request should receive the SAME response as this:

```
        GET
        api/reports/aws/storage/?group_by[region]=us-west-1, us-west-2
            &group_by[service]=AmazonS3, AmazonRDS
```

If the two responses are not equal, this test fails.


koku/api/query_handler.py
```diff
@@ -298,3 +298,44 @@ def _get_filter(self, delta=False):
...
+        Thereby removing the star when there is a filter provided.
+
+        Args:
+            parameters (QueryParameters): The parameters object
+
+        Returns:
+            parameters (QueryParameters): The parameters object
+
+        """
+        for group_by_key in parameters.parameters['group_by']:
```
chambridge: You should use a  `get` on the object otherwise you could get a *KeyError* when accessing a key that is not in the query parameters.
`for group_by_key in parameters.parameters.get('group_by', {}):`

koku/api/query_handler.py
```diff
@@ -298,3 +298,44 @@ def _get_filter(self, delta=False):
...
+            for group_by_value in parameters.parameters['group_by'][group_by_key]:
```
chambridge: Similar dictionary access comment as above:
`for group_by_value in parameters.parameters.get('group_by', {}).get(group_by_key):`

koku/api/query_handler.py
```diff
@@ -298,3 +298,44 @@ def _get_filter(self, delta=False):
...
+                if group_by_value == '*':
+                    # find if there is a filter[X]=Y that matches this group_by[X]=*
+                    for filter_key in parameters.parameters['filter']:
```
chambridge: Why do we need to loop through all of the filters and later check if the `filter_key` and `group_by_key` are equal?

Wouldn't it make more sense to check if the `group_by_key` was in the filters up front?
```
filter_value = parameters.parameters.get('filter', {}).get('group_by_key')
if filter_value:
    parameters.parameters['group_by'][group_by_key] = filter_value
```

koku/api/query_handler.py
```diff
@@ -298,3 +298,44 @@ def _get_filter(self, delta=False):
...
+                        if isinstance(parameters.parameters['filter'][filter_key], list):
+                            for filter_value in parameters.parameters['filter'][filter_key]:
+                                if filter_key == group_by_key:
+                                    parameters.parameters['group_by'][group_by_key] = [filter_value]
```
chambridge: I don't understand what this is suppose to do as it seems like it would just override the group_by_key entry on each  loop.

koku/api/tests_query_handler.py
```diff
@@ -0,0 +1,125 @@
...
+from api.iam.test.iam_test_case import IamTestCase
+from api.provider.test import create_generic_provider
+from api.report.aws.query_handler import AWSReportQueryHandler
+from api.report.aws.view import AWSInstanceTypeView
+from api.report.test import FakeAWSCostData
+from api.report.test.aws.helpers import AWSReportDataGenerator
+from api.utils import DateHelper
+
+
+class AWSReportQueryTest(IamTestCase):
```
chambridge: Feels like these tests could go here:
https://github.com/project-koku/koku/blob/764cdd03830d487ebb1a34ad69cc0527e6f0e584/koku/api/report/test/aws/tests_queries.py#L42

koku/api/tests_query_handler.py
```diff
@@ -0,0 +1,125 @@
...
+                self.assertEquals('eu-west-3', list_item['region'])
+        # Assert the first request contains only eu-west-3
+        for region_dict in data['data']:
+            # For each date, assert that the region is eu-west-3
+            for list_item in region_dict['regions']:
+                self.assertEquals('eu-west-3', list_item['region'])
+
+    def test_filter_to_group_by(self):
+        """Test the filter_to_group_by method."""
+        url = '?group_by[region]=*&filter[region]=eu-west-3&group_by[service]=AmazonEC2'
```
chambridge: You should add a test case whit multiple filters of the same value.
`url=?group_by[region]=*&filter[region]=eu-west-3&filter[region]=us-east-1`

You should add a test case without group_by in the query parameters:
`url=?filter[region]=eu-west-3`

I believe these two tests would uncover some of the comments above related to dictionary access and the filter list.

koku/api/query_handler.py
nbon12: Yes that’s the point.

koku/api/query_handler.py
nbon12: @chambridge Is your concern that you want to optimize in order to make this faster?

koku/api/tests_query_handler.py
nbon12: Yes that’s where it previously was located, but Andrew told me we have a 1:1 test-to-python-file organization. these tests are testing the koku/koku/api/query_handler.py file

koku/api/query_handler.py
chambridge: I don’t believe it would result in the desired result.

nbon12: @chambridge during my tests I didn’t notice it overriding the same group_by_key twice, since the if condition wouldn’t be true twice.

koku/api/query_handler.py
chambridge: Yes, this is a primary code path so I would like to avoid unnecessary cpu cycles.

koku/api/tests_query_handler.py
chambridge: Ok. Then another class name?

chambridge: Ok. Then another class name?

koku/api/tests_query_handler.py
nbon12: Will do

koku/api/tests_query_handler.py
chambridge: Then another class name for the test?

koku/api/query_handler.py
chambridge: This is why I suggested the test case with multiple filter by the same key.

koku/api/query_handler.py
nbon12: If group_by is not in the parameters wouldn’t we want an error to be thrown?
https://github.com/project-koku/koku/blob/master/koku/api/query_params.py#

koku/api/query_handler.py
nbon12: If group_by_key is not in the parameters wouldn’t we want an error to be thrown?
https://github.com/project-koku/koku/blob/master/koku/api/query_params.py#

koku/api/tests_query_handler.py
nbon12: Yep will change that

koku/api/query_handler.py
nbon12: Can we optimize later? I’d like Luke to start writing tests while I optimize further.

koku/api/tests_query_handler.py
nbon12: What is the desired output in the case of two filters?

koku/api/query_handler.py
chambridge: Any query can be run without a group_by present and defense coding practices have you not depend on another portion of code which could change outside of this to initialize group_by in the parameters object.

koku/api/tests_query_handler.py
nbon12: Would this query be an equivalent replacement for multiple filters?
`group_by[service]=eu-west-3&group_by[service]=us-east-1`

koku/api/query_handler.py
chambridge: I would suggest tackling what seems like an unnecessary loop no based on the above. 
The major focus of the current sprint is query performance improvement so adding extra time seems like the wrong choice at the moment.

Luke should still be able to write tests, the would just be expected to fail until this change goes in.

koku/api/tests_query_handler.py
chambridge: Yes. 

And the second is a sanity check related to the group_by access.  Should run with out error and essentially leaving the parameters untouched since it should skip your group_by cases.

chambridge:
I’d also look at the patch coverage to see what cases are missed.

dccurtis:
While it's always a great idea to run the IQE tests prior to merging its especially important when changing the query handler.  Please run the smoke and vortex tests as well.

`make docker-iqe-vortex-tests` and `make docker-iqe-smokes-tests`. 

I saw failures while running the vortex tests.

koku/api/query_handler.py
nbon12: @chambridge what is your time complexity goal for this function?

chambridge: You just need to avoid looping over all filter keys above. A loop is not necessary as you only need to proceed down the code path if the `group_by_key` is present in the filters.

Currently if you had
`?group_by[region]=*&filter[service]=EC2&filter[account]=1234`
Your code would loop through all the filters even though there is no work to be done.

In the next case:
`?group_by[region]=*&filter[service]=EC2&filter[account]=1234&filter[region]=us-east-1`
Your code would loop through 3 filters when only one is applicable, so can just look up the filter using the group_by key `region`.

chambridge:
@nbon12 Feel free to take a look at the following to fix https://github.com/project-koku/koku/pull/1552#discussion_r357439077:
https://github.com/project-koku/koku/commit/e2421db930e70b72a1c5b88f2d8498d800d36d16

But you should still run the **iqe** tests as suggested by @dccurtis

koku/api/tests_query_handler.py
nbon12: Added

koku/api/tests_query_handler.py
nbon12: renamed to QueryHandlerTest

koku/api/query_handler.py
nbon12: I removed this line

koku/api/query_handler.py
nbon12: Renamed

koku/api/query_handler.py
nbon12: I optimized this loop now.

koku/api/tests_query_handler.py
chambridge: So I could be wrong but per @adberglund  comment https://github.com/project-koku/koku/pull/1552#discussion_r357742705

I think this was suppose to land in koku/koku/api/report/test/aws/tests_queries.py

Right @adberglund ?

koku/api/query_handler.py
nbon12: added

koku/api/query_handler.py
```diff
@@ -326,8 +326,8 @@ def filter_to_order_by(self, parameters):  # noqa: C901
             parameters (QueryParameters): The parameters object
 
         """
-        for group_by_key in parameters.parameters['group_by']:
-            for group_by_value in parameters.parameters['group_by'][group_by_key]:
+        for group_by_key in parameters.parameters.get('group_by', {}):
+            for group_by_value in parameters.parameters.get('group_by', {}).get(group_by_key, {}):
```
chambridge: I don't think the default here on line 330 would be a dictionary for the second get. Plus you don't need the protection on the second line as the first loop on 329 checks that `'group_by'` exists and will give you a valid `group_by_value` if you get to 330.  I tried to show this update with my earlier branch before the break: https://github.com/project-koku/koku/commit/e2421db930e70b72a1c5b88f2d8498d800d36d16#diff-2fb28354acc7323fa4d295ec67e80c5aR330

koku/api/tests_query_handler.py
nbon12: moved to koku/koku/api/report/test/aws/tests_queries.py as per @chambridge suggestion

koku/api/query_handler.py
nbon12: @chambridge the line should be added now, is it still not visible?

chambridge: Yes, I see the update. I made a comment here: https://github.com/project-koku/koku/pull/1552/files#r368004234 that can be addressed, but I'll resolve this comment.

koku/api/tests_query_handler.py
chambridge: No you moved this to: koku/api/report/test/tests_queries.py

koku/api/query_handler.py
nbon12: Removed the protection from the second line

koku/api/report/test/tests_queries.py
```diff
@@ -141,3 +147,124 @@ def test_group_data_by_list_missing_units(self):
                     ],
                 }
                 self.assertEqual(expected, out_data)
+
+
+class AWSQueryHandlerTest(IamTestCase):
```
chambridge: Moved to wrong file, should be here:  koku/koku/api/report/test/aws/tests_queries.py

https://github.com/project-koku/koku/pull/1552#discussion_r357439804

The above discussion shows up strangely in my browser now the source file was removed so I'm just adding a comment here for clarity.

koku/api/tests_query_handler.py
chambridge: Captured this here (https://github.com/project-koku/koku/pull/1552/files#r368027773) since this is hard to view now the source was removed. and resolving this discussion.

nbon12: oops, it should be moved now. thanks for catching that @chambridge

koku/api/report/test/tests_queries.py
nbon12: moved to koku/api/report/test/aws/tests_queries.py

PR #1553, opened 2019-12-16
Display $/month/node cost under derived_cost in reports.

werwty:
Queue  update_charge_info task  when a cost_model is deleted.


werwty:
Demo: https://asciinema.org/a/9tUUONMOwpojMoYdWUjoID17L

Issue #1542, opened 2019-12-12
group_by[and:value] results in 500s for AWS/OCP on AWS 


dchorvat1:
I suspect that `account` isn't making it into the annotations in the query handler.

**Describe the bug**
Using group_by[and:value] in cost report requests for AWS and OCP on AWS providers returns 500 error. 
We get correct 200 response for Openshift and Azure cost reports

**To Reproduce**
1. GET `https://ci.cloud.redhat.com/api/cost-management/v1/reports/openshift/infrastructures/aws/costs/?group_by[and:project]=te&group_by[and:project]=ch`

2. GET  `https://ci.cloud.redhat.com/api/cost-management/v1/reports/aws/costs/?group_by[and:account]=7777&group_by[and:account]=4444`

3. See errors

**Expected behavior**
No errors

**Logs**
```
[2019-12-12 09:16:28,900] INFO None API: /api/cost-management/v1/reports/aws/costs/?delta=cost&filter[limit]=10&filter[offset]=0&filter[resolution]=monthly&filter[time_scope_units]=month&filter[time_scope_value]=-1&group_by[and:account]=7777&group_by[and:account]=4444&order_by[cost]=desc -- ACCOUNT: 6089719 USER: insights-qa
[2019-12-12 09:16:28,941] ERROR None Internal Server Error: /api/cost-management/v1/reports/aws/costs/
Traceback (most recent call last):
  File "/opt/app-root/lib/python3.6/site-packages/django/db/models/sql/query.py", line 1768, in add_fields
    join_info = self.setup_joins(name.split(LOOKUP_SEP), opts, alias, allow_many=allow_m2m)
  File "/opt/app-root/lib/python3.6/site-packages/django/db/models/sql/query.py", line 1504, in setup_joins
    names[:pivot], opts, allow_many, fail_on_missing=True,
  File "/opt/app-root/lib/python3.6/site-packages/django/db/models/sql/query.py", line 1420, in names_to_path
    "Choices are: %s" % (name, ", ".join(available)))
[...]

PR #1565, opened 2019-12-19
match parameters with prefixes

nbon12:
Addresses #1542 

Currently in master branch, the following query gets a 500 error. It also causes the annotations dict to return the following:

http://localhost:8000/api/cost-management/v1/reports/aws/costs/?group_by[and:account]=7777&group_by[and:account]=4444
```
'cost_units': Coalesce(F(currency_code), Value(USD)),
 'date': TruncDayString(F(usage_start))}
```

This second query does not cause an error, and the annotations dict appears to contain an 'account' entry.

http://localhost:8000/api/cost-management/v1/reports/aws/costs/?group_by[account]=7777&group_by[account]=4444
```
{'account': Concat(ConcatPair(F(usage_account_id), Value())),
 'cost_units': Coalesce(F(currency_code), Value(USD)),
 'date': TruncDayString(F(usage_start))}
```

Notice that the first query is missing 'account' in the output. This is because 'account' does not match 'and:account' in the parameters object: `self.parameters.get('group_by', {}).keys())` 

This PR proposes that we remove prefixes from the parameters object _before_ checking that the q_param is in the self.parameters.get('group_by', {}).keys()


koku/api/report/aws/query_handler.py
```diff
@@ -83,8 +83,11 @@ def annotations(self):
             annotations['usage_units'] = Coalesce(self._mapper.usage_units_key, Value(units_fallback))
         # { query_param: database_field_name }
         fields = self._mapper.provider_map.get('annotations')
+        prefix_removed_parameters_list = map(lambda x: x if ':' not in x else x.split(':', maxsplit=1)[1],
```
adberglund: Consider a list comprehension here. Not required, but I find them them easier to read than map, lambda. 

```
prefix_removed_params_list = [param_key.split(':', maxsplit=1)[1] if ':' in param_key else param_key for param_key in parameters.get('group_by', {}).keys()]
```

koku/api/report/aws/query_handler.py
```diff
@@ -83,8 +83,11 @@ def annotations(self):
...
+                                             self.parameters.get('group_by', {}).keys())
+        parameters_list = self.parameters.get('group_by', {}).keys()
```
adberglund: You don't need the parameters_list now that you created the prefix_removed_parameters_list

koku/api/report/aws/query_handler.py
```diff
@@ -83,8 +83,11 @@ def annotations(self):
...
         for q_param, db_field in fields.items():
-            if q_param in self.parameters.get('group_by', {}).keys():
+            if q_param in parameters_list or q_param in prefix_removed_parameters_list:
```
adberglund: I don't think checking both lists helps here. Should be `if q_param in prefix_removed_parameters_list`

koku/api/report/aws/query_handler.py
nbon12: @adberglund i thought so too, until the unit tests failed, and then i put them both there and now the unit tests pass

nbon12: @adberglund fixed.

koku/api/report/test/aws/tests_queries.py
```diff
@@ -1268,3 +1268,15 @@ def test_execute_query_with_tag_filter(self):
         for key in totals:
             result = data_totals.get(key, {}).get('value')
             self.assertEqual(result, totals[key])
+
+    def test_prefixed_logical_and_or(self):
+        """"Test group_by[and:account] prefixes."""
+        """Test execute_query for current month on daily breakdown."""
+        url = '?group_by[and:account]=7777&group_by[and:account]=4444'
```
chambridge: Instead of hardcoding two account numbers you should use ones from the unit test. In the setup one is created for you: `self.fake_aws.account_alias`

In the current case you are likely to only have 0 cost data, since those account probably don't exist for this tenant.

You can create additional accounts with the following steps for this test case:
```
redhatisgreat_fake_aws = FakeAWSCostData(self.provider, account_alias='redhatisgreat')
redhatisgreat_generator = AWSReportDataGenerator(self.tenant)
redhatisgreat_generator.add_data_to_tenant(redhatisgreat_fake_aws)

redhatisthefuture_fake_aws = FakeAWSCostData(self.provider, account_alias='redhatisthefuture')
redhatisthefuture_generator = AWSReportDataGenerator(self.tenant)
redhatisthefuture_generator.add_data_to_tenant(redhatisthefuture_fake_aws)
```
So you can reference the account_alias values from in this URL.
url = '?group_by[and:account]=redhat&group_by[and:account]=future'

So its important that the test case prove that the *and* condition is working.

Here you can run the individual group_by query for the `redhatisthefuture` account and `redhatisgreate` account on its own and collect that total cost.
`url = '?group_by[account]=redhatisthefuture'`
`url = '?group_by[account]=redhatisgreat'`

You can run a the default OR case
`url = '?group_by[account]=redhat&group_by[account]=future'`

Then you can run the and condition above:
url = '?group_by[and:account]=redhat&group_by[and:account]=future'
which should only match the total for `redhatisthefuture`.

nbon12: added

chambridge:
Thanks for updating the test case. Have you run the iqe tests also?
The grouping logic is pretty core to the reporting queries, so good to have the sanity check.

nbon12:
> Thanks for updating the test case. Have you run the iqe tests also?
> The grouping logic is pretty core to the reporting queries, so good to have the sanity check.
@chambridge 
Yes, Here's a screenshot of passing IQE smokes tests:
Screen Shot 2020-01-06 at 2 07 51 PM
The other IQE tests are still running.

PR #1573, opened 2019-12-23
Log a warning instead of an error when deleting manifest that does no…

nbon12:
Addresses #1513 
Log a -warning- INFO when deleting a manifest file that does not exist.

### DIFF
--- a/docs/source/testing.rst
+++ b/docs/source/testing.rst
@@ -124,3 +124,14 @@ AssertionError: no logs of level WARNING or higher triggered on api.provider.mod
 Ran 756 tests in 2025.258s
 FAILED (failures=1)
 ```
+
+# Gotchas:
+- The logger is disabled by default during unit tests. If you are building a unit test that asserts that a log occurs, you must re-enable the logger. For example:
+
+```
+import logging
+with self.assertLogs(logger='masu.external.downloader.aws.aws_report_downloader', level='WARN') as cm:
+            logging.disable(logging.NOTSET)
+            self.aws_report_downloader._remove_manifest_file("None")
+            self.assertEqual(['WARN: Could not delete manifest file at'], cm.output)
+```
--- a/koku/api/query_params.py
+++ b/koku/api/query_params.py
@@ -174,6 +174,12 @@ class QueryParameters:
         self._set_access_azure()
         self._set_access_ocp()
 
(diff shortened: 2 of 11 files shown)


adberglund:
@blentz This is part of a series of issues raised by Sentry that went through backlog refinement. There's several places that we were abusing log.error and it was causing static in our sentry alerts. Here, we're just trying to clean up a temporary file, so it it doesn't exists it's not really an error in our program, just a case we're already catching. I don't think the log message here hurts anything, although it could be updated to say that the file doesn't exist. 

I agree with your assessment, but I think the problem is actually in the report downloader's `get_report_context_for_date` methods where `_get_manifest` is called and effectively returns garbage. We then continue to try to process that garbage instead of stopping. We can open a separate issue to deal with that, but this one was just a quick "fix the logging" issue.

blentz:
@adberglund - Understood, but that doesn't really change my objection. I don't see how changing the logging priority is fixing anything. It may be silencing the noise generated by a new tool, but as you said - that's not the part that needs fixing. That's basically the same thing as just adding `pylint: disable` and ignoring the underlying issue.

I'd prefer to see us fix the root cause. There's no need to change this code at all if the problem is elsewhere.

adberglund:
@blentz This issue went through backlog refinement as-is, and as a team we decided we wanted to start by "fixing" the logging level. This accomplishes that. We can certainly also open a new issue to address the underlying issue. I believe it really only happens with QA runs tests, so I'd call it low priority, but we can have it in the backlog.

blentz:
I'm not asking to re-scope the issue or change what was decided during grooming. Apologies if that's how my comment sounded.

I'm asking to set the log level to a more appropriate level. Everything else I mentioned is broader context and doesn't need to affect this PR. Again, apologies if it sounded like I was asking for something bigger than that.

The issue opened by Sentry suggests `INFO`. I'm suggesting `DEBUG` or `pass`. `WARNING` is too high of a logging level. So, please pick between `INFO`, `DEBUG`, or `pass`, or help me understand what raises this to the level of `WARNING`.

Secondly, yes, please open a new issue to address the root cause in the `_get_manifest` method. I agree that is a low priority bugfix.

Once the log level is at the appropriate level, I think this PR is fine.

PR #1579, opened 2020-01-03
1465 batch summary by date range

adberglund:
## Summary
This breaks up summarization into smaller date chunks when summarizing over long periods (e.g. a month). This should allow the database to process and store in temporary tables less data and ultimately run faster. 

## Testing

Rank IQE API tests. Getting 4 failures, but the same failures as on master.


koku/masu/test/processor/ocp/test_ocp_report_summary_updater.py
```diff
@@ -239,7 +249,9 @@ def test_update_summary_tables_new_period_last_month(
 
         self.updater = OCPReportSummaryUpdater(self.schema, self.provider, self.manifest)
 
-        start_date = self.date_accessor.today_with_timezone('UTC')
+        start_date = datetime.datetime(
+            year=self.today.year, month=self.today.month, day=self.today.day
+        )
```
blentz: Is there some reason you're dropping the TZ-aware datetime object here? (This same pattern is repeated a few times in this PR)

PR #1587, opened 2020-01-06
Makefile and create_test_customer adjustments

maskarb:
Some adjustments to the Makefile and the create_test_customer script. 

For the makefile, added a missing parameter for reinitializing the db without providers. **Also, the db is cleared when IQE tests are run in docker.**

In the create_test_customer, the yaml loader can now grab environment variables. To use an env variable, it must be defined in the yaml like: `${ENV_VAR:default_value}` where `ENV_VAR` is the variable you want, and `default_value` is the default when the variable is undefined.

### DIFF
--- a/scripts/create_test_customer.py
+++ b/scripts/create_test_customer.py
@@ -45,33 +45,18 @@ koku:
 
 import argparse
 import os
-import re
 import sys
 from base64 import b64encode
 from json import dumps as json_dumps
 from uuid import uuid4
 
-from yaml import add_constructor, add_implicit_resolver, safe_load, SafeLoader
+from yaml import safe_load
 import psycopg2
 import requests
 
 BASEDIR = os.path.dirname(os.path.realpath(__file__))
 DEFAULT_CONFIG = BASEDIR + '/test_customer.yaml'
 SUPPORTED_PROVIDERS = ['aws', 'ocp', 'azure']
-
-ENV_VARIABLE_MATCHER = re.compile(r'\$\{([^}^{]+)\}')
-
-def variable_constructor(loader, node):
-    """Extract the matched value, return the environment variable or the default."""
-    value = node.value
-    match = ENV_VARIABLE_MATCHER.match(value)
-    env_var = match.group()[2:-1].split(':')
-    if os.getenv(*env_var):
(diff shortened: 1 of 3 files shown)


scripts/create_test_customer.py
```diff
@@ -45,23 +45,33 @@
...
+    """Extract the matched value, return the environment variable or the default."""
+    value = node.value
+    match = ENV_VARIABLE_MATCHER.match(value)
+    env_var = match.group()[2:-1].split(':')
+    if os.getenv(*env_var):
+        return os.getenv(*env_var)
+    raise ValueError("Environment variable is not defined and no default value given.")
+
+add_implicit_resolver('!ENV', ENV_VARIABLE_MATCHER, None, SafeLoader)
+add_constructor('!ENV', variable_constructor, SafeLoader)
```
blentz: Let's leverage `argparse` for this, as the script already uses it. In a perfect world, the script would expose most of the options in the YAML config as CLI args. You don't have to go quite that far with it for this PR. I'm okay if you just add the couple of options most important to you into the `argparse` setup and leave the rest for a future PR.

```suggestion
parser.add_argument('--whatever-param', default=os.environ.get('WHATEVER_PARAM', 'Some Default Value'))
```

scripts/create_test_customer.py
```diff
@@ -72,10 +67,10 @@ def __init__(self, conf):
         self.customer = self._config.get('customer')
         self.koku = self._config.get('koku')
 
-        self.endpoint_base = 'http://{}:{}/{}/v1/'.format(
+        self.endpoint_base = 'http://{}:{}{}/v1/'.format(
             self.koku.get("host"),
             self.koku.get("port"),
-            self.koku.get("prefix"))
+            self._config.get("api_prefix") or self.koku.get("prefix"))
```
maskarb: @blentz How about this?

blentz: This looks really good. It's definitely on the right track.

My goal is to have both the `argparse` config and the loaded YAML populate a single configuration object. That way, we don't have inconsistencies in picking which value to use. Additionally, a CLI arg can then override a defined YAML value, if needed.

Issue #1464, opened 2019-12-02
Consolidate global variable references to provider types


adberglund:
## User Story
As a developer I want a single source of truth for the value for provider type so that any change is automatically propagated through the codebase. 

## Assumptions
- The Provider model seems like a good place to host provider types (See https://github.com/project-koku/koku/blob/master/koku/api/provider/models.py#L107-L141)
- We need to convert masu global variables and any other provider type references not using those defined in the Provider model over to using the Provider model.

## Acceptance Criteria

- [ ] All other global variables referencing provider type are removed. 
- [ ] All provider type references use the provider model

lcouzens:
Verified
Commit: 712a742a78b0c0593169c8a8266f4392603477be

PR #1592, opened 2020-01-07
Consolidate provider type references (#1464)

abaiken:
This PR does the following: 
 - Changes `AZURE` to `Azure` in both the provider type model & in the source types throughout the code to keep consistency
- Uses the provider type model definitions anywhere that a provider type is referenced instead of a redefinition
- Make all provider types & source types case insensitive for api calls 
- Added unit tests to ensure that provider creation is case insensitive 

~~Corresponding IQE test MR is [here](https://gitlab.cee.redhat.com/insights-qe/hccm-plugin/merge_requests/467).~~ - (MR has become redundant since I have made the type case insensitive)  - Smoke tests are 👍  Need to fix 1 vortex test due to adding a custom filter - vortex test MR is [here](https://gitlab.cee.redhat.com/insights-qe/hccm-plugin/merge_requests/479)


chambridge:
Since you had to change IQE tests I was concerned that it may also impact the UI:
https://github.com/project-koku/koku-ui/blob/baf40844e79c77e1c8a1291bfbbace261b78dd88/src/store/providers/providersCommon.ts#L12

https://github.com/project-koku/koku-ui/blob/4a13d94c86d3c4cdde493ca633b01e41267312f9/src/pages/onboardingModal/sourceKind/form.tsx#L80

https://github.com/project-koku/koku-ui/blob/af7e0a654ce3e95b04ff947837c00f5b9eaf5442/src/api/providersQuery.ts#L5

I haven't looked through all the code changes, but could we handle the incoming provider type queries in a case-insensitive manner?

@dlabrecq @boaz0 Are there more cases in the UI code where this would need to be accounted for? And can we make that code more case insensitive?

abaiken:
@chambridge That’s a good point. I guess we could add both casings to the provider types model - but that seems kind of redundant. (I also made the change in the source types to be consistent) - so it would impact those as well. I am okay with reversing the casing change or I am open to other ways to make it case-insensitive. What are your thoughts?

chambridge:
I think the changes you are doing are fine; the case chosen is proper.

Its more not requiring the input to be `Azure` or `AZURE`, but support anything (e.g. `azure`). I think there is a limited set. For example, listing providers, the type could be case insensitive (like the name is today):
https://github.com/project-koku/koku/blob/master/koku/api/provider/view.py#L61

I can't really think of any other places. @adberglund any other thoughts?

dlabrecq:
We use AZURE, AWS, and OCP in our API requests. (Looks like Chris found them all?) We can change this, but would prefer to see consistency with each API.

This will break the UI, so we'll need to quickly change the UI to match

abaiken:
@chambridge  Gotcha! Let me do that! Thanks 😄

chambridge:
@abaiken  I think the other place might be in the serializer.
We would make type be case-insensitive:
https://github.com/project-koku/koku/blob/master/koku/api/provider/serializers.py#L217

`type = serializers.ChoiceField(choices=PROVIDER_CHOICE_LIST)` --> `type = serializers.ChoiceField(choices=LCASE_PROVIDER_CHOICE_LIST)`

But then need to set the value of type to the Enum strings we've chosen ( AWS, Azure, OCP, etc) in the create and update flows.

koku/api/report/ocp/serializers.py
```diff
@@ -103,7 +103,10 @@ def validate(self, data):
 
         if data.get('infrastructures'):
             infra_value = data['infrastructures']
-            data['infrastructures'] = [infra_value.upper()]
+            if infra_value != 'azure':
```
chambridge: Could we use `Provider.PROVIDER_AZURE.lower()` instead?

koku/api/report/ocp/serializers.py
```diff
@@ -103,7 +103,10 @@ def validate(self, data):
...
+                data['infrastructures'] = [infra_value.upper()]
+            else:
+                data['infrastructures'] = ['Azure']
```
chambridge: Could we use `Provider.PROVIDER_AZURE` instead?

koku/api/provider/models.py
```diff
@@ -105,13 +105,25 @@ class Meta:
...
 
+    PROVIDER_CASE_MAPPING = {
+        'aws': PROVIDER_AWS,
+        'ocp': PROVIDER_OCP,
+        'azure': PROVIDER_AZURE,
+        'gcp': PROVIDER_GCP,
+        'aws-local': PROVIDER_AWS_LOCAL,
+        'azure-local': PROVIDER_AZURE_LOCAL,
+        'gcp-local': PROVIDER_GCP_LOCAL,
+        'ocp-aws': 'OCP_AWS',
```
chambridge: Should we create constants for `'OCP_AWS'` and `'OCP_Azure'`, since these are used in other files?

koku/sources/test/test_storage.py
```diff
@@ -306,7 +308,7 @@ def test_add_provider_sources_auth_info_with_sub_id(self):
                             auth_header=self.test_header,
                             offset=3,
                             endpoint_id=test_endpoint_id,
-                            source_type='AZURE',
+                            source_type='Azure',
```
chambridge: use `Provider.PROVIDER_AZURE` instead?

koku/sources/test/test_kafka_listener.py
```diff
@@ -456,7 +456,7 @@ def test_sources_network_info_sync_azure(self):
 
         source_obj = Sources.objects.get(source_id=test_source_id)
         self.assertEqual(source_obj.name, source_name)
-        self.assertEqual(source_obj.source_type, 'AZURE')
+        self.assertEqual(source_obj.source_type, 'Azure')
```
chambridge: Use `Provider.PROVIDER_AZURE` instead?

koku/sources/test/api/test_view.py
```diff
@@ -46,7 +46,7 @@ def setUp(self):
                                  auth_header=self.request_context['request'].META,
                                  account_id=customer.get('account_id'),
                                  offset=1,
-                                 source_type='AZURE',
+                                 source_type='Azure',
```
chambridge: use `Provider.PROVIDER_AZURE` instead?

koku/sources/test/api/test_serializers.py
```diff
@@ -44,7 +44,7 @@ def setUp(self):
                                  auth_header=self.azure_request_context['request'].META,
                                  account_id=customer.get('account_id'),
                                  offset=1,
-                                 source_type='AZURE',
+                                 source_type='Azure',
```
chambridge: Use `Provider.PROVIDER_AZURE` instead?

koku/sources/storage.py
```diff
@@ -71,7 +71,7 @@ def screen_and_build_provider_sync_create_event(provider):
     provider_event = {}
     screen_map = {'AWS': _aws_provider_ready_for_create,
                   'OCP': _ocp_provider_ready_for_create,
-                  'AZURE': _azure_provider_ready_for_create}
+                  'Azure': _azure_provider_ready_for_create}
```
chambridge: Maybe we should replace the constants in this map with the provider constants.

koku/sources/kafka_listener.py
```diff
@@ -290,7 +290,7 @@ def sources_network_info(source_id, auth_header):
     elif source_type_name == SOURCES_AWS_SOURCE_NAME:
         source_type = 'AWS'
     elif source_type_name == SOURCES_AZURE_SOURCE_NAME:
-        source_type = 'AZURE'
+        source_type = 'Azure'
```
chambridge: `Provider.PROVIDER_AZURE` instead?

koku/sources/kafka_listener.py
```diff
@@ -213,7 +213,7 @@ def save_auth_info(auth_header, source_id):
                 raise SourcesHTTPClientError('Unable to find Cluster ID')
         elif source_type == 'AWS':
             authentication = {'resource_name': sources_network.get_aws_role_arn()}
-        elif source_type == 'AZURE':
+        elif source_type == 'Azure':
```
chambridge: use `Provider.PROVIDER_AZURE` instead?

koku/sources/api/serializers.py
```diff
@@ -95,7 +95,7 @@ def _update_billing_source(self, instance, billing_source):
             instance.save()
 
     def _update_authentication(self, instance, authentication):
-        if instance.source_type not in ('AZURE',):
+        if instance.source_type not in ('Azure',):
```
chambridge: use `Provider.PROVIDER_AZURE` instead?

koku/sources/api/serializers.py
```diff
@@ -83,7 +83,7 @@ def _validate_billing_source(self, provider_type, billing_source):
                 raise SourcesStorageError('Missing AZURE storage_account')
 
     def _update_billing_source(self, instance, billing_source):
-        if instance.source_type not in ('AWS', 'AZURE'):
+        if instance.source_type not in ('AWS', 'Azure'):
```
chambridge: Use provider constants here instead of hard coding?

koku/cost_models/serializers.py
```diff
@@ -289,6 +289,10 @@ def source_type_internal_value_map(self):
     def validate(self, data):
         """Validate that the source type is acceptable."""
         # The cost model has markup, no rates, and is for a valid non-OpenShift source type
+        source_type = data.get('source_type', '')
```
chambridge: why does this get set to `''` and then we check `if source_type` on the next line?
Is there a case where this data can be `None` if so should we be throwing a validation error?

koku/api/tests_query_params.py
```diff
@@ -28,13 +28,13 @@
...
 
-from api.models import Tenant, User
+from api.models import Provider, Tenant, User
 from api.query_params import QueryParameters, get_tenant
 from api.report.serializers import ParamSerializer
 from api.report.view import ReportView
 
 LOG = logging.getLogger(__name__)
-PROVIDERS = ['AWS', 'AZURE', 'OCP', 'OCP_AWS', 'OCP_All']
+PROVIDERS = [Provider.PROVIDER_AWS, Provider.PROVIDER_AZURE, Provider.PROVIDER_OCP, 'OCP_AWS', 'OCP_All']
```
chambridge: Maybe have constants for `'OCP_AWS'` and `'OCP_All'` in the Provider class?

koku/api/tests_query_params.py
```diff
@@ -28,13 +28,13 @@
...
 ACCESS_KEYS = {
     'aws': ['aws.account'],
     'azure': ['azure.subscription_guid'],
```
chambridge: maybe use provider constants `.lower()` for this map?

koku/api/tags/azure/openshift/queries.py
```diff
@@ -26,7 +26,7 @@ class OCPAzureTagQueryHandler(AzureTagQueryHandler, OCPTagQueryHandler):
     """Handles tag queries and responses for OCP-on-Azure."""
 
     data_sources = [{'db_table': OCPAzureCostLineItemDailySummary, 'db_column': 'tags'}]
-    provider = 'OCP_AZURE'
+    provider = 'OCP_Azure'
```
chambridge: Maybe have a constant for this in the Providers class, as it is used more than once across files?

koku/api/report/azure/openshift/view.py
```diff
@@ -29,7 +29,7 @@ class OCPAzureView(ReportView):
     """OCP+Azure Base View."""
 
     permission_classes = [AzureAccessPermission, OpenShiftAccessPermission]
-    provider = 'OCP_AZURE'
+    provider = 'OCP_Azure'
```
chambridge: Maybe have a constant in Providers class for this?

koku/api/report/azure/openshift/query_handler.py
```diff
@@ -28,7 +28,7 @@
 class OCPAzureReportQueryHandler(AzureReportQueryHandler):
     """Handles report queries and responses for OCP on Azure."""
 
-    provider = 'OCP_AZURE'
+    provider = 'OCP_Azure'
```
chambridge: Maybe have a constant in Providers class for this?

koku/api/report/azure/openshift/provider_map.py
```diff
@@ -33,7 +33,7 @@ def __init__(self, provider, report_type):
         """Constructor."""
         self._mapping = [
             {
-                'provider': 'OCP_AZURE',
+                'provider': 'OCP_Azure',
```
chambridge: Maybe have a constant in Providers class for this?

koku/api/provider/view.py
```diff
@@ -141,11 +127,17 @@ def get_queryset(self):
...
         """Create a Provider."""
+        provider_type = request.data.get('type', '')
+        if provider_type and Provider.PROVIDER_CASE_MAPPING.get(provider_type.lower()):
+            request.data['type'] = request.data.get('type', '').lower()
         return super().create(request=request, args=args, kwargs=kwargs)
 
     @never_cache
     def update(self, request, *args, **kwargs):
         """Update a Provider."""
+        provider_type = request.data.get('type', '')
```
chambridge: Why set provider type to `''`? Shouldn't we throw a validation error instead?

koku/api/provider/view.py
```diff
@@ -141,11 +127,17 @@ def get_queryset(self):
     @never_cache
     def create(self, request, *args, **kwargs):
         """Create a Provider."""
+        provider_type = request.data.get('type', '')
```
chambridge: Why set provider type to `''`? Shouldn't we throw a validation error instead?

koku/api/provider/test/tests_serializers.py
```diff
@@ -126,7 +131,7 @@ def test_create_all_providers(self):
     def test_create_provider_fails_user(self):
         """Test creating a provider fails with no user."""
         provider = {'name': 'test_provider',
-                    'type': Provider.PROVIDER_AWS,
+                    'type': Provider.PROVIDER_AWS.lower(),
```
chambridge: Why did all of these have to get set to `lower()`?

koku/api/provider/serializers.py
```diff
@@ -318,6 +322,8 @@ def create(self, validated_data):
     def update(self, instance, validated_data):
         """Update a Provider instance from validated data."""
         provider_type = validated_data['type']
+        provider_type = Provider.PROVIDER_CASE_MAPPING.get(provider_type, provider_type)
```
chambridge: Should this be?
`provider_type = Provider.PROVIDER_CASE_MAPPING.get(provider_type.lower())`

and then throw a validation error if provider type is None?

koku/api/provider/serializers.py
```diff
@@ -288,9 +290,11 @@ def create(self, validated_data):
         credentials = authentication.get('credentials')
         provider_resource_name = credentials.get('provider_resource_name')
         provider_type = validated_data['type']
+        provider_type = Provider.PROVIDER_CASE_MAPPING.get(provider_type, provider_type)
```
chambridge: Should this be?
`provider_type = Provider.PROVIDER_CASE_MAPPING.get(provider_type.lower())`

and then throw a validation error if provider type is None?

koku/api/provider/serializers.py
```diff
@@ -183,28 +183,28 @@ class OCPBillingSourceSerializer(ProviderBillingSourceSerializer):
...
-                              'GCP-local': GCPBillingSourceSerializer,
-                              'OCP': OCPBillingSourceSerializer,
+BILLING_SOURCE_SERIALIZERS = {Provider.PROVIDER_AWS: AWSBillingSourceSerializer,
+                              Provider.PROVIDER_AWS_LOCAL: AWSBillingSourceSerializer,
+                              Provider.PROVIDER_AZURE: AzureBillingSourceSerializer,
+                              Provider.PROVIDER_AZURE_LOCAL: AzureBillingSourceSerializer,
+                              Provider.PROVIDER_GCP: GCPBillingSourceSerializer,
+                              Provider.PROVIDER_GCP_LOCAL: GCPBillingSourceSerializer,
+                              Provider.PROVIDER_OCP: OCPBillingSourceSerializer,
                               'OCP_AWS': AWSBillingSourceSerializer,
```
chambridge: Maybe create a constant for these in the Provider class.

koku/api/provider/serializers.py
```diff
@@ -183,28 +183,28 @@ class OCPBillingSourceSerializer(ProviderBillingSourceSerializer):
...
+AUTHENTICATION_SERIALIZERS = {Provider.PROVIDER_AWS: AWSAuthenticationSerializer,
+                              Provider.PROVIDER_AWS_LOCAL: AWSAuthenticationSerializer,
+                              Provider.PROVIDER_AZURE: AzureAuthenticationSerializer,
+                              Provider.PROVIDER_AZURE_LOCAL: AzureAuthenticationSerializer,
+                              Provider.PROVIDER_GCP: GCPAuthenticationSerializer,
+                              Provider.PROVIDER_GCP_LOCAL: GCPAuthenticationSerializer,
+                              Provider.PROVIDER_OCP: OCPAuthenticationSerializer,
                               'OCP_AWS': AWSAuthenticationSerializer,
-                              'OCP_AZURE': AzureAuthenticationSerializer,
+                              'OCP_Azure': AzureAuthenticationSerializer,
```
chambridge: Maybe add a constant for these in the Provider class.

koku/api/provider/models.py
adberglund: I agree. We just have to be careful that those constants stay out of `PROVIDER_CHOICES` and `CLOUD_PROVIDER_CHOICES` as you can't actually have an OCP_AWS provider, but we still make the distinction when it comes to our reporting APIs. I would just add a note that `OCP_AWS` and `OCP_Azure` are not actually provider types.

scripts/create_test_customer.py
```diff
@@ -202,15 +203,13 @@ def create_provider_db(self, provider_type):
         conn.commit()
         conn.close()
 
-
     def create_providers_db(self, skip_providers):
         """Create a Koku Provider by inserting into the Koku DB."""
         if not skip_providers:
-            for provider_type in ['AWS', 'OCP', 'AZURE']:
+            for provider_type in ['AWS', 'OCP', 'azure']:
```
adberglund:
```suggestion
            for provider_type in ['AWS', 'OCP', 'Azure']:
```

koku/sources/test/test_storage.py
abaiken: I wasn't aware that sources and providers were so similar which is why I didn't utilize the provider type constants in the sources file - but it is fixed now 😄

koku/cost_models/serializers.py
abaiken: This was a mistake! I think I had it set to `''` before I checked for the `source type` and forgot to remove it. Good catch! - Fixed 👍

koku/api/provider/view.py
abaiken: This was a mistake! I think I had it set to '' before I checked for the source type and forgot to remove it. Good catch! - Fixed 👍

koku/api/provider/view.py
abaiken: This was a mistake! I think I had it set to '' before I checked for the source type and forgot to remove it. Good catch! - Fixed 👍

koku/api/provider/test/tests_serializers.py
abaiken: These were set to `.lower()` because I was lowercasing the type of the incoming request in the view (so the serializer expects it to be lowercase). I am open to changing this if you would like!

koku/api/provider/serializers.py
abaiken: I did remove the default of provider_type, but I didn't introduce a new validation error - because I found that it was already being caught. For example, if I create a provider with the type being `notaprovider` then I get back this response from postman which seems reasonable to me: 
```
{
    "errors": [
        {
            "detail": "notaprovider is not a valid source type.",
            "source": "type",
            "status": 400
        }
    ]
}
```

Issue #1466, opened 2019-12-02
ProviderManager should check the infrastructure id on the providers table


adberglund:
## User Story
As a providers API user I want the API to return results as quickly and efficiently as possible so that I can make quick use of the information. 

## Assumptions
- Right now we check the OCP on AWS table (we query that table and get a group by; if that table is large, the operation may take a very long time) which is not compatible with OpenShift on Azure or future providers.
- We recently introduced the concept of infrastructure on the providers table which can be queried directly to show the infrastructure type independent of cloud provider
  - This value is pre-computed and should be much faster to query.
  - https://github.com/project-koku/koku/blob/master/koku/api/provider/models.py#L163 
  - https://github.com/project-koku/koku/blob/master/koku/api/provider/models.py#L274

## Acceptance Criteria

- [ ] ProviderManager now accesses the information in the infrastructure map to show infrastructure_type

lcouzens:
Verified

PR #1593, opened 2020-01-07
make ProviderManager use new lookup table (see: #1466)

blentz:
This change updates the `ProviderManager`  to use the new `ProviderInfrastructureMap` lookup table, as requested in #1466.

## Testing
[Added by @adberglund]

Used nise to load in OpenShift and AWS data that should produce an OpenShift on AWS relationship in the api_providerinfrastructuremap table.

Once data was loaded and summarized I checked the providers API for the infrastructure, which was correctly identified as `AWS-local`

```
...
{
            "uuid": "d60954b4-1a65-4699-a400-e15e9525120a",
            "name": "Test OCP on AWS",
            "type": "OCP",
            "authentication": {
                "uuid": "61e3f060-b6a4-4747-aace-36afdd43ca38",
                "provider_resource_name": "my-ocp-cluster-1",
                "credentials": {
                    "provider_resource_name": "my-ocp-cluster-1"
                }
            },
            "billing_source": {
                "uuid": "8ded3999-61d8-476a-a41c-0e8c4586d4ca",
                "bucket": "",
                "data_source": {
                    "bucket": ""
                }
            },
            "customer": {
                "uuid": "9ad49623-d0a1-40f9-b552-879318542ccf",
                "account_id": "10001",
                "date_created": "2020-01-28T18:33:15.837750Z"
            },
            "created_by": {
                "uuid": "c682b21e-5953-487f-80cc-4ba80e955f2a",
                "username": "user_dev",
                "email": "user_dev@foo.com"
            },
            "created_timestamp": "2020-01-28T18:33:29.478728Z",
            "active": true,
            "infrastructure": "AWS-local",
            "cost_models": []
        }
...
```

DEBUG logs confirmed that the new table was being used for the lookup 

```
SELECT "api_providerinfrastructuremap"."id", "api_providerinfrastructuremap"."infrastructure_type", "api_providerinfrastructuremap"."infrastructure_provider_id" FROM "api_providerinfrastructuremap" WHERE "api_providerinfrastructuremap"."id" = 1; args=(1,)
```


koku/api/provider/models.py
```diff
@@ -280,7 +280,10 @@ class ProviderInfrastructureMap(models.Model):
     """
 
     infrastructure_type = models.CharField(
-        max_length=50, choices=Provider.CLOUD_PROVIDER_CHOICES
+        max_length=50,
+        choices=Provider.CLOUD_PROVIDER_CHOICES,
+        null=True,
+        blank=False
```
blentz: `null=True` refers to the Database allowing NULL values.
`blank=False` refers to the Serializer accepting `' '` (empty string) values.

This tightens-up the table to ensure that we only have one null value instead of two.

koku/api/provider/provider_manager.py
```diff
@@ -50,8 +49,8 @@ def __init__(self, uuid):
         self._uuid = uuid
         try:
             self.model = Provider.objects.get(uuid=self._uuid)
-        except (ObjectDoesNotExist, ValidationError) as e:
-            raise (ProviderManagerError(str(e)))
+        except (ObjectDoesNotExist, ValidationError) as exc:
+            raise ProviderManagerError(str(exc))
```
blentz: Lint change

koku/api/provider/provider_manager.py
```diff
@@ -198,5 +193,6 @@ def remove(self, request, customer_remove_context=False):
 
             LOG.info('Provider: {} removed by {}'.format(self.model.name, current_user.username))
         else:
-            err_msg = 'User {} does not have permission to delete provider {}'.format(current_user, str(self.model))
+            err_msg = 'User {} does not have permission to delete provider {}'.format(current_user,
+                                                                                      str(self.model))
```
blentz: Lint change

koku/api/provider/provider_manager.py
```diff
@@ -67,15 +66,11 @@ def get_name(self):
...
-        provider_accessor = ProviderAccessor(self.model.type)
-        try:
-            infra_type = provider_accessor.infrastructure_type(self._uuid, tenant)
-        except ProviderAccessorError as error:
-            LOG.error('Unable to determine infrastructure type. Reason: %s', str(error))
-            infra_type = 'Unknown-Error'
-        return infra_type
+        if self.model.infrastructure and self.model.infrastructure.infrastructure_type:
+            return self.model.infrastructure.infrastructure_type
+        return 'Unknown'
```
blentz: This is the core of the PR.

koku/api/provider/models.py
adberglund: The default for null (which is False) is actually what we want here. Having a record in this table with a null value for infrastructure type will break things. Can you switch this back?

koku/api/provider/provider_manager.py
nbon12: Is the infrastructure_type the name that the provider is running on?

blentz: Yes

koku/api/provider/models.py
blentz: I've removed the null.

PR #1597, opened 2020-01-09
Add human readable openapi.html to static files

boaz0:
Right now, we can see OpenAPI spec in its JSON format but it can be very confusing and not readable.
I thought about using [`redoc`](https://github.com/Redocly/redoc) to get a user-friendly HTML format.

image

## What did you change in the code? :thinking: 

I updated `make html` to create `openapi.html` if `npm` is installed as it's required to `redoc-cli`.

I also added a new route to `/openapi.html` to display openapi in HTML format. To avoid code duplication I moved some parts of `get_json` and `openapi` to a new method `static_response`.

Just let me know what you think. I personally think it's much more readable than JSON.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,7 @@ PYTHON	= $(shell which python)
 TOPDIR  = $(shell pwd)
 PYDIR	= koku
 APIDOC  = apidoc
+OPENAPI_SPEC_PATH = ./docs/source/specs/openapi.json
 
 # How to execute Django's manage.py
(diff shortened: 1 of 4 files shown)


chambridge:
@boaz0 you can view all the apps/services API specs here:
https://ci.cloud.redhat.com/beta/docs/api/

Cost management:
https://ci.cloud.redhat.com/beta/docs/api/cost-management

Users can click on the Help button in the chrome top nave then select API Documentation.
image

boaz0:
Oh... I didn't know that. That's actually what I was looking for.
Closing this.

Thanks @chambridge

PR #1599, opened 2020-01-09
Update testing.rst

nbon12:
- Add documentation about how to use PDB in the IQE docker container.
- Format using RST


docs/source/testing.rst
```diff
@@ -97,41 +97,62 @@ Smoke test:
...
-Run the server with service-ports: `docker-compose run —service-ports koku-server`
-Breakpoints will now be stopped in this terminal window.
+    1. Ensure all migrations are run.
+    2. Stop the server `docker-compose stop koku-server`
+    3. Run the server with service-ports: `docker-compose run —service-ports koku- 
+       server`
+
+Breakpoints should now be stopped at, in this terminal window.
+
+To test a specific file using tox, edit this line, for example::
```
abaiken: Should we move the `PDB in Docker` section below these points about using tox but above the `PDB in IQE` section? I'm not sure if it is related - I could be misunderstanding.

docs/source/testing.rst
```diff
@@ -97,41 +97,62 @@ Smoke test:
...
+
+    coverage run {toxinidir}/koku/manage.py test --noinput -v 2 {posargs:masu.test.database}`
```
abaiken: Also I'm not sure if it is documented elsewhere - but it might be worth documenting that you can run unit tests for a single file by providing the path to the file you want to unit test when running tox: 
```
tox -e py36 -- masu.test.external.downloader.azure.test_azure_services.AzureServiceTest
```

docs/source/testing.rst
nbon12: @abaiken 

I think I misnamed this section,  PDB in Docker is really: "Running PDB while Koku is deployed in Docker-compose" 

And "PDB in IQE" is really "Running PDB in the IQE docker container"
I can rename those

docs/source/testing.rst
```diff
@@ -97,41 +97,112 @@ Smoke test:
...
+since IQE itself is installed in this container, you may run the following command while inside the container to run a specific test::
+
+    iqe tests plugin hccm -k test_api_aws_storage_filtered_top --pdb
+
+or to run all tests::
+
+    iqe tests plugin hccm --pdb
+
+Any test that fails should start a PDB session.
+
```
maskarb: This section is in this file twice. Let's remove this section and leave the one down below.

```suggestion
```

docs/source/testing.rst
```diff
@@ -97,41 +97,112 @@ Smoke test:
...
+optionally you can, run all tests::
+
+    iqe tests plugin hccm --pdb
+
+Any test that fails should start a PDB session.
+
+
+
+
+
```
maskarb:
```suggestion
```

docs/source/testing.rst
```diff
@@ -97,41 +97,112 @@ Smoke test:
     $ cd koku/testing
     $ ./run_smoke_tests.sh
     
+=====================
+PDB in Docker
+=====================
```
myersCody: So, the extra `=` here are causing the the title bold syntax to bleed into the line below.  The `=` should only follow up to the end of the text. 

Fox example,
```
=============
PDB in Docker
=============
```

docs/source/testing.rst
```diff
@@ -97,41 +97,112 @@ Smoke test:
...
+To run IQE Smoke, Vortex or API tests, run one of the following commands, respectively::
+
+    make docker-iqe-smokes-tests
+    make docker-iqe-vortex-tests
+    make docker-iqe-api-tests
+
+
+====================
+Running PDB in IQE
+====================
```
myersCody: So, the extra = here are causing the the title bold syntax to bleed into the line below. Same comment as above.

docs/source/testing.rst
kholdaway: I agree with @abaiken.  We should not be editing the tox.ini file.  Let's go with her suggestion.

docs/source/testing.rst
abaiken: @nbon12 My comment was unclear - sorry!  I was more so asking if the section about tox belonged underneath the header `PDB in Docker` where it currently is. If the tox documentation wasn't related to PDB in Docker - I was suggesting moving the entire PDB section to be moved below it or perhaps we could move the tox documentation & the `Unit testing log messages` sections to be somewhere underneath the unit tests section [here](https://github.com/project-koku/koku/blob/f38de9de5fd72b201bef855e63a04ae749dcb8df/docs/source/testing.rst#unit-testing).

blentz: I think I agree with the direction @abaiken is going here. Let's look at the organization of the document sections and try to make sure it flows in a coherent way.

For example, the sections and sub-sections might be best organized like this:

- Debugging Options
  - Running PDB with docker-compose
  - Running PDB with IQE container
  - Running PDB inside openshift/crc
  - Alternative debugging ideas (e.g. Printing additional output, strategically placed sys.exit()s, etc.)

This way, we collect related information under the same heading so that it's easier to locate the information a reader wants to look at.

docs/source/testing.rst
```diff
@@ -26,27 +26,37 @@ Unit testing using OpenShift
...
+    with self.assertLogs(logger='masu.external.downloader.aws.aws_report_downloader', 
+    level='WARN') as cm:
+                logging.disable(logging.NOTSET)
+                self.aws_report_downloader._remove_manifest_file("None")
+                self.assertEqual(['WARN: Could not delete manifest file at'], cm.output)
 
-  - koku
-  - iqe
-  - hccm-plugin
+If you observe the following error in the tox tests, you may sometimes ignore it, due to the environment variable DEBUG=TRUE sometimes not being TRUE::
```
blentz: Let's avoid the term "tox tests" and use "unit tests" instead.

Tox is just the tool we use to run the unit tests. The phrase "unit tests" appears to be used more frequently is this doc, so let's standardize on that for consistency.

```suggestion
If you observe the following error when running the unit tests, it may be a false error. This can be due to some tests expecting the DEBUG setting to be TRUE::
```

Though, really, this is something I'd consider a bug that needs to be fixed. Tests shouldn't fail based on the value of DEBUG. They should either be skipped or succeed. I suggest opening a bug about this.

docs/source/testing.rst
```diff
@@ -92,46 +102,80 @@ settings.local.yaml:
...
-- The logger is disabled by default during unit tests. If you are building a unit test that asserts that a log occurs, you must re-enable the logger. For example:
-
-```
-import logging
-with self.assertLogs(logger='masu.external.downloader.aws.aws_report_downloader', level='WARN') as cm:
-            logging.disable(logging.NOTSET)
-            self.aws_report_downloader._remove_manifest_file("None")
-            self.assertEqual(['WARN: Could not delete manifest file at'], cm.output)
-```
+    1. Ensure all migrations are run.
```
myersCody:
```suggestion

1. Ensure all migrations are run.
```

docs/source/testing.rst
```diff
@@ -92,46 +102,80 @@ settings.local.yaml:
...
+    3. Run the server with service-ports: :code:`docker-compose run —service-ports koku- 
+       server`
+    4. set a breakpoint using :code:`import pdb; pdb.set_trace()`
+
+Breakpoints should now be stopped at, in this terminal window.
+
+
+Running PDB in IQE
+------------------
+To run PDB while using dockerized IQE:
```
myersCody:
```suggestion

To run PDB while using dockerized IQE: 
```

docs/source/testing.rst
```diff
@@ -92,46 +102,80 @@ settings.local.yaml:
...
+optionally you can, run all tests::
+
+    iqe tests plugin hccm --pdb
+
+Any test that fails should start a PDB session.
+
+Smoke testing with IQE
+======================
+
+Prerequisites:
```
myersCody:
```suggestion
Prerequisites:

```

docs/source/testing.rst
```diff
@@ -92,46 +102,80 @@ settings.local.yaml:
...
+This will selectively run only the AzureServiceTest module
+The argument can be vague or specified further, for example the following arguments would separately be considered valid:
+
+    - masu.test.external
+    - masu.test.external.downloader.azure.test_azure_services.AzureServiceTest.specific_test
+
+Debugging Options
+=================
+
+Running PDB in dockerized koku
```
blentz: Let's avoid "dockerized" and just use "in a container" or "in Docker" instead. I prefer "in a container" just because it's a bit more agnostic. Docker and Podman both run our containers, and OpenShift is likely to convert over to Podman completely at some point.

docs/source/testing.rst
blentz: Dockerized IQE ==> "IQE in a container"

docs/source/testing.rst
```diff
@@ -92,46 +102,80 @@ settings.local.yaml:
...
+
+
+Running PDB in IQE
+------------------
+To run PDB while using dockerized IQE: 
+    Start a shell session in the docker container that runs IQE::
+
+        koku/testing/run_test.sh bash
+
+To run a specific test, use the following command. Replace :code:`-k` flag's input to a specific test you want to run::
```
blentz: Similar to a previous comment above. This directive is complex and indirect. It should be presented in a more direct way. 

Saying "replace X with Y" has an implicit assumption that the reader knows more about the test suite than they may actually know. It's hard to approach this kind of a sentence as a newbie who may not know what to replace the example with. 

```suggestion
To run a specific subset of the integration test suite (IQE), you can specify a group of tests or a single test using the :code:`-k` flag. You can locate a list of test components <describe where to find them>. Here is one example::
```

docs/source/testing.rst
```diff
@@ -92,46 +102,80 @@ settings.local.yaml:
...
+
+    iqe tests plugin hccm -k test_api_aws_storage_filtered_top --pdb
+
+optionally you can, run all tests::
```
blentz: Running all of the tests should be the first thing discussed, not the last thing.

docs/source/testing.rst
```diff
@@ -92,46 +102,80 @@ settings.local.yaml:
...
+
+    iqe tests plugin hccm --pdb
+
+Any test that fails should start a PDB session.
```
blentz: Why should you start a PDB session?

docs/source/testing.rst
```diff
@@ -92,46 +102,80 @@ settings.local.yaml:
...
+
+Smoke testing with IQE
+======================
+
+Prerequisites:
+    - koku is deployed via docker-compose or other method,
```
blentz:
```suggestion
    - koku is running and accessible via the network 
```

docs/source/testing.rst
```diff
@@ -92,46 +102,82 @@ settings.local.yaml:
...
+
+To run IQE Smoke, Vortex or API tests, run one of the following commands, respectively::
+
+    make docker-iqe-smokes-tests
+    make docker-iqe-vortex-tests
+    make docker-iqe-api-tests
+
+Running IQE source code
+-------------------------------
+There need to be clones of the following git repositories:
```
abaiken: It would be helpful to link to the IQE documentation (https://gitlab.cee.redhat.com/insights-qe/hccm-plugin/tree/master) which tells you how to setup IQE & install the hccm-plugin.

docs/source/testing.rst
```diff
@@ -92,46 +102,82 @@ settings.local.yaml:
...
+
+  - koku
+  - iqe
+  - hccm-plugin
+
+You will need to edit the settings file to match your environment: ``koku.git/testing/conf/settings.local.yaml``
```
abaiken: It might be helpful to add an example of `settings.local.yaml`: 
```
local:
  hccm:
    ocp_dir: ~/${PATH}/koku/testing/pvc_dir/insights_local
    aws_dir: ~/${PATH}/koku/testing/local_providers/aws_local
    azure_dir: ~/${PATH}/koku/testing/local_providers/azure_local
    gcp_dir: ~/${PATH}/koku/testing/local_providers/gcp_local
    masu:
      path: api/cost-management/v1
      port: 5000
      hostname: localhost
      scheme: http
  main:
    hostname: localhost:8000
    scheme: http
    username: user_dev@foo.com
    password: password
  http:
    default_auth_type: basic
```
And then add a note to replace `${PATH}` with your local path to koku.

docs/source/testing.rst
nbon12: @abaiken added!

docs/source/testing.rst
nbon12: @abaiken 
I also found this settings.local.yaml file example in the hccm repo, so I ended up removing this line from this repo and linked to the hccm plugin repo itself. Does that resolve this?

PR #1603, opened 2020-01-09
Add nise documentation

nbon12:
This PR will soon be transferred to the nise repo and deleted


maskarb:
I think we ought to move this PR to the nise repository: https://github.com/project-koku/nise

PR #1609, opened 2020-01-10
Fix makefile target for Fedora.

myersCody:


Makefile
```diff
@@ -116,7 +116,7 @@ help:
 	@echo "  ocp_provider_from_yaml               Create ocp provider using a yaml file."
 	@echo "      cluster_id=<cluster_name>            @param - Required. The name of your cluster (ex. my-ocp-cluster-0)"
 	@echo "      srf_yaml=<filename>                  @param - Required. Path of static-report-file yaml (ex. '/ocp_static_report.yml')"
-	@echo "      ocp_name=<provider_name>             @param - defaults to cluser_id param"
+	@echo "      ocp_name=<provider_name>             @param - Required. The name of the provider. (ex. 'OCPprovider')"
```
abaiken: I would suggest making these `-`instead of underscores `ocp-provider-from-yaml` to follow the other make commands.

myersCody: Oh wow, Great Catch ash 👍

PR #1613, opened 2020-01-10
Updating README to provide an end to end scenario for new people

kholdaway:
This provides updates I needed to run an end to end scenario.  The intent was to help a new developer get an end to end flow running quickly.  This does not provide a deep dive or cover all scenarios. If there is a need for that I suggest adding content to a more advanced section.


README.rst
```diff
@@ -78,53 +78,97 @@ To activate the virtual environment run ::
 
     pipenv shell
 
-Preferred Environment
----------------------
 
-Please refer to `Working with Openshift`_.
```
blentz: Why is this getting removed?

README.rst
```diff
@@ -78,53 +78,97 @@ To activate the virtual environment run ::
...
+Quick Start
+-----------
+This will explain how to start the server and its dependencies using Docker, create AWS/OCP providers, and view reports.  This will not cover all API or scenarios but should give you an end to end flow.
 
-Alternative Environment
------------------------
-If deploying with Openshift seems overly complex you can try an alternate local environment where you will need to install and setup some of the dependencies and configuration.
+Settting up environment variables
```
blentz:
```suggestion
Setting up environment variables
```

README.rst
kholdaway: @blentz Great question!  It was not removed.  It was moved down since it's not the way most people develop.  It is not located in the `Running in a Production-like Environment` section.

README.rst
kholdaway: That header was removed.

kholdaway:
@blentz Please give specific examples.  Much of what was there was covered by docker-compose or make commands and therefore redundant.  I intentionally simplified it.  As stated in the intro to this PR, we may want to add an advanced section.  The quick start is not a deep dive and much of what was there was not needed.

I am willing to discuss specifics if you would like.  Additionally, what was there didn't get me to a state where I could run end to end flows.

Issue #1467, opened 2019-12-02
Date cap derived cost and markup calculations


adberglund:
## User Story
As a developer I want the same date ranges set for summarization to be enforced when calculating derived cost and markup so that consistent values are calculated. 

## Assumptions
- Currently the calculations for derived cost and markup on at least OpenShift do not use date ranges
  - https://github.com/project-koku/koku/blob/master/koku/masu/processor/ocp/ocp_report_charge_updater.py#L303
  - https://github.com/project-koku/koku/blob/master/koku/masu/processor/ocp/ocp_report_charge_updater.py#L179

## Acceptance Criteria

- [ ] A supplied date range is respected always and calculations are run only for those dates

lcouzens:
Verified

PR #1621, opened 2020-01-14
Date cap derived cost and markup calculations.

myersCody:
#1467

- the calculations for derived cost and markup on OpenShift will now respect the date ranged passed in
- Improved the speed in which certain test run locally

TODO:

- [x] Fix the iqe tests that were broken due to these changes. 

----

After a discussion with Andrew, I think I am going to ask for a review before fixing the iqe tests. I just want to make sure my changes in this pr are actually good before fixing the iqe tests. I mostly want to make sure I don't spend time fixing the tests for a pr that will need more changes later on anyway.


koku/masu/database/ocp_report_db_accessor.py
```diff
@@ -730,21 +745,6 @@ def populate_monthly_cost(self, node_cost, start_date=None, end_date=None):
             end_date (datetime, str): The end_date to calculate monthly_cost.
 
         """
-        if isinstance(start_date, str):
```
adberglund: Although I'm not convinced we want exactly this date handling here, for now don't delete this block of code in populate_monthly_cost. It's a funny special case where we allow the cost of a node per month, so the start and end are always supposed to be special.

PR #1624, opened 2020-01-16
1499 and 1504: metrics db down and OperationalError

maskarb:
#1504 

This PR reworks some logic for error logging when querying the DB during schema size metrics collection. During metrics collection, the log will say `failed` _only_ when the query was unsuccessful, otherwise it will say the query returned empty. I added a separate metrics collection that explicitly checks the db connection and increments a prometheus counter when the db is down.

This counter was also added to the `IdentityHeaderMiddleware`. If the Customer lookup failed with an OperationalError, the counter is incremented, LOGed, and the exception is re-raised resulting in 500 response.

`process_exception` was added to KokuTenantMiddleware. If the DB goes down while getting a view, an InterfaceError is raised. The `process_exception` returns 424 Failed Dependency when handling the InterfaceError.


koku/koku/metrics.py
```diff
@@ -20,20 +20,32 @@
 
 from celery.utils.log import get_task_logger
 from django.db import InterfaceError, OperationalError, connection
-from prometheus_client import Gauge
+from prometheus_client import Counter, Gauge
 
 from .celery import app
 
+DB_CONNECTION_ERRORS = Counter('db_connection_erros', 'Number of DB connection errors')
```
maskarb:
```suggestion
DB_CONNECTION_ERRORS = Counter('db_connection_errors', 'Number of DB connection errors')
```

koku/koku/middleware.py
```diff
@@ -61,7 +75,13 @@ class HttpResponseUnauthorizedRequest(HttpResponse):
...
     """
 
-    status_code = 401
+    status_code = HTTPStatus.UNAUTHORIZED
+
+
+class HttpResponseFailedDependency(HttpResponse):
+    """A subclass of HttpResponse to return a 424."""
+
+    status_code = HTTPStatus.FAILED_DEPENDENCY
```
kholdaway: Should we add a JSON error payload?  That would allow us to indicate that it was Postgres or RBAC (next PR) failing.  It could follow similar formatting to other error JSON bodies we provide (despite being 424).

koku/koku/middleware.py
```diff
@@ -211,23 +252,24 @@ def process_request(self, request):  # noqa: C901
...
-            logger.info(stmt)
+            LOG.info(stmt)
             try:
                 customer = Customer.objects.filter(account_id=account).get()
             except Customer.DoesNotExist:
                 customer = IdentityHeaderMiddleware._create_customer(account)
+            except OperationalError as err:
+                LOG.error(err)
+                DB_CONNECTION_ERRORS_COUNTER.inc()
+                return HttpResponseFailedDependency()
```
kholdaway: This is where we know it's a DB failure, so we could add the JSON body here.

koku/koku/middleware.py
kholdaway: You may not need to modify this class at all. The super constructor will do that. See https://github.com/project-koku/koku/blob/master/koku/api/sources/view.py#L89-L93

koku/koku/middleware.py
kholdaway: Message could be passed in like https://github.com/project-koku/koku/blob/master/koku/api/sources/view.py#L89-L93

koku/koku/middleware.py
kholdaway: looks good

koku/koku/middleware.py
kholdaway: Postman result.  Looks good to me.
```{
    "errors": [
        {
            "detail": "Database unavailable. Error: could not translate host name \"db\" to address: Name or service not known\n",
            "status": 424,
            "title": "Failed Dependency"
        }
    ]
}```

Issue #1626, opened 2020-01-20
Cannot add Provider using HTML form in Browsable API


maskarb:
**Describe the bug**
Adding a provider with the HTML form in the browsable API results in AttributeError.

**To Reproduce**
1. Go to [http://localhost:8000/api/cost-management/v1/providers/](http://localhost:8000/api/cost-management/v1/providers/)
2. Try to add a new provider using the HTML form. (See screenshot)
3. See error

```
koku_server       | [2020-01-20 14:25:03,256] ERROR None Internal Server Error: /api/cost-management/v1/providers/
koku_server       | Traceback (most recent call last):
koku_server       |   File "/opt/app-root/lib/python3.6/site-packages/django/core/handlers/exception.py", line 34, in inner
koku_server       |     response = get_response(request)
koku_server       |   File "/opt/app-root/lib/python3.6/site-packages/django/core/handlers/base.py", line 115, in _get_response
koku_server       |     response = self.process_exception_by_middleware(e, request)
koku_server       |   File "/opt/app-root/lib/python3.6/site-packages/django/core/handlers/base.py", line 113, in _get_response
koku_server       |     response = wrapped_callback(request, *callback_args, **callback_kwargs)
koku_server       |   File "/opt/app-root/lib/python3.6/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
koku_server       |     return view_func(*args, **kwargs)
koku_server       |   File "/opt/app-root/lib/python3.6/site-packages/rest_framework/viewsets.py", line 114, in view
koku_server       |     return self.dispatch(request, *args, **kwargs)
koku_server       |   File "/opt/app-root/lib/python3.6/site-packages/rest_framework/views.py", line 505, in dispatch
koku_server       |     response = self.handle_exception(exc)
koku_server       |   File "/opt/app-root/lib/python3.6/site-packages/rest_framework/views.py", line 465, in handle_exception
koku_server       |     self.raise_uncaught_exception(exc)
koku_server       |   File "/opt/app-root/lib/python3.6/site-packages/rest_framework/views.py", line 476, in raise_uncaught_exception
koku_server       |     raise exc
koku_server       |   File "/opt/app-root/lib/python3.6/site-packages/rest_framework/views.py", line 502, in dispatch
koku_server       |     response = handler(request, *args, **kwargs)
koku_server       |   File "/opt/app-root/lib/python3.6/site-packages/django/views/decorators/cache.py", line 44, in _wrapped_view_func
[...]

PR #1627, opened 2020-01-20
do not override provider_type if already in lowercase

maskarb:
#1626 

Small fix so HTML form in browsable API can be used. The HTML form sends the provider type in lowercase so there is no need to override the provider type during creation.

Testing:
1. Go to [http://localhost:8000/api/cost-management/v1/providers/](http://localhost:8000/api/cost-management/v1/providers/)
2. Add provider using HTML form (See screenshot in linked issue)
3. See provider created successfully.

Try to create provider by POSTing directly to API and see creation is successful.
Try PUT request in Browsable API and Postman and see success.

### DIFF
--- a/koku/api/provider/view.py
+++ b/koku/api/provider/view.py
@@ -141,21 +141,11 @@ class ProviderViewSet(mixins.CreateModelMixin,
     @never_cache
     def create(self, request, *args, **kwargs):
         """Create a Provider."""
-        provider_type = request.data.get('type')
-        if (provider_type
-                and Provider.PROVIDER_CASE_MAPPING.get(provider_type.lower())
-                and provider_type != provider_type.lower()):
-            request.data['type'] = request.data.get('type', '').lower()
         return super().create(request=request, args=args, kwargs=kwargs)
 
     @never_cache
     def update(self, request, *args, **kwargs):
         """Update a Provider."""
-        provider_type = request.data.get('type')
-        if (provider_type
-                and Provider.PROVIDER_CASE_MAPPING.get(provider_type.lower())
-                and provider_type != provider_type.lower()):
-            request.data['type'] = provider_type.lower()
         if request.method == 'PATCH':
             raise ProviderMethodException('PATCH not supported')
         user = request.user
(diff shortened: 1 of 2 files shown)


koku/api/provider/view.py
```diff
@@ -142,15 +142,19 @@ def get_queryset(self):
     def create(self, request, *args, **kwargs):
         """Create a Provider."""
         provider_type = request.data.get('type')
-        if provider_type and Provider.PROVIDER_CASE_MAPPING.get(provider_type.lower()):
+        if (provider_type
+                and Provider.PROVIDER_CASE_MAPPING.get(provider_type.lower())
+                and provider_type != provider_type.lower()):
             request.data['type'] = request.data.get('type', '').lower()
```
chambridge: Should this actually be:
```
request.data['type'] = Provider.PROVIDER_CASE_MAPPING.get(provider_type.lower())
```

koku/api/provider/view.py
```diff
@@ -142,15 +142,19 @@ def get_queryset(self):
...
 
     @never_cache
     def update(self, request, *args, **kwargs):
         """Update a Provider."""
         provider_type = request.data.get('type')
-        if provider_type and Provider.PROVIDER_CASE_MAPPING.get(provider_type.lower()):
+        if (provider_type
+                and Provider.PROVIDER_CASE_MAPPING.get(provider_type.lower())
+                and provider_type != provider_type.lower()):
             request.data['type'] = provider_type.lower()
```
chambridge: Should this actually be:
```
request.data['type'] = Provider.PROVIDER_CASE_MAPPING.get(provider_type.lower())
```

koku/api/provider/view.py
maskarb: The HTML form sends the `request.data` as a QueryDict which is immutable, so the `request.data['type'] = something` was throwing an AttributeError.

However, I agree that we should probably use the case_mapping here anyway.

maskarb: I take that back. We explicitly set provider_type to lowercase here, and then the serializer `create` gets the DB expected value. So here, it should be lowercase...unless we change the serializer.

abaiken: @maskarb  first of all - thanks for fixing this! I didn't realize my PR had broken the browsable api because I was using postman to test! Secondly ^ you're correct in how I do this. I transformed the type to lowercase in the view if it was populated and in the mapping table because the serializer expects it to be lowercase. I then did the change to the actual DB expected value in the serializer 👍

PR #1630, opened 2020-01-21
Missing Azure Provider check for cost export on creation

nbon12:
Cost usage source reachable should throw an exception if there is no cost report set up
Addresses #1594


chambridge:
@nbon12  Here is a branch extending yours with a commit that mocks the `AzureService `that you have been having trouble with: https://github.com/project-koku/koku/commit/cc86366da24555734880de3ea4bd39daa4d010f9

Let me know if you have an questions.

nbon12:
IQE smokes tests passed

koku/providers/azure/provider.py
```diff
@@ -99,9 +106,15 @@ def cost_usage_source_is_reachable(self, credential_name,
             storage_accounts = azure_client.storage_client.storage_accounts
             storage_account = storage_accounts.get_properties(resource_group,
                                                               storage_account)
+            azure_service = AzureService(subscription_id, tenant_id, client_id,
+                                         client_secret, resource_group, storage_account)
         except (AdalError, AzureException, ClientException, TypeError) as exc:
```
maskarb:
```suggestion
         except (AdalError, AzureException, AzureServiceError, ClientException, TypeError) as exc:
```
`AzureService` raises `AzureServiceError`. If you're adding to this `try` block, then we need to catch that exception. You will need to import `AzureServiceError` from `azure_service`.

koku/providers/azure/provider.py
```diff
@@ -99,9 +106,15 @@ def cost_usage_source_is_reachable(self, credential_name,
...
             raise ValidationError(error_obj(key, str(exc)))
 
+        if azure_service and not azure_service.describe_cost_management_exports():
+            message = 'No cost management export was found.'
```
maskarb:
```suggestion
            message = 'Cost management export was not found.'
```

koku/providers/azure/provider.py
```diff
@@ -99,9 +106,15 @@ def cost_usage_source_is_reachable(self, credential_name,
             storage_accounts = azure_client.storage_client.storage_accounts
             storage_account = storage_accounts.get_properties(resource_group,
                                                               storage_account)
+            azure_service = AzureService(subscription_id, tenant_id, client_id,
+                                         client_secret, resource_group, storage_account)
```
maskarb:
```suggestion
            azure_service = AzureService(**credential_name, resource_group_name=resource_group,
                                          storage_account_name=storage_account)
```
Instead of unpacking the `credential_name` dictionary (lines 94-97), you _could_ do this, but you don't have to if you don't want to.

koku/providers/azure/provider.py
nbon12: good _catch_.  I've added it. Resolve this conversation if you think it's done.

blentz: See: https://azure.github.io/azure-sdk-for-python/exceptions.html and https://github.com/Azure/msrest-for-python/blob/master/msrest/exceptions.py and https://github.com/Azure/azure-sdk-for-python/blob/master/sdk/core/azure-common/azure/common/exceptions.py

It should be sufficient to catch `ClientException` and `CloudError`, which are the base exception classes for almost everything Azure throws.

nbon12: @maskarb Added

nbon12: @blentz just saw your comment, how do you feel about leaving the redundant catch to be more explicit?

koku/providers/azure/provider.py
nbon12: I added your suggestion @maskarb

PR #1640, opened 2020-01-23
1549 squash db migrations

Red-HAP:
## Description

This was accomplished by using django's `squashmigrations --no-optimize` utility. The reason to leave out django's optimizer is that it really was not factoring out steps as well as it should have. 

Ex:

* Create A, Alter A.x -> could be optimized to just be Create A with the modified column in the create
* Create A Add A.y -> could be optimized to just be Create A with the new column in the create
* Create A, <other steps>, Drop A -> could be optimized to remove the create and all intermediate steps up to and including the drop

So the manual optimization was to deal with these intermediate steps as well as possible to not only cut down on the numbre of migration files to process, but also to cut down on the number of steps per file in order to cut down the overall migration time.

The api migration files and the reporting_common migrations files are interdependent, so I could not reduce them to one file per. I had to reduce them to 2 files per. This solved the interdependency issue.

Some extra steps had to be added in to make the new DB schemas look identical to the schemas produced by running the migrations from the master branch.

## Local Testing:

* `make reinitdb-with-providers` is successful
* `tox -e py36` passes
* Databases were compared using [dBeaver CE](https://dbeaver.io/download/) and no diferences were found
* Running all migrations from master with `make docker-reinitdb-with-providers`, then changing to this brach and running `make run-migrations` did not re-apply migrations as was expected


Red-HAP:
Removed replaced files and re-ran tox with success.

PR #1643, opened 2020-01-24
Update documentation and modify e2e-deploy script to work on mac

abaiken:
This PR adds documentation about running with e2e and modifies the script so that it should run on both Mac/Fedora (but will need someone with each to confirm). 

The main change is that the script now expects 3 environment variables to be set `OPENSHIFT_API_URL` `REGISTRY_REDHAT_IO_SECRETS` and `E2E_REPO`. I did this so that everyone wasn't having to ignore changes to a file that was not .gitignored. 

~~The only other question I have is what the `quay-cloudservices-push` secret should be. I will comment on the line - but this secret was required so i set it to the same value as the `rh-registry-pull-secret` and it worked for me - but I would rather it be what it is supposed to be.~~ - Quay secret was removed from buildfactory config and our script 👍  

Corresponding e2e PR and branch to run against - https://github.com/RedHatInsights/e2e-deploy/pull/1001


scripts/e2e-deploy.sh
```diff
@@ -121,19 +109,36 @@ echo "Adding registry.redhat.io secret."
...
+                                    --from-file=.dockerconfigjson=/dev/stdin \
+                                    -n ${project} \
+                                    --type=kubernetes.io/dockerconfigjson
+        ${OC} secrets link default rh-registry-pull-secret -n ${project} --for=pull
+        ${OC} secrets link builder rh-registry-pull-secret -n ${project}
+    done
+
+    # i'm not sure what the quay secret should be so i copied the above secret and just named it quay-cloudservices-push
+    for project in "${SECRETS_PROJECT}"; do
+        echo ${SECRET} | ${OC} create secret generic quay-cloudservices-push \
```
abaiken: My biggest question/concern is that I don't know what this secret (`quay-cloudservices-push`) should be set to. It was required so i set it to the same value as the `rh-registry-pull-secret` during debugging and it worked for me - but I would rather it be what it is supposed to be. If anyone has any ideas - let me know 😄

chambridge: @abaiken  My guess is that this is related to the two very new PRs that @bsquizz just added to the e2e-deploy repo:
https://github.com/RedHatInsights/e2e-deploy/pull/971
https://github.com/RedHatInsights/e2e-deploy/pull/990

Looks like its just another private registry secret based on the sanitized secret here:
https://github.com/RedHatInsights/e2e-deploy/blob/5e81236147a5064bf83b1df19d7a15da0b31b903/secrets/sanitized/quay-cloudservices-push.yml

I think what you have is okay for now, but we could follow up with Brandon for more clarity. Maybe beneficial for that repo to add some documentation about it so all teams have some more clarity.

abaiken: Ah! It looks like it is literally expecting the yaml - I will update the script to add the yaml file instead of unwrapping it to get the contents of the nested object. Thanks @chambridge

docs/source/openshift.rst
```diff
@@ -65,8 +65,38 @@ Installing and configuring `crc`_ is outside the scope of this document.  Please
...
+    - python
+    - base64
+
+3. Finally, either export the following environment variables or add them to a .env file and enter the IQE virtual environment::
+
+    OPENSHIFT_API_URL=CRC_OR_OKD_API_URL
+    REGISTRY_REDHAT_IO_SECRETS=PATH_TO_RH_REGISTRY_YAML
+    E2E_REPO=PATH_TO_LOCAL_E2E_REPO
+
+**Note:**  If you're using crc the ``OPENSHIFT_API_URL`` will be ``https://api.crc.testing:6443`` and if you're using OKD 3.11 it will be ``https://127.0.0.1:8443/``. The ``REGISTRY_REDHAT_IO_SECRETS`` will be the path to the  file distributed by https://access.redhat.com/terms-based-registry. If your account does not have access to the content, you should file an issue with https://redhat.service-now.com/help and reach out to a teammate to share the file so that you are unblocked.
```
blentz: This is worded in a very RH-specific way. Can you try to reword that last sentence in a way that avoids implying, "You must be a RH employee to contribute"?

```suggestion
If you do not have access to some of the repositories or resources discussed in this document, please contact a member of the Koku development team. Some resources mentioned are internal to Red Hat Associates. If you are unable to access those resources, we will work with you to identify suitable alternatives.
```

docs/source/openshift.rst
```diff
@@ -65,8 +65,38 @@ Installing and configuring `crc`_ is outside the scope of this document.  Please
...
+
+4. Now that you are inside of the IQE virtual environment, change directories to the scripts folder inside of the koku repo (``koku/scripts``). Make sure that you have initiated a bare bones OpenShift cluster (``oc cluster up`` if you are using OKD 3.11) and then run the e2e-deploy script. This will set up your secrets in the secrets project, pull down and build the required images for koku, and deploy the project.
```
blentz: I'd like to suggest tightening up the language a bit by removing some of the directional metaphors that doesn't translate well for non-native English speakers. Also, I'm removing the call-out for okd 3.11 in favor of just saying that we expect you to have a running cluster.

```suggestion
4. Once the IQE virtual environment is activated, change directories to the scripts folder inside of the koku repo (``koku/scripts``). Make sure that you have a running OpenShift cluster before proceeding. The setup script handles all configuration once the cluster is running. With your OpenShift cluster running,  execute the e2e-deploy script. This will set up three projects (``secrets``, ``buildfactory``, & ``hccm``); it will pull the required imagestreams and build the container images for koku; once the container pods are built, the script will deploy the project.
```

docs/source/openshift.rst
```diff
@@ -65,8 +65,38 @@ Installing and configuring `crc`_ is outside the scope of this document.  Please
...
+
+**Note:** If you are getting intermittent deployment failures that don't make sense, you could try increasing the memory and CPUs allotted to docker.
```
blentz: Let's offer up a little more direction on where they should look for failures.

```suggestion
**Note:** If you are getting intermittent deployment failures that don't have an obvious root cause in the pod's runtime logs or in the OpenShift cluster's Event logs, the most common reason is that there are insufficient resources to deploy all pods. Try increasing the memory and CPUs allocated to your openshift cluster or to the individual pods.
```

scripts/e2e-deploy.sh
```diff
@@ -21,7 +21,7 @@
 # and testing.
 #
 # Prerequisites:
-# - an OCP4 environment (e.g. Code-Ready Containers)
+# - an OCP4 environment (e.g. Code-Ready Containers) or OKD 3.11
```
blentz:
```suggestion
# - an OpenShift cluster 3.11+ or 4.x+ (e.g. ``oc cluster up``, ``minishift``, or ``crc``) 
```

scripts/e2e-deploy.sh
```diff
@@ -79,18 +70,15 @@ OC=$(which oc)
...
 
-########## END CONFIGURATION ##########
-########## You shouldn't need to edit below here. ##########
-
-pushd ${E2E_REPO}
+pushd $E2E_REPO
 
 ### validation
-if [ -z "${REGISTRY_REDHAT_IO_SECRETS}" ]; then
+if [ -z "$REGISTRY_REDHAT_IO_SECRETS" ]; then
```
blentz: Why is this being changed?  Both forms should be equivalent.

scripts/e2e-deploy.sh
```diff
@@ -121,19 +109,36 @@ echo "Adding registry.redhat.io secret."
...
 # oc wants the contents of the .data object, so we need to unwrap the outer layer
 # in order to load the pull secrets dockerconfigjson object into the secret.
-if [ -f ${REGISTRY_REDHAT_IO_SECRETS} ]; then
-    SECRET=$(cat ${REGISTRY_REDHAT_IO_SECRETS} | \
+if [ -f $REGISTRY_REDHAT_IO_SECRETS ]; then
+    if [[ "$OSTYPE" == "darwin"* ]]; then
+        # Mac OSX
+        SECRET=$(cat $REGISTRY_REDHAT_IO_SECRETS | \
+             python -c 'import yaml, sys; print(yaml.safe_load(sys.stdin).get("data").get(".dockerconfigjson"))' | \
+             base64 -D)
```
blentz: The only difference I'm seeing here is the flag being passed to base64. Can we DRY this up a bit?

scripts/e2e-deploy.sh
blentz: Based on what @bsquizz said on Friday, I think the tl;dr is that there's a secrets yaml file that they're side-loading to fill in the secrets. So, we just need to decide where we want to keep our secrets and do a similar thing.

scripts/e2e-deploy.sh
```diff
@@ -168,7 +173,7 @@ done
 
 ### deploy application
 echo "Creating HCCM application."
-${IQE} oc deploy -t templates -s hccm -e ${DEPLOY_ENV} hccm
+${IQE} oc deploy -t templates -s hccm -e dev hccm
```
blentz: I think this should be left as a variable in case somebody wants to run a different set of configs.

scripts/e2e-deploy.sh
abaiken: https://github.com/project-koku/koku/issues/1644 I filed an issue to use the new format 👍  But this was about the script requiring a secret that we actually don't use. I reached out to Brandon about it and he said he was going to remove it from the buildfactory config so we will no longer have to create it.

scripts/e2e-deploy.sh
abaiken: This was actually prompted by a change from e2e. You no longer provide the file path. Instead you pass in the environment that you want to pull the configs from. 

From Brandon's email:
```
*NOTE* A key change here is how ocdeployer is run ... now instead of
passing a specific file path, e.g. "-e env/ci.yml", to select the ci
environment you will need to use: "-e ci"
```

If you prefer we can create another environment variable that you can set to `dev`, `ci`, etc. and use that here.

scripts/e2e-deploy.sh
abaiken: I changed the script to use environment variables instead of setting the variables in the script and maybe it is a mac thing but in the `${REGISTRY_REDHAT_IO_SECRETS}` format - they weren't being picked up which prompted the change. I'm not a scripting expert so I'm definitely open to suggestions.

scripts/e2e-deploy.sh
bsquizz: You can still pass in a file for backward-compatibility.

But if you want `env/ci.yml` to be merged with `templates/hccm/env/ci.yml` you can just pass `-e ci` and it'll take care of it.

abaiken: Thanks @bsquizz!

abaiken: @blentz when I ran it with just the path to `env/ci.yml` I was getting failures that were resolved when I used `-e ci` instead. I'm definitely open to how you want to handle this!

docs/source/openshift.rst
```diff
@@ -65,8 +65,38 @@ Installing and configuring `crc`_ is outside the scope of this document.  Please
...
 
+Running locally in OpenShift using e2e-deploy
+---------------------------------------------
+The script ``scripts/e2e-deploy.sh`` handles setup and configuration of `crc`_ or OKD 3.11, including `Red Hat Registry Authentication`_. To use the script, complete the following steps.
+
+1. First, make sure that you have cloned and followed the setup instructions of the following repos::
+
+    - https://github.com/project-koku/koku
+    - https://github.com/RedHatInsights/e2e-deploy
+    - https://gitlab.cee.redhat.com/insights-qe/iqe-tests.git
```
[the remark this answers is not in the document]
abaiken: They were already listed in the script [here](https://github.com/project-koku/koku/blob/master/scripts/e2e-deploy.sh#L32) which is public - but I get that people are more likely to see/read docs. @chambridge thoughts?

scripts/e2e-deploy.sh
blentz: Gotcha. Okay. If that makes it work better for you, I'm fine with it. They're *supposed* to be syntactically equivalent, with the curly braces being "safer" for use in complex strings - e.g. `FOO = "${somevar}_plus_other_stuff"`.

But at any rate, thanks for the extra info. :slightly_smiling_face:

scripts/e2e-deploy.sh
blentz: @abaiken - For now, let's go with your change as-is. 

It sounds like this stuff is evolving and we probably don't need to spend the effort on making it perfect (only to have e2e change again underneath us. :wink: ).

Long-term, I foresee converting this script to python at some point just because Bash requires a lot of legwork to do some things that Python handles in the stdlib. (e.g. optparse/argparse for one.)

PR #1645, opened 2020-01-24
1617 aws compute api cost incorrect

Red-HAP:
## Description

I have tried several permutations trying to find the best performance while correcting the data issue. I think I have it with this. Execution time should also be a bit faster than the views that these are to replace.

## Local Testing

- [x] Migration is applied successfully
- [x] `make reinitdb-with-providers` is successful
- [x] tox tests pass (py36)


koku/reporting/migrations/0091_aws_compute_cost_correction.py
```diff
@@ -0,0 +1,301 @@
...
+         select date(usage_start) as usage_start,
+                instance_type,
+                sum(usage_amount) as usage_amount,
+                max(unit) as unit,
+                sum(unblended_cost) as unblended_cost,
+                sum(markup_cost) as markup_cost,
+                max(currency_code) as currency_code
+           from reporting_awscostentrylineitem_daily_summary
+          where usage_start >= date_trunc('month', now() - '1 month'::interval)
+            and usage_start < date_trunc('month', now() + '1 month'::interval)
```
adberglund: These date_truncs make way more sense than whatever I was trying to do 😂

Issue #1302, opened 2019-10-24
Masu date inconsistencies


nbon12:
## User Story
The masu code expects date objects, but various other parts of the code pass in datetime, date, and string objects. Example of [current code](https://github.com/project-koku/koku/blob/master/koku/masu/database/ocp_report_db_accessor.py#L305):
```def populate_line_item_daily_table(self, start_date, end_date, cluster_id):
        """Populate the daily aggregate of line items table.

        Args:
            start_date (datetime.date) The date to start populating the table.
            end_date (datetime.date) The date to end on.
            cluster_id (String) Cluster Identifier

        Returns
            (None)

        """
        # Cast start_date and end_date into date object instead of string
        if isinstance(start_date, str):
            start_date = datetime.datetime.strptime(start_date, '%Y-%m-%d').date()
            end_date = datetime.datetime.strptime(end_date, '%Y-%m-%d').date()
        if isinstance(start_date, datetime.datetime):
            start_date = start_date.date()
            end_date = end_date.date()
```

This is inconsistent and should be changed to one of the following options:
1) accept any type(date/datetime/str) and convert to date
2) only be passed datetime
3) only be passed date
4) only be passed string

Whoever implements this, please choose according to shared discussion with team!

Please use Type Hinting in any new functions: https://docs.python.org/3/library/typing.html

## Impacts
Masu functions

## Acceptance Criteria

- [ ] all unit tests pass

infinitewarp:
When we get to the database, we ultimately want to end as a `date` and this _is_ happening today. So, however we make this change, the end result must match the current behavior.

nbon12:
After conversation with Andrew:

- All tasks and dates we generate should be `date` type, not `datetime` or `str`.
- All `datetimes` we receive from the Reports (for example in the AWSReportProcessor) should be conserved as `datetime`

nbon12:
On hold due after conversation with Andrew that this work might not be needed, waiting to hear on status of whether story should be cancelled.

PR #1655, opened 2020-01-27
WIP: Masu date inconsistencies

nbon12:
Addresses Issue #1302

This PR converts objects into `date` objects, in the Masu application.
String dates are converted into `date`
datetime.datetime are converted into `date`


koku/masu/util/common.py
```diff
@@ -204,3 +208,30 @@ def date_range_pair(start_date, end_date, step=5):
         start_date = date + timedelta(days=1)
     if len(dates) != 1 and end_date not in dates:
         yield start_date.date(), end_date.date()
+
+
+def log_date_deprecation_warning(date) -> None:
+    """Log that an object should be a date."""
+    if not isinstance(date, datetime.date):
+        LOG.warn(date
```
nbon12: These warnings are a temporary measure that I plan to remove later after things are all converted to `date`

PR #1657, opened 2020-01-28
Sources: Close connection on InterfaceError

maskarb:
The thread controlling sources integration could not re-establish a connection to the koku-db if that connection was lost. This PR closes the thread's connection when it encounters an InterfaceError.

I also cleaned up some repetitive code by adding a `get_source` method.

Testing:
[sources.txt](https://github.com/project-koku/koku/files/4128801/sources.txt)


koku/sources/storage.py
```diff
@@ -150,6 +151,17 @@ def load_providers_to_delete():
     return providers_to_delete
 
 
+def get_source(source_id, err_msg):
```
nbon12: What type is source_id, and err_msg?
Can we use type hinting here? https://docs.python.org/3/library/typing.html

adberglund: Question with this: If the DB is down and we hit the close, I know it will reopen a new connection the next time it needs to, but what about the current call will we potentially drop info? Would it be worth thinking about a retry in this get source method that tries a few times to get the Sources.object.get... if we do hit the InterfaceError?

maskarb: I was thinking about that. I could add some retry logic.

Though, I need to look further into what happens to an unsuccessful message. There are a few flags here and there. Need to check on what they do and how they work. I think with what I have right now, the message would be lost, which would not be good.

maskarb: I'm not entirely sure I like the idea of type hinting. Hinting is certainly nice for a library, but we aren't building one.

I can add to the function docstring though.

adberglund: I don't mind doing this incrementally, if we want to merge this PR and then look at retry logic next.

nbon12: 👍

Issue #1589, opened 2020-01-07
GroupBy combinations break with tags


dlabrecq:
**Describe the bug**
Users want the ability to filter using multiple combinations. In the UI details page; for example, I may want to group_by project, then group_by node and/or tag.

It turns out that `group_by[project]=*&group_by[cluster]=ocp` works, but `group_by[project]=*&group_by[tag:environment]=dev` generates a 500 error.

**To Reproduce**

1. Go to: /api/cost-management/v1/reports/openshift/infrastructures/all/costs/?delta=cost&filter[limit]=10&filter[offset]=0&filter[resolution]=monthly&filter[time_scope_units]=month&filter[time_scope_value]=-1&group_by[project]=*&group_by[tag:environment]=*&order_by[cost]=desc

2. Go to: /api/cost-management/v1/reports/openshift/infrastructures/all/costs/?delta=cost&filter[limit]=10&filter[offset]=0&filter[resolution]=monthly&filter[time_scope_units]=month&filter[time_scope_value]=-1&group_by[or:project]=*&group_by[or:tag:environment]=*&order_by[cost]=desc

3. See 500 error

**Expected behavior**
Expecting any group_by combination of project, cluster, node, and/or tag etc. not to error.

**Screenshots**
For reference, please see the issue below showing our new UI filtering feature
https://github.com/project-koku/koku-ui/issues/1111

**Traceback**

```

[2020-01-07 15:53:16,049] INFO None API: /api/cost-management/v1/reports/openshift/infrastructures/all/costs/?delta=cost&filter[limit]=10&filter[offset]=0&filter[resolution]=monthly&filter[time_scope_units]=month&filter[time_scope_value]=-1&group_by[project]=*&group_by[or:tag:environment]=dev&order_by[cost]=desc -- ACCOUNT: 1460290 USER: jalberts@redhat.com
--
  | [2020-01-07 15:53:16,587] ERROR None Internal Server Error: /api/cost-management/v1/reports/openshift/infrastructures/all/costs/
  | Traceback (most recent call last):
  | File "/opt/app-root/lib/python3.6/site-packages/django/core/handlers/exception.py", line 34, in inner
  | response = get_response(request)
  | File "/opt/app-root/lib/python3.6/site-packages/django/core/handlers/base.py", line 115, in _get_response
  | response = self.process_exception_by_middleware(e, request)
  | File "/opt/app-root/lib/python3.6/site-packages/django/core/handlers/base.py", line 113, in _get_response
[...]

PR #1660, opened 2020-01-28
1589 bugfix tag groupby

adberglund:
## Summary
This addresses the issue in https://github.com/project-koku/koku/issues/1589 where for any `openshift/infrastructures/` API called if you first grouped by project, and then added a second group by for a tag value, we generated a 500 error. The fix is just using the provider map to identify that `pod_labels` should be used in this case instead of `tags`. 

## Testing
Ingested OCP on AWS and OCP on Azure data and tried the group by combination for each API endpoint. Also wrote unit test that runs through each API endpoint with a matrix of group bys with a tag and those tests now pass.


koku/api/report/test/azure/openshift/test_views.py
```diff
@@ -15,25 +15,42 @@
...
+from api.provider.test import create_generic_provider
+from api.report.test.azure.openshift.helpers import OCPAzureReportDataGenerator
+from api.utils import DateHelper
+from reporting.models import OCPAzureCostLineItemDailySummary
 
+URLS = [
+    reverse('reports-openshift-azure-costs'),
+    reverse('reports-openshift-azure-storage'),
+    reverse('reports-openshift-azure-instance-type'),
+    # 'openshift-azure-tags',  # TODO: uncomment when we do tagging
```
maskarb: Maybe we ought to uncomment this one?

PR #1666, opened 2020-01-29
update Makefile to use new syntax for oc-3.11.x

blentz:
This addresses @1636. I've updated the `oc cluster up` syntax to work with oc 3.10+.

It doesn't work if you're not logged in to the Red Hat Container Registry:
```
[root@localhost koku]# make oc-up
oc cluster up \
	--image=registry.redhat.io/openshift3/ose \
	--base-dir=/root/.oc/openshift.local.data
Getting a Docker client ...
Checking if image registry.redhat.io/openshift3/ose is available ...
Pulling image registry.redhat.io/openshift3/ose
E0129 14:34:55.458904    1241 helper.go:179] Reading docker config from /root/.docker/config.json failed: open /root/.docker/config.json: no such file or directory, will attempt to pull image registry.redhat.io/openshift3/ose anonymously
error: error pulling Docker image registry.redhat.io/openshift3/ose
make: *** [Makefile:420: oc-up] Error 1
```

It works once you've logged into the Red Hat Container Registry:
```
[root@localhost koku]# docker login registry.redhat.io
Username: blentz@redhat.com
Password: 
WARNING! Your password will be stored unencrypted in /root/.docker/config.json.
Configure a credential helper to remove this warning. See
https://docs.docker.com/engine/reference/commandline/login/#credentials-store

Login Succeeded
[root@localhost koku]# make oc-up
oc cluster up \
	--image=registry.redhat.io/openshift3/ose \
	--base-dir=/root/.oc/openshift.local.data
Getting a Docker client ...
Checking if image registry.redhat.io/openshift3/ose is available ...
Checking type of volume mount ...
Determining server IP ...
Checking if OpenShift is already running ...
Checking for supported Docker version (=>1.22) ...
Checking if insecured registry is configured properly in Docker ...
Checking if required ports are available ...
Checking if OpenShift client is configured properly ...
Checking if image registry.redhat.io/openshift3/ose is available ...
Starting OpenShift using registry.redhat.io/openshift3/ose ...
```


blentz:
After some internal discussion, there's some questions around whether to remove support for `oc cluster up` entirely, due to it being removed from OCP 4.x and due to difficulties Mac users were having with the 3.11 images.

We'd like to get input from @abaiken before making a final decision. You've got the most recent experiences here, so we'd like you to weigh in. The main question is whether it makes sense to lock this at 3.9, which works for our Mac-based developers or to push everyone away from `oc cluster up` and focus our efforts on supporting minishift and CRC.

abaiken:
I think we should eventually focus our efforts on supporting minishift and crc, but I think it is fine to leave the makefile on using `3.9` in the meantime since this might not be one of our top priorities. I did try to run with the changes - and ran into the same mac issues that @kholdaway was hitting so I agree that spending time/headaches trying to resolve the `oc cluster up` 3.11 issues seems like it just isn't worth it when we have other viable options. Thanks for the work that you put into this though @blentz!

PR #1667, opened 2020-01-29
AWS EC2 instance counts should be param gated

kholdaway:
Note that the original issue called for a `count` query parameter.  `count` conflicts with pagination, therefore I went with `compute_count`.  

### Setup
1. Generate some AWS instance data

### Test without query param
1. Retrieve data without query param: `/api/cost-management/v1/reports/aws/instance-types/`
2. Verify that the response totals do not have counts

### Test with query param set to false
1. Retrieve data with query param set to false: `/api/cost-management/v1/reports/aws/instance-types/?count=false`
2. Verify that the response totals do not have counts

### Test with query param set to true
1. Retrieve data with query param set to false: `/api/cost-management/v1/reports/aws/instance-types/?count=true`
2. Verify that the response totals do have counts

### IQE
This likely has an impact on IQE because the default behavior now excludes count.  Before this is merged, I will run and update the IQE tests.


koku/api/report/aws/query_handler.py
```diff
@@ -198,6 +198,13 @@ def execute_query(self):
         data = []
 
         with tenant_context(self.tenant):
+
+            if not self.parameters.parameters.get('compute_count'):
+                # Query parameter indicates count should be removed from DB queries
+                self._mapper.report_type_map.get('aggregates').pop('count', None)
```
blentz: The mapper object should be treated as read-only.

Instead of `.pop()`, can you use `.get()`?

kholdaway: Lot's of chat offline.  I migrated to what we chatted about.  Essentially using a deepcopy and not changing the maps.

PR #1673, opened 2020-01-30
update sources header

maskarb:
The old header here was resulting in 401 Unauthorized when making GET requests to sources endpoints. Updating the header gets koku-sources working again.


nbon12:
LGTM

PR #1683, opened 2020-02-03
Fix reporting_aws_compute_summary_by_account

adberglund:
* JOIN condition needed usage_account_id OR account_alias_id to match
* Previously failed when no alias existed and alias ID was NULL


koku/reporting/migrations/0092_auto_20200203_1758.py
```diff
@@ -0,0 +1,91 @@
...
+                            ) as x
+                    GROUP BY date(usage_start),
+                        usage_account_id,
+                        account_alias_id,
+                        instance_type
+                ) AS r
+                ON c.usage_start = r.usage_start
+                    AND (
+                            (c.usage_account_id = r.usage_account_id)
+                            OR (c.account_alias_id = r.account_alias_id)
```
adberglund: This OR condition is the actual change here

Issue #1691, opened 2020-02-04
Adding Provider should trigger a task to download data.


maskarb:
## User Story
As a customer I want to see data as soon as I add a Source/Provider. Currently, when we add a source, we must wait until a scheduled tasks kicks off and fetches data.

## Acceptance Criteria
- [ ] Adding a provider causes a data download task to begin.

nbon12:
Sprint Planning Meeting Minutes:

Andrew: @maskarb is already investigating this but we might as well refine it. Even 30 minutes later it can be blank, and we're not sure if it worked. Some of that is for a large customer, if it takes hours to download and fully process a report, it's going to be a while before we have data in there, I'm not sure if that's something we can really fix, but if it's someone trying the tool out, or a small customer to see if small examples work, for Cloud Accounts, it would be nice to trigger the download and process workflow as they add it, instead of waiting for the next time that thing runs, because in production right now, that's 12 hours, so that can be messy.

Chris: For a good user experience, look potentially at how the UI can show the status of the providers from the stat stuff, that came out from the openshift dedicated team not knowing we weren't gathering their data because we ran out of storage.

lcouzens:
Verified

PR #1692, opened 2020-02-05
WIP: Kickoff data download upon Provider save

maskarb:
#1691 

This PR would cause an immediate data ingest when a Source is created.

- Created a receiver on `post_save` signal for Sources. The receiver launches a `check_report_updates` task with the created provider uuid.

- The provider uuid is passed through to the `get_accounts_from_source` function. If this function receives a uuid, it returns the account information for the corresponding provider. If, for whatever reason, that uuid is not found in the list of providers, the function returns an empty list, and no data ingestion occurs. If a provider_uuid is not provided to the function (which is how things currently happen) the function returns the full list of providers like it does today.


adberglund:
👏

adberglund:
Regarding the concerns: The tasks should get enqueued, so I don't think that's a problem. If we have several large customers, we'll need to scale to support that anyway. 

Second concern is a big one, but theoretically we should have logic that does not duplicate ingestion (download/process) of data. It needs a fix right now, but soon we should have logic that prevents that.

koku/api/provider/provider_manager.py
```diff
@@ -184,6 +185,12 @@ def remove(self, request, customer_remove_context=False):
             raise ProviderManagerError(err_msg)
 
 
+@receiver(post_save, sender=Provider)
+def provider_post_save_callback(sender, instance, created, **kwargs):
+    """Kickoff report download on Provider.save."""
+    check_report_updates.delay(provider_uuid=instance.uuid)
```
adberglund: Do you know if you have information about the request here? If so you could check the request.method and if it's patch *not* call the task?

adberglund: Looks like there is the possibility of a request_finished signal that would likely have info about the request + method.

https://docs.djangoproject.com/en/3.0/ref/signals/#request-finished

maskarb:
Different implementation in [https://github.com/project-koku/koku/pull/1927](https://github.com/project-koku/koku/pull/1927)

PR #1700, opened 2020-02-05
1652 local docker db stats

Red-HAP:
## Description

Local DB container initialization changes to enable statement statistics. This change will operate in conjunction with the other changes for large data generation.

This was accomplished without needing to build a new image.

### Setup

`make docker-reinitdb`
Then use psql to connect to the local db

### Copleted Local Tests

- [ ] `show shared_preload_libraries;` should show `pg_stat_statements`
- [ ] `\dx` should show `pg_stat_statements`
- [ ] `select * from pg_stat_statements;` should return records if executed after the first two tests

### DIFF
--- a/pg_init/99_postgresql_conf.sh
+++ b/pg_init/99_postgresql_conf.sh
@@ -1,0 +1,42 @@
+#! /usr/bin/env bash
+
+echo "Updating ${PGDATA}/postgresql.conf..."
+cat >>${PGDATA}/postgresql.conf <<EOF
+
+# ==============================================
+# -----------------------------
+# PostgreSQL configuration file additions
+# These additions are to allow for local statement statistics
+# These configuration options were taken from PostgreSQL 10.6
+# -----------------------------
+
+shared_preload_libraries = 'pg_stat_statements'		# (change requires restart)
+
+pg_stat_statements.max = 5000          # max number of statements to track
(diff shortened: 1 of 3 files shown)


pg_init/99_postgresql_conf.sh
```diff
@@ -0,0 +1,46 @@
...
+					#   debug1
+					#   info
+					#   notice
+					#   warning
+					#   error
+					#   log
+					#   fatal
+					#   panic (effectively off)
+
+log_min_duration_statement = 5000	# -1 is disabled, 0 logs all statements
```
adberglund:
```suggestion
log_min_duration_statement = 2000	# -1 is disabled, 0 logs all statements
```

adberglund: This again is just to stay consistent with what the platform considers latency downtime today. I know they're tracking request/response time and that's certainly not all DB time, but this at least gets us close to that metric.

kholdaway:
Nice!

pg_init/99_postgresql_conf.sh
Red-HAP: This has been changed to 2000 ms

Red-HAP: Yep. Forgot to make that change initally.

Issue #1619, opened 2020-01-14
GroupBy categories


dlabrecq:
**Describe the bug**
Users want the ability to group_by categories. In the UI details page; for example, we want the ability to  group_by `tag:environment=dev` OR `tag:environment=qe`, but only show results for `cluster_1` or `cluster_2`.

In order to do that, we need to apply logical AND to each group_by category; project, cluster, tag, etc., while applying logical OR to the values belonging to the same group.

**Expected behavior**
Want to apply logical logical AND to group_by categories. For example, `group_by[cluster]` AND `group_by[project]` AND `group_by[tag:environment]`.

Want to apply logical OR to multiple values within the same group. For example, `group_by[cluster]=cluster_1` OR `group_by[cluster]=cluster_1`

**Screenshots**
<img width="1211" alt="Screen Shot 2020-01-14 at 1 37 16 PM" src="https://user-images.githubusercontent.com/17481322/72371752-01895c00-36d3-11ea-8dfc-ccd0585f2286.png">

**Additional context**
Complex filter chips
https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/filters#filter-chips

dlabrecq:
@nlcwong @chargio Without this functionality, we cannot properly implement tag filtering with the new PatternFly data toolbar.

chargio:
@dlabrecq I don't understand your request? 
Is it necessary to change the API? Or this is an additional issue for UI that needs to be prioritized?

dlabrecq:
This is an issue that needs to be prioritized in order to properly support tag filtering.

I don't believe the API necessarily needs to change, just how the back-end applies logical AND Vs OR. We simply need to apply logical AND between group categories. The group values can remain logical OR.

dlabrecq:
This is an issue that needs to be prioritized in order to properly support tag filtering in the UI.

I don't believe the API query params need to change, just how the back-end applies logical AND Vs OR. 

We simply need to apply logical AND between unique group categories; project, cluster, etc.. The group values can remain logical OR; cluster=cluster1, cluster=cluster2, etc..

nbon12:
Sprint Planning Meeting Minutes:
[...]

PR #1723, opened 2020-02-07
add unit tests for ReportQueryHandler

blentz:
As part of investigating #1619 , I found that there are no unit tests for any of the ReportQueryHandler methods. So, @adberglund asked me to add some tests to the methods related to #1619 . This PR does that.

I'm not exhaustively adding tests for all of the methods. I figure that's something we can iterate on over time.


koku/api/query_handler.py
```diff
@@ -327,7 +327,7 @@ def filter_to_order_by(self, parameters):  # noqa: C901
 
         """
         for group_by_key in parameters.parameters.get('group_by', {}):
-            for group_by_value in parameters.parameters['group_by'][group_by_key]:
+            for group_by_value in parameters.parameters.get('group_by')[group_by_key]:
```
adberglund:
```suggestion
            for group_by_value in parameters.parameters.get('group_by', {})[group_by_key]:
```

koku/api/query_handler.py
```diff
@@ -321,7 +321,7 @@ def filter_to_order_by(self, parameters):  # noqa: C901
 
         """
         for group_by_key in parameters.parameters.get("group_by", {}):
-            for group_by_value in parameters.parameters["group_by"][group_by_key]:
+            for group_by_value in parameters.parameters.get("group_by", {})[group_by_key]:
```
abaiken: I have a question. If the group_by defaults to the `{}`, directly accessing the `group_by_key` will cause a key error - Is this okay or should we do another get defaulted to another `{}`?

blentz: The default return value for `.get()` is `None`. 

The full signature of the `dict.get()` method is `.get(value, default=None)` and whatever the default is gets returned if an `AttributeError` is thrown by trying to access a key that's not in the dict.

myersCody: @abaiken correct me if I am wrong, but I think your point was that if he does a `.get("group_by", {})[group_by_key]`. If the `group_by` does not exist it, then you are guaranteed to get a key error because its an empty dictionary. I think the better alternative here would be to do something similar to how the `filter_value` is being handled below: https://github.com/project-koku/koku/blob/30e8dd5bc5467337a177065f47e6cbf4e9e2ea57/koku/api/query_handler.py#L328

abaiken: @blentz completely unrelated to your PR - can we simplify the two for loops here. Right now we are doing: 
```
for group_by_key in parameters.parameters.get('group_by', {}):
            for group_by_value in parameters.parameters['group_by'][group_by_key]:
```
but unless i'm misunderstanding - we could simplify it using: 
```
for group_by_key, group_by_value in parameters.parameters.get('group_by', {}).items():
```
I think this would make it more readable. Does that make sense to you all?  (I know this wasn't introduced by you - but I think I had a hard time understanding what was happening here bc of it)

blentz: @abaiken - I agree with you, the code itself likely can be simplified.

But, if it's alright with you, I'd like to handle fixing that looping in a separate PR. I'd like to keep this one to just adding the two missing tests.

koku/api/report/test/tests_queries.py
```diff
@@ -105,3 +114,137 @@ def test_group_data_by_list_missing_units(self):
...
+             a few attributes that are required by QueryHandler.__init__().
+         """
+
+        _mapper = Mock(
+            spec=ProviderMap,
+            _report_type_map=Mock(return_value=mapper, get=lambda x, y=None: mapper.get(x, y)),
+            _provider_map=Mock(return_value=mapper, get=lambda x, y=None: mapper.get(x, y)),
+        )
+
+    mock_params = Mock(
```
adberglund: There is an existing implementation that mocks out the QueryParams: https://github.com/project-koku/koku/blob/master/koku/api/iam/test/iam_test_case.py#L127 . And pattern for using the query handler here: https://github.com/project-koku/koku/blob/master/koku/api/report/test/aws/tests_queries.py#L67-L68. Might see if that can be reused.

koku/api/report/test/tests_queries.py
```diff
@@ -105,3 +114,137 @@ def test_group_data_by_list_missing_units(self):
...
+
+        fake_group = FAKE.word()
+        mapper = {"filter": {}, "filters": {}}
+        params = {"filter": {}, "group_by": {fake_group: ["and:" + FAKE.word(), "and:" + FAKE.word()]}, "order_by": {}}
+
+        rqh = create_test_handler(mapper, params)
+        output = rqh._set_operator_specified_filters(operator)
+        self.assertIsNotNone(output)
+
+        expected = Q(**{})
```
adberglund: Why is the expected here a Q with an empty dict?

koku/api/report/test/tests_queries.py
```diff
@@ -105,3 +114,137 @@ def test_group_data_by_list_missing_units(self):
...
+
+        fake_group = FAKE.word()
+        mapper = {"filter": {}, "filters": {}, "group_by_options": [fake_group]}
+        params = {"filter": {}, "group_by": {fake_group: ["or:" + FAKE.word(), "or:" + FAKE.word()]}, "order_by": {}}
+
+        rqh = create_test_handler(mapper, params)
+        output = rqh._set_operator_specified_filters(operator)
+        self.assertIsNotNone(output)
+
+        expected = Q(**{})
```
adberglund: Same question. Why an empty dict?

koku/api/report/test/tests_queries.py
blentz: It can't be reused because `AWSQueryHandler.__init__()` implements logic that doesn't exist in `ReportQueryHandler`, primarily the setting of the `_mapper` attribute. That's why I had to create a testable sub-class to mock-out some of the logic that we've delegated to the provider-specific sub-classes.

koku/api/report/test/tests_queries.py
blentz: Mostly because that's what was being returned, regardless of my inputs. 

I'm not sure if that's expected behavior, a bug in the method, or a problem with the inputs I'm using. 

One of the problems with not having unit tests on these functions written by the author of this code is that I'm forced to make some guesses about what should be tested based on the method's return output that I'm observing.

koku/api/query_handler.py
```diff
@@ -320,12 +320,11 @@ def filter_to_order_by(self, parameters):  # noqa: C901
...
         """
-        for group_by_key in parameters.parameters.get("group_by", {}):
-            for group_by_value in parameters.parameters["group_by"][group_by_key]:
-                if group_by_value == "*":
-                    # find if there is a filter[X]=Y that matches this group_by[X]=*
-                    # get filter value for current group_by_key
-                    filter_value = parameters.parameters.get("filter", {}).get(group_by_key)
-                    if filter_value:
-                        parameters.parameters["group_by"][group_by_key] = filter_value
+        for key, value in parameters.parameters.get("group_by", {}).items():
```
kholdaway: This method is fun with two dictionaries.  What could we do to clarify the meaning of key/value?  I think either we need an update to that comment or rename key/value to have a more specific meaning.

blentz: Look at the method's docstring. I think that contains the info you're looking for.

kholdaway: Looks good

kholdaway:
All the feedback I gave was addressed.  I removed the request changes flag and will allow Andrew to approve.

Issue #1616, opened 2020-01-13
Keep Only Current Month Line Item Data


adberglund:
**Describe the bug**
We recently made a config change to keep less raw data in the DB. Removing previous months data deletes our summary tables for that time period. If we want to keep less raw data, we either need to modify the expired data remover or update our DB models for the summary tables' on_delete option to not delete the summary tables when a bill is deleted. 

**Expected behavior**
We should be able to remove raw data and still display the summarized data in the UI

**Assumptions**
- Short term update e2e to have a 2 month config. [Done]

## Additional Information
- The Expired Data Remover class: https://github.com/project-koku/koku/blob/30208f4ccfd80db2270d6e523b0343d5d6139b19/koku/masu/processor/expired_data_remover.py#L45
- We essentially want a new method in each source-specific DBCleaner class, similar to the existing purge_expired_report_data, but one that removes all rows from the raw line item table for previous billing periods. 
  - Reference: https://github.com/project-koku/koku/blob/30208f4ccfd80db2270d6e523b0343d5d6139b19/koku/masu/processor/aws/aws_report_db_cleaner.py#L89-L91

nbon12:
Sprint Planning:
Andrew: "Sergio was showing the demo account during the all hands, and was asking why isn't there December data? When we ran out of disk space in the database. I made a choice to wipe a lot of the data there. We have an expired data remover class and celery scheduled task for beginning of every month. We ask it to keep a single months worth of data. Keeping summary data only is great but because of our foreign key usage, if you delete, it cascades and deletes summary data too. I'm inclined to call this a bug. Priority medium because we can fix it soon but in the meantime we can keep that setting to 2 months so we don't delete more. After we do that delete we end up processing the previous month anyway."
Chris: "In the short term we update e2e to have a 2 month config. I don't think it's important to solve this given other things we need to do. It's a good thing to work on but unless we start hitting thresholds again due to huge ramp up, we should be ok for this sprint."
Andrew: "In that case I moved priority to low because we have a quick fix. I'll make a PR right after this"

myersCody:
I have finished up the base of this issue, but still need to complete a part 2 essentially described in this comment by Andrew: https://github.com/project-koku/koku/pull/1727/files#diff-ceaeef4fc79d087d75a0859bd17d6b87R82

- [ ] Complete part 2

lcouzens:
Verified

PR #1727, opened 2020-02-10
Create method to keep only current month line item data.

myersCody:
#### How I tested
#1616
Note:
- If you want to use the django api you will need to change some environment variables. For example change the default months so that we can test the expired data here: https://github.com/project-koku/koku/blob/master/koku/masu/config.py#L113
- You will also need to set: `MASU_DEBUG=True`

```
make docker-reinitdb-with-providers
make load-test-customer-data start=2019-08-02 end=2020-02-11
```
(wait for masu to process)

Go ahead and check to make sure that the line item data was populated with:
```
(koku) bash-3.2$ PGPASSWORD=postgres psql postgres -U postgres -h localhost -p 15432
SELECT COUNT(id) FROM acct10001.reporting_azurecostentrylineitem_daily;
SELECT COUNT(id) FROM acct10001.reporting_ocpusagelineitem;
SELECT COUNT(id) FROM acct10001.reporting_awscostentrylineitem;
```
Example:
```
postgres=# SELECT COUNT(id) FROM acct10001.reporting_azurecostentrylineitem_daily;
 count
-------
  2968
(1 row)

postgres=# SELECT COUNT(id) FROM acct10001.reporting_ocpusagelineitem;
 count
--------
 611421
(1 row)

postgres=# SELECT COUNT(id) FROM acct10001.reporting_awscostentrylineitem;
 count
-------
 94227
(1 row)
```
1. Hit the masu expired_data endpoint: 
Now hit the masu expired_data with the new param created in the pr to trigger the deletion.
http://127.0.0.1:5000/api/cost-management/v1/expired_data/?line_items_only=True
Make sure to click the red delete button, or it will just do a simulated delete on the GET request.
2. Check the django logs:
Ctrl + F: `Calling purge_expired_line_item`

You should see something similar to:
```
koku_worker       | [2020-02-11 22:05:42,146] INFO e80796a7-7ffa-42a8-a861-8f312074f696 Calling purge_expired_lite_item for ocp
masu_server       | [2020-02-11 22:05:42,160] INFO None Expired data removal queued - schema_name: acct10001, Task ID: 22a03d2c-0ef1-4358-ad88-eadb622bd53a
koku_worker       | [2020-02-11 22:05:42,180] INFO e80796a7-7ffa-42a8-a861-8f312074f696 Removing (0, {'reporting.OCPUsageLineItem': 0}) usage period line items for usage period id 1
koku_worker       | [2020-02-11 22:05:42,180] INFO e80796a7-7ffa-42a8-a861-8f312074f696 Report line items usage period ID: 1 with interval start: 2019-08-01 00:00:00+00:00
koku_worker       | [2020-02-11 22:05:42,217] INFO e80796a7-7ffa-42a8-a861-8f312074f696 Removing (4320, {'reporting.OCPUsageLineItem': 4320}) usage period line items for usage period id 3
koku_worker       | [2020-02-11 22:05:42,217] INFO e80796a7-7ffa-42a8-a861-8f312074f696 Report line items usage period ID: 3 with interval start: 2019-09-01 00:00:00+00:00
koku_worker       | [2020-02-11 22:05:42,249] INFO e80796a7-7ffa-42a8-a861-8f312074f696 Removing (4464, {'reporting.OCPUsageLineItem': 4464}) usage period line items for usage period id 5
koku_worker       | [2020-02-11 22:05:42,249] INFO e80796a7-7ffa-42a8-a861-8f312074f696 Report line items usage period ID: 5 with interval start: 2019-10-01 00:00:00+00:00
koku_worker       | [2020-02-11 22:05:42,277] INFO e80796a7-7ffa-42a8-a861-8f312074f696 Removing (4320, {'reporting.OCPUsageLineItem': 4320}) usage period line items for usage period id 7
koku_worker       | [2020-02-11 22:05:42,278] INFO e80796a7-7ffa-42a8-a861-8f312074f696 Report line items usage period ID: 7 with interval start: 2019-11-01 00:00:00+00:00
masu_server       | [2020-02-11 22:05:42,287] INFO None DELETE: /api/cost-management/v1/expired_data/?line_items_only=True 200
masu_server       | [2020-02-11 22:05:42,289] INFO None "DELETE /api/cost-management/v1/expired_data/?line_items_only=True HTTP/1.1" 200 7891
koku_worker       | [2020-02-11 22:05:42,322] INFO e80796a7-7ffa-42a8-a861-8f312074f696 Removing (4320, {'reporting.OCPUsageLineItem': 4320}) usage period line items for usage period id 15
koku_worker       | [2020-02-11 22:05:42,323] INFO e80796a7-7ffa-42a8-a861-8f312074f696 Report line items usage period ID: 15 with interval start: 2019-08-01 00:00:00+00:00
koku_worker       | [2020-02-11 22:05:42,397] INFO e80796a7-7ffa-42a8-a861-8f312074f696 Removing (4320, {'reporting.OCPUsageLineItem': 4320}) usage period line items for usage period id 17
koku_worker       | [2020-02-11 22:05:42,397] INFO e80796a7-7ffa-42a8-a861-8f312074f696 Report line items usage period ID: 17 with interval start: 2019-09-01 00:00:00+00:00
koku_worker       | [2020-02-11 22:05:42,439] INFO e80796a7-7ffa-42a8-a861-8f312074f696 Removing (4464, {'reporting.OCPUsageLineItem': 4464}) usage period line items for usage period id 19
koku_worker       | [2020-02-11 22:05:42,440] INFO e80796a7-7ffa-42a8-a861-8f312074f696 Report line items usage period ID: 19 with interval start: 2019-10-01 00:00:00+00:00
koku_worker       | [2020-02-11 22:05:42,476] INFO e80796a7-7ffa-42a8-a861-8f312074f696 Removing (4320, {'reporting.OCPUsageLineItem': 4320}) usage period line items for usage period id 21
koku_worker       | [2020-02-11 22:05:42,476] INFO e80796a7-7ffa-42a8-a861-8f312074f696 Report line items usage period ID: 21 with interval start: 2019-11-01 00:00:00+00:00
koku_worker       | [2020-02-11 22:05:43,057] INFO e80796a7-7ffa-42a8-a861-8f312074f696 Removing (96629, {'reporting.OCPUsageLineItem': 96629}) usage period line items for usage period id 29
koku_worker       | [2020-02-11 22:05:43,057] INFO e80796a7-7ffa-42a8-a861-8f312074f696 Report line items usage period ID: 29 with interval start: 2019-08-01 00:00:00+00:00
koku_worker       | [2020-02-11 22:05:43,263] INFO e80796a7-7ffa-42a8-a861-8f312074f696 Removing (38217, {'reporting.OCPUsageLineItem': 38217}) usage period line items for usage period id 31
koku_worker       | [2020-02-11 22:05:43,263] INFO e80796a7-7ffa-42a8-a861-8f312074f696 Report line items usage period ID: 31 with interval start: 2019-09-01 00:00:00+00:00
koku_worker       | [2020-02-11 22:05:43,672] INFO e80796a7-7ffa-42a8-a861-8f312074f696 Removing (96458, {'reporting.OCPUsageLineItem': 96458}) usage period line items for usage period id 33
koku_worker       | [2020-02-11 22:05:43,672] INFO e80796a7-7ffa-42a8-a861-8f312074f696 Report line items usage period ID: 33 with interval start: 2019-10-01 00:00:00+00:00
koku_worker       | [2020-02-11 22:05:44,065] INFO e80796a7-7ffa-42a8-a861-8f312074f696 Removing (72193, {'reporting.OCPUsageLineItem': 72193}) usage period line items for usage period id 35
```

Now check to see the the line item data we saw earlier is gone:
```
SELECT COUNT(id) FROM acct10001.reporting_azurecostentrylineitem_daily;
SELECT COUNT(id) FROM acct10001.reporting_ocpusagelineitem;
SELECT COUNT(id) FROM acct10001.reporting_awscostentrylineitem;
```
Example:
```
qpostgres=# SELECT COUNT(id) FROM acct10001.reporting_azurecostentrylineitem_daily;
 count
-------
  1274
(1 row)

postgres=# SELECT COUNT(id) FROM acct10001.reporting_ocpusagelineitem;
 count
--------
 273076
(1 row)

postgres=# SELECT COUNT(id) FROM acct10001.reporting_awscostentrylineitem;
 count
-------
 40541
(1 row)
```


myersCody:
Whoops. Just kidding, moved pr back to WIP because its keep up to 3 months worth of line item data. Need to write a method to set a different expired_date for the new method.

Makefile
```diff
@@ -66,6 +66,8 @@ help:
 	@echo "  large-ocp-provider-testing            create a test OCP provider "large_ocp_1" with a larger volume of data"
 	@echo "                                          @param nise_config_dir - directory of nise config files to use"
 	@echo "  load-test-customer-data               load test data for the default providers created in create-test-customer"
+	@echo "                                          @param start - (optional) start date ex. 2019-08-02"
+	@echo "                                          @param end - (optional) end date ex. 2019-12-5"
```
adberglund: 🙇

koku/masu/api/expired_data.py
```diff
@@ -42,8 +43,19 @@ def expired_data(request):
         simulate = False
     LOG.info("Simulate Flag: %s", simulate)
 
+    params = request.query_params
+    acceptabools = ["true", "True", "false", "False"]
+    line_items_only = params.get("line_items_only", "False")
+    if line_items_only not in acceptabools:
```
adberglund: First off, ❤️ the variable name `acceptabools`. And since this is an internal-only API this is not remotely critical, but one thing you could do here is call `.lower()` (or upper) on the `line_items_only` variable and just make sure it's `true` or `false`. That way someone could do `tRuE` as the param and it would still pass.

myersCody: Glad you appreciate the puns 😂 The `.lower()` is a wonderful idea. Super simple and elegant but catches a lot of edge cases. 👍

blentz: I like the variable name, but if it's only being used once, it doesn't need to be assigned to a variable at all.

```
    if line_items_only not in ["true", "false"]:
```

Though, if you want to get _really_ clever, you can use the `json` module to parse the string into a native Python type for you. (It'll throw a `JSONDecodeError` if it's not "true" or "false" or some other valid JSON value. So, you'll need a try/except block around it.)


```
IPython 7.11.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import json                                                                                                                                        
In [2]: foo = "true"                                                                                                                                       
In [3]: json.loads(foo)                                                                                                                                    
Out[3]: True
```

koku/masu/processor/aws/aws_report_db_cleaner.py
```diff
@@ -42,6 +42,58 @@ def __init__(self, schema):
...
+                bill_objects = accessor.get_cost_entry_bills_query_by_provider(provider_uuid)
+            with schema_context(self._schema):
+                for bill in bill_objects.all():
+                    bill_id = bill.id
+                    removed_payer_account_id = bill.payer_account_id
+                    removed_billing_period_start = bill.billing_period_start
+
+                    if not simulate:
+                        del_count = accessor.get_lineitem_query_for_billid(bill_id).delete()
+                        LOG.info("Removing %s cost entry line items for bill id %s", del_count, bill_id)
```
blentz: This approach might have performance consequences that we might want to think about. A moderately large cost report can be tens of millions of lineitem rows, which may require that we lock the table to do the deletes. Or we may need to think about table partitioning or other strategies. (Some of this might be moot if we move away from Postgres in the medium-term.)

Have you tried running this delete while masu is simultaneously trying to ingest new data?  I'm curious if we'd see a deadlock or just poor performance.

I'll defer to @adberglund or @Red-HAP , as I think they've got more DB experience than I do. But, I'd like to at least sanity-check that this is a workable solution for the size of the data we're expecting to see. :slightly_smiling_face:

myersCody: I have tried running this delete while data is being ingested and what happens is that the delete task is queued until after the ingestion is completed. As for the performance concerns, I pretty much stole this logic from here: https://github.com/project-koku/koku/blob/master/koku/masu/processor/aws/aws_report_db_cleaner.py#L86

Which means that if we do have performance concerns about this method we should make sure to do it in both places, and it seems like there a lot of delete tasks that do something very similar in the `purge_expired_report_data` method, so it may be worth turning this into its own issue at that point to check on it.

koku/masu/api/expired_data.py
myersCody: Not gonna lie, I was just trying to keep the `acceptabools` variable 😂 , but I can totally make this change.

myersCody: So, I realized as I was making the change that I also use the acceptabools in the error message here: https://github.com/project-koku/koku/pull/1727/files#diff-b344b463fb32d3ad4607281872704ecdR51

So, I just kept it in, but I can also just write out in the string `['true', 'false']` if we do want to get rid of it.

koku/masu/processor/aws/aws_report_db_cleaner.py
adberglund: I think the task stayed queued because celery was only set to run 1 task at a time, if we were running multiple workers it could trigger. However, the lock is a row exclusive lock, so it will only lock the rows being deleted, not prevent new data from being inserted. There is a case that we do want to prevent: This task gets called and run in the middle of processing a report from last month (like say at the beginning of this month when we get updates to the previous month's report). So we probably want some condition to make sure we only run it after finishing processing a report from the previous month.

This brings up another important point that I missed last week. We need a place that this is regularly being called. Unless I missed that, I think the only way to trigger this today is via the API call. We could add the call to the `remove_expired_data` task with `line_items_only=True` as the last item of the chain here. 

https://github.com/project-koku/koku/blob/7acb90eb68c69268ea8ca9c616e5ff2bc9efe04f/koku/masu/processor/tasks.py#L233-L238

We could conditionally trigger that task when we've just processed data from a previous month. The month of the `start_date` in the `update_summary_tables` task would not be the current month.

myersCody: Andrew and I talked about this, and we decided it would be easier to review the trigger in tasks.py as a separate pr to this user story and to use this pr as a base.

adberglund: The DELETE lock type is `ROW EXCLUSIVE` which does not block access to the table, just the deleted rows. But we should make sure we only run this during certain times

PR #1735, opened 2020-02-11
Get indicator for tags on group by AWS account

Red-HAP:
## Description

Build a second query using the same initial filtering and add another filter criteria for the unique accounts returned in the report query results.

Get the data for the tag query, then set a new item in the results for the report query.

This was done to not pollute the count and group logic in the python code.

## Issue

Tried multiple variations on the theme using Django ORM, but it was generating SQL that was inefficient and grouping by the primary key of the table which was wrong. After fighting it for a week, I put in a static (mostly) query to get the data as efficiently as possible.

I don't expect this to be the final solution, but I was wanting to get this ticket moving.


koku/api/report/aws/query_handler.py
```diff
@@ -243,6 +247,16 @@ def execute_query(self):
...
                 )
 
+                # Setup tags exist query for the returned data
+                tag_indicator_query = query_table.objects.filter(self.query_filter)
+                tag_indicator_query = tag_indicator_query.annotate(
+                    account_alias=Coalesce(F(self._mapper.provider_map.get("alias")), "usage_account_id"),
+                    tags_exist_sum=Sum(ExpressionWrapper(Coalesce(F('tags'), {}) != {}, output_field=IntegerField))
+                )
+                tag_indicator_query.group_by = ['account_alias']
+                tag_indicator_query = tag_indicator_query.values('acount_alias', 'tags_exist_sum')
```
blentz:
```suggestion
                tag_indicator_query = tag_indicator_query.values('account_alias', 'tags_exist_sum')
```

koku/api/report/aws/query_handler.py
```diff
@@ -243,6 +247,16 @@ def execute_query(self):
                     account_alias=Coalesce(F(self._mapper.provider_map.get("alias")), "usage_account_id")
                 )
 
+                # Setup tags exist query for the returned data
+                tag_indicator_query = query_table.objects.filter(self.query_filter)
+                tag_indicator_query = tag_indicator_query.annotate(
+                    account_alias=Coalesce(F(self._mapper.provider_map.get("alias")), "usage_account_id"),
+                    tags_exist_sum=Sum(ExpressionWrapper(Coalesce(F('tags'), {}) != {}, output_field=IntegerField))
+                )
+                tag_indicator_query.group_by = ['account_alias']
```
adberglund: Using .values in the line below should handle the group by for you so you don't have to explicitly set it.

koku/api/report/aws/query_handler.py
```diff
@@ -213,6 +215,136 @@ def _build_sum(self, query, annotations):
...
+                "lte": "<=",
+                "gt": ">",
+                "gte": ">=",
+                "contains": "like",
+                "icontains": "ilike",
+                "startswith": "like",
+                "istartswith": "ilike",
+                "endswith": "like",
+                "iendswith": "ilike",
+            }
```
blentz: Your operator map is different in a couple places than how it's implemented in the Django Postgres backend. ( See: https://github.com/django/django/blob/master/django/db/backends/postgresql/base.py#L106-L121 )

Do you have an opinion on the differences between, for example, `ILIKE` versus `LIKE UPPER(%s)`?

Secondarily, do you think it's worthwhile to use the `DatabaseWrapper.operators` from Django's postgres backend instead of implementing our own version of it? It seems like it'll lower the maintenance overhead to use Django's version.

Red-HAP: I know it's a bit different. In this case, at this time, both ILIKE and UPPER(col) LIKE UPPER(val) are both going to break indexing equally. 

I'm working on means to fix this on the ocp query performance ticket. Some significant work there should translate to AWS, Azure, etc.

Issue #1734, opened 2020-02-11
Assess current implementation of Tags APIs


adberglund:
## User Story
As a customer, I want tags to return quickly so that I can filter and group by at ease.

## Impacts
API, UI

## Assumptions
- Currently the tags APIs get tags from the line item daily summary table instead of the much smaller tag summary tables. https://github.com/project-koku/koku/blob/master/koku/api/tags/aws/queries.py#L27
- If it is easy to switch over to using the tag summary table, then we should go ahead and do that. 
- The date filters on the tags APIs will likely be the thing that causes trouble with using the tag summary tables. We can either
  1) rework the filters in the tag query handler to look at the bill's start and end dates. 
  2) add in usage_start and usage_end columns to the tag summary tables and populate them using the bill start/end dates so the filters work (potentially) as-is.
 

## Acceptance Criteria

- [x] We determine if we should be using the tag summary tables
- [x] Tag queries work against the tag summary tables
- [x] No regressions

lcouzens:
Verified

PR #1736, opened 2020-02-12
1734 Use tag summary tables for tags endpoints

maskarb:
#1734 

Recently, the slow performance of the tags endpoints was causing timeouts. This PR uses the TagSummary tables to populate the tags endpoints, rather than an expensive query of the line-item tables.

Testing:
Ingest data with tags, ensure tags endpoint still works.


koku/api/tags/queries.py
```diff
@@ -195,71 +212,71 @@ def get_tag_keys(self, filters=True):
...
-                tag_keys = [tag.get(source.get("db_column")) for tag in tag_keys]
+                for dikt in tag_keys:
+                    merged_data[dikt.get("key")].extend(dikt.get("values"))
+
+                if source.get("type"):
+                    self.append_to_final_data_with_type(final_data, merged_data, source)
+                    merged_data = defaultdict(list)
+
+            if not source.get("type"):
+                self.append_to_final_data_without_type(final_data, merged_data)
```
maskarb: Need to think of better logic here. If we were to look at OCP and non-OCP tables, this logic would explode, and we would likely drop some tags somewhere.

maskarb: Need to think of better logic here. If we were to look at OCP and non-OCP tables at the same time, this logic would explode, and we would likely drop some tags somewhere.

koku/api/query_handler.py
```diff
@@ -145,8 +145,6 @@ def get_resolution(self):
             (String): The value of how data will be sliced.
 
         """
-        if self.resolution:
-            return self.resolution
```
blentz: Why is this being removed?

koku/api/tags/queries.py
```diff
@@ -80,6 +70,11 @@ def __init__(self, parameters):
             parameters    (QueryParameters): parameter object for query
 
         """
+        if parameters.parameters.get("filter", {}).get("time_scope_value") == "-10":
+            parameters.parameters["filter"]["time_scope_value"] = "-1"
+            parameters.parameters["filter"]["time_scope_units"] = "month"
+            parameters.parameters["filter"]["resolution"] = "monthly"
```
blentz: `get_filter()` can be used to simplify this.

koku/api/query_handler.py
maskarb: Since I'm setting self.resolution in the __init__ method, the lines below (150-160) would not have been called resulting in attribute errors.

koku/api/tags/queries.py
```diff
@@ -80,6 +70,9 @@ def __init__(self, parameters):
             parameters    (QueryParameters): parameter object for query
 
         """
+        if parameters.get_filter("time_scope_value") == "-10":
```
adberglund: The initial reason to get rid of 10/30 days is the tags summary table is keyed off of the billing period (so -1 and -2 / monthly are naturally supported). 

This complicates the logic a little, but we could do some date checking here. If 10/30 days ago is in last month, we could return tags for both months. I think the tag query handler would just have to have its own `def `_get_date_delta` method that returns the correct date ranges? 

The date comparison can be done using our date helper class: https://github.com/project-koku/koku/blob/4a1a08abeca0c7db845a55da4e7bf10c41b4ef97/koku/api/utils.py#L26

It already has convenience methods for n_days_ago and month start/stops.

koku/api/tags/queries.py
maskarb: These lines have been removed.

PR #1740, opened 2020-02-12
Remove stale data from tenants

adberglund:
## Summary
A recent change to update the tag summary table with the bill id / report period id is failing in CI due to stale data. There should not be a case now where we hold on to data that does not have a bill/report period associated with it. The expired data remover hinges on data having these columns populated, so this is stale data that has never been cleaned out. So this blanket data migration **should** be safe as no current data is missing these columns.


koku/reporting/migrations/0095_auto_20200212_1606.py
```diff
@@ -0,0 +1,43 @@
...
+
+            DELETE FROM reporting_ocpusagelineitem_daily_summary
+            WHERE report_period_id IS NULL;
+
+            DELETE FROM reporting_ocpstoragelineitem
+            WHERE report_period_id IS NULL;
+
+            DELETE FROM reporting_ocpstoragelineitem_daily
+            WHERE report_period_id IS NULL;
+            """
```
Red-HAP: Here is a list of other reporting tables that have report_period_id. Do these tables need to be considered as well?
```
                      table_name                      |   column_name    
------------------------------------------------------+------------------
 reporting_ocpawscostlineitem_daily_summary           | report_period_id
 reporting_ocpawscostlineitem_project_daily_summary   | report_period_id
 reporting_ocpazurecostlineitem_daily_summary         | report_period_id
 reporting_ocpazurecostlineitem_project_daily_summary | report_period_id
 reporting_ocpcosts_summary                           | report_period_id
 reporting_ocpusagereport                             | report_period_id
```

adberglund: Will update.

Red-HAP: The only other question that I have is: if this data is not supposed to be in the system, should it be blocked with a NOT NULL constraint on that FK?

adberglund: Capturing this in a follow-up PR. We should do the NOT NULLs, but it also breaks MANY tests, and this is blocking QE now, so we should get it out as the first step.

PR #1748, opened 2020-02-14
Updated jenkins file to meet new e2e updates

lcouzens:
Updating smoke deployment, basically the platform service set was getting too bulky and slowing down smoke deployments + using up quotas. Long story short most apps only need a small subset of services so they are getting split up in e2e. This PR tweaks are requirements so we only deploy the pieces we require for testing.

### DIFF
--- a/Jenkinsfile-smoke.groovy
+++ b/Jenkinsfile-smoke.groovy
@@ -9,7 +9,7 @@ if (env.CHANGE_ID) {
     execSmokeTest (
         ocDeployerBuilderPath: "hccm/koku",
         ocDeployerComponentPath: "hccm",
-        ocDeployerServiceSets: "ingress,inventory,platform-mq,hccm",
+        ocDeployerServiceSets: "ingress,platform-mq,hccm",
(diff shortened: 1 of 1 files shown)


Jenkinsfile-smoke.groovy
```diff
@@ -9,7 +9,7 @@ if (env.CHANGE_ID) {
     execSmokeTest (
         ocDeployerBuilderPath: "hccm/koku",
         ocDeployerComponentPath: "hccm",
-        ocDeployerServiceSets: "hccm,platform,platform-mq",
+        ocDeployerServiceSets: "ingress,inventory,platform-mq,hccm",
```
lcouzens: Also my guess is we don't actually need inventory either right? We can probably remove that too if I am correct @adberglund

abaiken: @lcouzens  I don't think we need inventory - but @adberglund would know better than me!

adberglund: Yeah, inventory can go.

PR #1749, opened 2020-02-14
Add script to clean out testing dirs

adberglund:
## Summary
If there's data sitting in our testing directories, it can cause false failures in our IQE tests. This clears out data before running.

### DIFF
--- a/scripts/clear_testing.py
+++ b/scripts/clear_testing.py
@@ -1,0 +1,58 @@
+#
+# Copyright 2020 Red Hat, Inc.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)


scripts/clear_testing.py
```diff
@@ -0,0 +1,58 @@
...
+
+    paths_to_clear = [f"{testing_path}/{directory}" for directory in TESTING_DIRS]
+
+    for path in paths_to_clear:
+        try:
+            print(f"Checking {path}")
+            dirs_to_remove = [f.path for f in os.scandir(path) if f.is_dir()]
+            for directory in dirs_to_remove:
+                print(f"Removing {directory}")
+                shutil.rmtree(directory)
```
blentz: If you're using `rmtree`, you only need to remove the `local_providers` and `pvc_dir` directories. It'll recursively remove everything in those directories.

Second, you may want to add a call to `os.mkdir()` to recreate the removed paths.

adberglund: Each of those directories gets mounted as a volume, so I'd like to not delete them, just their contents.

blentz: The most direct option is `git clean -- <path>`

These paths are in the repository, so you can have git do all of the work for you without needing to specify anything beyond the base `testing` path.

PR #1766, opened 2020-02-17
fix make help & DEMO_ACCOUNTS

myersCody:
After rebuilding docker-compose and bringing it up I ran into a DEMO_ACCOUNTS error.
```
koku_worker       |     self._wrapped = Settings(settings_module)
koku_worker       |   File "/opt/app-root/lib/python3.6/site-packages/django/conf/__init__.py", line 157, in __init__
koku_worker       |     mod = importlib.import_module(self.SETTINGS_MODULE)
koku_worker       |   File "/opt/app-root/lib64/python3.6/importlib/__init__.py", line 126, in import_module
koku_worker       |     return _bootstrap._gcd_import(name[level:], package, level)
koku_worker       |   File "/koku/koku/koku/settings.py", line 392, in <module>
koku_worker       |     DEMO_ACCOUNTS = ENVIRONMENT.json("DEMO_ACCOUNTS", {})
koku_worker       |   File "/opt/app-root/lib/python3.6/site-packages/environ/environ.py", line 173, in json
koku_worker       |     return self.get_value(var, cast=json.loads, default=default)
koku_worker       |   File "/opt/app-root/lib/python3.6/site-packages/environ/environ.py", line 290, in get_value
koku_worker       |     value = self.parse_value(value, cast)
koku_worker       |   File "/opt/app-root/lib/python3.6/site-packages/environ/environ.py", line 346, in parse_value
koku_worker       |     value = cast(value)
koku_worker       |   File "/opt/rh/rh-python36/root/usr/lib64/python3.6/json/__init__.py", line 354, in loads
koku_worker       |     return _default_decoder.decode(s)
koku_worker       |   File "/opt/rh/rh-python36/root/usr/lib64/python3.6/json/decoder.py", line 339, in decode
koku_worker       |     obj, end = self.raw_decode(s, idx=_w(s, 0).end())
koku_worker       |   File "/opt/rh/rh-python36/root/usr/lib64/python3.6/json/decoder.py", line 357, in raw_decode
koku_worker       |     raise JSONDecodeError("Expecting value", s, err.value) from None
```

### DIFF
--- a/koku/koku/settings.py
+++ b/koku/koku/settings.py
@@ -389,7 +389,4 @@ SOURCES_CLIENT_BASE_URL = "http://{}:{}{}/v1".format(
 )
 
 # Demo Accounts list
-try:
-    DEMO_ACCOUNTS = ENVIRONMENT.json("DEMO_ACCOUNTS", {})
-except ValueError:
-    DEMO_ACCOUNTS = {}
+DEMO_ACCOUNTS = ENVIRONMENT.json("DEMO_ACCOUNTS", {})
--- a/koku/koku/settings.py
+++ b/koku/koku/settings.py
@@ -389,4 +389,7 @@ SOURCES_CLIENT_BASE_URL = "http://{}:{}{}/v1".format(
 )
 
 # Demo Accounts list
-DEMO_ACCOUNTS = ENVIRONMENT.json("DEMO_ACCOUNTS", {})
+try:
+    DEMO_ACCOUNTS = ENVIRONMENT.json("DEMO_ACCOUNTS", {})
+except ValueError:
+    DEMO_ACCOUNTS = {}
--- a/.env.example
+++ b/.env.example
@@ -23,4 +23,3 @@ KOKU_API_HOSTNAME=koku-hccm.apps-crc.testing #localhost for docker-compose
 KOKU_PORT=8000
 MASU_API_HOSTNAME=masu-hccm.apps-crc.testing #localhost for docker-compose
 MASU_PORT=5000
-DEMO_ACCOUNTS={}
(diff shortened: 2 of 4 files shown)


chambridge:
What about: https://github.com/project-koku/koku/blob/master/docker-compose.yml#L39


‘’’

- DEMO_ACCOUNTS=${DEMO_ACCOUNTS-{}}
‘’’

myersCody:
> What about: https://github.com/project-koku/koku/blob/master/docker-compose.yml#L39
> 
> DEMO_ACCOUNTS=${DEMO_ACCOUNTS-{}}

Ahhhh. That is a good idea, let me give that a try real quick.

koku/koku/settings.py
```diff
@@ -389,4 +389,7 @@
 )
 
 # Demo Accounts list
-DEMO_ACCOUNTS = ENVIRONMENT.json("DEMO_ACCOUNTS", {})
+try:
+    DEMO_ACCOUNTS = ENVIRONMENT.json("DEMO_ACCOUNTS", {})
```
chambridge: Did I just mess this up and it should be:
DEMO_ACCOUNTS = ENVIRONMENT.json("DEMO_ACCOUNTS", default={})

myersCody: I also tried the `default={}` there and it still came out a little wonky. I think it is using the built in .json in the dir(Environment), and when the DEMO_ACCOUNT doesn't exit it just throughs a value error.

chambridge: Same with default=“{}”?
Just wondering if it’s gonna parse it and needs a string instead of a dict

myersCody:
> What about: https://github.com/project-koku/koku/blob/master/docker-compose.yml#L39
> 
> DEMO_ACCOUNTS=${DEMO_ACCOUNTS-{}}

@chambridge I confirmed this works too, if we add it to both instances. So, we can do that too if you prefer it there instead of the .env

chambridge:
I think the docker-compose update would be nice in addition so nobody needs to update their .env

Thanks for trying it all out.

Issue #491, opened 2018-11-28
Total charge/usage/request not matching total of calculated items listed when using multiple group_bys


lcouzens:
**Note** 
This is an OLD issue, but as of the last comment is still a potential issue. Surfacing again because we are now supporting multiple group bys more readily in the UI.

**Describe the bug**
Total value is not matching listed items calculated total value when using group_by[node] with other group_by's

**To Reproduce**
1. Add ocp provider with multiple nodes
2. Add Rates to provider
3. api/v1/reports/inventory/ocp/cpu/??group_by[node]=*&group_by[cluster]=*&group_by[project]=*&filter[resolution]=daily&filter[time_scope_value]=-1&filter[time_scope_units]=month
4. Add up listed item values and compare with overall totals

**Expected behavior**
calculated listed items should match totals

**Context (please complete the following information):**
 - OS: Fedora 28
 - Browser: firefox 63.03
 - Commit Hash from status endpoint: "332008b5f7960302b6defd5fde55dd78f58fdd47"

**Additional context**
Working with Nise data added as a local ocp provider.

dccurtis:
Example:

```
GET /api/v1/reports/inventory/ocp/cpu/?filter[resolution]=monthly&filter[time_scope_value]=-1&filter[time_scope_units]=month&group_by[node]=*&group_by[cluster]=*&group_by[project]=*
HTTP 200 OK
Allow: OPTIONS, GET
Content-Type: application/json
Vary: Accept

{
    "group_by": {
        "cluster": [
            "*"
        ],
        "project": [
            "*"
        ],
        "node": [
            "*"
        ]
    },
    "filter": {
        "resolution": "monthly",
        "time_scope_value": "-1",
        "time_scope_units": "month"
    },
    "data": [
        {
            "date": "2018-11",
            "nodes": [
                {
                    "node": "node_pick",
                    "clusters": [
                        {
                            "cluster": "my-ocp-cluster-1",
                            "projects": [
                                {
                                    "project": "possible_qa",
                                    "values": [
                                        {
                                            "node": "node_pick",
                                            "date": "2018-11",
                                            "cluster": "my-ocp-cluster-1",
                                            "project": "possible_qa",
                                            "usage": 106.377409,
                                            "request": 192.76767,
                                            "limit": null,
                                            "charge": 0.0
                                        }
                                    ]
                                }
                            ]
                        }
                    ]
                },
                {
                    "node": "node_these",
                    "clusters": [
                        {
[...]

PR #1767, opened 2020-02-18
Fix _group_data_by_list method to not overwrite duplicate keys

myersCody:
#491

I wrote a python script to help test this one. You can run it on master, then on my branch to see how my change has made it to where the subtotals will add up to the totals. 
https://gist.github.com/myersCody/36ae96daa07356e14a5702a3cf63ff06

- [x] IQE smoke tests passed


adberglund:
@lcouzens I don't know if you want to steal code from @myersCody 's gist in this PR, for adding up totals with lots of group bys? Maybe you already have it, but it's there for the taking.

PR #1777, opened 2020-02-18
(1639) add celery task to clean worker volume weekly

abaiken:
This PR adds a weekly celery task (scheduled to run on Sunday at midnight) to delete any files on the volume that do not belong to manifests that we need to process. 

Testing:
--------
1. Remove the `if not settings.DEVELOPMENT` condition and edit the schedule for the `masu.celery.tasks.clean_volume` task in `koku/koku/celery.py` to run every minute: 
```
app.conf.beat_schedule["clean_volume"] = {
    "task": "masu.celery.tasks.clean_volume",
    "schedule": crontab(minute="*")
}
```
2. Bring up the project as normal and look at the koku-worker logs: 
```
make docker-up
docker-compose -f logs koku-worker
```
3. Populate the manifest table: 
first make sure that you have the following in your env file 
```
API_PATH_PREFIX='/api/cost-management'
NISE_REPO_PATH=/path/to/nise
KOKU_API_HOSTNAME=127.0.0.1
KOKU_PORT=8000
MASU_API_HOSTNAME=127.0.0.1
MASU_PORT=5000
```
and then run the following: 
```
make docker-reinitdb-with-providers
make load-test-customer-data
```
4. Now exec into your db and view all of the records for the manifest table:
```
docker exec -it koku_db psql -U postgres
SELECT * FROM reporting_common_costusagereportmanifest;
```
5. Now we are going to insert a fake record into the manifest table that has not finished processing - however modify the below to have an existing provider id - (replace `9be19863-cbfc-472d-bab2-b47ddd93c8ac` with your valid provider_id) : 
```
INSERT INTO reporting_common_costusagereportmanifest(assembly_id, manifest_creation_datetime, manifest_updated_datetime, num_processed_files, num_total_files, provider_id, billing_period_start_datetime) VALUES ('234567', '2020-02-19 22:09:20.202402+00', '2020-02-19 22:10:04.688277+00', 0, 1, '9be19863-cbfc-472d-bab2-b47ddd93c8ac', '2020-02-01 00:00:00+00');
```
Now - relook at the entries for the reporting_common_costusagereportmanifest table and make sure that your new entry is there. 

6. Create some random files in your PVC_DIR - this is in `koku/testing/pvc_dir/` if you are running using docker - you can create files at the top level or in subdirectories. Also create a file that is prefixed with the assembly_id from the entry that you created - (`234567` if you copied my insert): 
```
(koku-iU2DMmAw) bash-3.2$ pwd
/Users/ashleybaiken/cmdevelopment/koku/testing/pvc_dir
(koku-iU2DMmAw) bash-3.2$ touch 234567.csv
(koku-iU2DMmAw) bash-3.2$ touch fakedata.csv
```
7. Finally check the koku-worker logs - you should see something like the following: 
```
koku_worker       | [2020-02-19 14:49:00,046: INFO/ForkPoolWorker-1] [masu.celery.tasks.clean_volume(636d25dc-470d-4540-8a9e-1ebcbed01344 via 636d25dc-470d-4540-8a9e-1ebcbed01344)] Cleaning up the volume at /testing/pvc_dir 
koku_worker       | [2020-02-19 14:49:00,101: INFO/ForkPoolWorker-1] [masu.celery.tasks.clean_volume(636d25dc-470d-4540-8a9e-1ebcbed01344 via 636d25dc-470d-4540-8a9e-1ebcbed01344)] The following files were deleted: 
koku_worker       | [2020-02-19 14:49:00,101: INFO/ForkPoolWorker-1] [masu.celery.tasks.clean_volume(636d25dc-470d-4540-8a9e-1ebcbed01344 via 636d25dc-470d-4540-8a9e-1ebcbed01344)] ['/testing/pvc_dir/fakerdata.csv', '/testing/pvc_dir/fakerr/fakest.csv']
```
8. Check to make sure that the file you created with the assembly_id prefix is still there and that the other files have been discarded.


koku/masu/celery/tasks.py
```diff
@@ -312,3 +315,36 @@ def vacuum_schemas():
...
+
+@app.task(name="masu.celery.tasks.clean_volume", queue_name="clean_volume")
+def clean_volume():
+    """Clean up the volume in the worker pod."""
+    LOG.info("Cleaning up the volume at %s " % Config.PVC_DIR)
+    # get the billing months to use below
+    months = DateAccessor().get_billing_months(Config.INITIAL_INGEST_NUM_MONTHS)
+    db_accessor = ReportManifestDBAccessor()
+    # this list is initialized with the .gitkeep file so that it does not get deleted
+    assembly_ids_to_exclude = [".gitkeep"]
```
myersCody: I am not really sure the context of the issue you are working on, but I was wondering if this was something that may be worth defaulting in the `masu/config.py`? So that way we could change assembly_ids we want to exclude on the fly in openshift. The answer could totally be no, but I figured it would be worth suggesting just in case. 🤷‍♂

koku/masu/celery/tasks.py
```diff
@@ -312,3 +315,36 @@ def vacuum_schemas():
...
         LOG.info("Scheduling VACUUM task for %s", schema_name)
         vacuum_schema.delay(schema_name)
+
+
+@app.task(name="masu.celery.tasks.clean_volume", queue_name="clean_volume")
+def clean_volume():
+    """Clean up the volume in the worker pod."""
+    LOG.info("Cleaning up the volume at %s " % Config.PVC_DIR)
+    # get the billing months to use below
+    months = DateAccessor().get_billing_months(Config.INITIAL_INGEST_NUM_MONTHS)
```
myersCody: So, this would only ever clean out this month and the previous month right? So, if I had something in my volume from December it wouldn't be touched?

abaiken: It calculates the months that it uses to query the db by looking at the `INITIAL_INGEST_NUM_MONTHS` variable (which defaults to 2 - so unless you change that variable - the months to query for the latest seen manifest would be January and February). However the manifest ids that are returned from the query that uses the months are the ones that we don't want to delete - so on the contrary, if you had something in your volume from December, it would be deleted. The only way that files will not be deleted is if they are associated with the `assembly_ids` of the manifests that were most recently seen and have not been processed (which are filtered by month).

koku/masu/celery/tasks.py
abaiken: This list is actually populated below in line 332. However, I initialized it with `.gitkeep` inside because if you look at the repo (https://github.com/project-koku/koku/tree/master/testing/pvc_dir) the only thing in testing/pvc_dir is the `.gitkeep` file - the volume is mapped to our `koku/testing/pvc_dir` when you develop using docker. So if we were to run this locally it deletes all files that do not have one of the "assembly_ids" in them and this was my way of keeping the `.gitkeep` file from being deleted. I am open to doing this differently or renaming the list to make it more clear but I'm not sure that moving it `masu/config` is more clear. That being said I'm definitely open to other suggestions/ arguments for moving it there.

koku/masu/celery/tasks.py
myersCody: Ohhh. Gotcha nah this perfect, thanks for the breakdown. Just wanted to make sure I understood what was happening.

koku/masu/celery/tasks.py
myersCody: Nah, I wasn't sure if it would be needed or not just thought I would through it out there and get your opinion on if it would be worth it on not. The whole volume mapping explained it pretty well. Thanks ash!

koku/masu/celery/tasks.py
abaiken: No worries! It's always hard to follow the code when you haven't been heads down on the issue! Thanks for your comments/questions 😄

koku/koku/celery.py
```diff
@@ -112,6 +111,20 @@ def task(self, *args, **kwargs):
...
 app.conf.beat_schedule["db_metrics"] = {"task": "koku.metrics.collect_metrics", "schedule": crontab(minute="*/15")}
 
+
+# optionally specify the weekday and time you would like the clean volume task to run
+CLEAN_VOLUME_DAY_OF_WEEK = ENVIRONMENT.get_value("CLEAN_VOLUME_DAY_OF_WEEK", default="sunday")
+CLEAN_VOLUME_UTC_TIME = ENVIRONMENT.get_value("CLEAN_VOLUME_UTC_TIME", default="00:00")
+CLEAN_HOUR, CLEAN_MINUTE = CLEAN_VOLUME_UTC_TIME.split(":")
+if CLEAN_VOLUME_DAY_OF_WEEK:
+    clean_schedule = crontab(day_of_week=CLEAN_VOLUME_DAY_OF_WEEK, hour=int(CLEAN_HOUR), minute=int(CLEAN_MINUTE))
+else:
```
adberglund: You could probably just get rid of this else since `CLEAN_VOLUME_DAY_OF_WEEK` gets a default value assigned above. That should hopefully fix the codecov patch coverage problem too.

koku/masu/database/report_manifest_db_accessor.py
```diff
@@ -111,3 +117,30 @@ def get_manifest_list_for_provider_and_bill_date(self, provider_uuid, bill_date)
...
+            .annotate(
+                row_number=Window(
+                    expression=RowNumber(),
+                    partition_by=F("provider_id"),
+                    order_by=F("manifest_creation_datetime").desc(),
+                )
+            )
+            .order_by("row_number")
+        )
+        for manifest in manifests:
```
adberglund: There's one more thing we need to check here. I fiddled with the ORM query to try to get this done in SQL, but it's annoyingly difficult to figure out, so in this loop you want to only get the manifests where `row_number == 1`. When row number is `1` that means it is the most recent manifest.

abaiken: AHHH! That actually makes a ton of sense. Thank you for catching that!

PR #1778, opened 2020-02-18
OCP Query Performance

Red-HAP:
## Description

This PR contains changes to django models, supporting code changes and some direct DB changes via SQL in a migration.

The crux of these changes falls into two categories:

**Structure Changes**

1. Change to usage_start, usage_end fields in the OCP reporting tables (excluding the actual line item tables)
2. Addition of varchar_pattern_ops to indexes on namespace, node fields that are not arrays
3. Addition of the `pg_trgm` PostgreSQL extension
4. GIN indexes on namespace, node columns (non-array) to allow for `like` operations to utilize indexes if the search term contains 3 or mar characters. (like, ilike will normally _not_ use indexing)

**Code Changes**

* Code utilizing `usage_start`, `usage_end` has been changed to check for datetime and convert to date
* Code hit as part of the changes as found by the tox tests changed to handle datetime and date and use date as necessary

## Notes

This is passing the unit tests, but should be given a thorough review. I'm hoping that, if this does necessitate a regression, that it's targeted.

**This makes model changes.**

If this PR succeeds, then queries to tables filtering on usage_start or usage_end, the values should be date vs datetime.

## Goodies

* Fixed the `pg_data` volume mount so it will persist properly when the db container is stopped or removed.
* Added a grafana container to docker-compose.yml and a Dockerfile to initialize grafana.
* Added a makefile target `backup-local-db-dir` to make a quick-n-dirty backup of the `pg_data` diretory.
* Added a makefile target `restore-local-db-dir` to restore from the created backup
* Added a makefile target `docker-up-db-monitor` to build and up grafana

**Grafana Usage:**

1. `docker-compose up -build -d grafana`
2. point your browser to `localhost:3001`
3. username/password : `admin`/`admin12`

You should see a dashboard with 5 panels. You can see queries as point-in-time reports

This should be beneficial as the research into query performance is carried out.

The monitor does its best to filter out its own queries from the lower panels. The lower panels will only be beneficial when tracking long-running queries as that data becomes ephemeral as connections are closed.


docker-compose.yml
```diff
@@ -124,6 +124,18 @@ services:
     volumes:
       - ./pgadmin_servers.json:/pgadmin4/servers.json
 
+  grafana:
+    container_name: koku_grafana
+    build:
+        context: grafana
+        dockerfile: Dockerfile-grafana
+    ports:
+        - 3000:3000
```
maskarb: Could we change port number?

The sources-api uses port 3000, so I can envision some issues if we had sources-api set up locally.

Red-HAP: Ah.

abaiken:
@Red-HAP This looks great to me! I think Grafana is going to be a really beneficial. Query performance definitely isn't my expertise so someone else should look this over to give a 👍. My only suggestion is that it might be useful to add a make command for bringing up grafana (`docker-compose up -build -d grafana`) and add it to the make help as well.

Pipfile.lock
```diff
@@ -77,11 +77,11 @@
         },
         "azure-storage-blob": {
             "hashes": [
-                "sha256:b628e2f8a8470a52895e96c43c3321542be3246c61ff9ed4f3856fe87a04a406",
-                "sha256:f6f39a2c297c0ed5db2fbf480ac73a00c1ace94a364d203d2e00081371c27ada"
+                "sha256:200c7045f765c2b79c3b13715e0058ed956286e79270cf6c5b6285083dbc55b9",
```
adberglund: I see individual hash changes, but the overall lock file hash did not update. Did you run `make requirements`?

koku/reporting/migrations/0099_ocp_performance.py
```diff
@@ -0,0 +1,416 @@
...
+            reporting_ocpazurecostlineitem_project_daily_summary.unit_of_measure AS unit,
+            reporting_ocpazurecostlineitem_project_daily_summary.pretax_cost AS unblended_cost,
+            reporting_ocpazurecostlineitem_project_daily_summary.project_markup_cost,
+            reporting_ocpazurecostlineitem_project_daily_summary.pod_cost,
+            reporting_ocpazurecostlineitem_project_daily_summary.currency AS currency_code
+           FROM reporting_ocpazurecostlineitem_project_daily_summary
+          WHERE reporting_ocpazurecostlineitem_project_daily_summary.usage_start >= date_trunc('month'::text, date_trunc('month'::text, now()) - '1 day'::interval day)) lids
+          with no data;
+
+create index mv_reporting_ocpallcostlineitem_prj_daily_summary_namespace_ix
```
adberglund: This is an extremely silly nitpick, and since this is a migration I really don't care here, but in the future for SQL that we're going to keep in the code base and read, can we CAPS keywords? The mix of CAPS and lower here is confusing to read. Please don't put any effort into changing this PR, just in the future since the existing SQL in the repo uses CAPS, let's do that.

Pipfile.lock
Red-HAP: yes

koku/reporting/migrations/0099_ocp_performance.py
Red-HAP: OK. It's just a habit since PG will lower everything anyway.

Issue #1623, opened 2020-01-15
Clean up manifest and report tables when expiring data


adberglund:
## Summary
We have two tables in the public schema that track the manifest and files for a report for a source: `reporting_common_costusagereportmanifest` and `reporting_common_costusagereportstatus`. Those tables grow indefinitely right now and we don't really need a record for a report or manifest that has been deleted from our system. We can remove manifests and file records from a billing period when we remove the data for that billing period.

## Assumptions
- This can be done in the `ExpiredDataRemover` 

## Additional Information
- The Manifest model: https://github.com/project-koku/koku/blob/46b5c98364cf27b19e0530cc9a50cc79753b6900/koku/reporting_common/models.py#L24
- The report status model: https://github.com/project-koku/koku/blob/46b5c98364cf27b19e0530cc9a50cc79753b6900/koku/reporting_common/models.py#L44
- Expired Data Remover: https://github.com/project-koku/koku/blob/46b5c98364cf27b19e0530cc9a50cc79753b6900/koku/masu/processor/expired_data_remover.py#L45
- Just deleting the `CostUsageReportManifest` should cascade deletes to the `CostUsageReportStatus` table.

## Acceptance Criteria

- [ ] Report file records are deleted when you delete data for a billing period. 
- [ ] Manifest records are deleted when you delete data for a billing period.

nbon12:
Andrew: "One thing we're not doing is cleaning up the manifest and report status tables."

nbon12:
https://github.com/project-koku/koku/blob/46b5c98364cf27b19e0530cc9a50cc79753b6900/koku/masu/processor/aws/aws_report_db_cleaner.py#L46

lcouzens:
Verified!!

PR #1788, opened 2020-02-19
Feature/#1623 - Delete expired CostUsageReportManifests

nbon12:
Addresses issue #1623 

Cleans up CostUsageReportManifest objects by deleting all CostUsageReportManifest objects of a provider type with a billing_period_start_datetime older than the ExpiredDataRemover's calculated expiration_date. Records that occur on or after the expiration_date are kept.

This deletion happens in the ExpiredDataRemover's remove() method, after other data is purged.

- The deletion respects the `simulate` flag of the ExpiredDataRemover.remove() method

Additional Features:

- [x]  when calling ExpiredDataRemover.remove(provider_uuid=X)  the method should remove all of the CostUsageReportManifests associated with the provider_uuid
- [x] when calling ExpiredDataRemover.remove() without a provider_uuid, the method should remove all of the CostUsageReportManifests associated with the provider type of the ExpiredDataRemover.
- [x] When calling ExpiredDataRemover.remove(line_items_only=true) then CostUsageReportManifests should not be removed.
- [x]  Calling remove(simulate=True) should log a generic message about the query parameters that were used.
- [x] add testing instructions - for inserting test data, querying, triggering the delete, and querying after the delete

## Testing Instructions:
* Set INITIAL_INGEST_NUM_MONTHS=8
- Sometimes 8 months don't actually get injested, to make sure 8 months actually get ingested, you can manually change INITIAL_INGEST_NUM_MONTHS in the config.py file like so:
For now you can manually temporarily change that var here:
https://github.com/project-koku/koku/blob/master/koku/masu/config.py#L117
```
INITIAL_INGEST_NUM_MONTHS = int(os.getenv("INITIAL_INGEST_NUM_MONTHS", "8"))
```
* Run `make load-test-customer-data start=2019-11-01` replace the start-date with something old. the start date should be more than 2 months ago, to simulate old records that should be expired.
* Observe in the database old records in the public.reporting_common_costusagereportmanifest table. In this screenshot, the current date was March 2020, and manifests exist with the billing_period_start_datetime from October 2019. These associate with a Test AWS-local provider.
Query:
```
SELECT * FROM public.reporting_common_costusagereportmanifest ORDER BY id ASC LIMIT 100;
```
copy the provider_id of one of the older records (one of the records where billing_period_start_datetime is older than 2 months ago)

Screen Shot 2020-03-06 at 9 45 08 AM

Note the provider_uuid type in the provider_table, in this case, it is a Test AWS Provider. Copy this provider_uuid:

```
SELECT * FROM public.api_provider;
```

Screen Shot 2020-03-06 at 10 03 57 AM
* Trigger the delete by calling this endpoint: `GET` `http://127.0.0.1:5000/api/cost-management/v1/expired_data/?line_items_only=False` The response should contain async jobs for expired data removal.

* Wait for celery to run through all of it's tasks.

* In the koku_worker logs, we expect to see logs about expired_data removal:

```
koku_worker       | [2020-03-06 16:38:01,402: INFO/ForkPoolWorker-1] [masu.processor.tasks.remove_expired_data(6ccf30cf-1bcd-4820-b1a5-c7f5292e7091 via 7a6a7cdf-61ba-4eca-9a7b-134dd2eab32d)] Removed Data:
koku_worker       |  [{'usage_period_id': 25, 'interval_start': '2019-10-01 00:00:00+00:00'}, {'usage_period_id': 27, 'interval_start': '2019-11-01 00:00:00+00:00'}, {'usage_period_id': 29, 'interval_start': '2019-12-01 00:00:00+00:00'}, {'usage_period_id': 31, 'interval_start': '2020-01-01 00:00:00+00:00'}, {'usage_period_id': 33, 'interval_start': '2020-02-01 00:00:00+00:00'}]

```

* Query the public.reporting_common_costusagereportmanifest table, and observe that all older records have been purged:
```
SELECT * FROM public.reporting_common_costusagereportmanifest WHERE provider_id = <insert provider_uuid that you copied from before here>
```

Screen Shot 2020-03-06 at 11 40 46 AM (2)

---


koku/masu/processor/expired_data_remover.py
```diff
@@ -151,4 +154,7 @@ def remove(self, simulate=False, provider_uuid=None, line_items_only=False):
                 removed_data = self._cleaner.purge_expired_line_item(expired_date=expiration_date, simulate=simulate)
             else:
                 removed_data = self._cleaner.purge_expired_report_data(expired_date=expiration_date, simulate=simulate)
+            # Separately remove expired CostUsageReportManifests
+            CostUsageReportManifest.objects.filter(billing_period_start_datetime__lt=expiration_date).delete()
```
myersCody: So, what you have here probably works right now because I have not finished part 2 of my issue to add a trigger for the line items only data removal here: https://github.com/project-koku/koku/issues/1616

However, when I do this `.remove()` method will be called for both when we delete a report, but also for when we want to clean out the line items raw data in the database. Which means now with this way this is organized whenever we delete the raw line item data we will also be deleting the `CostUsageReportManifest` which for sure doesn't sound like what we want. I think we may want to move this at the very least under: https://github.com/project-koku/koku/pull/1788/files#diff-ec0942de17133f6841f909b02e083437R156

**Side Question**: 

So, the way that this method is dynamically created the `self._cleaner.purge_expired_report_data` will actually be called based off the provider type. So, this line will actually trigger one of the following:
- https://github.com/project-koku/koku/blob/master/koku/masu/processor/aws/aws_report_db_cleaner.py#L98
- https://github.com/project-koku/koku/blob/master/koku/masu/processor/ocp/ocp_report_db_cleaner.py#L94
- https://github.com/project-koku/koku/blob/master/koku/masu/processor/azure/azure_report_db_cleaner.py#L97

So, I am not as familiar with the `CostUsageReportManifest`, but I think this is deleting all manifest before the start date regardless of provider type, every time a report is deleted. Is that what we want?  If not we may want to set the provider, flag here too:
```
CostUsageReportManifest.objects.filter(provider=model, billing_period_start_datetime__lt=expiration_date)
```
If we are gonna group by provider, it may be better to move this into each `purge_expired_report_data` method I linked above since it is already being called base on provider report deleting. That would allow you to lean on some of the logic already in those methods too.

myersCody: whooops, sorry @nbon12, I started reviewing before I saw you move it back to WIP. Hopefully my comments are helpful though.

nbon12: @myersCody how about we separate the deletion of the expired CostUsageReportManifests into a separate task?

nbon12: @myersCody regarding the side question, the answer is yes; we want to delete all CostUsageReportrManifests regardless of the provider type, according to @adberglund

adberglund: @nbon12 @myersCody So looking at the way we call this, we only ever call the expired data remover for 1 provider at a time so when we do remove the manifest objects we'll want to do what Cody suggested and filter on the provider.

myersCody: Hmm. I think at the very least we want to be careful to not delete the manifests when line items only raw data is being deleted. The thought being just cause we removed the raw data cause it was summarized, doesn't mean that we have deleted the report itself from the system. We should add this as a separate method inside of the `ExpiredDataRemover`, and just make sure to only call it when line_items_only is false inside of the .remove() method. Additionally, I think that since ExpiredDataRemover gets called for a single provider at a time, we should just go ahead and only delete a single provider at a time. I reached out to Andrew and he agrees that it makes sense to make the removal of the manifest provider based since it will just live in this class anyway. Also, we can do a quick bluejeans to talk about this too if you would like. I feel like I have some background in the line items only stuff since I picked up that issue that would probably be helpful to know here.

koku/masu/processor/expired_data_remover.py
```diff
@@ -154,4 +157,10 @@ def remove(self, simulate=False, provider_uuid=None, line_items_only=False):
                 removed_data = self._cleaner.purge_expired_line_item(expired_date=expiration_date, simulate=simulate)
             else:
                 removed_data = self._cleaner.purge_expired_report_data(expired_date=expiration_date, simulate=simulate)
+                if not line_items_only:
+                    expiration_date = self._calculate_expiration_date()
+                    CostUsageReportManifest.objects.filter(
+                        provider__type=self._provider, billing_period_start_datetime__lt=expiration_date
```
myersCody: Do we want to do an equal to the expiration date here or a before? The equal would only give billing periods that start on on the expiration date, where as the `<` would give you all billing periods with a start date that come before the expiration date.

koku/masu/test/processor/test_expired_data_remover.py
```diff
@@ -136,3 +139,86 @@ def test_remove_items_only(self, mock_purge):
...
+            """
+            remover = ExpiredDataRemover(self.schema, provider_type)
+            expiration_date = remover._calculate_expiration_date()
+            this_month = datetime.today().replace(day=1)
+            day_before_cutoff = expiration_date - relativedelta.relativedelta(days=1)
+
+            # Record A
+            manifest_creation_datetime = this_month
+            manifest_updated_datetime = manifest_creation_datetime + relativedelta.relativedelta(days=2)
+            record_a_uuid = uuid4()
```
myersCody: These could be done in a for loop.

koku/masu/processor/expired_data_remover.py
nbon12: I believe the __lt suffix does the same thing as `<` @myersCody https://docs.djangoproject.com/en/3.0/topics/db/queries/#filters-can-reference-fields-on-the-model

koku/masu/test/processor/test_expired_data_remover.py
```diff
@@ -136,3 +139,86 @@ def test_remove_items_only(self, mock_purge):
...
+                "num_processed_files": 1,
+                "num_total_files": 1,
+                "provider_id": provider_uuid,
+            }
+            manifest_entry_3 = CostUsageReportManifest(**day_of_expiration_data)
+            manifest_entry_3.save()
+
+            remover.remove()
+            # Check if record A and C still exist. B should be deleted.
+            record_a = CostUsageReportManifest.objects.filter(assembly_id=record_a_uuid)
```
myersCody: I think you can do a `.count()` to return the number of rows that match this inside of doing the len(repord_a)

koku/masu/processor/expired_data_remover.py
myersCody: Ohhh gotcha 👍 That is my bad then, completely missed that.

koku/masu/test/processor/test_expired_data_remover.py
```diff
@@ -136,3 +139,86 @@ def test_remove_items_only(self, mock_purge):
...
+
+        def insert_and_assert_delete_cost_usage_manifests(provider_type, provider_uuid):
+            """
+            Insert CostUsageReportManifests for a specific provider.
+
+            Assert that the expired records for that provider are deleted.
+            """
+            remover = ExpiredDataRemover(self.schema, provider_type)
+            expiration_date = remover._calculate_expiration_date()
+            this_month = datetime.today().replace(day=1)
```
myersCody: Can we rename this to `current_month`?

koku/masu/test/processor/test_expired_data_remover.py
nbon12: @myersCody I changed it to be in a loop now.

koku/masu/test/processor/test_expired_data_remover.py
nbon12: @myersCody added.

koku/masu/test/processor/test_expired_data_remover.py
nbon12: @myersCody changed

koku/masu/processor/expired_data_remover.py
```diff
@@ -154,4 +157,10 @@ def remove(self, simulate=False, provider_uuid=None, line_items_only=False):
                 removed_data = self._cleaner.purge_expired_line_item(expired_date=expiration_date, simulate=simulate)
             else:
                 removed_data = self._cleaner.purge_expired_report_data(expired_date=expiration_date, simulate=simulate)
+                if not simulate:
+                    CostUsageReportManifest.objects.filter(
```
adberglund: We talked about using the ReportManifestDBAccessor for this. Also didn't we say we wanted to support simulate and have it tell us how many records would have been deleted if it had been real?

nbon12: @adberglund added

koku/masu/database/report_manifest_db_accessor.py
```diff
@@ -146,3 +146,31 @@ def get_last_seen_manifest_ids(self, bill_date):
...
+        Args:
+            provider_type   (String) the provider type to delete
+            billing_period_start_datetime (datetime.datetime) delete all manifests older than this date, exclusive.
+        """
+        recordsToDelete = CostUsageReportManifest.objects.filter(
+            provider__type=provider_type, billing_period_start_datetime__lt=billing_period_start_datetime
+        ).count()
+        CostUsageReportManifest.objects.filter(
+            provider__type=provider_type, billing_period_start_datetime__lt=billing_period_start_datetime
+        ).delete()
```
myersCody: When you call `.delete()` it will return the number of items it deletes. I would take a look at what they do here:  https://github.com/project-koku/koku/blob/master/koku/masu/processor/azure/azure_report_db_cleaner.py#L79

The way simulate was implemented in the past is that they would have a general log that would be called regardless of the deletion occurred that would highlight what vars were used in the query. So like:
```
LOG.info(
                        "Cost usage report manifest removed for provider type: %s with billing period: %s",
                        provider_type,
                        billing_period_start_datetime,
                    )
```
And then the count was only used in the log if the delete actually occurred. 
```
if not simulate:
     delete_count = CostUsageReportManifest.objects.filter(
            provider__type=provider_type, billing_period_start_datetime__lt=billing_period_start_datetime
        ).delete()
        LOG.info("Removing %s cost usage items for billing period start %s", del_count, bill_info)
```

koku/masu/processor/expired_data_remover.py
nbon12: @adberglund, given @myersCody recent comment below, do you agree that simulate should only show a generic message not not show how many records were deleted?

adberglund: Yes, that sounds just fine.

koku/masu/database/report_manifest_db_accessor.py
nbon12: @myersCody added

koku/masu/processor/expired_data_remover.py
nbon12: @adberglund added

koku/masu/processor/expired_data_remover.py
nbon12: @myersCody added

koku/masu/processor/expired_data_remover.py
```diff
@@ -154,4 +157,15 @@ def remove(self, simulate=False, provider_uuid=None, line_items_only=False):
...
+                with ReportManifestDBAccessor() as manifest_accessor:
+                    if not simulate:
+                        manifest_accessor.delete_cost_usage_reports_older_than(self._provider, expiration_date)
+                    else:
+                        LOG.info(
+                            """Simulated removing CostUsageReportManifest for
+                            provider type: %s before billing period: %s""",
+                            self._provider,
+                            expiration_date,
+                        )
```
myersCody:
```suggestion
                    if not simulate:
                        manifest_accessor.delete_cost_usage_reports_older_than(self._provider, expiration_date)
                    LOG.info(
                        """Removing CostUsageReportManifest for
                        provider type: %s before billing period: %s""",
                        self._provider,
                        expiration_date)
```

myersCody: So, this is my first time using a multiple line suggestion. I would still make sure the spacing is correct cause python really won't be happy if its off 😂 But essentially we should log the query info regardless if its simulate or not.

nbon12: @myersCody added

koku/masu/database/report_manifest_db_accessor.py
nbon12: @myersCody changed

koku/masu/processor/expired_data_remover.py
nbon12: @myersCody this multi-line suggestion is very nice.

myersCody:
I know I have brought this up in the past, but I strongly feel like we also need to remove the manifest in this else condition too: https://github.com/project-koku/koku/blob/556993b5d74352de3b14ea46d65170d4aca4ec70/koku/masu/processor/expired_data_remover.py#L153

The reason I want to bring this up again is because I discovered when I was adding my trigger in task.py for delete line items only that the `_removed_expired_data` is pretty much ony called with the provider_uuid. Its what let me know that I wasn't correctly capping the date for when a provider_uuid was provided: https://github.com/project-koku/koku/blob/c8464540462411fcccf90aae09bb76de16d06cab/koku/masu/processor/tasks.py#L170

nbon12:
> I know I have brought this up in the past, but I strongly feel like we also need to remove the manifest in this else condition too:
> 
> https://github.com/project-koku/koku/blob/556993b5d74352de3b14ea46d65170d4aca4ec70/koku/masu/processor/expired_data_remover.py#L153
> 
> The reason I want to bring this up again is because I discovered when I was adding my trigger in task.py for delete line items only that the `_removed_expired_data` is pretty much ony called with the provider_uuid. Its what let me know that I wasn't correctly capping the date for when a provider_uuid was provided:
> 
> https://github.com/project-koku/koku/blob/c8464540462411fcccf90aae09bb76de16d06cab/koku/masu/processor/tasks.py#L170

@myersCody  changed

koku/masu/database/report_manifest_db_accessor.py
```diff
@@ -146,3 +146,32 @@ def get_last_seen_manifest_ids(self, bill_date):
             if not processed:
                 assembly_ids.append(manifest.assembly_id)
         return assembly_ids
+
+    def delete_cost_usage_reports_older_than(self, provider_type, billing_period_start_datetime):
```
myersCody: So, as an optional suggestion you really could move these two methods into one:

```
def delete_expired_report_manifests(self, expiration_date, provider_uuid=None, provider_type=None):
    """ Doco Stringo"""
    if (provider_type is None) and (provider_uuid is None):
        Log.warning("some message saying you have to pass in either the uuid or type")
        raise Error
    elif provider_type and provider_uuid:
        Log.warning('some message saying you can NOT pass in both.')
        raise Error
    base_msg = "Removed %s CostUsageReportManifest(s) for provider %s (%s) that had a billing period start date before %s"
    if provider_type:
        delete_count = CostUsageReportManifest.objects.filter(
            provider__type=provider_type, billing_period_start_datetime__lt=billing_period_start_datetime
        ).delete()
        msg = base_msg % (delete_count, 'type', provider_type, expiration_date)
    else:
        delete_count = CostUsageReportManifest.objects.filter(
            provider_id=provider_uuid, billing_period_start_datetime__lt=billing_period_start_datetime
        ).delete()
        msg = base_msg % (delete_count, 'uuid', provider_uuid, expiration_date)
    LOG.info(msg)
```

koku/masu/database/report_manifest_db_accessor.py
```diff
@@ -146,3 +146,32 @@ def get_last_seen_manifest_ids(self, bill_date):
...
+        ).delete()
+        LOG.info(
+            "Removed %s CostUsageReportManifest(s) for provider type %s that had a billing period start date before %s",
+            delete_count,
+            provider_type,
+            billing_period_start_datetime,
+        )
+
+    def delete_provider_cost_usage_reports_older_than(self, provider_uuid, billing_period_start_datetime):
+        delete_count = CostUsageReportManifest.objects.filter(
```
myersCody: if you don't do the optional suggestion above you will need a doc string here.

koku/masu/processor/expired_data_remover.py
```diff
@@ -154,6 +157,10 @@ def remove(self, simulate=False, provider_uuid=None, line_items_only=False):
                     )
             else:
                 removed_data = self._cleaner.purge_expired_report_data(simulate=simulate, provider_uuid=provider_uuid)
+                with ReportManifestDBAccessor() as manifest_accessor:
+                    # Remove expired CostUsageReportManifests
+                    expiration_date = self._calculate_expiration_date()
+                    manifest_accessor.delete_provider_cost_usage_reports_older_than(provider_uuid, expiration_date)
```
myersCody: we still need the `if not simulate:` logic here.

koku/masu/database/report_manifest_db_accessor.py
nbon12: added

koku/masu/processor/expired_data_remover.py
nbon12: added

koku/masu/processor/expired_data_remover.py
```diff
@@ -164,5 +178,17 @@ def remove(self, simulate=False, provider_uuid=None, line_items_only=False):
                         expired_date=expiration_date, simulate=simulate
                     )
             else:
+                # Remove expired CostUsageReportManifests
+                expiration_date = self._calculate_expiration_date()
```
myersCody: The expiration date is already being calculated here: https://github.com/project-koku/koku/blob/466c5c31e6b25f23622839c78fec148ece713b5a/koku/masu/processor/expired_data_remover.py#L172

So you probably don't need to recalculate it here.

nbon12: added

myersCody:
@nbon12 You wouldn't happen to have the queries you used in psql to confirm that the data is being deleted correctly? If so would you mind adding them to description so that I can run them on my end and check that the outcome is coming out as expected. Also, @lcouzens will most likely need them as well for when he goes to confirm this issue when it gets moved to the QA state.

nbon12:
@myersCody I'm currently working on writing a test query, inserting test data with nise, and querying that it was deleted.

nbon12:
awaiting PR review

kholdaway:
@nbon12 Would you mind providing a text version of the SQL queries you use instead of screenshots? The testing instructions will be difficult without that

nbon12:
@kholdaway SQL text version is has been added

myersCody:
Hi @nbon12, I am having some trouble with your testing scenario, I may be doing something wrong. However, I am running a basic server setup with: `make docker-reinitdb-with-sources; sleep 2; make load-test-customer-data start=2019-12-01`

However, when I log into the database and try to query for the initial data point, I already can't see any manifest that have a start date older than two months before I even trigger your code:
```
postgres=# SELECT * FROM public.reporting_common_costusagereportmanifest;
 id |             assembly_id              |  manifest_creation_datetime   |   manifest_updated_datetime   | billing_period_start_datetime | num_processed_files | num_total_files |             provider_id              |                 task                 |  manifest_completed_datetime
----+--------------------------------------+-------------------------------+-------------------------------+-------------------------------+---------------------+-----------------+--------------------------------------+--------------------------------------+-------------------------------
  1 | 4111499e-d542-4d64-97dc-b2aa1c154150 | 2020-03-17 12:21:44.049823+00 | 2020-03-17 12:21:58.23023+00  | 2020-02-01 00:00:00+00        |                   1 |               1 | 91424913-7fba-4662-bef0-535669646812 | b6a473c8-c651-4ec7-ab7d-033beee5267a | 2020-03-17 12:29:01.029539+00
  2 | 7b604f5a-797c-43a9-9904-46d4f4657da9 | 2020-03-17 12:21:58.381497+00 | 2020-03-17 12:22:04.416566+00 | 2020-03-01 00:00:00+00        |                   1 |               1 | 91424913-7fba-4662-bef0-535669646812 | 2a21ce9c-6002-4ded-b923-8752059c68fc | 2020-03-17 12:29:02.764201+00
  3 | 31979e8e-5a60-4b20-8eef-6f2692f497be | 2020-03-17 12:22:04.586958+00 | 2020-03-17 12:22:05.980012+00 | 2020-02-01 00:00:00+00        |                   1 |               1 | 23c97d97-da79-4181-9d34-d78e4eff7e5c | 3b3c9bcc-a7df-4236-a3cc-6d3b61892d93 | 2020-03-17 12:29:02.819458+00
  4 | 3285d75c-e582-4eb4-8e3f-f1b2051bac3a | 2020-03-17 12:22:06.065549+00 | 2020-03-17 12:22:07.280671+00 | 2020-03-01 00:00:00+00        |                   1 |               1 | 23c97d97-da79-4181-9d34-d78e4eff7e5c | 4857b7e1-c16c-4a83-90e7-5a8284131052 | 2020-03-17 12:29:02.895975+00
  5 | 9f69e92a-9534-45f4-a079-b8fc43c15a7b | 2020-03-17 12:22:07.377033+00 | 2020-03-17 12:22:18.480249+00 | 2020-02-01 00:00:00+00        |                   2 |               2 | f7b3e10e-11c1-4aa1-93d1-d1f34e1377be | 46ebbb2f-15d4-4558-9ff1-eda2d13ba384 | 2020-03-17 12:29:02.927036+00
  6 | ebe39c27-8440-49f0-92d6-bcbacf8f8562 | 2020-03-17 12:22:18.583883+00 | 2020-03-17 12:22:25.437842+00 | 2020-03-01 00:00:00+00        |                   2 |               2 | f7b3e10e-11c1-4aa1-93d1-d1f34e1377be | 4366cfc4-2820-446e-b8dd-2a467866a7b6 | 2020-03-17 12:29:02.989+00
  7 | 1cf25652-0821-47d6-bad1-743259669b23 | 2020-03-17 12:22:25.518665+00 | 2020-03-17 12:22:35.503632+00 | 2020-02-01 00:00:00+00        |                   2 |               2 | 910e38ad-9aad-4d7c-8340-936bca2a9bff | 9cd85136-cbfd-4f0b-ad98-51ecbb7230d9 | 2020-03-17 12:29:03.015809+00
  8 | aeb73d51-b8e2-409f-9953-73a17d211851 | 2020-03-17 12:22:35.59435+00  | 2020-03-17 12:22:41.41701+00  | 2020-03-01 00:00:00+00        |                   2 |               2 | 910e38ad-9aad-4d7c-8340-936bca2a9bff | 7b81538f-1a42-4eb8-a82b-98f20ccb7345 | 2020-03-17 12:29:03.067335+00
  9 | 4c593540-da17-4453-8314-4e5e1fb30bc5 | 2020-03-17 12:22:41.512474+00 | 2020-03-17 12:23:43.537634+00 | 2020-02-01 00:00:00+00        |                   2 |               2 | 81c74fbd-7878-41f0-9ab1-e89a015763ed | d9d4075e-3e26-4ed1-960c-01cd3ba86bf6 | 2020-03-17 12:29:03.095828+00
 10 | 99c36498-fa5d-4bd3-8333-1c4298fa718e | 2020-03-17 12:23:43.657829+00 | 2020-03-17 12:24:04.39953+00  | 2020-03-01 00:00:00+00        |                   2 |               2 | 81c74fbd-7878-41f0-9ab1-e89a015763ed | da9dbfc6-3455-4ad7-8227-30274580cb81 | 2020-03-17 12:29:03.14167+00
(10 rows)

postgres=#
```
It almost feels like something is already triggering your manifest delete code when I run: `make docker-reinitdb-with-sources; sleep 2; make load-test-customer-data start=2019-12-01`

nbon12:
@myersCody I suspect that's because INITIAL_NUM_MONTHS=8 isn't getting into the container.
"We can worry about why the env is not being picked up later. For now you can manually temporarily change that var here:

https://github.com/project-koku/koku/blob/master/koku/masu/config.py#L117

    `INITIAL_INGEST_NUM_MONTHS = int(os.getenv("INITIAL_INGEST_NUM_MONTHS", "8"))`

myersCody:
@nbon12 I tested your code and it all looks good, now all you got to do is get your test passing, and you should be good to go.

koku/masu/database/report_manifest_db_accessor.py
```diff
@@ -146,3 +146,39 @@ def get_last_seen_manifest_ids(self, bill_date):
             if not processed:
                 assembly_ids.append(manifest.assembly_id)
         return assembly_ids
+
+    def delete_cost_usage_reports_older_than(self, provider_type, billing_period_start_datetime):
```
myersCody: So, I debated on leaving this comment for a while, but on second thought I don't think the name of your functions accurately portraying what your code is doing here. 

You are not actually deleting the cost usage report data here, you are deleting the cost usage report manifest instead. I am totally open to better names for these, but my suggestion is:

```
def purge_expired_report_manifest(self, provider_type, billing_period_start_datetime):
```
I think this will match the naming format of the function that is actually deleting the cost usage reports `self._cleaner.purge_expired_report_data`

koku/masu/database/report_manifest_db_accessor.py
```diff
@@ -146,3 +146,39 @@ def get_last_seen_manifest_ids(self, bill_date):
...
+            provider__type=provider_type, billing_period_start_datetime__lt=billing_period_start_datetime
+        ).delete()
+        LOG.info(
+            "Removed %s CostUsageReportManifest(s) for provider type %s that had a billing period start date before %s",
+            delete_count,
+            provider_type,
+            billing_period_start_datetime,
+        )
+
+    def delete_provider_cost_usage_reports_older_than(self, provider_uuid, billing_period_start_datetime):
```
myersCody: For this one I think: `purge_provider_expired_report_manifest`. 

Again I am totally open to suggestions or conversation around this. I just don't want the functions to be reused someplace else later on with people thinking it deletes the cost usage report.

koku/masu/database/report_manifest_db_accessor.py
nbon12: sounds good, I plan fix the unit tests first, then re-naming this.

koku/masu/database/report_manifest_db_accessor.py
```diff
@@ -146,3 +146,39 @@ def get_last_seen_manifest_ids(self, bill_date):
             if not processed:
                 assembly_ids.append(manifest.assembly_id)
         return assembly_ids
+
+    def delete_cost_usage_reports_older_than(self, provider_type, billing_period_start_datetime):
+        """
+        Deletes Cost usage Reports older than billing_period_start_datetime.
```
dccurtis:
```suggestion
        Deletes Cost usage Report Manifests older than billing_period_start_datetime.
```

koku/masu/database/report_manifest_db_accessor.py
```diff
@@ -146,3 +146,39 @@ def get_last_seen_manifest_ids(self, bill_date):
...
+
+        Args:
+            provider_type   (String) the provider type to delete associated manifests
+            billing_period_start_datetime (datetime.datetime) delete all manifests older than this date, exclusive.
+        """
+        delete_count = CostUsageReportManifest.objects.filter(
+            provider__type=provider_type, billing_period_start_datetime__lt=billing_period_start_datetime
```
dccurtis: Are we sure we want to delete all manifests the match the desired provider's type instead of the specific provider being removed?

dccurtis: on second look. It looks like this is called when a provider_uuid is not supplied in `ExpiredDataRemover.remove()`

koku/masu/database/report_manifest_db_accessor.py
dccurtis: I think anything that lets the caller know it's report manifests being deleted and that one function is for a specific provider uuid and another is for a provider type is fine.  You could even use the parameter names to communicate some of this.

koku/masu/database/report_manifest_db_accessor.py
```diff
@@ -1,5 +1,5 @@
 #
-# Copyright 2018 Red Hat, Inc.
+# Copyright 2020 Red Hat, Inc.
```
dccurtis:
```suggestion
# Copyright 2018 Red Hat, Inc.
```

dccurtis: We only change these for when the file was created.  @blentz can keep me honest here.

koku/masu/processor/expired_data_remover.py
```diff
@@ -1,5 +1,5 @@
 #
-# Copyright 2018 Red Hat, Inc.
+# Copyright 2020 Red Hat, Inc.
```
dccurtis:
```suggestion
# Copyright 2018 Red Hat, Inc.
```

koku/masu/test/processor/test_expired_data_remover.py
```diff
@@ -1,5 +1,5 @@
 #
-# Copyright 2018 Red Hat, Inc.
+# Copyright 2020 Red Hat, Inc.
```
dccurtis:
```suggestion
# Copyright 2018 Red Hat, Inc.
```

koku/masu/test/processor/test_expired_data_remover.py
```diff
@@ -137,6 +142,179 @@ def test_remove_items_only(self, mock_purge):
...
+            # Check if the log message exists in the log output:
+            self.assertTrue(
+                any(match is not None for match in [re.search(expected_log_message, line) for line in cm.output]),
+                "Expected to see log message: "
+                + expected_log_message
+                + "in the list of log messages"
+                + " but the list of log messages was instead : "
+                + str(cm.output),
+            )
+        # Re-enable log suppressionjjn
```
dccurtis:
```suggestion
        # Re-enable log suppression
```

koku/masu/database/report_manifest_db_accessor.py
nbon12: @myersCody @dccurtis I renamed it.

koku/masu/database/report_manifest_db_accessor.py
nbon12: @myersCody I renamed it.

abaiken: @nbon12 I like the name that Cody suggested better (`purge_expired_report_manifest`). I feel that it matches the pre-existing pattern and clearly explains what the function is doing. I also feel like the parameter `expired_date` was a good suggestion.

nbon12: @abaiken The pre-existing pattern among other db_accessors is that they only have `get` methods and CRUD operations, I don't see any business logic in the db_accessors, and it seems to be more of a middleware layer that only has knowledge of how to interact with the database. The DB_accessor layer is meant to replace the Django ORM syntax, not to contain any business logic. Isn't the db_accessor middleware layer not supposed to have a concept of the business logic of what's expired and what's not expired? All of the other db_accessors only have `get`, and `delete`, and `update`. Do you want the db_accessor to include the non-CRUD terminology 'expired'? @adberglund

nbon12: @abaiken I renamed it to purge_expired_report_manifest and expired_date

Issue #1795, opened 2020-02-20
OCP on All API access should not 403 is user has subset of cloud RBAC


adberglund:
**Describe the bug**
Currently access to the OpenShift on All APIs fails if a user does not have view access for AWS **AND** Azure. 

**To Reproduce**
1. Create a user and assign to an RBAC role with OpenShift view and AWS view
2. See error on OpenShift on Infra dashboard in UI

**Expected behavior**
If the user has one or the other we should allow access to the API with the appropriate subset of data. 

**Additional context**
Current access control: https://github.com/project-koku/koku/blob/6d31d2fde02608ef7660707e597af25d90017067/koku/api/query_params.py#L173

dchorvat1:
verified

PR #1802, opened 2020-02-21
1795 openshift on all permissions

maskarb:
#1795 

The general idea here:
1. Access to the ocp-on-all view is granted as long as a user has access to a single provider type. This access is determined using the new `openshift_all_access`.
2. When setting access for the ocp-on-all reporting view, the data is filtered on all accounts the user has access to. If, for example, a user has access to all AWS accounts and a subset of the Azure accounts, we look up all AWS account numbers in the DB and append to the azure account number filter list. 

Testing:
2. `make load-test-customer-data start=2020-01-01 end=2020-02-24`
3. Check these endpoints, see data:
    - reports/openshift/infrastructures/all/costs/
    - reports/openshift/infrastructures/aws/costs/
    - reports/openshift/infrastructures/azure/costs/
4. Apply the following diff (This user has access to AWS/OCP but not Azure):
```
diff --git a/koku/koku/middleware.py b/koku/koku/middleware.py
index 38656124..4f4653cb 100644
--- a/koku/koku/middleware.py
+++ b/koku/koku/middleware.py
@@ -216,8 +216,8 @@ class IdentityHeaderMiddleware(MiddlewareMixin):  # pylint: disable=R0903
     def _get_access(self, user):
         """Obtain access for given user from RBAC service."""
         access = None
-        if user.admin:
-            return access
+        # if user.admin:
+        #     return access
         access = self.rbac.get_access_for_user(user)
         return access
 
@@ -284,13 +284,14 @@ class IdentityHeaderMiddleware(MiddlewareMixin):  # pylint: disable=R0903
             user.req_id = req_id
 
             cache = caches["rbac"]
-            user_access = cache.get(user.uuid)
+            # user_access = cache.get(user.uuid)
+            user_access = None
             if not user_access:
                 try:
                     user_access = self._get_access(user)
                 except RbacConnectionError as err:
                     return HttpResponseFailedDependency({"source": "Rbac", "exception": err})
-                cache.set(user.uuid, user_access, self.rbac.cache_ttl)
+                # cache.set(user.uuid, user_access, self.rbac.cache_ttl)
             user.access = user_access
             request.user = user
 
diff --git a/koku/koku/rbac.py b/koku/koku/rbac.py
index 6a5480b4..d4339d7f 100644
--- a/koku/koku/rbac.py
+++ b/koku/koku/rbac.py
@@ -225,12 +225,15 @@ class RbacService:  # pylint: disable=too-few-public-methods
             self.protocol, self.host, self.port, self.path
         )
         headers = {"x-rh-identity": user.identity_header.get("encoded")}
-        acls = self._request_user_access(url, headers)
-        if isinstance(acls, list) and len(acls) == 0:  # pylint: disable=len-as-condition
-            return None
+        # acls = self._request_user_access(url, headers)
+        # if isinstance(acls, list) and len(acls) == 0:  # pylint: disable=len-as-condition
+        #     return None
 
-        processed_acls = _process_acls(acls)
-        return _apply_access(processed_acls)
+        # processed_acls = _process_acls(acls)
+        # return _apply_access(processed_acls)
+
+        processed_acls = {'aws.account': {'read': ['*']}, 'azure.subscription_guid': {'read': []}, 'openshift.cluster': {'read': ['*']}, 'openshift.node': {'read': ['*']}, 'openshift.project': {'read': ['*']}, 'provider': {'read': ['*'], 'write': ['*']}, 'rate': {'read': [], 'write': []}}
+        return processed_acls
 
     def get_cache_ttl(self):
```
5. Check reports/openshift/infrastructures/azure/costs endpoint and see 403 permission error
6. Check aws/cost and all/cost, and see matching values.

[ocp-on-all.txt](https://github.com/project-koku/koku/files/4247051/ocp-on-all.txt)


koku/api/common/permissions/__init__.py
```diff
@@ -1 +1,25 @@
-# noqa
+#
+# Copyright 2019 Red Hat, Inc.
```
maskarb:
```suggestion
# Copyright 2020 Red Hat, Inc.
```

chambridge:
@adberglund How are we capturing changes to RBAC `Cost Cloud Viewer` or split into multiple roles? Different issue?

@chargio Should we just update `Cost Cloud Viewer` to include Azure? Or do we want separate roles (I think we just go with the minor update and let customers create their own customer roles).

koku/api/common/permissions/openshift_all_access.py
```diff
@@ -0,0 +1,43 @@
...
+from rest_framework import permissions
+
+from api.common.permissions import RESOURCE_TYPES
+
+
+class OpenshiftAllAccessPermission(permissions.BasePermission):
+    """Determines if a user can view Openshift on All data."""
+
+    def has_permission(self, request, view):
+        """Check permission to view AWS data."""
```
maskarb:
```suggestion
        """Check permission to view OCP-on-ALL data."""
```

chargio:
Viewing all would be preferred, the customer can restrict their views if they need more details.

I would also add a viewer role, that has access to everything but in read-only mode, including cost models et al.

chambridge:
Does this limit the view correctly for a user that has RBAC access to just a limited set of AWS accounts or Azure subscription_guids?  I wasn't sure if that part was already taken care of as I know with the ALL api we are using AWS jargon, so subscription_guid has to be put into the list of filtered accounts.

maskarb:
@chambridge good question. 

I've made an adjustment such that the subscription-guid is converted to "account". 

However, if I specify access on a specific aws and azure account, we raise a [PermissionDenied here](https://github.com/project-koku/koku/blob/1795-OCPonAll-user-access/koku/api/query_params.py#L332).

On a different note, if I specify access to an aws or azure account and read access to everything else, I'm excluding everything but the specified account.

Will need to address both of these issues.

koku/api/common/permissions/test/tests_ocp_all_access.py
```diff
@@ -0,0 +1,50 @@
...
+
+class OCPAllAccessPermissionTest(TestCase):
+    """Test the OCP-on-All access permissions."""
+
+    def test_has_perm_with_access_on_get(self):
+        """Test that a user with at least 1 access can execute."""
+        accessPerm = OpenshiftAllAccessPermission()
+        s = ACCESS_KEYS.keys()
+        for key in chain.from_iterable(combinations(s, r) for r in range(1, len(s) + 1)):
+            print(key)
```
adberglund:
```suggestion
```

koku/api/query_params.py
```diff
@@ -146,35 +186,33 @@ def _set_access(self, filter_key, access_key, raise_exception=True):
...
-            self._set_access(name, f"openshift.{name}", raise_exception=exc)
-
-    def _set_access_ocp_aws(self):
+    def _set_access_ocp_all(self, provider, filter_key, access_key, raise_exception=True):
         """Alter query parameters based on user access."""
-        self._set_access_aws()
-        self._set_access_ocp()
+        access_list = self.access.get(access_key, {}).get("read", [])
+        access_filter_applied = False
+        if ReportQueryHandler.has_wildcard(access_list):
```
adberglund: We should accommodate the special case where both `AWS` and `Azure` have wildcards. When both (or all infrastructure/cloud providers in the future) have a wildcard, we can just not filter then.

adberglund: We should accommodate the special case where both `AWS` and `Azure` have wildcards. When both (or all infrastructure/cloud providers in the future) have a wildcard, we can just not filter then.

adberglund: I was so adamant about this I commented twice.

PR #1810, opened 2020-02-25
1507 error processing failed

maskarb:
This PR adds 424 workflow for 500 response from RBAC.

If RBAC returns a 500 response, increment the `RBAC_CONNECTION_ERROR_COUNTER` and raise an `RbacConnectionError` which is handled in middleware and returned as a 424 response.

### DIFF
--- a/koku/koku/rbac.py
+++ b/koku/koku/rbac.py
@@ -16,6 +16,7 @@
 #
 """Interactions with the rbac service."""
 import logging
+from json.decoder import JSONDecodeError
 
 import requests
 from prometheus_client import Counter
@@ -183,7 +184,7 @@ class RbacService:  # pylint: disable=too-few-public-methods
         }
         return rbac_conn_info
 
-    def _request_user_access(self, url, headers):
+    def _request_user_access(self, url, headers):  # noqa: C901
         """Send request to RBAC service and handle pagination case."""
         access = []
         try:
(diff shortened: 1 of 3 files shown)


koku/koku/rbac.py
```diff
@@ -193,11 +194,17 @@ def _request_user_access(self, url, headers):
...
+        if response.status_code >= status.HTTP_500_INTERNAL_SERVER_ERROR:
+            msg = ">=500 Response from RBAC"
+            LOGGER.warning(msg)
+            RBAC_CONNECTION_ERROR_COUNTER.inc()
+            raise RbacConnectionError(msg)
+
         if response.status_code != status.HTTP_200_OK:
             try:
                 error = response.json()
                 LOGGER.error("Error requesting user access: %s", error)
```
chambridge: I would think we would change this to a warning to avoid sentry triggers.
```suggestion
                LOGGER.warning("Error requesting user access: %s", error)
```

koku/koku/rbac.py
```diff
@@ -193,11 +194,17 @@ def _request_user_access(self, url, headers):
...
-            except ValueError as res_error:
+            except (JSONDecodeError, ValueError) as res_error:
                 LOGGER.error("Error processing failed, %s, user access: %s", response.status_code, res_error)
```
chambridge: I would think we would change this to a warning to avoid sentry triggers.
```suggestion
                LOGGER.warning("Error processing failed, %s, user access: %s", response.status_code, res_error)
```

PR #1812, opened 2020-02-25
Remove preferences API in favor of platform level preferences.

chambridge:
* Remove api, serializer, and model
* Add migration
* Update openapi.json spec

Want to only publish relevant APIs for long-term support.


docs/source/users.rst
```diff
@@ -1,7 +1,7 @@
 User Management
 ===============
 
-This section covers the customer owner, adding and removing users, email on-boarding and user preferences.
+This section covers the customer owner, adding and removing users.
 
 Definitions
```
maskarb: Should the `User Preferences` section below be removed? (Lines 22 - 32)

chambridge: yep. done.

Issue #1760, opened 2020-02-17
Optimize Pricelist Calculation Code


adberglund:
## User Story
As a developer, I want to optimize the price list calculation code to utilize the database as much as possible, so that calculations are as quick as possible and less calls are made against the database.

## Impacts
- Worker

## Assumptions
- We need the ability to get the memory or cpu or volume or node specific values for a particular cost type, but we probably don't want to explode out the number of columns in this table to handle each new cost type.
- Will explore using a JSON dictionary for costs that have sub values to split by resource contributing the cost. 
- We can leverage the Python code in existence to grab the rates for each resource type, but instead of calculating in Python then loading the data into the DB, we can template some SQL to do the math in the DB and populate the new columns. 

## Implementation Details
- OpenShift DailySummary model will need to be altered to have the following columns
  - infrastructure_raw_cost
  - infrastructure_project_raw_cost
  - infrastructure_usage_cost
  - infrastructure_markup_cost
  - infrastructure_monthly_cost
  - supplementary_raw_cost
  - supplementary_project_raw_cost
  - supplementary_usage_cost
  - supplementary_markup_cost
  - supplementary_monthly_cost
- OpenShift DailySummary model will need to be altered to remove the following columns
  - pod_charge_cpu_core_hours
  - pod_charge_memory_gigabyte_hours
  - persistentvolumeclaim_charge_gb_month
  - infra_cost
  - project_infra_cost
  - markup_cost
  - project_markup_cost
  - monthly_cost

## Acceptance Criteria
- [ ] Verify that existing test cases (unit and QE) still pass

nbon12:
Sprint Planning Meeting Notes:

@adberglund to populate this issue with more detail
Andrew: "Right now there's a lot of stuff done in code that might be able to move over."

lcouzens:
Verified

PR #1815, opened 2020-02-26
Rename charge -> cost model cost

adberglund:
## Summary
This is a small first step toward #1760. I didn't want to cloud the upcoming PRs with name changes, so creating a distinct PR for name changing here.

### DIFF
--- a/koku/masu/openapi.json
+++ b/koku/masu/openapi.json
@@ -23,8 +23,8 @@
         "description": "Operations about summarizing report data"
       },
       {
-        "name": "Update Charge",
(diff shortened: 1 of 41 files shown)


koku/masu/openapi.json
```diff
@@ -414,7 +414,7 @@
         "UpdateChargeGetResponse": {
```
abaiken:
```suggestion
        "UpdateCostModelCostGetResponse": {
```

abaiken: If we update the above line - we also have to update line 335 in the openapi.json as well.

PR #1837, opened 2020-03-02
Updated grafana config db for smaller query intervals

Red-HAP:
## Description

Alter the localdev grafana config db to include 1 second and 2 second query intervals. This affects only the local DB monitoring through grafana.

## Review/Test

There is no way to really review these changes via github as the change is in the binary config database.

To review these changes you will have to follow these steps:

1. `git checkout 1818-dev-grafana-interval`
2. `docker-compose down`
3. `docker-compose up --build -d grafana`
4. Open a new browser tab/window and go to `http://localhost:3001`
5. Login with user **admin** and password **admin12**
6. Validate that the refresh interval (upper right corner) shows **5s**
7. Click on the refresh interval select, you should see intervals for **1s** and **2s**.
8. If you want to see stuff change in the monitor, select either the **1s** or **2s** option to see the most change.
9. In your terminal, execute `make docker-reinitdb` and immediately return to the monitor.
10. Enjoy
11. Logout button is in the lower-left corner. Or just close the window and shut down the container.

## Validations

* efresh interval defaults to **5s** on login then 
* The **1s** and **2s** options should appear as refresh interval selections


grafana/Dockerfile-grafana
```diff
@@ -1,2 +1,7 @@
 FROM grafana/grafana:latest
-COPY --chown=grafana:grafana grafana.db /var/lib/grafana/grafana.db
+USER 0
```
blentz: Small readability nit.

```suggestion
USER root
```

grafana/Dockerfile-grafana
```diff
@@ -1,2 +1,7 @@
...
+RUN apk add sqlite
```
blentz: Just a general note: Please try to stick with using RHEL or CentOS-based containers.

Because this is more of a developer productivity thing, I won't block this PR on it. But, try to use RHEL/UBI or CentOS when you can. And, at some point in the future when we have some spare time, if you can update this image, we should do that. :slightly_smiling_face:

Red-HAP: Dang! You're right. I completely missed that. I shall have to try to find a RHEL container or build one.
If you're not going to block the PR (since this is a dev-only tool) then I'll put in another ticket for that work..

grafana/Dockerfile-grafana
Red-HAP: got it

PR #1843, opened 2020-03-03
Sources proxy for update

maskarb:
This PR adds back the PATCH method in the `SourcesProxyViewSet` which was removed in #1823. Without this proxy, a PATCH call during source creation would fail to send a signal to the koku-sources pod to complete the source creation.

Testing:
1. Post to sources
```
POST 3000/api/v1.0/sources

{"name":"mskarbek-aws","source_type_id":"1"} 
```

2.  Post to endpoints
```
POST 3000/api/v1.0/endpoints

{"role": "aws", "default": true, "source_id":"1"}
```

3. Post to  applications
```
POST 3000/api/v1.0/applications

{"application_type_id":"2","source_id":"1"}
```

4. Post to authentications
```
POST 3000/api/v1.0/authentications

{"authtype":"arn","password":"arn:aws:iam::111111111111:role/CostManagement","resource_id":"1","resource_type":"Endpoint"}
```

5.  Patch to cost-management sources
```
PATCH 8000/api/cost-management/v1/sources/1/

{"billing_source": {"bucket": "cost-usage-bucket"} }
```

See PATCH request in koku and koku-sources logs. See completed source creation in koku-sources logs.


adberglund:
@maskarb @chambridge Let's definitely get this back in now so we're working again, but we could also consider using something like kafka or our own rabbitmq instance to pass a message to koku-sources ultimately. Since this is ultimately message passing, no?

chambridge:
@maskarb and I chatted, I suggested that we could update the sources serializer update to directly perform the provider update instead of leveraging the post_save signal.

adberglund:
sounds :atom: ic

PR #1849, opened 2020-03-04
Sources: iteration

maskarb:
- Add `provider_linked` flag to the sources endpoints. Used to indicate that the provider was created for the source.


koku/sources/api/view.py
```diff
@@ -210,7 +212,12 @@ def stats(self, request, pk=None):
...
         schema_name = create_schema_name(account_id)
         source = self.get_object()
-        manager = ProviderManager(source.source_uuid)
-        tenant = Tenant.objects.get(schema_name=schema_name)
-        stats = manager.provider_statistics(tenant)
+        stats = None
+        try:
+            manager = ProviderManager(source.source_uuid)
+        except ProviderManagerError:
+            pass
```
chambridge: what is the response here? Should we error in some way or return an empty payload
```suggestion
            stats = {}
```

maskarb: Since I set `stats = None` above, it just returns an empty payload. I think returning nothing is good enough. Or I could just add the `provider_linked` flag?

chambridge: sometimes empty payload ` ` is not super friendly if json is the expected response.

maskarb: I made `stats = {}` and added the `provider_linked` flag here as well.

chambridge: 👍 works for me

PR #1828, opened 2020-02-28
1760 openshift cost model cost calculation updates

adberglund:
## Summary 
This updates or cost model cost calculation code to honor the new supplementary and infrastructure cost types allowed in cost models. It also calculates cost based on usage costs, monthly costs (a special case form of usage costs), and markup. The OCP provider map was also updated to use the new database fields. 

## Testing

- A cost model was applied to an "On Prem" OpenShift cluster with cpu, memory, and volume costs assigned as supplementary. The node cost per month was set to infrastructure. 
- A cost model purely with a 10% markup was associated with the OCP on AWS cluster, which is calculated only from the infrastructure raw costs. 
- The OCP on Azure cluster has no cost model, and just the raw costs from the Azure report

```
GET /api/cost-management/v1/reports/openshift/costs/?filter[time_scope_value]=-1&group_by[cluster]=*
```
```
HTTP 200 OK
Allow: GET, HEAD, OPTIONS
Content-Type: application/json
Vary: HTTP_X_RH_IDENTITY, Accept

{
    "meta": {
        "count": 1,
        "filter": {
            "time_scope_value": "-1"
        },
        "group_by": {
            "cluster": [
                "*"
            ]
        },
        "order_by": {},
        "total": {
            "infrastructure": {
                "raw": {
                    "value": 667.34183987,
                    "units": "USD"
                },
                "markup": {
                    "value": 63.3013545044,
                    "units": "USD"
                },
                "usage": {
                    "value": 4000.0,
                    "units": "USD"
                },
                "total": {
                    "value": 4730.6431943744,
                    "units": "USD"
                }
            },
            "supplementary": {
                "raw": {
                    "value": 0.0,
                    "units": "USD"
                },
                "markup": {
                    "value": 0.0,
                    "units": "USD"
                },
                "usage": {
                    "value": 129808.07971804855,
                    "units": "USD"
                },
                "total": {
                    "value": 129808.07971804855,
                    "units": "USD"
                }
            },
            "cost": {
                "raw": {
                    "value": 667.34183987,
                    "units": "USD"
                },
                "markup": {
                    "value": 63.3013545044,
                    "units": "USD"
                },
                "usage": {
                    "value": 133808.07971804854,
                    "units": "USD"
                },
                "total": {
                    "value": 134538.72291242296,
                    "units": "USD"
                }
            }
        }
    },
    "links": {
[...]

PR #1889, opened 2020-03-10
Nise unit test data gen

adberglund:
## Summary
- Replace the test data generators with a 1-time nise data load when creating the test database. 
- Add a keepdb option that will keep the test database between test runs to cut down on local test build time.
- Switch masu tests to subclassing Django's `TestCase`. This means every test is inside a transaction and rolled back, so you can modify the DB at will and changes will always roll back after each test. This is also significantly faster than the previous test setup/teardown for masu tests.

## Notes
- This commented out tests for the charge updater, which are about to be re-written in https://github.com/project-koku/koku/pull/1828.
- The bulk of the change here are in `koku/koku/koku_test_runner.py` and `koku/api/report/test/utils.py`, the rest is just unit test updates to fix assumptions made when loading little bits of test generator data.


maskarb:
7 minute test run.... :bow:

Issue #1890, opened 2020-03-11
Sources: Intermittent source creation error


maskarb:
**Describe the bug**
In CI, adding or editing a Source can result in a `Configuration unsuccessful` error. I've tried with an AWS source, and _sometimes_ see the attached error. This issue is a timing error. When creating a source, we send 4 POST requests to the sources-api endpoint, followed by a PATCH to cost-management/sources endpoint. The last post to sources-api contains the authentication information. If there is a lag from the time the source-api POST is processed, the authentication information will not land in the cost-management Sources table before the PATCH is called resulting in the error.

With the switch away from the SourcesProxyViewSet, we create the Provider when we receive the PATCH request. We need to check that we have all the necessary information in the Sources table _before_ we create the Provider.

**To Reproduce**
1. Go to [https://ci.cloud.redhat.com/beta/settings/sources](https://ci.cloud.redhat.com/beta/settings/sources)
2. Create a new AWS source
3. See error

OR

1. Create the source.
2. Edit source to remove cost-management application
3. Re-add cost management to the source.

**Expected behavior**
Creating a source or reattaching cost-management to a source should complete successfully.

**Screenshots**
![Screen Shot 2020-03-11 at 9 54 40 AM](https://user-images.githubusercontent.com/19751919/76424404-94b2db00-637e-11ea-98a1-57dcc3eab670.png)

lcouzens:
I believe this is the same thing our tests sometimes hit. https://ibutsu.cloud.paas.psi.redhat.com/results/5e69f91175a3335eae1cc26f

lcouzens:
This looks good for the most part, it didnt resolve the automation issues though. However I think I need to investigate those a little deeper and figure it out. Will raise a new issue for that after further investigation.

PR #1892, opened 2020-03-11
Sources: Use celery task to create Provider

maskarb:
#1890 This PR addresses the timing issue noted in #1890. All provider create/update events are handled in Celery.

- Use celery to create/update Provider. This has the added benefit of increasing our response time when PATCHing a source.
- Add `status` field to Sources table. Status is added to table during Provider creation or update. This status is sent to sources-api when a provider is created or updated.

I noticed that the status was not making its way to the sources-api when a Source was edited. This PR address that issue.


adberglund:
If we're using celery to create the provider, we might consider looking at task priority https://docs.celeryproject.org/en/latest/userguide/routing.html#rabbitmq-message-priorities or setting up a celery worker to run priority tasks. Just in the event that our normal workers are backed up processing reports. Something to discuss before we do anything about it. @chambridge Thoughts?

chambridge:
Is the focus of this to reduce latency? We could certainly investigate priority, but if this is mostly an impact from a PATCH, I'm kinda okay with seeing how we are overall. I'm not sure this API is going to be hit a ton by customers.

chambridge:
I like the transition to a uniform code flow, but I'm still not clear on what happens if a source is created during an ingestion cycle ... is there the potential for this to take hours of waiting to create the provider? If not priority, should we spin up a `sources-worker` with our e2e templates to handle this queue of tasks (seems a bit resource heavy, but should work)?

adberglund:
@chambridge @maskarb That's essentially what I was getting at with this potentially taking hours to process. Totally not unheard of to have a worker that handles specific queues for celery, and if we have one for this we can give it significantly fewer resources than the current workers.

maskarb:
The celery tasks are added to a `sources` queue. I've created a `koku-sources-worker` pod which only handles the `sources` queue in this PR:
[https://github.com/RedHatInsights/e2e-deploy/pull/1223](https://github.com/RedHatInsights/e2e-deploy/pull/1223)

PR #1896, opened 2020-03-12
Updating openapi.json with basic_auth

chargio:
Adding basic_auth description in all elements so the API request for authentication when needed.
Please review, I tried to do my best identifying all changes needed and not making mistakes.


chargio:
I would suggest we change the AP definition to 0.9.0, and start upgradign it with new changes.

docs/source/specs/openapi.json
```diff
@@ -416,44 +451,31 @@
                 }
             }
         },
-        "/reports/aws/costs/": {
-            "get": {
+        "/providers/": {
```
lcouzens: Ah I missed this, providers was depricated, this should be removed.

chargio:
I will close the PR and create a new one on Monday. I must be used the openapi.json from production

Issue #1833, opened 2020-03-02
Enable Setting to Select OpenShift Tag Keys to Summarize On


adberglund:
## User Story
As a customer, I want to curate the set of labels used to summarize and filter on from OpenShift, so that I am not bombarded with useless labels that also slow down my report queries.

## Impacts
- API
- Docs

## Assumptions:
- Our settings API can generate the necessary forms driven fields needed for the settings API

## Implementation Details
- The platform introduced application level setting with the following application:
https://github.com/RedHatInsights/settings-frontend

- We need to create an API endpoint /api/cost-management/v1/settings/ that correctly handles the GET and POST conditions needed to store the filtered set of OpenShift tag keys.

- The GET will return JSON that creates Data Driven Forms. These forms would include the current set of enabled tag keys, all available tags (multi-select or table with check boxes?)

- The POST passes back the new set of tag keys for summarization.

## Acceptance Criteria:
- [ ] Verify that Cost Management shows up in the Application level settings
- [ ] Verify that the user is presented with a form with all existing OpenShift tag keys
- [ ] Verify that the user can save OpenShift tag keys to be summarized
- [ ] Verify that visiting the setting after the initial setup presents the user with a way of knowing what tag keys are currently enabled.
- [ ] Verify that updating the tag keys changes the enabled list (add new and remove existing)

adberglund:
We can iterate on this and build out the backend infrastructure and APIs immediately. We will need input from @nlcwong on what the UI should look like via Data Driven Forms (see https://github.com/RedHatInsights/settings-frontend)

nbon12:
Sprint Planning Meeting Notes:
Andrew: We did Node and tag inheritance stuff. That was a precursor to some of this work. Now we're getting to a point where we give the users the ability to select a subset of the openshift tags they want to use for summarization. We need some input from @nlcwong on what we should be telling the data driven form to build. This is tag settings, for the data driven forms. I think iterating on this is a great idea. There's a lot of back-end work we can lay the foundation and then change the json. 
Natalie: Why do we need Data Driven Forms?
Andrew: That's the settings endpoint Insights Uses.
Chris: This is for Application Settings.
Andrew: This would be creating our application settings.
Natalie: What would we have in Settings?
Andrew: Tag keys that we want to summarize for on OpenShift. It's configuration Settings. Since we no longer have Preferences API, we now have an Insights Platform-driven Settings. We want to be using their Settings.
[...]

PR #1897, opened 2020-03-12
Create settings api endpoint to enable application level settings.

chambridge:
* Added GET and POST APIs to handle application level settings
* GET queries available + enable OCP tag keys for display
* POST handles changes to the enabled set of OCP tag keys

- [x] Need to add unit tests
- [x] Need to update OpenAPI spec

Screen Shot 2020-03-12 at 6 49 46 PM

@dayleparker This is the OpenShift Tag Key enablement feature that will live in Application Settings that will need documentation. I would appreciate any input on the wording (see image above). We can also iterate as needed if its easier to work with once in *CI*
@chargio Feel free to also provide comments.


dayleparker:
Hi @chambridge -
Sorry for the slow follow-up. I reviewed this with the related issues (ie. #1833) and the wording looks good, though I'd suggest these edits:

- If possible, it would be good to change the top heading "Applications settings" to "Application settings" (singular), as I think we are just configuring this for cost management.
-  Add missing 'for' --> "Enable your OpenShift label names to be used as tag keys **for** report grouping and filtering."

Question - what happens if you don't enable an OpenShift tag key? Is enabling a tag key mandatory to see it in reports in the cost mgmt app? I can definitely add this to docs, but not sure if anything additional needs to be mentioned on this screen.

FYI I've also created a docs JIRA where I'll track adding the config instructions and ask for your review later: https://projects.engineering.redhat.com/browse/CRCDOCS-185

chambridge:
@dayleparker 

> If possible, it would be good to change the top heading "Applications settings" to "Application settings" (singular), as I think we are just configuring this for cost management.

The title is because from this view in "Settings" you can potentially alter several applications settings (Insights, Cost-Management, Custom Policies).
image

> Add missing 'for' --> "Enable your OpenShift label names to be used as tag keys for report grouping and filtering."
I'll open a bug to update the wording to include the missing "for".

> what happens if you don't enable an OpenShift tag key?
You wont see any tag keys in the *Group By* option for the _OpenShift Details_ page
image

> Is enabling a tag key mandatory to see it in reports in the cost mgmt app?
No, you just wont see data grouped and filtered by tags for OpenShift, you would still see the cluster, node, and project based views.

dayleparker:
Thanks for the extra info Chris!

PR #1900, opened 2020-03-13
OCP  on all materialized views

Red-HAP:
## Description

There are now 8 more materialized views built on top of the existing reporting_ocpallcostlineitem_daily_summary and the project variant. These views segment the data from the base views by compute, storage, database, network. Selection flags will be checked in the OCPAllProviderMap to determine which view models should be utilized.


koku/reporting/migrations/0102_ocp_all_materialized_views.py
```diff
@@ -0,0 +1,465 @@
...
+            index=GinIndex(fields=["namespace"], name="cost_summary_namespace_idx"),
+        ),
+        migrations.AddIndex(
+            model_name="ocpazurecostlineitemdailysummary",
+            index=GinIndex(fields=["namespace"], name="ocpazure_namespace_idx"),
+        ),
+        migrations.RunSQL(
+            sql="""
+DROP MATERIALIZED VIEW IF EXISTS reporting_ocpallcostlineitem_daily_summary_database;
+CREATE MATERIALIZED VIEW reporting_ocpallcostlineitem_daily_summary_database AS
```
adberglund: The idea for these materialized views is to precompute the group bys for each API type and reduce down the data returned to just the essentials. The issue referenced a few PRs that had examples for what we did for AWS. Want to walk through those to see what we need here?

koku/reporting/migrations/0102_ocp_all_materialized_views.py
```diff
@@ -0,0 +1,465 @@
+# Generated by Django 2.2.10 on 2020-03-09 22:04
+from django.contrib.postgres.indexes import GinIndex
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
```
adberglund: I'm not seeing the create models migrations here. How are you making migrations with Django? Those should have been automatically generated.

koku/reporting_common/data/source_service_product.json
```diff
@@ -0,0 +1,58 @@
...
+        "service_category": "network",
+        "product_codes": [
+            "AmazonVPC",
+            "AmazonCloudFront",
+            "AmazonRoute53",
+            "AmazonAPIGateway"
+        ]
+    },
+    {
+        "source": "AWS",
```
adberglund: The APIs don't currently look at product_codes for compute and storage, so I think this opens up the possibility of returning results inconsistent with the filters on the APIs.

koku/api/report/all/openshift/provider_map.py
```diff
@@ -28,13 +28,50 @@
...
+    ("costs", "network"): NETWORK_SUMMARY,
+    ("costs", "database"): DATABASE_SUMMARY,
+    ("storage_by_project", None): STORAGE_SUMMARY,
+    ("instance_type_by_project", None): STORAGE_SUMMARY,
+    ("costs_by_project", "network"): NETWORK_SUMMARY,
+    ("costs_by_project", "database"): DATABASE_SUMMARY,
+}
+
+
+def resolver(report_type, report_subtype=None):
```
adberglund: These are fine, but just a different implementation than what was referenced in the PRs in the issue. So we should probably decide now which route we want to go down for the other issues and either change this or the existing AWS implementation.

koku/reporting_common/models.py
```diff
@@ -111,3 +113,20 @@ class Meta:
 
     region = models.CharField(max_length=32, null=False, unique=True)
     region_name = models.CharField(max_length=64, null=False, unique=True)
+
+
+class SourceServiceProduct(models.Model):
```
adberglund: I see that this table provides some flexibility to not re-write the materialized views if we ever change out the products we are considering for network or database, but I think it also causes some issues for compute and storage since we don't look at product type for those today. 

See: https://github.com/project-koku/koku/blob/e32b5d9c8724d19db5d1d79a09eb4210a1d3e67f/koku/api/report/all/openshift/provider_map.py#L265-L268 

and

https://github.com/project-koku/koku/blob/e32b5d9c8724d19db5d1d79a09eb4210a1d3e67f/koku/api/report/all/openshift/provider_map.py#L421

koku/reporting/migrations/0102_ocp_all_materialized_views.py
Red-HAP: @adberglund Yes, If I've missed the point, a walk-through will be invaluable.

koku/reporting/migrations/0102_ocp_all_materialized_views.py
Red-HAP: The models were created by hand as unmanaged. That is the pattern I saw for the mat-view stuff I looked over. Was this incorrect?

adberglund: That is correct, but I believe Django will still create a `migrations.CreateModel` in a migration even with managed=False. I think that keeps it flagged as a model that has already been migrated. Was wondering if you were manually writing migrations, or using `make make-migrations` which uses the Django manage command to auto-generate migrations (that you can then add RunSQLs to).

PR #1903, opened 2020-03-16
Update openapi.json with security description

chargio:
Added security --> basic_auth to all the paths available.
Modified the version of the API to 0.9.0 as it is still tech preview (even if the endpoint is v1).

I've followed the previous format adding security before responses.

Then made sure that all the path included this using:
1. Verified in http://editor.swagger.io/ that there is only one definition per path (no errors)
2. Verified that there is one per path using the following code:
with open('openapi.json') as json_data:
    d = json.load(json_data)
    pprint(d)
op = d['paths']
for x,y in op.items():
    print("Path =", x)
    for z1, z2 in y.items():
       sec = z2.get('security')
       print(z1)
       if not sec:
        print("ERRRO", z2.get('security'))
       else:
           print("OK")


chargio:
Other warnings (not addressed by the PR but reported by swagger):
CustomerOut:  Definition was declared but never used in this document.
UserOut: Definition was declared but never used in this document.

docs/source/specs/openapi.json
```diff
@@ -54,6 +54,9 @@
                 ],
                 "summary": "Obtain server status",
                 "operationId": "getStatus",
+                "security": [{
+                    "basic_auth": []
+                }],
```
adberglund:
```suggestion
```

adberglund: server status is a no-auth endpoint

docs/source/specs/openapi.json
```diff
@@ -1,8 +1,8 @@
 {
     "openapi": "3.0.0",
     "info": {
-        "description": "The API for Cost Management.  You can find out more about     Cost Management at [https://github.com/project-koku/](https://github.com/project-koku/).",
-        "version": "1.0.0",
+        "description": "The API for Project Koku and OpenShift cost management. You can find out more about Project Koku at [https://github.com/project-koku/](https://github.com/project-koku/).",
+        "version": "0.9.0",
```
lcouzens: Version should be higher not lower here.

docs/source/specs/openapi.json
```diff
@@ -54,6 +54,9 @@
                 ],
                 "summary": "Obtain server status",
                 "operationId": "getStatus",
+                "security": [{
```
lcouzens: Status endpoint doesn't require authentication.

docs/source/specs/openapi.json
lcouzens: Beat me to it

docs/source/specs/openapi.json
chargio: Normally, I believe we used 1.0.0 because it was the default, but we are not GA and we are planning on changing the API, so defining this as 1.0.0 is misleading and could cause problems in the future.

chargio: I am ok with 1.0.0, I just don't want customers opening bugs because we change the API from 1.0.0 to 1.0.1

docs/source/specs/openapi.json
chargio: Followed suggestion and deleted this, the endpoint does not ask for authentication any more

docs/source/specs/openapi.json
chargio: @adberglund what do you think?

I agree with Luke that this is a problem, but it would allow us to update the minor number every time we have a minor update.

The way I see it:
Version 1.0.0 would be consistent with the endpoint and will not downgrade existing version. Problem: using semantic versioning that would mean we are new minor versions of this API will not break the current implementation and we know that is wrong.

Version 0.9.0 would not be consistent with the endpoint /v1/ and will downgrade the existing version. However, it would allow us to refer to the api as beta until we are ready to release it as 1.0.0, meanwhile we should update the API version to 0.9.x, 0.10.x

This would only be done until GA of the API. Once we promote to version 1 we would need a wider strategy, as there should be two endpoints, the production endpoint and the beta endpoint, with different versioning. Version 2 would only be needed once we need a major update on the API.

So it would be:
- Endpoint /v1/ --> For production. Stable branch
- Endpoint /v1beta/ --> For production beta. Beta branch, it is expected some bugs that will break the API but that should be fixed before moving the API to /v1/

lcouzens: After our discussion I am happy to go with 1.0.1 for now.

PR #1911, opened 2020-03-17
cost models API always return cost type

adberglund:
## Summary
Have the new cost_type field always returned by the API with a default of supplementary cost for older cost models.


koku/cost_models/serializers.py
```diff
@@ -374,6 +374,9 @@ def to_representation(self, cost_model_obj):
         rep = super().to_representation(cost_model_obj)
         rates = rep["rates"]
         for rate in rates:
+            # Add a default for older cost models
+            if not rate.get("cost_type"):
+                rate["cost_type"] = CostModelMetricsMap.SUPPLEMENTARY_COST_TYPE
```
chambridge: Should there have been a migration to cover this? I get that rates are JSON values but could use [Run Python ](https://docs.djangoproject.com/en/3.0/ref/migration-operations/#runpython) in a migration to add the default.

I'm okay with this but we are going to have to have the same default if not found in the cost calculation logic if this isn't present.

adberglund: Yeah, you are right. Will handle this via a migration. This was actually laziness on my part.

PR #1926, opened 2020-03-20
configure dbstats to use prometheus pushgateway

blentz:
This fixes the db schema size stats that get pushed to prometheus.

As an added bonus, this sets up pushing stats to a prometheus pushgateway. So, this pattern can be used for any future stats we want to push from a celery task.

Here's a screenshot showing the stats are working again:
Screenshot from 2020-03-24 17-33-33


maskarb:
I added the PushGateway service to Docker-compose. To see stats using that service, hit the `localhost:9091` endpoint.

PR #1927, opened 2020-03-20
1691 Provider create triggers data ingest

maskarb:
#1691

This PR would cause an immediate data ingest when a Provider is created.
- The Provider.save() method kicks off a celery task if the Provider is new or if the authentication or billing source is updated.
- The provider uuid is passed through to the get_accounts_from_source function. If this function receives a uuid, it returns the account information for the corresponding provider. If, for whatever reason, that uuid is not found in the list of providers, the function returns an empty list, and no data ingestion occurs. If a provider_uuid is not provided to the function (which is how things currently happen) the function returns the full list of providers like it does today.

Testing scenarios:
1. Create a valid source, watch `koku-worker` for data ingest upon Provider creation.
2. Create an invalid source, see that the provider is not created and no data is ingested.
3. Create a valid source, see data ingest. Patch to invalid source, see that data is not ingested.
4. Create a valid source, see data ingest. Patch source to another valid source, see data ingest.
5. Create an invalid source, see that no data is ingested. Patch to valid source, see data is ingested.
6. Create multiple sources and see that the task only checks for updates for each Provider as they are created and not _all_ providers in the db.


koku/api/provider/models.py
```diff
@@ -160,6 +161,29 @@ class Meta:
...
+
+        # Commit the new/updated Provider to the DB
+        super().save(*args, **kwargs)
+
+        # Local import of task function to avoid potential import cycle.
+        from masu.celery.tasks import check_report_updates
+
+        # Start check_report_updates task after Provider has been committed.
+        if should_ingest and self.active:
+            transaction.on_commit(lambda: check_report_updates.delay(provider_uuid=self.uuid))
```
maskarb: This is where the magic happens. Once the Provider is committed, this task is started.

PR #1936, opened 2020-03-23
Add missing OpenAPI endpoints for OpenShift on All Infrastructures

chambridge:
OpenAPI spec was missing endpoints for OpenShift on All; both reports and tag endpoints.

### DIFF
--- a/docs/source/specs/openapi.json
+++ b/docs/source/specs/openapi.json
@@ -1080,6 +1080,204 @@
                 }]
             }
         },
+        "/reports/openshift/infrastructures/all/costs/": {
+            "get": {
+                "tags": [
+                    "OpenShift Reports"
+                ],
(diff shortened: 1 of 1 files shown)


chargio:
I would suggest we change the minor number in the opeapi.json version

chambridge:
I think we should be `1.0.0` until we GA (as this isn't a true contract until that point), but I'll let @adberglund weigh in before I merge.

chargio:
That is a discussion we need to have.
@lcouzens was proposing keep 1.0 as beta and GA with 2.0, so we can increment versions before that moment.

adberglund:
I don't have a strong opinion here, but don't mind starting the uptick once we're GA. I don't think GA should be v2, that essentially indicates separate code paths in the APIs and real differences or breaking changes in functionality.

Issue #1394, opened 2019-11-12
User Story: Store shared static data in python data structures rather than in the database


werwty:
## User Story

As a developer, I would like to be able to add and edit shared data without having to create a database migration.

Currently, there are many cases in Koku where shared static data is stored in JSON files and loaded into the database via migrations.
I believe these are due to the fact that koku and masu used to be separate applications and could only access these from the database.

However, with the masu/koku merger, we should reconsider and make these into python objects (see https://github.com/project-koku/koku/blob/master/koku/masu/celery/export.py#L6 as an example).
As python objects, these would be more transparent to update/read.

Some examples of where this occurs:

- [x] [Sub-Task 1934](https://github.com/project-koku/koku/issues/1934) - Store shared static data in python data structures rather than in the database for 0029_cloud_account_seeder.py
    - [x] [Sub-Task 2004](https://github.com/project-koku/koku/issues/2004) - Cloud Accounts store in a python constant instead of JSON (Revisit Python/JSON structure for consistency)
- [ ] [Sub-Task 1913](https://github.com/project-koku/koku/issues/1913) - Store shared static data in python data structures rather than in the database for 0018_auto_20190923_1838.py
- [x] [Sub-Task 1914](https://github.com/project-koku/koku/issues/1914) - Store shared static data in python data structures rather than in the database for 0010_costmodelmetricsmap.py
    - [x] [Bug introduced by 1914 solutions](https://github.com/project-koku/koku/issues/2016) - ProgrammingError: relation "cost_models_metrics_map" does not exist

## Impacts
API

## Assumptions

- Making these changes should not impact the API in any way.

## Acceptance Criteria

- [x] Database tables {cost_models_metrics_map, api_cloudaccount, reporting_common_reportcolumnmap, ...} storing these common objects are removed.
- [x] cloud account settings are loaded by a non-db migration mechanism
- [x] All API calls still work as before and conform to platform API response guidelines

nbon12:
@werwty is the link in your description still pointing to the correct code? https://github.com/project-koku/koku/blob/master/koku/masu/celery/export.py#L6

nbon12:
Planning and Estimation:

- I plan to delete the database seeder for cloud_account.
- I plan to store the following JSON into a .json file into a new directory /koku/api/cloud_accounts/data/cloud_accounts.json

```
[
    {
        "name" : "AWS Account ID",
        "value" : "589173575009"
    }
]
```
[...]

PR #1938, opened 2020-03-23
WIP: Feature/#1394 - return static JSON from a .json instead of storing it in database

nbon12:
This PR addresses issue #1394 
It removes the cloud_account table from the database.
It replaces it with a Python object that stores cloud_account.

Nick Bonilla  4:44 PM
I plan to move the hard coded value into an environment variable using the env() function. The following code would transform from:
cloud_account = CloudAccount.objects.create(
    name="AWS", value="589173575009", description="Cost Management's AWS account ID"
)
cloud_account.save()
To this:
cloud_account = CloudAccount.objects.create(
    name=env('COST_MANAGEMENT_CLOUD_ACCOUNT_TYPE), value="env('COST_MANAGEMENT_CLOUD_ACCOUNT_ID')"
)
cloud_account.save()
Sounds good?

chambrid  4:46 PM
that would still be storing it in the db. I think the goal is to be more like https://github.com/project-koku/koku/blob/master/koku/api/metrics/data/cost_models_metric_map.json
Where the data is stored in a file and served back instead of the db (edited) 

Nick Bonilla  5:14 PM
so the JSON would look like this, and we'll return this JSON straight through the API?
[
    {
        "name" : "AWS Account ID",
        "value" : "589173575009"
    }
]

chambrid  6:35 PM
If that is what the current output looks like then, yes that seems right.

chambrid  4:55 PM
There are several reasons, but I believe these are the primary drivers of those issues 

- avoid the network hop to the dB and reduce dB load
- avoid migrations to the dB which can impact startup time of the app
- simplify the code flow 

chambrid  4:49 PM
I can take a look later or tomorrow, but you may need to manually handle the output yourself and alter it based on the incoming query params. since the data is static it should be almost an if/else based on the offset and limit value where the data object is the content you have or an empty list.

Nick Bonilla  10:32 AM
Hi Chris, we mentioned talking about adding a .JSON file, but Kevan says to add a python object instead of a JSON, do you agree that we should not add a .JSON file?
10:32
for reference here's what we discussed:

chambrid  10:41 AM
I think Kevan’s point is fine. Just storing it in a Python object similar to our provider maps (https://github.com/project-koku/koku/blob/master/koku/api/report/ocp/provider_map.py#L38); also make sense and saves on pulling the file from static storage and related I/O. Plus just alter your PR from loading a file to returning a dictionary in the get_json method.
koku/api/report/ocp/provider_map.py:38
        self._mapping = [
<https://github.com/project-koku/koku|project-koku/koku>project-koku/koku | Added by GitHub


koku/api/cloud_accounts/views.py
```diff
@@ -15,23 +15,49 @@
...
 
-    def get_queryset(self):
-        """Override default get_queryset to filter on name."""
-        queryset = CloudAccount.objects.all()
-        cloud_account = self.request.query_params.get("name", None)
-        if cloud_account is not None:
-            queryset = queryset.filter(name=cloud_account)
-        return queryset
+
+def get_json(path):
```
kholdaway: You should just store this as python not JSON.  The first thing you did is load it into python anyway.  This will also help with query params.

blentz: @kholdaway - This is following the same pattern here that we already use for serving up the `openapi.json`. 

The cloud account info needs to be stored in a human-editable format so that we can change it when needed. This data should not be pickled or stored in any other non-editable format.

koku/api/cloud_accounts/views.py
```diff
@@ -15,23 +15,49 @@
...
+
+@api_view(["GET"])
+@permission_classes((permissions.AllowAny,))
+@renderer_classes([BrowsableAPIRenderer, JSONRenderer])
+def cloudaccounts(request):
+    """Provide the openapi information."""
+    data = get_json(OPENAPI_FILE_NAME)
+    paginator = Paginator(data, 25)
+    page_number = request.GET.get("page")
+    page_obj = paginator.get_page(page_number)
```
chambridge: This needs to follow the input and output formats of the platform specification for pagination.
https://platform-docs.cloud.paas.psi.redhat.com/rest/pagination.html

Here is a pointer for our common flow that uses in DB objects that meets the above specs:
https://github.com/project-koku/koku/blob/master/koku/api/common/pagination.py

Defaults if no query param should match this implementation. Errors should be provided for bad inputs to the query parameters (you could use a serializer to do this like we do with the report APIs).
Its also preferred that you use `request.query_params` for readability when accessing limit and offset. https://www.django-rest-framework.org/api-guide/requests/#query_params

koku/api/cloud_accounts/views.py
```diff
@@ -15,23 +15,49 @@
...
+    if data:
+        return Response(list(page_obj))
+    return Response(status=status.HTTP_404_NOT_FOUND)
```
chambridge: Output format should match with the spec provided above. A 404 doesn't make sense as you will return the above spec with an empty *data* field (i.e. `"data": []`)

koku/api/cloud_accounts/views.py
nbon12: @kholdaway where do you want this Python object to live?

nbon12: @kholdaway I plan to store this Python object in /cloud_accounts/cloud_accounts_data.py 
Let me know if you think the location should change.

kholdaway: @blentz There is a query parameter so I would think we may need to filter.  Is that incorrect?  Also, the code did a json.load immediate (python again anyway)

kholdaway: see @blentz https://github.com/project-koku/koku/pull/1938/files#diff-60437d138c668df7d7e3346cc6fd97c4R65-R72

koku/api/cloud_accounts/views.py
```diff
@@ -15,23 +15,117 @@
...
+            LOG.exception(exc)
+    return json_data
+
+
+@api_view(["GET"])
+@permission_classes((permissions.AllowAny,))
+@renderer_classes([BrowsableAPIRenderer, JSONRenderer])
+def cloudaccounts(request):
+    """Provide the openapi information."""
+    data = CloudAccountsDictionary()._mapping
```
kholdaway: Just remove CloudAccountsDictionary class.  It provides no value over a constant.

koku/api/cloud_accounts/views.py
blentz: @kholdaway - To me, this is about mantaining separation of concerns. This contents of the json is data, not code. So, storing it as json is an appropriate solution and allows us to change it on the fly (e.g. with an openshift mounted secret file).

Hard-coding this information into the code means that it's less flexible and can't be changed without going through the entire release process. 

My suggestion to Nick to go this route is based on my years of experience operating webservices at scale and knowing these are the kinds of knobs that it's good to leave exposed to maintain service flexibility. During my time in OpenShift Operations, we moved AWS accounts more than once. So, I'm just trying to get us a solution that we won't have to change later.

However, I don't feel the need to argue the point. I worked with Nick yesterday to remove the json file and just use a hard-coded version as you and Chris were requesting. So, the point is moot. As soon as he pushes his latest changes, you'll have the implementation you were asking for.

kholdaway: @blentz Do we have a plan to co-located all configuration files?  @nbon12 is going to do this for several more similar issues.  If there is a defined configuration area then I think I could agree to the separation of concerns.  We should make sure the documentation covers that configuration.

kholdaway:
This PR should be resubmitted as it was reverted due to dismissed reviews.

PR #1942, opened 2020-03-24
Missing sources migration

kholdaway:
While working on another PR, this migration showed up when I tried to generate my migrations.  It's not related to my other PR so I split this out.


chambridge:
Looks like this is also getting captured in this PR: https://github.com/project-koku/koku/pull/1940

kholdaway:
This was coved by #1940

PR #1962, opened 2020-03-27
1705 suggested changes

adberglund:
## Summary 

- Use Django's make migrations to add models to the migration 
- Added cost summary views for no group by, group by account, service, and region
- Renamed views for consistent naming across sources
- Updated SQL for views to be consistent with API filters (this part is particularly nuanced, since it isn't immediately clear how each API endpoint selects what is compute or storage, et cetera) 
   - For reference storage filters are here: https://github.com/project-koku/koku/blob/1ea8d150a86909bed05ae5ef5552a2b2dd020290/koku/api/report/all/openshift/provider_map.py#L265-L267
   - CPU filters are here: https://github.com/project-koku/koku/blob/1ea8d150a86909bed05ae5ef5552a2b2dd020290/koku/api/report/all/openshift/provider_map.py#L421
- Removed project summary views (we should hold on to the idea for these, but they're not in use on the dashboard, so we can investigate those as we look at optimizing the details page)
- Removed index changes -- let's keep those in a separate PR
- Cleaned up extraneous code in query params
- Removed service map table for now. I like the idea but I think we need a spike to look at Azure and AWS APIs to get product services programmatically. This is also a bigger change than it appears, as the compute and storage entries do not line up directly with the filtering our APIs do and could thus generate different data sets in the view than in the daily summary table. 
   - AWS pricing API: https://docs.aws.amazon.com/cli/latest/reference/pricing/get-products.html
   - Azure product API maybe????: https://docs.microsoft.com/en-us/partner-center/develop/get-a-list-of-products


koku/api/query_params.py
```diff
@@ -32,6 +32,7 @@
 from api.report.queries import ReportQueryHandler
 from reporting.models import OCPAllCostLineItemDailySummary
 
+
```
adberglund:
```suggestion
```

koku/reporting/provider/all/openshift/models.py
```diff
@@ -102,6 +102,268 @@ class Meta:
...
+    unblended_cost = models.DecimalField(max_digits=30, decimal_places=15, null=True)
+
+    markup_cost = models.DecimalField(max_digits=30, decimal_places=15, null=True)
+
+    currency_code = models.CharField(max_length=10, null=True)
+
+
+# ===============
+#  Project views
+# ===============
```
adberglund:
```suggestion
```

adberglund:
These changes were made in https://github.com/project-koku/koku/pull/1974

PR #1971, opened 2020-03-28
WIP: 1704 ocp on azure views

adberglund:
## Summary 
OCP on Azure views.


koku/api/report/azure/openshift/query_handler.py
```diff
@@ -53,6 +59,72 @@ def __init__(self, parameters):
...
+        # { query_param: database_field_name }
+        fields = self._mapper.provider_map.get("annotations")
+        for q_param, db_field in fields.items():
+            annotations[q_param] = Concat(db_field, Value(""))
+        if (
+            "project" in self.parameters.parameters.get("group_by", {})
+            or "and:project" in self.parameters.parameters.get("group_by", {})
+            or "or:project" in self.parameters.parameters.get("group_by", {})
+        ):
+            annotations["project"] = Concat("namespace", Value(""))
```
Red-HAP, line 80: Am I reading this right that you're concatenating an empty string to namespace?

koku/api/report/azure/openshift/query_handler.py
```diff
@@ -53,6 +59,72 @@ def __init__(self, parameters):
         super().__init__(parameters)
         # super() needs to be called before _get_group_by is called
 
+    @property
+    def annotations(self):
```
blentz, line 63: I really like this idea. I think it might be better to put it in the ProviderMap class than in the QueryHandler, though.

My reasoning is to keep the QueryHandler focused on running the report query and letting the ProviderMap handle organizing the fields to be queried.

What do you think?

koku/api/report/azure/openshift/query_handler.py
```diff
@@ -53,6 +59,72 @@ def __init__(self, parameters):
...
+            "project" in self.parameters.parameters.get("group_by", {})
+            or "and:project" in self.parameters.parameters.get("group_by", {})
+            or "or:project" in self.parameters.parameters.get("group_by", {})
+        ):
+            annotations["project"] = Concat("namespace", Value(""))
+
+        return annotations
+
+    @property
+    def query_table(self):
```
blentz, line 85: This is also something I'd consider keeping in the ProviderMap class.

koku/api/report/azure/openshift/query_handler.py
adberglund: You know, I copied this from the ocp query handler, and didn't think twice. I will fix this to a `Coalesce` here and there.

PR #1974, opened 2020-03-30
OCP on ALL materialized views

Red-HAP:
With much thanks to Andrew, here are the changes for the OCP on ALL materialized views


koku/reporting/migrations/0103_ocpallcomputesummary_ocpallcostsummary_ocpallcostsummarybyaccount_ocpallcostsummarybyregion_ocpallco.py
```diff
@@ -0,0 +1,361 @@
...
+
+    dependencies = [
+        (
+            "reporting",
+            "0103_azurecomputesummary_azurecostsummary_azurecostsummarybyaccount_azurecostsummarybylocation_azurecosts",
+        )
+    ]
+
+    operations = [
+        migrations.CreateModel(
```
adberglund: Can you rename the actual file to be 104 instead of 103?

Issue #1914, opened 2020-03-18
Store shared static data in python data structures rather than in the database for 0010_costmodelmetricsmap.py


kholdaway:
## Sub-Task

As a developer, I would like to be able to add and edit shared data without having to create a database migration.

Currently, there are many cases in Koku where shared static data is stored in JSON files and loaded into the database via migrations.
I believe these are due to the fact that koku and masu used to be separate applications and could only access these from the database.

However, with the masu/koku merger, we should reconsider and make these into python objects (see https://github.com/project-koku/koku/blob/master/koku/masu/celery/export.py#L6 as an example).
As python objects, these would be more transparent to update/read.

Some examples of where this occurs:
- https://github.com/project-koku/koku/blob/master/koku/api/migrations/0010_costmodelmetricsmap.py#L11
    Same as above.

## Impacts
API

## Assumptions

- Making these changes should not impact the API in any way.

## Acceptance Criteria

- [ ] Database tables {cost_models_metrics_map, api_cloudaccount, reporting_common_reportcolumnmap, ...} storing these common objects are removed.
- [ ] JSON settings are loaded by a non-db migration mechanism
- [ ] All API calls still work as before and conform to platform API response guidelines

nbon12:
Planning and Estimation:

I plan to delete the area of code that inserts the cost model metrics map into the database.
I plan to load the cost model metrics map JSON directly into the View, and serve this View.
I plan to return this JSON straight through the API, changing the formatting to match the current formatted JSON response. 
I plan to alter the cost model metrics map tests to work without a database model and with a View.
I estimate this to take 5 days. Any changes to this plan may change the estimate.
After a PR is initially submitted and review begins, any additional review comment that requests for additional changes to be made will take an estimated 1 day per each request for additional change.

nbon12:
Acceptance criteria: 

- Cost_models_metric_map currently performs querying based on source_type param. This new solution should also perform querying based on source_type param.
- The current Cost Models Metric Map performs a translation from OCP --> OpenShift Container Platform before returning the JSON to the client. This translation should also exist in the new solution.

PR #1986, opened 2020-03-30
Store shared static data in python data structures rather than in the database for 0010_costmodelmetricsmap.py

nbon12:
This PR removes the CostModelMetricsMap model, and replaces it with the usage of static JSON and `dict`.
Addresses: #1914 

## Functional Testing

1. Go to http://localhost:8000/api/cost-management/v1/metrics/
2. Observe that the list of 7 metrics appear
Screen Shot 2020-04-01 at 3 10 44 PM

3. enter a query param ?limit=1
4. observe the output was limited to 1 array item
5. enter a query param ?offset=1
6. Observe the second element in the array.
7. enter a query param ?source_type=OCP
8 observe all output is of source_type OCP
9. enter query param ?source_type=AWS
10. observe an empty data array in the response

## Functional that test cost-models still works
1. Go to http://localhost:8000/api/cost-management/v1/cost-models/
2. Add a cost model in the browseable API:
 Name: hi
Description: hiho
Source type: AWS
Value: 1
Unit: %
3. Observe that the cost model was added.
Screen Shot 2020-04-01 at 3 10 28 PM

## Functional test invalid JSON
1. Locally edit the following file so that the JSON is invalid: https://github.com/project-koku/koku/blob/master/koku/api/metrics/data/cost_models_metric_map.json
2. Set DEBUG = False in your Django settings
3. Go to http://localhost:8000/api/cost-management/v1/cost-models/ observe an error 500
Screen Shot 2020-04-01 at 4 20 56 PM


koku/api/metrics/views.py
```diff
@@ -45,14 +48,25 @@ def get_queryset(self):
...
         return super().list(request=request, args=args, kwargs=kwargs)
+
+    def get_json(self, path):
+        """Obtain API JSON data from file path."""
+        json_data = None
+        with open(path) as json_file:
+            try:
+                json_data = json.load(json_file)
+            except (IOError, json.JSONDecodeError) as exc:
+                LOG.exception(exc)
```
adberglund: We'll want to fail and provide the proper HTTP code if we can't load the JSON in. If that happens this API has nothing to return.

koku/api/metrics/constants.py
```diff
@@ -0,0 +1,41 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+"""Model for our cost model metric map."""
+OCP_METRIC_CPU_CORE_USAGE_HOUR = "cpu_core_usage_per_hour"
```
adberglund: I would consider doing 1 of 2 things. 

1) Move these constant definitions to the `__init__.py` for metrics so they are importable from metrics or 
2) Alias the import `from api.metrics import constants` to `from api.metrics import constants as metric_constants`

I lean slightly towards 1, but either way the goal is improved readability/understandability in the places where you import these constants. Just importing something as "constants" doesn't provide the context that they are metric specific constants.

koku/cost_models/serializers.py
```diff
@@ -96,20 +101,31 @@ def validate(self, data):
             return data
 
 
+def get_json(path):
```
adberglund: Copy pasted from metrics/views. Would consider making it a utility function instead of writing the same code in 2 places. But more than that in this case we know the JSON file we are trying to load. I would add this kind of error handling if we had user input or different possibly malformed JSON files coming in, but in this case we know it should be a good JSON file so I'd probably swap this entire function with just a 1-line variable definition 

`data = json.load(json_file)`

koku/cost_models/serializers.py
```diff
@@ -96,20 +101,31 @@ def validate(self, data):
...
     metric = serializers.DictField(required=True)
-    cost_type = serializers.ChoiceField(choices=CostModelMetricsMap.COST_TYPE_CHOICES)
+    cost_type = serializers.ChoiceField(choices=constants.COST_TYPE_CHOICES)
     tiered_rates = serializers.ListField(required=False)
 
     @property
     def metric_map(self):
         """Return a metric map dictionary with default values."""
-        metrics = CostModelMetricsMap.objects.values("metric", "default_cost_type")
+        metrics = get_json(os.path.join(BASE_DIR, "api/metrics/data/cost_models_metric_map.json"))
```
adberglund: You've already got a constant defined in another file for this location, I'd use that.

koku/cost_models/serializers.py
```diff
@@ -294,10 +310,10 @@ class Meta:
     def metric_map(self):
         """Map metrics and display names."""
         metric_map_by_source = defaultdict(dict)
-        metric_map = CostModelMetricsMap.objects.all()
+        metric_map = get_json(os.path.join(BASE_DIR, "api/metrics/data/cost_models_metric_map.json"))
```
adberglund: You've already got a constant defined in another file for this location, I'd use that.

koku/cost_models/serializers.py
```diff
@@ -294,10 +310,10 @@ class Meta:
...
 
         for metric in metric_map:
-            metric_map_by_source[metric.source_type][metric.metric] = metric
+            metric_map_by_source[metric.get("source_type", None)][metric.get("metric", None)] = metric
```
adberglund: If either of those .get() calls returns a None this will probably blow up. Since you're hard referencing the key with `[]` on metric_map_by_source

koku/cost_models/test/tests_serializers.py
```diff
@@ -363,11 +357,11 @@ def test_get_metric_display_data_openshift(self):
         """Test the display data helper function for OpenShift metrics."""
         serializer = CostModelSerializer(data=None)
 
-        for metric_choice in CostModelMetricsMap.METRIC_CHOICES:
+        for metric_choice in constants.METRIC_CHOICES:
             response = serializer._get_metric_display_data(Provider.PROVIDER_OCP, metric_choice[0])
-            self.assertIsNotNone(response.label_measurement_unit)
-            self.assertIsNotNone(response.label_measurement)
-            self.assertIsNotNone(response.label_metric)
+            self.assertIsNotNone(response.get("label_measurement_unit", None))
```
adberglund: None is the default return value for .get() so you don't actually need to include it.

koku/api/metrics/constants.py
nbon12: I chose option #2 @adberglund , it's updated now.

koku/cost_models/serializers.py
nbon12: This comment can be resolved due to a slack conversation:

> Nick Bonilla  15:42
> ok, so we don't need error handling, is what I'm drawing from this, but we do need a unit test to ensure the json contains the required data






> aberglun:bulbasaur:
> 15:43
> Yeah, I would feel good about that validation.

koku/cost_models/serializers.py
nbon12: @adberglund is this comment asking for it to eliminate the function entirely?

adberglund: Given what we talked about, what do you think?

koku/cost_models/test/tests_serializers.py
nbon12: updated

koku/cost_models/serializers.py
nbon12: updated

koku/cost_models/serializers.py
nbon12: yes, I think you are asking to eliminate the function entirely.

koku/cost_models/test/tests_serializers.py
nbon12: updated

koku/cost_models/serializers.py
nbon12: and to replace it with this: https://github.com/project-koku/koku/pull/1986/files#diff-a5fd0aef40ad2b68f502d6854df834b8R51

nbon12: is that correct?

koku/api/metrics/views.py
nbon12: resolved as per "so I'd probably swap this entire function with just a 1-line variable definition"

koku/cost_models/serializers.py
nbon12: updated.

koku/api/metrics/views.py
```diff
@@ -45,14 +51,29 @@ def get_queryset(self):
 
         Restricts the returned data to provider_uuid if supplied as a query parameter.
         """
-        queryset = CostModelMetricsMap.objects.all()
+        try:
+            with open(COST_MODEL_METRICS_FILE_NAME) as json_file:
+                queryset = json.load(json_file)
```
kholdaway, line 56: so this is no longer queryset.  Let's give it a name that matches what it is now.  QuerySet is a specific django type.

koku/api/metrics/views.py
```diff
@@ -45,14 +51,29 @@ def get_queryset(self):
...
 
         return queryset
 
     @vary_on_headers(RH_IDENTITY_HEADER)
     def list(self, request, *args, **kwargs):
         """Obtain the list of CostModelMetrics for the tenant."""
         return super().list(request=request, args=args, kwargs=kwargs)
+
+
+class CostModelMetricMapJSONException(APIException):
```
kholdaway, line 73: What benefit does this class have over just returning a status Response with 500 and an error?

nbon12: Andrew said: "Take a look at https://github.com/project-koku/koku/blob/master/koku/cost_models/view.py#L87-L93 so that’s the form of error message we want to return when something goes wrong.
koku/cost_models/view.py:87-93
```
class CostModelProviderQueryException(APIException):
    """Rate query custom internal error exception."""

    def __init__(self, message):
        """Initialize with status code 500."""
"
```

koku/api/metrics/views.py
nbon12: updated @kholdaway

koku/cost_models/migrations/0002_auto_20200318_1233.py
```diff
@@ -1,15 +1,18 @@
 # Generated by Django 2.2.11 on 2020-03-18 12:33
```
dccurtis, line 1: I need to think this through a little more and even try it for myself. But I think you might have to do this metrics loading in a new migration rather than modify an existing one.  I'll get back to you on this.

koku/api/metrics/views.py
kholdaway: 👍

koku/api/metrics/views.py
kholdaway: 👍

koku/api/metrics/views.py
```diff
@@ -36,23 +43,40 @@ class CostModelMetricsMapViewSet(mixins.ListModelMixin, viewsets.GenericViewSet)
...
+        """
+        Get JSON from the cost_model_metrics_map.json.
 
-        Restricts the returned data to provider_uuid if supplied as a query parameter.
+        Filter on source_type
         """
-        queryset = CostModelMetricsMap.objects.all()
+        try:
+            with open(COST_MODEL_METRICS_FILE_NAME) as json_file:
+                cost_model_metrics_map = json.load(json_file)
```
chambridge, line 57: Can we consolidate the calls to load this JSON file?
Its done in 3 different places:
https://github.com/project-koku/koku/pull/1986/files#diff-999179f93d3cba58861a615b1899b2c4R303
https://github.com/project-koku/koku/pull/1986/files#diff-999179f93d3cba58861a615b1899b2c4R117

Preferably it would all go to one method and the data would be stored in a variable on the first load and then returned from the loaded variable on subsequent calls to that method.

nbon12: @chambridge I'll plan to store it in a constant instead of loading it from JSON.

nbon12: @chambridge see updated code. It's now a native python constant.

koku/cost_models/migrations/0002_auto_20200318_1233.py
adberglund: @dccurtis We talked through this. It's a data migration that will break from now on with the model gone. I probably would have taken it a step forward and just had this migration do nothing at all.

dccurtis: yeah I just tested this on master, pulled this branch in, and restarted koku.  Worked fine. 👍

koku/api/metrics/constants.py
```diff
@@ -0,0 +1,117 @@
...
+    (SUPPLEMENTARY_COST_TYPE, SUPPLEMENTARY_COST_TYPE),
+)
+
+SOURCE_TYPE_MAP = {
+    Provider.PROVIDER_OCP: "OpenShift Container Platform",
+    Provider.PROVIDER_AWS: "Amazon Web Services",
+    Provider.PROVIDER_AZURE: "Microsoft Azure",
+}
+
+cost_model_metric_map = [
```
chambridge, line 52: Constants are typically all caps.

koku/api/metrics/serializers.py
```diff
@@ -14,41 +14,13 @@
...
-
-class CostModelMetricMapSerializer(serializers.ModelSerializer):
-    """Serializer for the CostModelMetricsMap model."""
-
-    class Meta:
-        """Metadata for the serializer."""
 
-        model = CostModelMetricsMap
-        exclude = ("id",)
+class CostModelMetricMapSerializer(serializers.Serializer):
```
chambridge, line 21: What does this class do? If the `to_representation just hands back the data that's given it seems like this is just the `serializers.Serializer` class and we don't need it at all?

koku/api/metrics/views.py
```diff
@@ -15,19 +15,27 @@
...
 
 from api.common import RH_IDENTITY_HEADER
-from api.metrics.models import CostModelMetricsMap
+from api.metrics import constants as metric_constants
 from api.metrics.serializers import CostModelMetricMapSerializer
+from koku.settings import BASE_DIR
 
 LOG = logging.getLogger(__name__)
 
+COST_MODEL_METRICS_FILE_NAME = os.path.join(BASE_DIR, "api/metrics/data/cost_models_metric_map.json")
```
chambridge: Seems like this can be deleted, right?

koku/api/metrics/views.py
```diff
@@ -36,23 +44,42 @@ class CostModelMetricsMapViewSet(mixins.ListModelMixin, viewsets.GenericViewSet)
...
+            if source_type:
+                # Filter on source type
+                cost_model_metric_map_copy = list(
+                    filter(lambda x: x.get("source_type") == source_type, cost_model_metric_map_copy)
+                )
+            # Convert source_type to human readable.
+            for metric_map in cost_model_metric_map_copy:
+                metric_map["source_type"] = metric_constants.SOURCE_TYPE_MAP[metric_map["source_type"]]
+        except KeyError:
+            LOG.debug("Malformed JSON", exc_error=True)
```
chambridge, line 64: Feels like this should be surfaced if its happening, since this is an in code constant now.
```suggestion
            LOG.error("Malformed JSON", exc_error=True)
```

koku/api/metrics/views.py
dccurtis, line 37: As far as I can tell I don't think we need this any more and can also remove the `/metrics/data/cost_models_metric_map.json` file

nbon12: removed

koku/cost_models/serializers.py
```diff
@@ -398,9 +400,9 @@ def to_representation(self, cost_model_obj):
             display_data = self._get_metric_display_data(cost_model_obj.source_type, metric.get("name"))
             metric.update(
                 {
-                    "label_metric": display_data.label_metric,
-                    "label_measurement": display_data.label_measurement,
-                    "label_measurement_unit": display_data.label_measurement_unit,
+                    "label_metric": display_data.get("label_metric", ""),
+                    "label_measurement": display_data.get("label_measurement", ""),
+                    "label_measurement_unit": display_data.get("label_measurement_unit", ""),
```
chambridge, line 405: I don't think we should be passing back empty string here.
Seems more like we want to check that these values are not None before we update the metric and if any are None it seems like a 500 because the constant should have these items.

koku/api/metrics/views.py
```diff
@@ -36,23 +40,42 @@ class CostModelMetricsMapViewSet(mixins.ListModelMixin, viewsets.GenericViewSet)
 
     """
 
-    queryset = CostModelMetricsMap.objects.all()
     serializer_class = CostModelMetricMapSerializer
```
chambridge, line 43: Can this just be the `Serializer` class  `from rest_framework.serializers import Serializer`? as mentioned above

koku/api/metrics/serializers.py
nbon12: Hi Chris, when I deleted the serializer, Django gave me an error that a serializer_class was needed in the ViewSet, or to override get_serializer_class(). When I overrided the get_serializer_class(), the data came back as empty. What do you suggest instead?

chambridge: Did you try this https://github.com/project-koku/koku/pull/1986#discussion_r402409433 and it didn't work? Or just not provide a serializer?

nbon12: nevermind, I'll plan to use the default django Serializer instead of this class

myersCody:
Wait, I am a little confused didn't we decide in the last pr that we would save these constants in a json file, but in this pr we have switched the constants to a python module. Did we decide that a python module is better? I am totally down either, but it would be great if they were consistent.

koku/api/metrics/serializers.py
nbon12: @chambridge yes, I tried https://github.com/project-koku/koku/pull/1986#discussion_r402409433

chambridge: ok. I guess keep the existing serializer. I guess the `to_representation` method is what is making this work.

koku/api/metrics/views.py
chambridge: ok. thanks for trying that. Would have preferred to simplify this for readability, but looks like it needs to stay as is.

koku/cost_models/serializers.py
nbon12: added

koku/api/metrics/views.py
nbon12: deleted

koku/api/metrics/constants.py
nbon12: updated

kholdaway:
@nbon12 I added #2004 as a subtask to #1394.  Splitting that out is a great idea since it is not related to this PR.  It is, however, part of the work for the story.

koku/api/migrations/0013_auto_20200226_1953.py
```diff
@@ -8,16 +8,7 @@
...
-    CostModelMetricsMap.objects.all().delete()
-
-    data = pkgutil.get_data("api", "metrics/data/cost_models_metric_map.json")
-
-    data = json.loads(data)
-
-    for entry in data:
-        map = CostModelMetricsMap(**entry)
-        map.save()
+    pass
```
chambridge, line 11: @adberglund  Can you take a look?

If we are going to make this change it makes more sense to me to delete the method and the *RunPython* call:
https://github.com/project-koku/koku/pull/1986/commits/f1b23c08e09f91f7630f82e7fbacbc78b6aeeda9#diff-6776daf81424264056a8d93f5f0a077dR26

koku/api/migrations/0018_auto_20200326_0102.py
```diff
@@ -7,16 +7,7 @@
...
-    CostModelMetricsMap.objects.all().delete()
-
-    data = pkgutil.get_data("api", "metrics/data/cost_models_metric_map.json")
-
-    data = json.loads(data)
-
-    for entry in data:
-        map = CostModelMetricsMap(**entry)
-        map.save()
+    pass
```
chambridge, line 10: @adberglund  Can you take a look, similar to the note above.

Why not just remove both the method and the operation call:
https://github.com/project-koku/koku/pull/1986/commits/f1b23c08e09f91f7630f82e7fbacbc78b6aeeda9#diff-3dee2e30ad652d9aba5aa0b9f1427813R17

nbon12: updated @chambridge.

I removed the method and operation call.

koku/api/migrations/0013_auto_20200226_1953.py
nbon12: @chambridge I updated the PR.

I deleted the pre-existing method and RunPython call.

PR #2023, opened 2020-04-06
Create an aws account hierarchy crawler 

myersCody:
This pull request containers the account hierarchy crawler that will obtain the organizational unit tree and save it to the database. 

## Testing

**Step One**:
In order to test his pr you will first need to modify the celery.py crawler here: https://github.com/project-koku/koku/pull/2023/files#diff-f4a924a349e0b16cd71f8dd368d923d1R128

To be:
```
pp.conf.beat_schedule["crawl_account_hierarchy"] = {
    "task": "masu.celery.tasks.crawl_account_hierarchy",
    "schedule": crontab(minute="*"),
}
```

This will change the beat to be every minute instead of what it is set or in production which is currently once a day. 

**Step Two**:
After that you should get a hold with me for a postman query that contains some sensitive information for an aws account that has some organizational units in it.  After you post the aws account the crawler will start crawling for the information like so: 
```
koku_worker       | [2020-04-06 16:49:51,746] INFO 55c4213b-17a0-4960-afe9-c657f718b0a4 Polling for
koku_worker       |  schema_name: acct10001
koku_worker       |  provider: AWS
koku_worker       |  account (provider uuid): b07c3edb-5119-4245-b187-a0ce0070a75d
koku_worker       | [2020-04-06 16:49:51,746: INFO/ForkPoolWorker-1] [masu.celery.tasks.crawl_account_hierarchy(55c4213b-17a0-4960-afe9-c657f718b0a4 via 55c4213b-17a0-4960-afe9-c657f718b0a4)] Account hierarchy crawler found 1 accounts to scan
koku_worker       | [2020-04-06 16:49:51,747: INFO/ForkPoolWorker-1] [masu.celery.tasks.crawl_account_hierarchy(55c4213b-17a0-4960-afe9-c657f718b0a4 via 55c4213b-17a0-4960-afe9-c657f718b0a4)] Starting account hierarchy crawler for type AWS with provider_uuid (b07c3edb-5119-4245-b187-a0ce0070a75d)
koku_worker       | [2020-04-06 16:49:51,772: INFO/ForkPoolWorker-1] Found credentials in environment variables.
koku_worker       | [2020-04-06 16:49:51,995] INFO 55c4213b-17a0-4960-afe9-c657f718b0a4 Starting aws organizations session for crawler.
koku_worker       | [2020-04-06 16:49:52,303] INFO 55c4213b-17a0-4960-afe9-c657f718b0a4 Obtained the root identifier: r-jtfp
koku_worker       | [2020-04-06 16:49:52,303] INFO 55c4213b-17a0-4960-afe9-c657f718b0a4 Saving account or org unit: unit_name=Root, unit_id=r-jtfp, unit_path=r-jtfp, account_id=None
koku_worker       | [2020-04-06 16:49:52,337] INFO 55c4213b-17a0-4960-afe9-c657f718b0a4 Obtaining accounts for organizational unit: r-jtfp
koku_worker       | [2020-04-06 16:50:21,808] INFO 55c4213b-17a0-4960-afe9-c657f718b0a4 Saving account or org unit: unit_name=Root, unit_id=r-jtfp, unit_path=r-jtfp, account_id=0002
koku_worker       | [2020-04-06 16:50:21,824] INFO 55c4213b-17a0-4960-afe9-c657f718b0a4 Saving account or org unit: unit_name=Root, unit_id=r-jtfp, unit_path=r-jtfp, account_id=002
koku_worker       | [2020-04-06 16:50:21,843] INFO 55c4213b-17a0-4960-afe9-c657f718b0a4 Saving account or org unit: unit_name=Root, unit_id=r-jtfp, unit_path=r-jtfp, account_id=003
koku_worker       | [2020-04-06 16:50:21,864] INFO 55c4213b-17a0-4960-afe9-c657f718b0a4 Saving account or org unit: unit_name=Root, unit_id=r-jtfp, unit_path=r-jtfp, account_id=008
```

Note: Be careful, some times when you post it, the beat to download will start, if this happens you will want to run `docker-compose down` & `docker-compose up` again.


koku/masu/celery/tasks.py
```diff
@@ -370,3 +372,34 @@ def clean_volume():
...
+        if account.get("provider_type") == Provider.PROVIDER_AWS:
+            crawler = AWSOrgUnitCrawler(account)
+
+        if crawler:
+            LOG.info(
+                "Starting account hierarchy crawler for type {} with provider_uuid ({})".format(
+                    account.get("provider_type"), account.get("provider_uuid")
+                )
+            )
+            crawler.crawl_account_hierarchy()
```
blentz, line 397: It looks like this is going to do a serial crawl of each account. So, as the number of accounts increase, the time it takes for this to run will increase.

Before we try to optimize that, do you have a sense for how long each account takes to crawl? Like, will we need to worry about this task's runtime at 100 accounts or 10,000 accounts?

At some point, we'll probably need to adjust this, but before we worry about that, I just want to do a quick "gut check" for whether that needs to happen now versus 9-12 months from now.

koku/masu/external/accounts/hierarchy/account_crawler.py
```diff
@@ -0,0 +1,37 @@
...
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+"""Account Hierarchy Abstract class."""
+import logging
+
+LOG = logging.getLogger(__name__)
+
+
+class AccountCrawler:
```
blentz, line 23: It looks like you're intending this to be an abstract class. So, you probably want to sub-class this off of `abc.ABC`.

That will ensure the class automatically raises `NotImplemented` exceptions for unimplemented methods.

```suggestion
from abc import ABC

class AccountCrawler(ABC):
```

koku/masu/external/accounts/hierarchy/aws/aws_org_unit_crawler.py
```diff
@@ -0,0 +1,147 @@
...
+    def _crawl_org_for_acts(self, ou, prefix):
+        """
+        Recursively crawls the org units and accounts.
+
+        Args:
+            ou (dict): A return from aws client that includes the Id
+        """
+
+        # Save entry for current OU
+        self._save_aws_org_method(ou.get("Name", ou.get("Id")), ou.get("Id"), prefix, None)
```
blentz, line 105: If I'm reading this right, every time we recurse, we hit the DB to save the record.

On a sufficiently cclientomplex organization, I'm a bit concerned this is going to have negative performance ramifications. I'd like @adberglund or @Red-HAP to weigh in on whether the potential performance impact is something we should address now or later. 

I think the solution is relatively simple - we'd need to cache all data and batch-upload them once the entire tree has been walked. I just don't want to make you rework that right now unless the potential impact seems big enough.

koku/masu/celery/tasks.py
myersCody: Yeah, during local testing I used the openshift dedicated aws account, which has about 2000 accounts associated with it. It was pretty fast on my mac taking about 3-4 minutes on average to collect the entire hierarchy tree. The benefit of the save method we are using is that we are doing a get_or_create. So the hardest hitting run will always be the first run and all the runs after that is mostly just adding new information. 

That being said, I would be open to any performance suggestions.

koku/masu/external/accounts/hierarchy/account_crawler.py
myersCody: This is a great suggestion @blentz ,  I am working on it now. Just double checking that I did right at the moment. 😂

koku/masu/celery/tasks.py
blentz: Some back of the napkin math says that should cover us for a few thousand accounts of similar complexity. So, I think we'll be okay for a while. I see no need to optimize this right now.

koku/masu/external/accounts/hierarchy/aws/aws_org_unit_crawler.py
```diff
@@ -0,0 +1,147 @@
...
+        LOG.info(
+            "Saving account or org unit: unit_name=%s, unit_id=%s, unit_path=%s, account_id=%s"
+            % (unit_name, unit_id, unit_path, account_id)
+        )
+        with schema_context(self.schema):
+            row, created = AWSOrganizationalUnit.objects.get_or_create(
+                org_unit_name=unit_name, org_unit_id=unit_id, org_unit_path=unit_path, account_id=account_id
+            )
+            if created:
+                row.created_timestamp = DateAccessor().today()
```
adberglund, line 138: You can let Django handle this with a default value so you don't have to set it. (Totally optional, this is okay as-is)

koku/reporting/provider/aws/models.py
```diff
@@ -761,3 +761,17 @@ class Meta:
...
+
+    org_unit_name = models.CharField(max_length=250, null=False, unique=False)
+
+    org_unit_id = models.CharField(max_length=50, null=False, unique=False)
+
+    org_unit_path = models.TextField(null=False, unique=False)
+
+    account_id = models.CharField(max_length=50, null=True, unique=False)
+
+    created_timestamp = models.DateField(null=True)
```
adberglund, line 777: Optional: You can have the timestamp auto set using https://docs.djangoproject.com/en/3.0/ref/models/fields/#django.db.models.DateField.auto_now_add

koku/masu/external/accounts/hierarchy/aws/aws_org_unit_crawler.py
```diff
@@ -0,0 +1,147 @@
...
+        """
+        parent_id = ou.get("Id")
+        LOG.info("Obtaining accounts for organizational unit: %s" % parent_id)
+        child_accounts = self._depaginate(
+            function=self.client.list_accounts_for_parent, resource_key="Accounts", ParentId=parent_id
+        )
+        for act_info in child_accounts:
+            self._save_aws_org_method(ou.get("Name", ou.get("Id")), ou.get("Id"), prefix, act_info.get("Id"))
+
+    def _crawl_org_for_acts(self, ou, prefix):
```
adberglund, line 96: This is silly, but when I can spare to I try to spell out the full word for readability later on. If I see accounts, I know later what I meant. If I see acts, I actually think of the _Acts of the Apostles_, a book in the Bible, so I have to go through another step to figure out it means accounts.

koku/masu/external/accounts/hierarchy/aws/aws_org_unit_crawler.py
adberglund: I think in this case we're okay. Should be relatively lightweight compared to something like report processing

koku/masu/external/accounts/hierarchy/aws/aws_org_unit_crawler.py
myersCody: Nah, thats a really good point. Readability is super important, I think it makes more sense to change it to accounts. 👍

PR #2036, opened 2020-04-07
WIP: Sources: create decorator functions

maskarb:
This PR creates a decorator function which allowed us to remove the `while True` loops from 3 kafka listener functions. With this decorator in place, the message processing functions can be more thoroughly unit tested.


maskarb:
#2044 made these decorators obsolete.

PR #2044, opened 2020-04-08
Sources hardening

dccurtis:
NOTE: This branch includes pending changes from PR https://github.com/project-koku/koku/pull/2031

This is a general hardening PR where the major changes include:
* Removing the `process_messages` Asyncio task which used to place filtered messages on a queue for further processing later.  Instead we are now processing the messages when we get them.
* `AIOKafkaConsumer` is now using manual offset commits.  This gives us the ability to commit a message on a kafka partition only after we have completed processing.  This closes the window where we could go down before processing messages that we have already removed from the message bus.  Any errors related to database, Sources API network, and KafkaErrors are not committed until successful.
* Asyncio exception handler is established to handle exceptions that occur in the process loop.  We are currently catching exceptions related to database, rabbit, and Sources API errors.  Any other errors will be handled by this exception handler.  For now I am simply logging the exception so we will get visibility of the problem in our monitoring tools.  In a subsequent change we can mark the service as dead here and make the OpenShift liveness probe fail in that case to force a pod restart.  For now we will have to manually restart the pod when this occurs.  This is an improvement from today where we can silently fail in some situations.
* Sources PATCH for billing_source and authentication return a 424 if rabbit is down
* The PROCESS_QUEUE is now a PriorityQueue to ensure that messages are re-queued and executed in the correct order.
* Source types other than AWS, Azure, and OpenShift are now filtered.  Issue open against platform for this.


koku/sources/kafka_listener.py
```diff
@@ -315,17 +314,20 @@ def sources_network_info(source_id, auth_header):
...
 
 
-def cost_mgmt_msg_filter(msg):
+def cost_mgmt_msg_filter(msg_data):
     """Verify that message is for cost management."""
-    msg_data = copy.deepcopy(msg)
     event_type = msg_data.get("event_type")
     auth_header = msg_data.get("auth_header")
+
+    if event_type in ((KAFKA_APPLICATION_DESTROY, KAFKA_SOURCE_DESTROY)):
```
maskarb, line 322:
```suggestion
    if event_type in (KAFKA_APPLICATION_DESTROY, KAFKA_SOURCE_DESTROY):
```
A nit-pick

koku/sources/kafka_listener.py
```diff
@@ -319,110 +356,122 @@ async def process_messages(app_type_id, msg_pending_queue):  # noqa: C901; pragm
...
 
     Returns:
         None
 
     """
-    LOG.info("Waiting to process incoming kafka messages...")
-    while True:
-        msg_data = await msg_pending_queue.get()
+    # if msg.get("source_id") == 1:
+    #     raise Exception("TESTING")
```
maskarb, line 369:
```suggestion
```

koku/sources/kafka_listener.py
```diff
@@ -186,8 +192,8 @@ def get_sources_msg_data(msg, app_type_id):
             else:
                 LOG.debug("Other Message: %s", str(msg))
         except (AttributeError, ValueError, TypeError) as error:
-            LOG.error("Unable load message. Error: %s", str(error))
-            raise SourcesIntegrationError("Unable to load message")
+            LOG.error("Unable load message: %s. Error: %s", str(msg.value), str(error))
```
maskarb, line 195:
```suggestion
            LOG.error(f"Unable to load message: {msg.value}. Error: {error}")
```

PR #2050, opened 2020-04-13
Summary table index performance

Red-HAP:
## Description

This branch adds indexes to base summary tables that the materialized views are built from. These indexes will help queries that are more complex run faster as they tend to bypass the materialized views for the data.

The provider maps were changed for a handful of providers to always use icontains vs contains. Part of that is to make the compares consistent as well as reduce the number of indexes needed.

This is passing tox and smoke.


koku/reporting/provider/all/openshift/models.py
```diff
@@ -38,6 +38,8 @@ class Meta:
             GinIndex(fields=["tags"], name="ocpall_tags_idx"),
             models.Index(fields=["product_family"], name="ocpall_product_family_idx"),
             models.Index(fields=["instance_type"], name="ocpall_instance_type_idx"),
+            # A functional index named "ocpall_product_code_ilike" was created manually via RunSQL migration operation
```
adberglund, line 41: Would you mind adding similar comments to the other model files that have functional indexes getting created?

Red-HAP: @adberglund 
Excellent catch! I have added comments.

PR #2062, opened 2020-04-15
Organizations api 

myersCody:
http://127.0.0.1:8000/api/cost-management/v1/organizations/aws/


koku/api/organizations/aws/view.py
```diff
@@ -0,0 +1,31 @@
...
+"""AWS Ogranization views."""
+from api.models import Provider
+from api.organizations.aws.queries import AWSOrgQueryHandler
+from api.organizations.serializers import AWSOrgQueryParamSerializer
+from api.organizations.view import OrganizationView
+
+
+class AWSOrgView(OrganizationView):
+    """AWS Org Base View."""
+
```
maskarb, line 26: Do we need to add permissions here? 

For example, here are the permission to view AWS reports:
[https://github.com/project-koku/koku/blob/9607c1dd895fb027d496359f80e8c0601ff0346f/koku/api/report/aws/view.py#L26-L29](https://github.com/project-koku/koku/blob/9607c1dd895fb027d496359f80e8c0601ff0346f/koku/api/report/aws/view.py#L26-L29)

myersCody: @maskarb So I model the org api after the tag api where it is set in the parent class here: https://github.com/project-koku/koku/pull/2062/files#diff-c78a72dda0d219079a94b76ad01950d5R26

However, if we want to set the permissions in the child view instead of the parent it would be a pretty simple switch. Which do you think is better?

maskarb: 🤷 to be honest, I'm kinda surprised we AllowAny for the tags views.

I think the question is, if someone only has access to see Azure cost reports, why should they be allowed to see org units in AWS? Maybe @chambridge or @adberglund could weigh in here.

adberglund: I think we should discuss this. One option is another RBAC group for org unit. Options I see

1) limit this on org unit via new RBAC group
2) limit this on existing accounts and only show org units for which the user has account permissions 
3) Some combination of the 2?

myersCody: Just as an update, I am gonna jump on a call later with Andrew and will bring this to confirm which route the permissions should go for the org api.

adberglund: Also, keep in mind this is going into a broader feature branch, so we can merge this jsut fine with the AllowAny permission, and discuss and get the issue written for RBAC control.

blentz: My $0.02 is that there's an MVP version and a longer-term vision we can aim for.

The MVP, to me, is that org-unit visibility should be restricted to org-admins.

Longer-term, I think we could attempt to map between org-units and RBAC, or we could provide a UI for customers to tell us the mapping we should do. That way we could enable users within the org to view the bits that they ought to have access to.

This may be more of a platform feature than something we want to do ourselves. In that case, I think we could help drive that feature within the platform.

chambridge: I think option (1) makes the most sense to me from the above comment https://github.com/project-koku/koku/pull/2062#discussion_r408998082.

We can broaden it from here, but I don't want to expose organization information through an implicit access to account. I think the intent is really the other direction where you can see AWS account info based on the org unit you are given permission via RBAC to see.

I always envisioned this feature would add us including a `aws.organization` resource type similar to `aws.account` today. And if we get that we would have to do the mapping of org unit hierarchy to collect all the accounts down the tree that user can view data for.

chargio:
Org units work as bundle accounts, as a bigger version of parent / children account.
From the business perspective, if you see some org unit, you should have access to every account or org unit that is part of it. RBAC should be able to be expressed like that: Org unit 1, and get access to everything below.
Right now, RBAC supports accounts one by one (or using *). Some user that is restricted to the parent account does not see information about the children account. The easiest way for the MVP should be to do the same, where you are asking people to explicitly name Org Units that are visible (and use wildcards if they want to add several of them).
The preferred way for me in the future for RBAC should be to make Org Units the default. You see everything below a branch of the tree, and those accounts that do not use org units should then be considered as having a "root" one, and only that, so you see that account in isolation. When you have configured OU, you then see every account or OU under the tree. If instead of OU, we use some similar component in AWS or other cloud (like GCP folders or LDAP groups), we continue with that behavior. We can call it OU (@dayleparker ), or call it groups or folders, whatever we think is more generic.

PR #2069, opened 2020-04-16
Slice values for tags API endpoint

adberglund:
## Summary
This is a temporary fix to limit the amount of value data coming from our tags APIs. Pagination only effects the number of keys, so value lists can grow indefinitely. There's a current situation in production where this API is causing the UI to come to a halt.


koku/api/tags/queries.py
```diff
@@ -107,10 +107,17 @@ def _format_query_response(self):
 
         """
         output = copy.deepcopy(self.parameters.parameters)
+        if not self.parameters.parameters.get("key_only"):
+            self._slice_tag_values_list()
         output["data"] = self.query_data
 
         return output
 
+    def _slice_tag_values_list(self, n=49):
```
adberglund, line 116:
```suggestion
    def _slice_tag_values_list(self, n=50):
```

Issue #2002, opened 2020-04-02
Unused SIUnitScale


nbon12:
As a developer I want the codebase to not have unused variables for the purpose of maintainability and clarity.

The following Model does not seem to be used anywhere in the codebase, this JSON is also unused, and not loaded anywhere.

Investigate if the following model and JSON is needed:

- https://github.com/project-koku/koku/blob/master/koku/reporting_common/models.py#L85
- https://github.com/project-koku/koku/blob/master/koku/reporting_common/data/si_scale.json

![Screen Shot 2020-04-02 at 12.04.59 PM.png](https://images.zenhubusercontent.com/5d66e2814877f8000104a00b/1584720c-7d32-441e-a97f-980e8b568dbd)

Acceptance Criteria:

- [ ] Delete the model and the Object if it's not needed. If it is needed, leave it alone.

PR #2077, opened 2020-04-17
Delete SIUnitScale Model

nbon12:
Addresses #2002 

This PR deletes the unused SIUnitScale model and JSON document.

Functional Testing - There should be no change to the API responses.


maskarb:
Testing:

1. remove the DB
2. reinitialize the DB
3. `make run-migrations` is successful.

PR #2089, opened 2020-04-21
Use github-actions alongside Travis

maskarb:
This PR puts in place a few github-actions: 1. unit tests, 2. check manifest, 3. pre-commit. 

Some benefits:
1. Moving to github-actions would enable us to drop Travis. It's one less third party application to rely on
2. pre-commit, manifest, and unittests run in separate workflows so it's more obvious when one of these fails
3. We can cache dependencies so we can save a smidge of time during tests.

Some negatives:
1. The UI is not quite as polished as Travis. Watching the logs in real-time can be kinda aggravating. Sometimes you can click a drop down and nothing happens, so it's best to wait for the run to finish before looking.
2. If one Python version fails, we must run both versions again (unlike Travis where you can selectively rebuild a run).
3. Codecov (big shocker). When I was running this PR in my fork, codecov was struggling to report. But, that could also be related to the settings of my fork 🤷 

The `koku_test_runner.py` modifications were made to suppress `print` statements from Nise. These statements caused `Broken pipe` exceptions while the db populator is running.

To see the pre-commit and manifest check workflows in action, see this PR: [https://github.com/maskarb/koku/pull/2](https://github.com/maskarb/koku/pull/2 )


dccurtis:
oh the irony!  Waiting on Travis ⏳

dccurtis:
Thanks for taking the initiative to check into this and great pro/con summary.  In my opinion the cons are worth the tradeoff of having a more reliable PR sanity check process.

I think it's worth giving this a try to see how it works out, knowing we could always switch back once we get more experience with it.

Would be good to get other's opinions on this.

nbon12:
I agree with Doug.

blentz:
I haven't really felt much pain with Travis. So, I'm not sure I see a big need to change it. 

That said, if there's a majority consensus on making the change, I don't really have any objections to it.

The only thing in the diff that I can see that we'd need to change is that `ubuntu-latest` isn't our target platform. We need this to test against RHEL. So, if it can't use RHEL (or at the very least, CentOS or Fedora - something that feeds into the RHEL pipeline) , I don't think we can switch to it.

maskarb:
@blentz 

I understand wanting to use RHEL, but we don't use it in Travis either because it's not a supported environment (we use bionic there, so I should change this PR to Ubuntu-18.04 instead of 16.04). This PR is mostly a like-for-like change. We'll just get all of our test results in Github instead of Travis.

blentz:
@maskarb  In the docs, I see support for "Self-hosted runners". While Github itself only supports Ubuntu, we can stand up our own build hosts. And RHEL is in the list of supported OSes for self-hosting. (See: https://help.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners#supported-operating-systems-for-self-hosted-runners )

That's a Pro and is something that I'm not sure Travis supports. So, for me, that a benefit that makes doing this change worthwhile. We can stand up our own runner as a follow-on task to making this change.

Issue #2075, opened 2020-04-17
Source status unavailable error messages should be actionable by a user to make their source available.


dccurtis:
## User Story
As user, I want to have useful, actionable error messages displayed in the UI so I understand why my source is not available.

## Impacts
API, UI, Docs

## API Details

Source Status is pushed to the platform in the following situations:
* The Sources client provides a `source-status` endpoint which the platform calls periodically to get source status.
* Source create and edits

`SourcesHTTPClient().set_source_status()` is used to push the status to the platform

## UI Details
No changes needed to the UI.  Ensuring that we are pushing appropriate user facing strings to the platform is all that's required.

## Acceptance Criteria

- [ ] All possible source status strings are known and have been reviewed by a Docs engineer 
- [ ] Functional tests have been run to demonstrate that all status strings are populated in the platform sources database for source create, update, and source-status API call

adberglund:
@dccurtis Do you want to add a link to the message spreadsheet here?

dccurtis:
Sheet used to review messages with docs: https://docs.google.com/spreadsheets/d/1BgweFAhU3RAU30Bj1rzYQcVxwOBWS1WUBYyjlvD-bFY/edit?usp=sharing

lcouzens:
Verified

PR #2095, opened 2020-04-23
Sources status cleanup

dccurtis:
Addresses #2075 

The goal is to continue to log the same error messages but have the capability to control what error messages are pushed to platform sources since they will be visible in the UI

Thanks for the help @dayleparker and @nlcwong on this!

**Testing**
1.  Add AWS source with missing bucket.
2. Add wrong bucket
3. Add wrong AWS role ARN.
4. Add Azure source without subscription id, resource group, and storage account
5.  Add Azure subscription ID to source from step 4.
6. Add only Azure resource group and storage account to an Azure source.
7. Create Azure Source with incorrect subscription id.
8. Add Azure source with incorrect storage account.
9. Add Azure source with incorrect resource group.
10. Add Azure source with incorrect client ID.
11. Add Azure account with incorrect client secret.
12.  Add an Azure account for an account where the cost management export has not been created yet.

**Testing Results**

[sources_msg_cleanup_ut.txt](https://github.com/project-koku/koku/files/4541388/sources_msg_cleanup_ut.txt)


koku/sources/sources_error_message.py
```diff
@@ -0,0 +1,90 @@
...
+    """Sources Error Message for Sources API service."""
+
+    def __init__(self, error):
+        """Initialize the message generator."""
+        self._error = error
+
+    def azure_client_errors(self, message):
+        """Azure client error messages."""
+        scrubbed_message = ProviderErrors.AZURE_GENERAL_CLIENT_ERROR_MESSAGE
+        if "http error: 401" in message:
```
dccurtis, line 37: Yes, we should figure out if there are specific Azure exceptions that are thrown for all of these rather than the broad `AzureException` so we don't have to do this error-prone string examining.  I'm looking into it.

maskarb:
Do we need to do something with this message too?

[https://github.com/project-koku/koku/pull/2095/files#diff-8479cce197116d004af40fe1d10a95a0R117-R120](https://github.com/project-koku/koku/pull/2095/files#diff-8479cce197116d004af40fe1d10a95a0R117-R120)

dccurtis:
Yes good reminder that it was merged.  I'll update

dccurtis:
here's the example with the new compression format change. @dayleparker approved ✅ 

```{
    "meta": {
        "count": 1,
        "limit": 100,
        "offset": 0
    },
    "links": {
        "first": "/api/v1.0/applications/?offset=0",
        "last": "/api/v1.0/applications/?offset=0"
    },
    "data": [
        {
            "application_type_id": "2",
            "availability_status": "unavailable",
            "availability_status_error": "Cost management requires that AWS Cost and Usage Reports use GZIP compression format.",
            "created_at": "2020-04-29T14:52:59Z",
            "id": "1",
            "source_id": "1",
            "updated_at": "2020-04-29T14:53:09Z",
            "tenant": "10001"
        }
    ]
}
```

PR #2097, opened 2020-04-23
add RBAC classes to Tags endpoint

blentz:
This adds the existing RBAC permissions classes to the Tags API views.

Functional testing:

CLI for user with permissions `{"aws.account": {"read": ["*"]}}"`:
```
$ curl http://koku-hccm.apps-crc.testing/api/cost-management/v1/tags/aws/
{"meta":{"count":4,"key_only":false,"filter":{"time_scope_value":"-10","time_scope_units":"day","resolution":"daily"},"group_by":{},"order_by":{}},"links":{"first":"/api/cost-management/v1/tags/aws/?limit=100&offset=0","next":null,"previous":null,"last":"/api/cost-management/v1/tags/aws/?limit=100&offset=0"},"data":[{"key":"version","values":["prod","master","gamma","beta"]},{"key":"storageclass","values":["gamma","epsilon","delta","charlie","bravo"]},{"key":"environment","values":["dev"]},{"key":"app","values":["cost","catalog","analytics"]}]} 

$ curl http://koku-hccm.apps-crc.testing/api/cost-management/v1/tags/azure/
{"errors":[{"detail":"You do not have permission to perform this action.","source":"detail","status":403}]}
```


chambridge:
@blentz Was the following code already filtering the users results?
https://github.com/project-koku/koku/blob/163b0a80a0400dd0ca5fdc53dfb1d9d62775c395/koku/api/query_params.py#L378-L392

For example a user only had access to one aws account but there were 30 aws accounts, they only see the tags associated with that account.

blentz:
@chambridge I'm not sure. I'll try to test it and see.

blentz:
With some help from @adberglund , we've validated that the code mentioned by @chambridge *is* successfully filtering results for the tag enpoint, too.

PR #2100, opened 2020-04-24
1998 db vacuum options

Red-HAP:
## Vacuum options

Instead of doing an arbitrary VACUUM FULL, these changes attempt to alter the autovacuum options on large tables. There's a scale involved here.

If these changes bear out, then the PG startup options should probably be added to production (Another reboot event)

## Description

This change sets a task that will run periodically to tune the autovacuum_vacuum_scale_factor setting on tables that meet a certain size threshold. This threshold is set as a table and will be evaluated from the largest threshold value to the smallest. Each threshold value in the table will have an associated scale factor value. These values should be developed with the intent of having the autovacuum process run more frequently on larger tables that likely will see more write traffic.

Currently, the vacuum processes can run for long periods of time which will attempt to process a table's dead tuples and keep things humming as smoothly as it can. The issue is that when there are a large number of tuples to process or if the table is getting a lot of write traffic, the autovacuum process can take an inordinate amount of time to complete. The current means of handling this was to periodically vacuum an entire schema. This can also run for an inordinate amount of time. The purpose of this change is to make vacuuming more efficient and somewhat automatically reactive.

The tuning table can be set with the `AUTOVACUUM_TUNING` environment variable which should be a JSON array in the form of:

```
[[live-tuple-threshold, autovacuum-scale-factor], ...]
```

This setting will completely overwrite the default table.


koku/masu/processor/tasks.py
```diff
@@ -372,3 +373,79 @@ def vacuum_schema(schema_name):
...
+    xlrg_scale = Decimal("0.01")
+    lrg_scale = Decimal("0.02")
+    med_scale = Decimal("0.05")
+    no_scale = Decimal("100")
+    zero = Decimal("0")
+    reset = Decimal("-1")
+    scale_factor = zero
+    xlrg_table = 10000000
+    lrg_table = 1000000
+    med_table = 100000
```
blentz, line 416: Does it make sense to expose any of these as tunable during runtime?  (e.g. as an environment variable) I think these are probably the knobs we'll want to fine-tune over time. So, exposing them would make that easier.

To keep it manageable, we could pass JSON as an environment variable, that way we can expose all of these settings with a single variable.

Red-HAP: That's a good idea. If we can modify the env at runtime then, sure.

blentz: We can. If the data is exposed as an environment variable, then the DeploymentConfig sets the variable. So, we can change it at the DeploymentConfig level, which redeploys a new pod with the updated environment variables. It's a rolling update, so there's no downtime. But it means service operators can fine-tune the values without needing a new release.

koku/masu/processor/tasks.py
```diff
@@ -372,3 +374,94 @@ def vacuum_schema(schema_name):
...
+    # initialize settings
+    scale_table = [(10000000, Decimal("0.01")), (1000000, Decimal("0.02")), (100000, Decimal("0.05"))]
+    alter_count = 0
+    no_scale = Decimal("100")
+    zero = Decimal("0")
+    reset = Decimal("-1")
+    scale_factor = zero
+
+    # override with environment
+    # This environment vaiable's data will be a JSON string in the form of:
```
blentz, line 416:
```suggestion
    # This environment variable's data will be a JSON string in the form of:
```

Issue #2107, opened 2020-04-27
Improve testability of RBAC


blentz:
## User Story
As a koku developer, I want to be able to define non-admin test users in my local dev environment so that I can test RBAC changes without requiring deploying the platform's RBAC service.

### Background

While working on #2064, I found that we don't currently have an easy way to define a test user with a specific set of permissions.

For development, the dev_middleware defines an org-admin with unrestricted access.
For production, the middleware expects to contact an external RBAC service.

We don't really have anything for testing. There's no way to say, "Create test user Joe with read-only AWS access, but read-write Azure access," and then use that test user in either functional or unit tests.

### Related issues

This PR documents workarounds that were required to test that change:
https://github.com/project-koku/koku/pull/1802#issue-378509474

The goal of this work is to eliminate the need to hack apart the RBAC code like this PR needed to do.

## Impacts
API functional and unit tests

## Role
Koku developer

## Assumptions
- The dev_middleware all-access user is not changed. We still need a "root" account for general development use.
- The RBAC middleware for production use shouldn't need to be changed. The primary RBAC integration should continue to work. 

## Acceptance Criteria

- [ ] There is a way to define a test user with custom permissions for functional testing
- [ ] There is a way to define a test user with custom permissions for use during unit tests

PR #2113, opened 2020-04-29
enable custom identity header (see: #2107)

blentz:
(See: #2107) This PR adds the ability to inject a custom Identity header using the `DEVELOPMENT_IDENTITY` environment variable.

This enables developers to develop and test with custom identity headers to facilitate seeing different API responses based on differing permissions.

Examples:

org_admin user:
```
[2020-05-05 19:32:57,810] INFO None Identity: {"identity": {"account_number": "10001", "type": "User", "user": {"username": "user_dev", "email": "user_dev@foo.com", "access": {"aws.account": {"read": ["111111111111"]}}}}, "entitlements": {"cost_management": {"is_entitled": "True"}}}
[2020-05-05 19:32:57,846] INFO None {'method': 'GET', 'path': '/api/cost-management/v1/status/', 'status': 200, 'request_id': None, 'account': None, 'username': None, 'is_admin': False}
10.128.0.1 - - [05/May/2020:19:32:57 +0000] "GET /api/cost-management/v1/status/ HTTP/1.1" 200 3873 "-" "kube-probe/1.16+"
[2020-05-05 19:33:01,819] INFO None Identity: {"identity": {"account_number": "10001", "type": "User", "user": {"username": "user_dev", "email": "user_dev@foo.com", "access": {"aws.account": {"read": ["111111111111"]}}}}, "entitlements": {"cost_management": {"is_entitled": "True"}}}
[2020-05-05 19:33:01,820] INFO None {'method': 'GET', 'path': '/api/cost-management/v1/tags/aws/', 'request_id': None, 'account': '10001', 'username': 'user_dev', 'is_admin': None}
[2020-05-05 19:33:01,979] INFO None {'method': 'GET', 'path': '/api/cost-management/v1/tags/aws/', 'status': 200, 'request_id': None, 'account': '10001', 'username': 'user_dev', 'is_admin': None}
10.128.0.1 - - [05/May/2020:19:33:01 +0000] "GET /api/cost-management/v1/tags/aws/ HTTP/1.1" 200 625 "-" "curl/7.66.0"
```

limited access user succeeds where they have access:
```
[2020-05-05 19:37:25,100] INFO None Identity: {"identity": {"account_number": "10001", "type": "User", "user": {"username": "user_dev", "email": "user_dev@foo.com", "access": {"aws.account": {"read": ["111111111111"]}}}}, "entitlements": {"cost_management": {"is_entitled": "True"}}}
[2020-05-05 19:37:25,100] INFO None {'method': 'GET', 'path': '/api/cost-management/v1/tags/aws/', 'request_id': None, 'account': '10001', 'username': 'user_dev', 'is_admin': None}
[2020-05-05 19:37:25,118] WARNING None DEVELOPMENT is Enabled. Bypassing access lookup for user: {'identity': {'account_number': '10001', 'type': 'User', 'user': {'username': 'user_dev', 'email': 'user_dev@foo.com', 'access': {'aws.account': {'read': ['111111111111']}}}}, 'entitlements': {'cost_management': {'is_entitled': 'True'}}}
[2020-05-05 19:37:25,147] INFO None {'method': 'GET', 'path': '/api/cost-management/v1/tags/aws/', 'status': 200, 'request_id': None, 'account': '10001', 'username': 'user_dev', 'is_admin': None}
10.128.0.1 - - [05/May/2020:19:37:25 +0000] "GET /api/cost-management/v1/tags/aws/ HTTP/1.1" 200 625 "-" "curl/7.66.0"
```

limited access user gets permission denied:
```
[2020-05-05 19:38:09,810] INFO None Identity: {"identity": {"account_number": "10001", "type": "User", "user": {"username": "user_dev", "email": "user_dev@foo.com", "access": {"aws.account": {"read": ["111111111111"]}}}}, "entitlements": {"cost_management": {"is_entitled": "True"}}}
[2020-05-05 19:38:09,811] INFO None {'method': 'GET', 'path': '/api/cost-management/v1/tags/azure/', 'request_id': None, 'account': '10001', 'username': 'user_dev', 'is_admin': None}
[2020-05-05 19:38:09,834] WARNING None DEVELOPMENT is Enabled. Bypassing access lookup for user: {'identity': {'account_number': '10001', 'type': 'User', 'user': {'username': 'user_dev', 'email': 'user_dev@foo.com', 'access': {'aws.account': {'read': ['111111111111']}}}}, 'entitlements': {'cost_management': {'is_entitled': 'True'}}}
[2020-05-05 19:38:09,866] INFO None {'method': 'GET', 'path': '/api/cost-management/v1/tags/azure/', 'status': 403, 'request_id': None, 'account': '10001', 'username': 'user_dev', 'is_admin': None}
[2020-05-05 19:38:09,867] WARNING None Forbidden: /api/cost-management/v1/tags/azure/
10.128.0.1 - - [05/May/2020:19:38:09 +0000] "GET /api/cost-management/v1/tags/azure/ HTTP/1.1" 403 107 "-" "curl/7.66.0"
```


koku/koku/dev_middleware.py
```diff
@@ -36,14 +42,31 @@ def process_request(self, request):  # pylint: disable=no-self-use
...
+                request.user = user
+            else:
+                identity_header = {
+                    "identity": {
+                        "account_number": "10001",
+                        "type": "User",
+                        "user": {"username": "user_dev", "email": "user_dev@foo.com", "is_org_admin": True},
+                    },
+                    "entitlements": {"cost_management": {"is_entitled": True}},
+                }
```
Red-HAP, line 68: Just a question here, but do we _not_ want to mock a user for `request.user` here?

blentz: I'm inserting new behavior while leaving the previous behavior unchanged. My main reasoning is just that I don't know if my chosen approach is a good one. If it is a good idea, I can certainly tighten up the logic.

It feels like a bit of a kludge to me to be using Mock outside of the unit tests. I needed an object that would take a dict as input and expose the dict's keys as class attributes. `unittest.Mock` was a solution that works a little better than `collections.UserDict` and is less work than sub-classing `dict`.

I see Mock as "not the best choice, but an expedient choice to show a working example of the idea."

Red-HAP: You could always steal `dicta` from here: https://github.com/project-koku/nise/blob/9347f801a5d6799ff8795f76c482704a5f63cc84/utility/generate_static_ocp_settings.py#L50

koku/koku/settings.py
```diff
@@ -143,6 +143,7 @@
 
 DEVELOPMENT = ENVIRONMENT.bool("DEVELOPMENT", default=False)
 if DEVELOPMENT:
+    DEVELOPMENT_IDENTITY = ENVIRONMENT.json("DEVELOPMENT_IDENTITY", default={})
```
chambridge, line 146: What happens if we hit the default?
Do we the default to be https://github.com/project-koku/koku/pull/2113/files#diff-86a0cdc27a1ea2f22801c8f80dbaea42R62-R68
and then we don't have an if/else?

blentz: There's two if/else conditions, one in `settings.py` and one in `dev_middleware.py`

In `settings.py`, the logic is, "if `DEVELOPMENT` is set, look for the `DEVELOPMENT_IDENTITY`. This way, there's no chance to sneak a forged identity into the app without setting the `DEVELOPMENT` env var.

In `dev_middleware.py`, the logic is, "if `DEVELOPMENT_IDENTITY` is True (empty dicts are False), then use the forged identity."

Does this better explain the behavior?  Do you want it to do something else, instead?

chambridge: Sorry. Maybe I wasn't clear if some one didn't update their `.env` file with the `DEVELOPMENT_IDENTITY` you showed the `.env.example` file the `ENVIRONMENT.json()` method will take `{}` as the default. Feels like our code would break. 

AIso with this change I don't think we can make it down the `else` path in `dev_middleware.py`, since you always set a value when `DEVELOPMENT` is set.

So I was suggesting why not make `default={}` --> `default=[what's in the .env.example]` then you can delete the else condition from the `dev_middleware.py` Additionally no one needs to update their `.env` unless they want to take advantage of the RBAC override feature.

blentz: I believe I addressed that in my response. In Python, empty dicts == False.

For example:
```
ipython
/usr/lib/python3.7/site-packages/IPython/core/interactiveshell.py:925: UserWarning: Attempting to work in a virtualenv. If you encounter problems, please install IPython inside the virtualenv.
  warn("Attempting to work in a virtualenv. If you encounter problems, please "
Python 3.7.7 (default, Mar 13 2020, 10:23:39) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.11.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: bool({})                                                                                                                                            
Out[1]: False
```

So, if we take the default value of `{}` in from the first conditional, it is a False value for the second conditional, which means we revert to the old behavior.

Am I missing something there?

chambridge: My apologies @blentz , too much multi-tasking and I read right past the "(empty dicts are False)" without comprehending.

To your point:

> So, what I need here is less about the exact mechanics of how this particular code can be made better and more an answer to the question of should we do it this way at all?

Yes, I'm in favor of this as a mechanism.

PR #2115, opened 2020-04-30
Allowing for OCP single manifest, multiple payloads

dccurtis:
When ingesting data from a large OCP cluster, the payloads need to be split in order to be within the ingress service file size limit.

Changes to korekuta that need to be made are to use a common manifest across multiple payloads.  That way when the summarization tasks are kicked off it won't summarize until all line item report data has completed processing.

**Testing**
1. Create and ingest multi-file OCP payload.
2. Verify processing is successful.
3. Repeat steps 1-3 but for local OCP ingest.  Verify API results match.

**Test Results**
[ocp_multi_payload_ut.txt](https://github.com/project-koku/koku/files/4566196/ocp_multi_payload_ut.txt)

Smoke Tests Passed

**Testing Failure Path**
Test should focus on ensuring that logic can handle when a report processing fails in the middle of processing a manifest.  The expected behavior is that failures should gracefully update stats, report to ingress service that download was successful and API should not change.

Note: A subsequent change could be considered to do manual kafka commits (like sources-client) to where we would re-download/process report failures.  Since we get reports every 45 minutes I think for now it's ok to make sure that subsequent payloads process successfully.

1. Generate/Ingest reports.  Sample API response
2. Re-generate/upload same yaml file but bring down db after second file processing last_completed_datetime
3. Bring DB up, Re-generate/upload report.  Verify that processing was successful and API matches #1

**Test Results**
[ocp_multi_error_path_ut.txt](https://github.com/project-koku/koku/files/4576205/ocp_multi_error_path_ut.txt)

### DIFF
--- a/koku/masu/processor/_tasks/process.py
+++ b/koku/masu/processor/_tasks/process.py
@@ -19,7 +19,6 @@ from os import path
 
 import psutil
 from celery.utils.log import get_task_logger
-from django.db import transaction
 
 from masu.database.provider_db_accessor import ProviderDBAccessor
 from masu.database.report_manifest_db_accessor import ReportManifestDBAccessor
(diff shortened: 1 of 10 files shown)


dccurtis:
Corresponding nise change: https://github.com/project-koku/nise/pull/225/files

koku/masu/processor/_tasks/process.py
```diff
@@ -76,22 +75,21 @@ def _process_report_file(schema_name, provider, provider_uuid, report_dict):
...
-            if manifest:
-                manifest.num_processed_files += 1
-                manifest.save()
-                manifest_accesor.mark_manifest_as_updated(manifest)
-            else:
-                LOG.error("Unable to find manifest for ID: %s, file %s", manifest_id, file_name)
+    with ReportManifestDBAccessor() as manifest_accesor:
+        manifest = manifest_accesor.get_manifest_by_id(manifest_id)
+        if manifest:
+            manifest.num_processed_files += 1
```
maskarb, line 85: Since you're here, I think you can knock out #2118 while you're at it 😄 

Instead of incrementing, I'm pretty sure we just need to actually count the number of files that have been processed, like this:
```suggestion
            manifest.num_processed_files = CostUsageReportStatus.objects.filter(manifest_id=manifest_id, last_completed_datetime__isnull=False).count()
```

koku/masu/processor/tasks.py
```diff
@@ -85,59 +86,64 @@ def get_report_files(
...
-            # Skip processing if already in progress.
-            if started_date and not completed_date:
-                expired_start_date = started_date + datetime.timedelta(hours=Config.REPORT_PROCESSING_TIMEOUT_HOURS)
-                if DateAccessor().today_with_timezone("UTC") < expired_start_date:
-                    LOG.info(
-                        "Skipping processing task for %s since it was started at: %s.", file_name, str(started_date)
-                    )
-                    continue
-
+    with transaction.atomic():
```
maskarb, line 89: I think I'm regretting agreeing to this idea. If I'm understanding this idea correctly, we are processing the _entire_ manifest within an atomic block. Let's say we had a particularly large manifest, say, 50 files, each ridiculously huge. And then, let's suppose we lost our DB connection whilst processing the 50th file, raising a `cursor already closed` error. Wouldn't this atomic block mean that we would have to process every file a second time?

dccurtis: yea thats a good point.  Maybe we should move the atomic block to be after  `for report_dict in reports:`.  That way we would only have to re-process the report (file) that failed.

PR #2124, opened 2020-05-06
Remove TravisCI

maskarb:
1. Add badge for the Unit Test action
2. Remove TravisCI


abaiken:
This LGTM except we might want to keep the script to check the manifest (just move its location) and update the make command that calls it.

maskarb:
whoops, forgot about the Makefile. Good catch!

Issue #2125, opened 2020-05-07
Source status pushes are failing due to RBAC.


dccurtis:
**Describe the bug**
Source status is pushed on two different paths.
1. When a source create/edit happens.
2. Responding to source status update requests by the platform's availability checker.

For platform source availability status requests we are currently using the request header that platform-sources is using to call us (internal API).

The request header that the platform is using to call us for source status does not include the `is_org_admin` value so all of our status pushes are failing.

Production example:
```
[2020-05-06 20:19:25,603] INFO None "POST /api/cost-management/v1/source-status HTTP/1.1" 204 0
[2020-05-06 20:19:25,765] WARNING None Unable to assume role with ARN arn:aws:iam::12345:role/CostManagement.
Traceback (most recent call last):
  File "/opt/app-root/src/koku/providers/aws/provider.py", line 51, in _get_sts_access
    assumed_role = sts_client.assume_role(RoleArn=provider_resource_name, RoleSessionName="AccountCreationSession")
  File "/opt/app-root/lib/python3.6/site-packages/botocore/client.py", line 316, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/opt/app-root/lib/python3.6/site-packages/botocore/client.py", line 626, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.exceptions.ClientError: An error occurred (AccessDenied) when calling the AssumeRole operation: User: arn:aws:iam::9999999:user/cmaas-dev is not authorized to perform: sts:AssumeRole on resource: arn:aws:iam::12345:role/CostManagement
[2020-05-06 20:19:25,896] WARNING None Unable to push source status. Reason: Unable to set status for Source 86972. Reason: Status code: 401. Response: {"errors":[{"status":401,"detail":"Unauthorized"}]}.
```

**To Reproduce**
1. Create a source in CI or production.
2. Wait for platform to call internal availability status endpoint and observe the failure to push status

**Expected behavior**
All internal status pushes should be successful.

**Additional context**
A discussion is on going with the platform sources team around this.  For now I want to fix the status endpoint, but I can imagine changes for all internal communication coming once we figure out the right direction.

lcouzens:
Verified

PR #2126, opened 2020-05-07
Generate org_admin authorization header for source status pushes.

dccurtis:
Addresses #2125 

See issue for background on this as well as recent discussion on the sources slack channel.  For now I'm going to fix just the availability checker status push path.  The platform is open to changes around how services communicate internally to address concerns around relying on RBAC status of the user that created the source for handling source-application events.

**Testing**
1. Create AWS source with bad bucket
2. Request status from our internal endpoint with a similar header to what the platform is using.
3. Verify status was received successfully in platform sources.

**Test Results**

[source_status_header_ut.txt](https://github.com/project-koku/koku/files/4593706/source_status_header_ut.txt)


koku/sources/sources_http_client.py
```diff
@@ -243,23 +253,48 @@ def build_source_status(self, error_obj):
...
+
+            return {"x-rh-identity": cost_internal_header}
+        except (binascii.Error, json.JSONDecodeError, TypeError, KeyError) as error:
+            LOG.error(f"Unable to build internal status header. Error: {str(error)}")
+
     def set_source_status(self, error_msg, cost_management_type_id=None):
         """Set the source status with error message."""
+        status_header = self.build_status_header()
+        if not status_header:
+            return False
```
maskarb, line 281: Do we not want to raise `SourcesHTTPClientError` here?

dccurtis: Yeah.. I thought about that.  I didn't want it to continuously retry since this could be thought of as a "best effort" status push.

maskarb: Sounds good to me!

PR #2131, opened 2020-05-08
add RbacPermissions decorator

blentz:
This is the second half of #2107 . This adds the `@RbacPermissions` decorator class. This decorator allows us to inject custom permission sets into the `User` portion of a `Request` object. This is useful for unit testing View endpoints for results filtered by access controls.

### DIFF
--- a/koku/koku/dev_middleware.py
+++ b/koku/koku/dev_middleware.py
@@ -67,7 +67,7 @@ class DevelopmentIdentityHeaderMiddleware(MiddlewareMixin):  # pylint: disable=t
             }
 
         """
-        if hasattr(request, "META") and hasattr(settings, "DEVELOPMENT_IDENTITY"):
+        if hasattr(request, "META") and (hasattr(settings, "DEVELOPMENT_IDENTITY") and settings.DEVELOPMENT_IDENTITY):
             identity_header = settings.DEVELOPMENT_IDENTITY
 
             user_dict = identity_header.get("identity", {}).get("user")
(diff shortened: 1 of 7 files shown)


koku/koku/dev_middleware.py
```diff
@@ -67,7 +67,7 @@ def process_request(self, request):  # pylint: disable=no-self-use
             }
 
         """
-        if hasattr(request, "META") and settings.DEVELOPMENT_IDENTITY:
+        if hasattr(request, "META") and hasattr(settings, "DEVELOPMENT_IDENTITY"):
             identity_header = settings.DEVELOPMENT_IDENTITY
```
Red-HAP, line 71: This change allows for a falsey DEVELOPMENT_IDENTITY which the former code was apparently checking. Does this case still need to be handled?

koku/koku/dev_middleware.py
```diff
@@ -67,7 +67,7 @@ def process_request(self, request):  # pylint: disable=no-self-use
             }
 
         """
-        if hasattr(request, "META") and settings.DEVELOPMENT_IDENTITY:
+        if hasattr(request, "META") and hasattr(settings, "DEVELOPMENT_IDENTITY"):
```
blentz, line 70:
```suggestion
        if hasattr(request, "META") and (hasattr(settings, "DEVELOPMENT_IDENTITY") and settings.DEVELOPMENT_IDENTITY):
```

koku/koku/dev_middleware.py
blentz: Fixed! Nice catch! ^_^

Issue #2043, opened 2020-04-08
AWS Organizational unit crawler should also record delete timestamps


kholdaway:
## User Story
The original implementation of the AWS organizational unit crawler does not account for a deleted account/organizational unit.  There are cases where knowing this could be useful.  
- API to query organization units/accounts for a time period will include deleted nodes
- Rolling off old data could also determine certain tree data is no longer needed.

## Impacts
AWS Org Unit crawler celery beat.  AWS Organizational Unit table will need to record a deleted timestamp.  The crawler will need to compute the org unit tree for the previous day, recurse the current tree, and note which org units/accounts are not seen today. It should then update the database to reflect the deletion of nodes. 

### Sample Table
| id | org_unit_name | org_unit_id|path.to.org_unit_name | account_id | created_timestamp |deleted_timestamp|
 | --- | --- | --- | --- | --- |--- |--- |
| 1 | root |24332| root | NULL | 2020.03.01T00:00:00Z |NULL|
| 2 | root |24332| root | account_1 | 2020.03.01T00:00:00Z |NULL|
| 3 | sub_org_unit_1 | 25332 | root.sub_org_unit_1 | NULL | 2020.03.01T00:00:00Z |NULL|
| 4 | sub_org_unit_1 | 25332 | root.sub_org_unit_1 | 12345 | 2020.03.01T00:00:00Z |NULL|
| 6 | sub_org_unit_1 | 25332 | root.sub_org_unit_1 | 12346 | 2020.03.01T00:00:00Z |NULL|
| 7 | sub_org_unit_2 | 26332 | root.sub_org_unit_2 | NULL | 2020.03.01T00:00:00Z |NULL|
| 8 | sub_org_unit_2 | 26332 | root.sub_org_unit_2 | 12345 | 2020.03.02T00:00:00Z |NULL|
| 9 | sub_org_unit_2 | 26332 | root.sub_org_unit_2 | 12345 | 2020.03.02T00:00:00Z |NULL|
| 3 | sub_org_unit_1 | 26332 | root.sub_org_unit_2 | NULL | 2020.03.01T00:00:00Z  |2020.03.02T00:00:00Z|

**Note: the last entry in the table is an update not new insert**
## Acceptance Criteria

- [ ] Deleted org unit/accounts are marked with a deleted_timestamp

Issue #1917, opened 2020-03-18
Add group by and filter option for org unit to AWS, OCP on AWS reports options


adberglund:
## User Story
As a user viewing AWS data I want to be able to view costs by organizational unit so I can understand the cost associated with each unit (e.g. a cost center or app team, et cetera).

## Impacts
- API

## Assumptions
- We need to update serializers to allow grouping by and filtering by organizational unit
- We'll need to support a rollup feature where we roll up cost based on a path "icontains" the org unit name and a non-rolled up where the org unit name is searched. (See https://github.com/project-koku/koku/issues/1894 for DB table design)

## API Details

| API  | Detail|
 | --- | --- |
 | **Method** | GET |
 | **Path** | /api/v1/reports/aws/*/ * |
 | **Path** | /api/v1/reports/openshift/infrastructures/aws/*/ |
 | **Body** |  |
 | **Response** | |

## Acceptance Criteria

- [ ] AWS reports can be grouped by specific org unit 
- [ ] AWS reports can be grouped by a rollup of all accounts/org units under an  org unit

Issue #1916, opened 2020-03-18
Add AWS organizations API


adberglund:
## User Story
As a user viewing AWS data I want to be able to view costs by organizational unit so I can understand the cost associated with each unit (e.g. a cost center or app team, et cetera)

## Impacts
- API

## Assumptions
- The AWS organization unit table exists and we can pull data from it (can mock data while initially building this out)
- See table structure detailed in https://github.com/project-koku/koku/issues/1894

## API Details

| API  | Detail|
 | --- | --- |
 | **Method** | GET/ |
 | **Path** | /api/cost-management/v1/organizations/aws/ |
 | **Body** |  |
 | **Response** | |

## API Details
- Return the organization name
- Return the path to the org
- Return a list of current accounts for that org unit

## Acceptance Criteria

- [ ] API returns required fields

Issue #1915, opened 2020-03-18
Correlate AWS costs with Organization Unit


adberglund:
## User Story
As a user viewing AWS data I want to be able to view costs by organizational unit so I can understand the cost associated with each unit (e.g. a cost center or app team, et cetera).

## Impacts
- Worker

## Assumptions
- We have already processed and stored AWS organizational units in our database
- We will need to add a foreign key to the org unit table in our AWS and OCP on AWS daily summary tables.
- The worker daily summary SQL will need an update to add the foreign key relationship for org units

## Acceptance Criteria

- [ ] AWS daily summary data has the correct link to an org unit the account is associated with

nbon12:
Sprint Planning Meeting Minutes:
Andrew: "These set of issues around AWS organizational units fall out of the realm of feature freeze. This is an internal Red Hat ask. Red Hat is in the process of renegotiating entire AWS contract organization-wide. Through negotiating what they do, Red Hat has to reduce the number of payer accounts. In order to maintain visibility they're using AWS organization units. There's a mandate we ingest organization data in Cost Management, and provide access to that data to group/slice cost data by org unit. Cody and Kevan are working on ingesting that data. This issue is the next step to correlate AWS costs with org units."

PR #2139, opened 2020-05-12
AWS org unit feature 

abaiken:
Connected issues: 
* Add AWS organizations API https://github.com/project-koku/koku/issues/1916
* AWS Organizational unit crawler should also record delete timestamps https://github.com/project-koku/koku/issues/2043
* Correlate AWS costs with Organization Unit https://github.com/project-koku/koku/issues/1915
* Add group by and filter option for org unit to AWS, OCP on AWS reports options https://github.com/project-koku/koku/issues/1917 - this is only implemented for the AWS reports in this PR. 

Marking WIP until a couple of cleanup items & questions are taken care of! Feel free to give feedback at this time.

---- 
# How to test
1. New `group_by[org_unit_id]` & `filter[org_unit_id]` for the aws reports endpoint.
We have added a make command that will populate a aws organizational data tree inside of the database for you. Therefore all you have to do is make sure that the server is up and ready and run:
```
 make load-aws-org-unit-tree
```
After the org tree has been upload you can now check the new group by/filter with the following links:
http://127.0.0.1:8000/api/cost-management/v1/reports/aws/costs/?group_by[org_unit_id]=R_001
http://127.0.0.1:8000/api/cost-management/v1/reports/aws/costs/?filter[org_unit_id]=OU_002

2. The new organizations endpoint. 
After running the `make load-aws-org-unit-tree` the data in the organizations endpoint will also be populated. So can view this data through the following links:
- http://127.0.0.1:8000/api/cost-management/v1/organizations/aws/?key_only=True

**Note:** The organizations api is disassociated with the costs. Therefore, if you are seeing an account in the org api, but not the cost api it likely means that the account did not have any cost associated with it. 

3. The org unit crawler. 
Testing the crawler is a little bit more complicated because you will need a real aws account with aws organizational units in it. If you do not have one, message me or @myersCody and we can provide you with one. 

After you have an aws account you will then need to manually change the crawler to run more frequently than once a day:
```
app.conf.beat_schedule["crawl_account_hierarchy"] = {
    "task": "masu.celery.tasks.crawl_account_hierarchy",
    "schedule": crontab(minute="*"),
}
```
You may still want to be careful here because when you start the process a download task may begin to run. If you are using a large account like we were  while developing this you will for sure not want to download these massive files. Therefore you may want to stop the downloading with a 
```
docker-compose down && docker-compose up -d
```
If at any point, if anyone has questions feel free to reach out to @abaiken


koku/api/organizations/serializers.py
```diff
@@ -0,0 +1,130 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+"""Tag serializers."""
```
abaiken, line 17:
```suggestion
"""Organizations serializers."""
```

koku/api/organizations/aws/queries.py
```diff
@@ -0,0 +1,69 @@
...
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+"""AWS Tag Query Handling."""
+from api.models import Provider
+from api.organizations.queries import OrgQueryHandler
+from api.report.provider_map import ProviderMap
+from reporting.provider.aws.models import AWSOrganizationalUnit
+
+
+class AWSOrgProviderMap(ProviderMap):
```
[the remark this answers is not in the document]
myersCody: Yeah, I had alot of internal debate at the time about where to put this and eventually just decided to put it here. However, I am down to move it, but initally I was like "eh its 12 lines of code" 😂

koku/api/organizations/serializers.py
myersCody: Thanks for taking care of these @abaiken Can you tell I stole the serializers from the Tap endpoint 😂

koku/api/report/aws/query_handler.py
```diff
@@ -164,6 +167,107 @@ def query_table(self):
...
+        sub_orgs_dict = {}
+        query_data_results = {}
+        query_sum_results = []
+        org_unit_applied = False
+        if "org_unit" in self.parameters.parameters.get("group_by"):
+            if self.parameters.report_type != "costs":
+                # since we only have the org unit group_by available for cost reports
+                # raise a validation error if it is a different report type
+                error = {"org_unit": _("Unsupported parameter or invalid value")}
+                raise serializers.ValidationError(error)
```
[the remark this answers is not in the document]
abaiken: Great catch - I moved this validation to the serializer.

koku/api/report/aws/query_handler.py
```diff
@@ -164,6 +167,107 @@ def query_table(self):
...
+                        if day.get("values"):
+                            each_day["sub_orgs"].append(
+                                {
+                                    "org_name": org_name,
+                                    "org_id": sub_orgs_dict.get(org_name),
+                                    "date": day.get("date"),
+                                    "values": day.get("values"),
+                                }
+                            )
+        return query_data
```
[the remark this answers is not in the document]
abaiken: Absolutely!

koku/api/report/aws/query_handler.py
```diff
@@ -350,7 +454,29 @@ def __resolve_conditions(condition, alias="t", where=None, values=None):
...
+        """
+        for key in sum2:
+            if key in sum1:
+                for next_key in sum2[key]:
+                    if next_key in sum1[key]:
+                        for final_key in sum2[key][next_key]:
+                            if final_key in sum1[key][next_key] and final_key == "value":
+                                sum2[key][next_key][final_key] = (
+                                    sum2[key][next_key][final_key] + sum1[key][next_key][final_key]
+                                )
```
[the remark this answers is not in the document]
abaiken: This is also a good suggestion! @myersCody actually pushed up a change to clean this function up a little bit with recursion - take a look at what he did.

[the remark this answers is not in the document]
blentz: I think @myersCody 's solution is adequate. I think set intersections might be _slightly_ faster. But I also think it's academic.

Set intersections are O(n) and the recursive approach looks like O(n^m) depending on how complex the structure is we need to recurse down. However, that point is moot when we know the structure is 3 levels deep.

So, short story made long, his solution is fine by me. :slightly_smiling_face: :+1:

koku/api/report/aws/serializers.py
```diff
@@ -96,6 +108,13 @@ def validate_group_by(self, value):
 
         """
         validate_field(self, "group_by", GroupBySerializer, value, tag_keys=self.tag_keys)
+        # Additionally, since we only have the org_unit_id group_by available for cost reports
+        # we must explicitly raise a validation error if it is a different report type
+        if "org_unit_id" in self.initial_data.get("group_by", {}).keys():
+            request = self.context.get("request")
+            if "costs" not in request.path:
```
dccurtis, line 115: @abaiken probably removing this is all that would be needed if we ever want to support `org_unit_id` filter/group_by in the instance-type or storage API?

koku/koku/celery.py
```diff
@@ -139,6 +139,13 @@ def task(self, *args, **kwargs):
     }
 
 
+# Beat used to crawl the account hierarchy
+app.conf.beat_schedule["crawl_account_hierarchy"] = {
+    "task": "masu.celery.tasks.crawl_account_hierarchy",
+    "schedule": crontab(hour=0, minute=0),
```
dccurtis, line 145: Does it make sense to make this configurable with an ENV?

dccurtis:
It looks like the materialized view stuff should be fine since the `query_table` is being used in `execute_individual_query` but curious if you sanity checked that to make sure we're not slowing the query handler.

koku/api/query_params.py
```diff
@@ -70,6 +70,7 @@ def __init__(self, request, caller, **kwargs):
 
         self.kwargs = kwargs
         self.request = request
+        self.caller = caller
```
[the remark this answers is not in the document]
abaiken: Yes! This is used in the query_handler on the parameters instance in order to configure rbac. When grouping by `org_unit_id` - we are actually combining 1. a query that groups by accounts and filters on `org_unit_id` and 2. A query filtering on each sub org with no group by - Since we are manipulating the query params outside of what the user has passed in, rbac is initially not configured so we have to call that [here](https://github.com/project-koku/koku/pull/2139/files#diff-9fa2e6770cbc217b865ad5a5f33ce5bcR214).

koku/api/report/aws/serializers.py
abaiken: Possibly - but we would have to make sure that the total sum [function](https://github.com/project-koku/koku/pull/2139/files#diff-9fa2e6770cbc217b865ad5a5f33ce5bcR450) that adds together the query sums works for the structure of the query sums returned by the storage and instance-type endpoints. The filter works for those endpoints - it is only the group by that is disabled.

koku/koku/celery.py
abaiken: I think that we originally started down that route but after discussing with @adberglund we decided that this was definitely a daily task and would not make sense for it to be changed. The only value that I could see an ENV adding would be for local testing. But I'll loop @kholdaway in here too since he was part of that decision.

abaiken:
> It looks like the materialized view stuff should be fine since the `query_table` is being used in `execute_individual_query` but curious if you sanity checked that to make sure we're not slowing the query handler.

I did not sanity check this! Tomorrow I can reach out and see how you would go about testing this.

koku/api/query_params.py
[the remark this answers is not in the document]
dccurtis: Thanks for pointing that out!

koku/koku/celery.py
dccurtis: I was thinking through the ordering of org updates and cost usage report ingestion and when the user would actually see the data.

Did you consider adding the work that crawl_account_hierarchy does to be part of the regular aws processing flow?  This is going to be difficult to add automated tests for.  Putting it in the processing flow would also give the user a refreshed view of the organization. 

If it must stay as a beat schedule. Maybe add the task to the /download admin endpoint.

kholdaway: We can talk through that tomorrow. I am not sure it’s related to that flow at all. Ordering doesn’t matter. We take a snapshot of the tree daily. It can change before or after that snapshot. So controlling order doesn’t seem worth coupling it together. 

For test we created a tool to inject a tree into the database. I would say we leave it as is and not add a flag with low value in production.

dccurtis: Sounds great.  Admittedly, I have very limited knowledge of AWS org units, so that would be helpful!

kholdaway: Also, we touched base with Luke and he said he wouldn't be testing with real AWS accounts.  I forgot to mention that detail too.  I am not opposed to tweaking it if needed.  As Ashley said, when we chatted with Andrew he said let's add it when we know why we need it.

PR #2145, opened 2020-05-15
add django-extensions, remove drf-nested-routers

blentz:
This PR adds `django-extensions` and removes `drf-nested-routers`.

I'm removing nested-routers because we stopped using it quite a while ago. So, this removes an unused dependency. We can always reinstall it later.

I'm adding the extensions module because I found it useful in doing a performance investigation recently. The `runprofileserver` command is really handy.

Here's the commands that are added by this module:

```
[django_extensions]
    admin_generator
    clean_pyc
    clear_cache
    compile_pyc
    create_command
    create_jobs
    create_template_tags
    delete_squashed_migrations
    describe_form
    drop_test_database
    dumpscript
    export_emails
    find_template
    generate_password
    generate_secret_key
    graph_models
    mail_debug
    merge_model_instances
    notes
    passwd
    pipchecker
    print_settings
    print_user_for_session
    reset_db
    reset_schema
    runjob
    runjobs
    runprofileserver
    runscript
    runserver_plus
    set_default_site
    set_fake_emails
    set_fake_passwords
    shell_plus
    show_template_tags
    show_templatetags
    show_urls
    sqlcreate
    sqldiff
    sqldsn
    sync_s3
    syncdata
    unreferenced_files
    update_permissions
    validate_templates

```


abaiken:
Do we need to run `make requirements` to remove this from the [rtd_requirements.txt](https://github.com/project-koku/koku/blob/master/docs/rtd_requirements.txt#L34)?

PR #2146, opened 2020-05-15
Validating source cost-model associations for cost-model creation/update

dccurtis:
Throwing an exception when a cost-model update or create is attempted and a source on the `source_uuids` list is already associated with a cost model.

**Testing**
1. Create two OCP sources (Source 1 and Source 2).
2. Create a cost model (Cost Model 1) with Source 1.
3. Create a second cost model (Cost Model 2) with both OCP sources.  Verify creation fails.
4. Remove Source 1 from Cost Model 1 and attempt to create Cost Model 2 again.  Verify it's successful and both sources are associated.
5. Update Cost Model 1 with Source 1 and verify that it fails.
6. Remove Source 1 from Cost Model 2, then attempt to update Cost Model 1 with Source 1 again.  Verify it's successful.

**Test Results**
[cost_model_no_alter_ut.txt](https://github.com/project-koku/koku/files/4636070/cost_model_no_alter_ut.txt)


dccurtis:
@lcouzens ran smokes against this, no failures.  Any other non-smoke tests that you think could be impacted?

lcouzens:
@dccurtis Thats correct, I think all the cost model tests are part of smoke so that should do it. That said I probably want to add a few new cases round this change at some point, but we will work that in after its merged.

dccurtis:
@boaz0 when you start working https://issues.redhat.com/browse/COST-54, using this branch might be helpful for the backend.

dccurtis:
Adding @boaz0 as a reviewer so we know to not merge this until frontend is ready.

PR #2155, opened 2020-05-21
Cost 133 upgrade postgres

Red-HAP:
## Description

This branch switches from using the postgres:10.6 image to using postgres:12.

After testing, it appears that koku-server repeatedly goes into funk-mode when the database is re-initialized via the makefile (down, delete, up, init) so the make targets were changed a bit to restart the koku server when the database goes down and comes back up.

A _slight_ bit of scope-creep here as the DB monitor user secret now uses a generated password vs a hardcoded one.

This is passing tox testing and IQE testing (with blessings from Luke).

JIRA ticket: [COST-133](https://issues.redhat.com/browse/COST-133)


docker-compose.yml
```diff
@@ -107,7 +107,7 @@ services:
 
   db:
     container_name: koku_db
-    image: postgres:10.6
+    image: postgres:latest
```
adberglund, line 110: Should we use the 12 tag explicitly so this doesn't bump on us before we're ready?

Red-HAP: Yeah. I had pulled an image labeled 12 that ended up a minor version behind what I knew was available. So I pulled out too big of a hammer.

PR #2160, opened 2020-05-22
Handle invalid source type flows for provider accessor.

chambridge:
* Create a check to raise error if source type was invalid on provider accessor creation
* Refactor error_msg function to common location.


dccurtis:
That error_obj refactor is long over due.  Thanks! 👍  Did some quick sanity testing with bad aws and azure sources and it's looking good so far.

PR #2164, opened 2020-05-25
Wrap create-sources task errors correctly to fix AttributeError

boaz0:
put the last nail in  #2151 coffin :scream: 
jk :smiley:

### DIFF
--- a/koku/api/provider/serializers.py
+++ b/koku/api/provider/serializers.py
@@ -374,20 +374,20 @@ class ProviderSerializer(serializers.ModelSerializer):
             instance.save()
             raise validation_error
 
-        bill, __ = ProviderBillingSource.objects.get_or_create(**billing_source)
-        auth, __ = ProviderAuthentication.objects.get_or_create(**authentication)
-        dup_queryset = (
-            Provider.objects.filter(authentication=auth).filter(billing_source=bill).filter(customer=customer)
-        )
-        if dup_queryset.count() != 0:
-            conflict_provder = dup_queryset.first()
-            message = (
-                f"Cost management does not allow duplicate accounts. "
-                f"{conflict_provder.name} already exists. Edit source settings to configure a new source."
-            )
-            LOG.warn(message)
-
         with transaction.atomic():
+            bill, __ = ProviderBillingSource.objects.get_or_create(**billing_source)
+            auth, __ = ProviderAuthentication.objects.get_or_create(**authentication)
+            dup_queryset = (
+                Provider.objects.filter(authentication=auth).filter(billing_source=bill).filter(customer=customer)
+            )
+            if dup_queryset.count() != 0:
+                conflict_provder = dup_queryset.first()
+                message = (
+                    f"Cost management does not allow duplicate accounts. "
+                    f"{conflict_provder.name} already exists. Edit source settings to configure a new source."
+                )
+                LOG.warn(message)
+
             for key in validated_data.keys():
                 setattr(instance, key, validated_data[key])
 
--- a/koku/api/provider/serializers.py
+++ b/koku/api/provider/serializers.py
@@ -15,6 +15,8 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 #
 """Provider Model Serializers."""
+import logging
+
 from django.conf import settings
 from django.db import IntegrityError
 from django.db import transaction
@@ -29,6 +31,9 @@ from api.provider.models import Provider
 from api.provider.models import ProviderAuthentication
 from api.provider.models import ProviderBillingSource
 from providers.provider_access import ProviderAccessor
+from providers.provider_errors import ProviderErrors
+
+LOG = logging.getLogger(__name__)
 
 PROVIDER_CHOICE_LIST = [
     provider[0]
@@ -322,12 +327,17 @@ class ProviderSerializer(serializers.ModelSerializer):
         auth, __ = ProviderAuthentication.objects.get_or_create(**authentication)
 
         # We can re-use a billing source or a auth, but not the same combination.
-        unique_count = (
-            Provider.objects.filter(authentication=auth).filter(billing_source=bill).filter(customer=customer).count()
+        dup_queryset = (
+            Provider.objects.filter(authentication=auth).filter(billing_source=bill).filter(customer=customer)
         )
-        if unique_count != 0:
-            error = {"Error": "A Provider already exists with that Authentication and Billing Source"}
-            raise serializers.ValidationError(error)
+        if dup_queryset.count() != 0:
+            conflict_provider = dup_queryset.first()
+            message = (
+                f"Cost management does not allow duplicate accounts. "
+                f"{conflict_provider.name} already exists. Edit source settings to configure a new source."
+            )
+            LOG.warn(message)
+            raise serializers.ValidationError(error_obj(ProviderErrors.DUPLICATE_AUTH, message))
 
         provider = Provider.objects.create(**validated_data)
         provider.customer = customer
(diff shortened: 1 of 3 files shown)


maskarb:
What you have here is probably be a good solution. Looks like we have another one that should be changed while we're at it ([line 349](https://github.com/project-koku/koku/blob/master/koku/api/provider/serializers.py#L349)):

```python
        if instance.type != provider_type:
            error = {"Error": "The Provider Type cannot be changed with a PUT request."}
            raise serializers.ValidationError(error)
```

koku/api/provider/serializers.py
```diff
@@ -326,8 +326,9 @@ def create(self, validated_data):
             Provider.objects.filter(authentication=auth).filter(billing_source=bill).filter(customer=customer).count()
         )
         if unique_count != 0:
-            error = {"Error": "A Provider already exists with that Authentication and Billing Source"}
-            raise serializers.ValidationError(error)
+            key = "Error"
+            message = "A Provider already exists with that Authentication and Billing Source"
```
dccurtis, line 330: @spank50 These are a few source error message strings that were missed as part of our user facing error string cleanup.  That work is captured here: https://docs.google.com/spreadsheets/d/1BgweFAhU3RAU30Bj1rzYQcVxwOBWS1WUBYyjlvD-bFY/edit#gid=0

The `Improved error message` column is what Dayle, Natalie and I landed on.  This might give you a sense of how these error messages are worded.

Do you have any suggestions on what the text should be?  The situation that needs to be covered is when a source (provider) already exists in cost management with the same account configuration.  Additionally we shouldn't be using the Provider name for user facing strings.  It should be Source.

Let me know if you have any questions on this.  Thanks!

koku/api/provider/serializers.py
```diff
@@ -346,8 +347,9 @@ def update(self, instance, validated_data):
         provider_type = Provider.PROVIDER_CASE_MAPPING.get(provider_type)
         validated_data["type"] = provider_type
         if instance.type != provider_type:
-            error = {"Error": "The Provider Type cannot be changed with a PUT request."}
-            raise serializers.ValidationError(error)
+            key = "Error"
```
dccurtis, line 350: I'm pretty sure this is a left over from when we supported providers and sources.

koku/api/provider/serializers.py
dccurtis: @boaz0 It would be great if we could internally log the source that already existed and then make a more general user facing string that would be the language that we use from @spank50 's guidance.

koku/api/provider/serializers.py
boaz0: I'll remove this then...

koku/api/provider/serializers.py
boaz0: sure thing

boaz0: @dccurtis done :+1: 
waiting on @spank50 response.

spank50: Based on the examples provided in the spreadsheet, I suggest the following wording. I do have some questions first.
- Is this an AWS error, Azure error, or both?
- The authentication/billing source is edited in AWS/Azure, and the user knows how to edit and remove the duplicate?
- Is "Authentication and Billing Source" one thing, or is it two things, "Authentication Source and Billing Source"? 
If one thing:
"Cost management does not allow a duplicate source for authentication and billing. Edit your [AWS/Azure] source and remove the duplicate information."

dccurtis: Thanks for the feedback @spank50!  I should have given you more background before, but here we go!

This error can occur for all providers (AWS, Azure, and OCP).  The terms `authentication` and `billing_source` are generic internal names that help define an account.

Each provider type has a different definition of what `authentication` and `billing_source` means:

**AWS**
Authentication: roleARN
Billing Source: S3 Bucket

**Azure**
Authentication: Client ID, Tenant ID, and Subscription ID 
Billing Source: Storage Resource Group, Storage Account

**OCP**
Authentication: Cluster ID
Billing Source: N/A

A duplicate source would happen if:
1. The user entered identical values for both "authentication" and "billing source" for the respective source type.
2. The name of the source would be different than the one that already exists.  This is because platform sources guarantees that source names are globally unique.

When the user creates a source they are prompted to enter the specific source values (roleARN, s3 bucket, client id, etc) and not authentication or billing source (internal names).

The user facing source error strings are generic error strings at this point in time and they don't provide too many specifics as to what the problem actually is.  I'm definitely open to enhancing this in the future, but today they are generic error messages.

Now for this case I think there is a lot of value in telling the user what pre-existing source already has the account settings since then they would be able to take action on fixing the problem (removing the old one, edit the old one, bail on creating the new one, etc).

@boaz0 made some changes to allow for us to log the existing source name.  I'm thinking we could use this in the user facing string.

With all this being said.  What do you think about something like:

"Cost management does not allow duplicate accounts. [Source Name] already exists.  Edit source settings to configure a new source."

@nlcwong might be interested this duplicate source situation.

spank50: Thanks for the explanation @dccurtis. Your suggested wording looks good to me.

boaz0: Done :+1:

koku/api/provider/serializers.py
```diff
@@ -326,8 +331,12 @@ def create(self, validated_data):
             Provider.objects.filter(authentication=auth).filter(billing_source=bill).filter(customer=customer).count()
         )
         if unique_count != 0:
-            error = {"Error": "A Provider already exists with that Authentication and Billing Source"}
-            raise serializers.ValidationError(error)
+            message = (
+                f"Cost management does not allow duplicate accounts. "
+                f"{validated_data['name']} already exists. Edit source settings to configure a new source."
```
dccurtis, line 336: Very close!
This name should be the first source that was created.  You'll have to get that name from the `Provider.objects.filter(authentication=auth).filter(billing_source=bill).filter(customer=customer)` query.

Here's an example:

Source message was successfully pushed to platform sources.  Just need to fix the name so the user will know what source they have to edit or remove.

`GET http://localhost:3000/api/v1.0/applications/?filter[source_id]=2`
```
{
    "meta": {
        "count": 1,
        "limit": 100,
        "offset": 0
    },
    "links": {
        "first": "/api/v1.0/applications/?filter%5Bsource_id%5D=2&offset=0",
        "last": "/api/v1.0/applications/?filter%5Bsource_id%5D=2&offset=0"
    },
    "data": [
        {
            "application_type_id": "2",
            "availability_status": "unavailable",
            "availability_status_error": "Cost management does not allow duplicate accounts. Test OCP Source 2 already exists. Edit source settings to configure a new source.",
            "created_at": "2020-05-27T19:54:38Z",
            "id": "2",
            "source_id": "2",
            "updated_at": "2020-05-27T19:54:39Z",
            "tenant": "10001"
        }
    ]
}
```

dccurtis: In this example I already had a source created with the same cluster-id.  Name was `Test OCP Source 1`

boaz0: OK it was tricky than I thought. I couldn't execute queries inside the atomic block (which is a little odd because `ProviderBillingSource.objects.get_or_create(**billing_source)` seems like one). Anyway, I moved all that block out of the atomic block.

dccurtis: Looks great!  I'm going to try to move that transaction back before the get_or_create since we definitely want the to rollback if there are any errors and can push a change to the branch if that's successful.

boaz0:
Thanks @dccurtis for helping out.
I am merging this in 3.. 2.. 1.. :bomb: :boom:

PR #2168, opened 2020-05-27
Fix bugs with moving account & ou and filtering for organization endpoint.

myersCody:
Jira Issues:
https://issues.redhat.com/browse/COST-160
https://issues.redhat.com/browse/COST-157

This pr fixes some bugs in the organizational endpoint. 

### COST-160
This bug is because I didn't catch accounts and sub_ou moving from one organizational unit (ou) to another ou properly. Therefore, the account was showing up in its previous ou and the new ou. The same goes for child organizational unit moving from one parent to another.

#### How to test
Therefore to test that this has been fixed you will need to bring up the server and run `make load-aws-org-unit-tree`

After the tree has been populated in the database, we can look at this endpoint: http://127.0.0.1:8000/api/cost-management/v1/organizations/aws/

###### Check Moving Account
The `account 003` was moved from `OU_002` to `OU_003`. Therefore, we just need to check that `account 003` no longer shows up in `OU_002`, but does show up in `OU_003`

###### Check Moving OU
The child organizational unit or sub_ou `OU_005` was moved from `OU_002` to `OU_001`. Therefore, we just need to check that `OU_005` is not in `OU_OO2`'s sub_orgs but is in `OU_001`

**Note:** You can check how the accounts and ou are moved by looking at the [aws_org_tree.yml](https://github.com/project-koku/koku/blob/master/scripts/aws_org_tree.yml#L28) file.

### COST-157
This bug was that the sub_orgs were not showing up when using `?filter[org_unit_id]=OU_001`. I managed to fix this as a result of fixing 160. To test this one you just need to make sure the sub_orgs now show up when you hit the endpoint with the filter. If you already have run `make load-aws-org-unit-tree` you should be able to go to: http://127.0.0.1:8000/api/cost-management/v1/organizations/aws/?filter[org_unit_id]=OU_001 and see:
```
        {
            "org_unit_id": "OU_001",
            "org_unit_name": "Dept OU_001",
            "org_unit_path": "R_001&OU_001",
            "level": 1,
            "sub_orgs": [
                "OU_005"
            ],
            "accounts": [
                "account 001",
                "account 002"
            ]
        }
```

### DIFF
--- a/koku/api/organizations/queries.py
+++ b/koku/api/organizations/queries.py
@@ -18,7 +18,6 @@
 import copy
 import logging
 
-from django.contrib.postgres.aggregates import ArrayAgg
 from django.db.models import F
 from django.db.models import Q
 from django.db.models.functions import Coalesce
@@ -177,17 +176,17 @@ class OrgQueryHandler(QueryHandler):
 
         return composed_filter
 
-    def _get_sub_ou_list(self, data):
+    def _get_sub_ou_list(self, data, org_ids):
         """Get a list of the sub org units for a org unit."""
         level = data.get("level")
         level = level + 1
         unit_path = data.get("org_unit_path")
+        final_sub_ou_list = []
         with tenant_context(self.tenant):
             for source in self.data_sources:
                 # Grab columns for this query
                 account_info = source.get("account_alias_column")
                 level_column = source.get("level_column")
-                org_id = source.get("org_id_column")
                 org_path = source.get("org_path_column")
                 # Build filters
                 filters = QueryFilterCollection()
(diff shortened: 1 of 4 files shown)


koku/api/organizations/queries.py
```diff
@@ -273,12 +264,16 @@ def get_org_units(self):
                 org_unit_query = org_unit_query.exclude(deleted_timestamp__lte=self.start_datetime)
                 org_unit_query = org_unit_query.exclude(created_timestamp__gte=self.end_datetime)
                 val_list = [org_id, org_name, org_path, level]
+                org_unit_query = org_unit_query.order_by(f"{org_id}", f"-{created_field}").distinct(f"{org_id}")
+                org_ids = org_unit_query.values_list("id", flat=True)
+                org_id_list.extend(org_ids)
```
abaiken, line 269: I think that we will want to filter these org_ids based on the ones that we have access to so that a user with restricted access cannot see orgs inside of the `sub_orgs` list that they shouldn't.

PR #2171, opened 2020-05-28
Try and force install of importlib-metadata

chambridge:
* Explicitly call out importlib-metadata


chambridge:
Tried branch doesn't work. Closing this out.

PR #2174, opened 2020-05-28
Update openapi.json to document organizations endpoint 

abaiken:
Related issue: https://issues.redhat.com/browse/COST-141

This PR adds openapi spec documentation for the organizations endpoint and also fixes some pre-existing syntax errors for the tags key parameter documentation. It also documents the `org_unit_id` filter and group_by. 

To test: 
1. Copy the openapi.json contents (or open the file) into http://editor.swagger.io/
2. View that there are no syntax errors and check out the newly documented organizations endpont

#### Also - sidenote, can someone make sure the filters/group by I added look okay - I'm less familiar with documenting those using swagger/openapi

### DIFF
--- a/docs/source/specs/openapi.json
+++ b/docs/source/specs/openapi.json
@@ -446,6 +446,9 @@
                     },
                     {
                         "$ref": "#/components/parameters/ReportQueryLimit"
+                    },
+                    {
+                        "$ref": "#/components/schemas/OrganizationFilter"
                     }
                 ],
                 "responses": {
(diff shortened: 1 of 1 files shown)


docs/source/specs/openapi.json
```diff
@@ -426,6 +426,55 @@
...
+                        "description": "A paginated list of Organization objects",
+                        "content": {
+                            "application/json": {
+                                "schema": {
+                                    "$ref": "#/components/schemas/OrganizationPagination"
+                                }
+                            }
+                        }
+                    },
+                    "500": {
```
dccurtis, line 465: I think you need to include the 403 response too

abaiken: I can add a 403 response. I didn't because I don't think that this endpoint returns a 403 unless you are actually adding a filter for an organization that you do not have access to (but it will never be returned without the filter). 

Also - we should probably go back and update the other endpoints because I don't see 403's anywhere in our openapispec and I suspect that other endpoints return them as well.

abaiken: @dccurtis added 😁

PR #2179, opened 2020-05-29
missed nise cli arg update

maskarb:
Testing the Nise version update:

```
make clear-testing
make docker-reinitdb
make create-test-customer
make load-test-customer-data 
```

Check the generated report files. Ensure they are not empty.


boaz0:
That's odd - why the unit tests fail for that "small" change.

PR #2190, opened 2020-06-04
Cost 140 cache tenant customer user

lrlevine6:
Added caching for users, tenants and customers to the middleware

### DIFF
--- a/koku/koku/middleware.py
+++ b/koku/koku/middleware.py
@@ -20,6 +20,7 @@ import logging
 from http import HTTPStatus
 from json.decoder import JSONDecodeError
 
+from cachetools import TTLCache
 from django.conf import settings
 from django.core.cache import caches
 from django.core.exceptions import PermissionDenied
@@ -46,6 +47,11 @@ from koku.metrics import DB_CONNECTION_ERRORS_COUNTER
 from koku.rbac import RbacConnectionError
 from koku.rbac import RbacService
 
+TIME_TO_CACHE = 3  # in seconds (15 minutes)
+MAX_CACHE_SIZE = 128
+
+user_cache = TTLCache(maxsize=MAX_CACHE_SIZE, ttl=TIME_TO_CACHE)
+
 
 LOG = logging.getLogger(__name__)
 MASU = settings.MASU
@@ -70,7 +76,6 @@ def is_no_entitled(request):
 
 class HttpResponseUnauthorizedRequest(HttpResponse):
     """A subclass of HttpResponse to return a 401.
-
     Used if identity header is not sent.
     """
 
@@ -98,10 +103,13 @@ class HttpResponseFailedDependency(JsonResponse):
 
 class KokuTenantMiddleware(BaseTenantMiddleware):
     """A subclass of the Django-tenant-schemas tenant middleware.
-
     Determines which schema to use based on the customer's schema
     found from the user tied to a request.
     """
+
+    def __init__(self, get_response=None):
+        super().__init__(get_response)
+        self.tenant_cache = TTLCache(maxsize=MAX_CACHE_SIZE, ttl=TIME_TO_CACHE)
 
     def process_exception(self, request, exception):  # pylint: disable=R0201,R1710
         """Raise 424 on InterfaceError."""
@@ -118,7 +126,9 @@ class KokuTenantMiddleware(BaseTenantMiddleware):
             if hasattr(request, "user") and hasattr(request.user, "username"):
                 username = request.user.username
                 try:
-                    User.objects.get(username=username)
+                    # User.objects.get(username=username)
+                    if username not in user_cache:
+                        user_cache[username] = User.objects.get(username=username)
                 except User.DoesNotExist:
                     return HttpResponseUnauthorizedRequest()
                 if not request.user.admin and request.user.access is None:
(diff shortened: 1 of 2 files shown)


chambridge:
First, nothing above looks wrong. I just have some questions for my own understanding.

Why the difference in implementation between tenant, customer, and user caches? I'm trying to understand why not all as a constant etc? Do you how the cache will behave depending on the gunicorn workers (is it per worker)? Are the middleware classes instantiated essentially once when the server starts?

lrlevine6:
> Why the difference in implementation between tenant, customer, and user caches? I'm trying to understand why not all as a constant etc? Do you how the cache will behave depending on the gunicorn workers (is it per worker)? Are the middleware classes instantiated essentially once when the server starts?

How I had implemented was so there are three different caches with one for user, tenant and customer. The tenant cache and customer cache are created similarly where when the `KokuTenantMiddleware` class is initiated the tenant class will be initiated and when the `IdentityHeaderMiddleware` class is initiated the customer class was initiated since these objects were reference in the scope of these classes. The user tenant class was determined to be global since it is referenced in both the `KokuTenantMiddleware` and `IdentityHeaderMiddleware` classes. It is initiated when the server starts.

koku/koku/middleware.py
```diff
@@ -98,12 +103,17 @@ def __init__(self, dikt):
...
     """A subclass of the Django-tenant-schemas tenant middleware.
-
     Determines which schema to use based on the customer's schema
     found from the user tied to a request.
     """
 
+    def __init__(self, get_response=None):
+        """ Create the tenant cache"""
+        super().__init__(get_response)
+        self.tenant_cache = TTLCache(maxsize=MAX_CACHE_SIZE, ttl=TIME_TO_CACHE)
```
adberglund, line 113: I don't know how cachetools works, but, is this okay as an instance variable? My gut tells me it should be a class attribute that all instances of the middleware share.

lrlevine6: I think that either would work based on my understanding. The tenant and customer cache are not being created with every request as that would defeat the purpose of the cache. Technically there would be more tenant and customer cache than the one user class though. I can change these to be class attribute so there is only one instance of each if that is preferred!

dccurtis:
> First, nothing above looks wrong. I just have some questions for my own understanding.
> 
> Why the difference in implementation between tenant, customer, and user caches? I'm trying to understand why not all as a constant etc? Do you how the cache will behave depending on the gunicorn workers (is it per worker)? Are the middleware classes instantiated essentially once when the server starts?

@chambridge 
Each Gunicorn worker is a separate process running the Django application.  So any in-memory type cache will be scoped only to a single worker.  So if we have 3 workers running per pod, then that would be 3 separate caches that would have to be populated.  So some inconsistent latency is expected until all 3 worker caches are populated.

Since we only have 3 workers per-pod and the TTL is quite long, this might not be a big deal for this application.

Maybe we should open a follow on issue to investigate this further and see if `memcached ` would give us a cache that spans all workers.  If so, we could migrate our rbac cache at the same time

kholdaway:
@dccurtis I think memcached would be slower due to network calls.  In fact, Jag started with Redis and was surprised to learn it didn't speed up performance much.  He then went to this local cache approach.

dccurtis:
> @dccurtis I think memcached would be slower due to network calls. In fact, Jag started with Redis and was surprised to learn it didn't speed up performance much. He then went to this local cache approach.

Yes that's what I remembered as well from the video.  But with `memcached` wouldn't it be a daemon process running within the pod itself?  I can see the network factor with Redis, but I"m not sure that would apply for memcached.

kholdaway:
I wouldn't include this as part of this PR, but after looking at the Django Memcached integration I think it could be worth a look.  See https://docs.djangoproject.com/en/2.2/topics/cache/ (since we are on version 2).  First, if we have 3 workers and Django is not sticky with its sessions then a single user's data may land in a cache for all 3 worker middleware objects.  There could be a small overhead for the Unix socket request to the daemon but could end up saving memory.

koku/koku/middleware.py
```diff
@@ -136,37 +144,38 @@ def process_request(self, request):
...
+                user = User.objects.get(username=tenant_username)
+                customer = user.customer
+                schema_name = customer.schema_name
+            try:
+                tenant = model.objects.get(schema_name=schema_name)
+            except model.DoesNotExist:
+                tenant = model(schema_name=schema_name)
+                tenant.save()
+            KokuTenantMiddleware.tenant_cache[tenant_username] = tenant
+            LOG.info(f"Tenant added to cache: {tenant_username}")
```
dccurtis, line 159:
```suggestion
            LOG.debug(f"Tenant added to cache: {tenant_username}")
```

koku/koku/middleware.py
```diff
@@ -118,7 +124,9 @@ def process_request(self, request):
             if hasattr(request, "user") and hasattr(request.user, "username"):
                 username = request.user.username
                 try:
-                    User.objects.get(username=username)
+                    if username not in USER_CACHE:
+                        USER_CACHE[username] = User.objects.get(username=username)
+                        LOG.info(f"User added to cache: {username}")
```
dccurtis, line 129:
```suggestion
                        LOG.debug(f"User added to cache: {username}")
```

PR #2191, opened 2020-06-05
COST-83: Adding source_uuid to report endpoints

myersCody:
Jira Issue: 
https://issues.redhat.com/browse/COST-83

For this issue I need to:
- Update to AWS, Azure, OpenShift, OpenShift on AWS/Azure/All daily summary models to include source uuid.
- Update daily summary SQL for all of the above mentioned to add source UUID
- Update provider maps to include source uuid.

--------

## Testing
For this pr, the best way to test is to pull up the server and load it with this test data:
```
make docker-reinitdb-with-sources
make create-test-customer
make load-test-customer-data start=2020-05-01
```
This will run the migrations that I added to add the source uuid to all of the daily summary models. It will will also run the sql I changed in `masu/database/sql` to populate the daily summary models when a report is processed.

After that to check the provider map changes you will need to get request every endpoint that now has the source_uuid data (which is now every report endpoint). 

I know what you are thinking, "but Cody I don't want have to figure out every url for each endpoint". I already had to figure that out for when I functionally testing so I got you, [ List of endpoints](https://gist.github.com/myersCody/23b285ff86928b435e1f561dbd83340c). That list contains a mapping of urls to materialize views, hopefully it helps cause you can quick click through the links.


boaz0:
@myersCody GJ! I will be testing this today and will give you a review.
Thanks!

PR #2196, opened 2020-06-08
bugfixes for scripts, part 2 (COST-106)

blentz:
Part 2 of scripting fixes.

This PR ensures that `load_test_customer.sh` uploads all of the test buckets when using OCP.

I've also changed the `e2e-deploy` script to not fail during the main deployment loop. Failures there are often correctable after-the-fact. But, it's important that the script runs to completion so that the local OCP is fully configured.

### DIFF
--- a/scripts/load_test_customer_data.sh
+++ b/scripts/load_test_customer_data.sh
@@ -34,15 +34,17 @@ KOKU_PATH=$1
 START_DATE=$2
 END_DATE=$3
 
-if [ -z "${KOKU_API_HOSTNAME}" ]; then
-    echo "Please set the KOKU_API_HOSTNAME environment variable. Exiting."
-    exit 2
-fi
+### validation
+function check_var() {
+    if [ -z ${!1:+x} ]; then
+        echo "Environment variable $1 is not set! Unable to continue."
+        exit 2
+    fi
+}
 
-if [ -z "${MASU_API_HOSTNAME}" ]; then
-    echo "Please set the MASU_API_HOSTNAME environment variable. Exiting."
-    exit 2
-fi
+check_var KOKU_API_HOSTNAME
+check_var MASU_API_HOSTNAME
+check_var NISE_REPO_PATH
 
 KOKU_API=$KOKU_API_HOSTNAME
 MASU_API=$MASU_API_HOSTNAME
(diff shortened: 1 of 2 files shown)


scripts/load_test_customer_data.sh
```diff
@@ -158,7 +155,11 @@ git checkout -- "$NISE_REPO_PATH/examples/ocp_on_azure/ocp_static_data.yml"
 if [[ $USE_OC == 1 ]]; then
     WORKER_POD="${KOKU_WORKER_POD_NAME:-koku-worker-0}"
     oc rsync --delete $KOKU_PATH/testing/pvc_dir/insights_local ${WORKER_POD}:/tmp
-    oc rsync --delete $KOKU_PATH/testing/local_providers/aws_local/* ${WORKER_POD}:/tmp/local_bucket
+    for SOURCEDIR in $(ls -1d $KOKU_PATH/testing/local_providers/aws_local*)
+    do
+        DESTDIR="${WORKER_POD}:$(echo $SOURCEDIR | sed 's#'$KOKU_PATH'#/tmp' | sed 's/aws_local/local_bucket/')"
```
Red-HAP, line 160: Don't you need a slash here?
changing `sed 's#'$KOKU_PATH'#/tmp` to `sed 's/#'$KOKU_PATH'#/tmp`

blentz: Nope. See: https://www.gnu.org/software/sed/manual/html_node/Regexp-Addresses.html#Regexp-Addresses

```
\%regexp%

    (The % may be replaced by any other single character.)

    This also matches the regular expression regexp, but allows one to use a different delimiter than /. This is particularly useful if the regexp itself contains a lot of slashes, since it avoids the tedious escaping of every /. If regexp itself includes any delimiter characters, each must be escaped by a backslash (\).

    The following commands are equivalent. They print lines which start with ‘/home/alice/documents/’:

    sed -n '/^\/home\/alice\/documents\//p'
    sed -n '\%^/home/alice/documents/%p'
    sed -n '\;^/home/alice/documents/;p'

```

Red-HAP: Ah. The `/tmp` threw me.

PR #2208, opened 2020-06-16
COST-83 (part two): Add a migration to populate the source_uuid

myersCody:
##### What problem am I trying to fix
This pr is to address a comment that came up during the QE process regarding nulls being returned in the array for source_uuid.

After some debugging, we discovered that if there is aready data in the database for the daily summary tables when the migrations run for source_uuid, it created the source_uuid field with the default value of null.

However, the actual values for the source_uuid are currently only populated when the sql in masu is run here: https://github.com/project-koku/koku/pull/2191/files#diff-cb553554f25fc3ee6ad997ccb4411e71R41-R42

Therefore, the source_uuid would only be populated when the download endpoint gets hit again, but it doesn't go back recreate the old data inside of the reporting_awscostentrylineitem_daily_summary table, only adds the new stuff which properly handles the source_uuid field.

#### The solution:
To add another migration file that runs after the adding of the source_uuid column to populate the source_uuid.

### How to test
⚠️ **Special Note:** To test this issue you will need to have pg_data setup in the volume. Therefore, if you are mac user that has `#DATABASE_DATA_DIR=psql` set in your env, that will need to commented out and your env reset.

We will need to check out a commit prior to the sources_uuid migration being run; however, at the time there was still a bug with creating aws sources locally.

So in order to test this I needed to cherry pick a certain commit over. However, instead of having y'all do that, I went ahead and did it in a branch to save y'all the trouble.
```
git fetch origin to_test_migrations
make docker-down && make docker-up && docker-compose logs -f koku-server koku-worker masu-server
make clear-testing && make docker-reinitdb-with-sources && make load-test-customer-data
```
From this point on we need to make sure we don't run any commands that would reset our database.
```
make docker-down
git fetch origin issues/cost-83-pt2
git checkout issues/cost-83-pt2
```
On the server terminal, we will want to run:
```
make docker-down && make docker-up && docker-compose logs -f koku-server koku-worker masu-server
```

After that we just need to check the summary tables in the migration to make sure the source_uuid was populated correctly.


koku/reporting/migrations/0115_populate_source_uuid.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [("reporting", "0114_adding_source_uuid")]
+
+    operations = [
+        migrations.RunSQL(
+            sql="""
+UPDATE acct10001.reporting_awscostentrylineitem_daily_summary as aws
```
maskarb, line 12: `acct10001`

What about all the other accounts?

myersCody: Whoops. Great catch @maskarb, I totally forgot to take these out for when I was doing manual testing.

PR #2210, opened 2020-06-17
Scaffold task to convert csv to parquet files.

chambridge:
* Adds pandas & pyarrow for CSV consumption and parquet conversion
* Makefile command to start minimal set of containers for local data consumption

Associated with: https://issues.redhat.com/browse/COST-143

Testing steps:
- Set `.env` with values for: ENABLE_S3_ARCHIVING, S3_BUCKET_NAME, S3_ACCESS_KEY, S3_SECRET
- Run `make docker-up-min`
- Run `make docker-reinitdb-with-sources` to create sources
- Run `make make load-test-customer-data`

Results:
S3 bucket will have the following structure:
data/csv/(account)/(provider_uuid)/(year)/(month)/(csv files)
data/parquet/(account)/(provider_uuid)/(year)/(month)/(parquet files)

- AWS data (CSV and parquet) only keeps the most recent data for the month
- Azure data (CSV and parquet) only keeps the most recent data for the month
- OCP data is special cased since it contains overlapping time windows. As reports are downloaded they are split into daily CSV files for each type (pod_usage, storage_usage, node_labels). Follow on upload will overwrite the overlapping data in S3 as file naming is structured on (report_type).(date).csv. The parquet files are created from the daily CSV reports related to the incoming manifest.

convert_parquet


koku/masu/external/downloader/azure/azure_report_downloader.py
```diff
@@ -29,6 +29,7 @@
 from masu.external.downloader.downloader_interface import DownloaderInterface
 from masu.external.downloader.report_downloader_base import ReportDownloaderBase
 from masu.util.aws.common import copy_local_report_file_to_s3_bucket
+from masu.util.aws.common import get_path_prefix
```
dccurtis, line 32: small nit.  Does it make sense to put these functions in `masu.util.common` since they are used in all downloaders?

dccurtis: on second look.. I can see the reason for AWS now.  Copying to S3 is an AWS thing...

chambridge: Sure, could move them. I think that's a fair point.

I was thinking of it more regarding AWS S3 tech, than how its related to the provider types.

koku/masu/external/downloader/ocp/ocp_report_downloader.py
```diff
@@ -112,6 +120,41 @@ def get_report_for(self, date_time):
 
         return reports
 
+    def detect_type(self, data_frame):
```
dccurtis, line 123: Would be cool to consolidate this report detection to an OCP utility file one day:
https://github.com/project-koku/koku/blob/368f317a025b905210af92c3cddfeeca5bf143dc/koku/masu/processor/ocp/ocp_report_processor.py#L147

koku/masu/processor/tasks.py
```diff
@@ -494,3 +511,99 @@ def autovacuum_tune_schema(schema_name):
...
+            LOG.error(message)
+        if not provider_uuid:
+            message = "missing required argument: provider_uuid"
+            LOG.error(message)
+        if not provider_type:
+            message = "missing required argument: provider_type"
+            LOG.error(message)
+        return
+
+    if not settings.ENABLE_S3_ARCHIVING:
```
dccurtis, line 561: Maybe move this to the top of this function so we don't log any error messages if `ENABLE_S3_ARCHIVING` is turned off?

koku/masu/processor/tasks.py
```diff
@@ -494,3 +511,99 @@ def autovacuum_tune_schema(schema_name):
...
+    s3_parquet_path = get_path_prefix(account, provider_uuid, cost_date, Config.PARQUET_DATA_TYPE)
+
+    file_keys = get_file_keys_from_s3_with_manifest_id(request_id, s3_csv_path, manifest_id, context)
+    files = [os.path.basename(file_key) for file_key in file_keys]
+    if not files:
+        msg = "S3 archiving feature is enabled, but no files to process."
+        LOG.info(log_json(request_id, msg, context))
+        return
+
+    if provider_type == Provider.PROVIDER_AWS or provider_type == Provider.PROVIDER_AWS_LOCAL:
```
dccurtis, line 589: Why is this restricted to just AWS?  Would we ever want to clear OCP or Azure files?

koku/masu/external/kafka_msg_handler.py
```diff
@@ -328,6 +329,8 @@ def extract_payload(url, request_id, context={}):  # noqa: C901
     report_meta["provider_uuid"] = account.get("provider_uuid")
     report_meta["provider_type"] = provider_type
     report_meta["schema_name"] = schema_name
+    report_meta["account"] = schema_name[4:]
```
dccurtis, line 332: This is for `acct1234` -> `1234`?  I guess this won't ever change as long as we don't touch `create_schema_name`

dccurtis: 🤦 I just realized this isn't actually part of the change, I was expanding the view to see more of the function.  Sorry!

koku/masu/external/kafka_msg_handler.py
```diff
@@ -546,29 +549,39 @@ def process_report(report):
...
         "file": report.get("current_file"),
         "compression": UNCOMPRESSED,
         "manifest_id": manifest_id,
         "provider_uuid": provider_uuid,
     }
-    return _process_report_file(schema_name, provider_type, provider_uuid, report_dict)
+    result = _process_report_file(schema_name, provider_type, provider_uuid, report_dict)
+    if start_date:
+        start_date_str = start_date.strftime("%Y-%m-%d")
+        convert_to_parquet.delay(request_id, account, provider_uuid, provider_type, start_date_str, manifest_id)
```
dccurtis, line 583: So `convert_to_parquet` is done on the manifest level right?  Does it make sense to do this after summary (in `summarize_manifest`) since then we would know that all report files have been uploaded and we do this just one time?

koku/masu/external/kafka_msg_handler.py
```diff
@@ -546,29 +549,39 @@ def process_report(report):
...
+    start_date = report.get("date")
 
     report_dict = {
         "file": report.get("current_file"),
         "compression": UNCOMPRESSED,
         "manifest_id": manifest_id,
         "provider_uuid": provider_uuid,
     }
-    return _process_report_file(schema_name, provider_type, provider_uuid, report_dict)
+    result = _process_report_file(schema_name, provider_type, provider_uuid, report_dict)
```
dccurtis, line 580: I think we're going to have to pull the `copy_local_report_file_to_s3_bucket` logic out of the downloader or use the OCP downloader here and call it.  Re-using the downloader logic would help to make it so OCP local could still work...

The recent multi-file OCP payload logic change made it so the OCP downloader isn't actually calling `download_file` anymore.  It's only used for OCP-local.  Sorry for the confusion on that!

koku/masu/external/kafka_msg_handler.py
dccurtis: Here's an example of how we are using a part of the downloader in the real OCP-ingress service flow:
https://github.com/project-koku/koku/blob/368f317a025b905210af92c3cddfeeca5bf143dc/koku/masu/external/kafka_msg_handler.py#L109

Once the worker scaling changes are in I think we can come back to this and clean it up...

koku/masu/processor/tasks.py
```diff
@@ -148,6 +160,11 @@ def get_report_files(
                 raise processing_error
 
     WorkerCache().remove_task_from_cache(cache_key)
+    if start_date:
+        start_date_str = start_date.strftime("%Y-%m-%d")
+        convert_to_parquet.delay(
```
dccurtis, line 165: Similar question from the listener flow.  Should we call this after summary or should it be called for every file?  I'm happy to sort it out when I merge the worker changes, just want to understand if after summary it can be done or not.

koku/masu/external/downloader/azure/azure_report_downloader.py
dccurtis: yeah I can see that.  Maybe split the difference? Put `get_path_prefix` in common and leave `copy_local_report_file_to_s3_bucket` in the aws util?

koku/masu/processor/tasks.py
chambridge: So you saying this plus Andrew's pause in a previous conversation made me go back and look at Azure too (thought it was only spitting out data for that data, but its daily data for the month). Looks like both Azure/AWS reports are the entire month, so we only want the data from the current report so we can remove the previously stored data as we are processing the new batch. OCP data provide daily granularity. So this should be updated to included Azure too.

dccurtis: Think you have OCP and Azure reversed maybe?  Azure is daily.

chambridge: Nope. not reversed. The data in Azure is daily values for the month (report contains multiple days). With the division of OCP data into daily chunked CSVs here: https://github.com/project-koku/koku/pull/2210/files#diff-6918b2e09ba9007181c66df408cc6072R132-R157

koku/masu/external/kafka_msg_handler.py
chambridge: That value is only used in the context logging and is to highlight a problem where sources isn't giving use data. I can allow it to be `None`, but I don't think it will log well. `schema_name` should always have a value with `acct` prefix in this case or we won't have found a matching provider.

koku/masu/external/kafka_msg_handler.py
dccurtis: Another option might be to consolidate and move the `copy_local_report_file_to_s3_bucket` logic from the downloaders and put it into something that could be called in or after `_get_report_files`.  Then that same logic could also be called from the listener.

koku/masu/external/downloader/azure/azure_report_downloader.py
chambridge: Done.

koku/masu/external/downloader/ocp/ocp_report_downloader.py
chambridge: Done. Let me know if it doesn't work for you.

koku/masu/processor/tasks.py
chambridge: Done.

koku/masu/processor/tasks.py
chambridge: Logic updated here.

koku/masu/external/kafka_msg_handler.py
chambridge: Feel free to resolve if you are ok with the above.

koku/masu/external/kafka_msg_handler.py
chambridge: I'm extracting out a function `create_daily_archives` that can be called in the `download_file` and the `extract_payload` flows.

koku/masu/processor/tasks.py
chambridge: Our end goal would be to summarize data from the converted parquet files. Ideally this wouldn't happen for every file as it will convert each file in the manifest to parquet.

However, for now I think it could be done after summary.

dccurtis: Yes, sounds like it could be a parallel task to our summary task.  Then one day we can just turn off our summary task or call the new summary_from_parquet logic once all manifest file line item is complete.

koku/masu/external/kafka_msg_handler.py
chambridge: Update pushed.

koku/masu/processor/tasks.py
chambridge: I moved this call to work/look like the `summarize_manifest` task that gets called.

koku/masu/external/kafka_msg_handler.py
```diff
@@ -536,6 +547,44 @@ def summarize_manifest(report_meta):
...
+    schema_name = report_meta.get("schema_name")
+    manifest_id = report_meta.get("manifest_id")
+    provider_uuid = report_meta.get("provider_uuid")
+    schema_name = report_meta.get("schema_name")
+    provider_type = report_meta.get("provider_type")
+    start_date = report_meta.get("date")
+
+    with ReportManifestDBAccessor() as manifest_accesor:
+        manifest = manifest_accesor.get_manifest_by_id(manifest_id)
+        if manifest.num_processed_files == manifest.num_total_files:
```
dccurtis, line 576: This looks like the same check we do for when to call `summarize_reports.delay`. I'll take it as a TODO for me to consolidate this once I merge the worker changes.  Changes will be needed anyway since I'm moving away from `num_processed_files`

koku/masu/external/kafka_msg_handler.py
```diff
@@ -629,6 +685,9 @@ async def process_messages(msg, loop=EVENT_LOOP):
             async_id = await loop.run_in_executor(pool, summarize_manifest, report_meta)
```
dccurtis, line 685: Think we need to include the `report_id` here.

chambridge: Fixed

PR #2211, opened 2020-06-17
Remove Sources Worker

dccurtis:
Addresses https://issues.redhat.com/browse/COST-162

This is an idea I've been tinkering with to use RPC instead of the old proxy to allow for sources-client to do all database activity when handling PATCH requests for AWS and Azure sources.

If the sources-client does the database saves, the database callbacks work and it makes the sources-worker irrelevant.  Once **all** provider CRUD and database activity is centralized to a single service (sources-client) all source events become serialized and we won't experience the race conditions we've been experiencing with QE tests lately.


maskarb:
Corresponding e2e PR: [https://github.com/RedHatInsights/e2e-deploy/pull/1891](https://github.com/RedHatInsights/e2e-deploy/pull/1891)

maskarb:
Testing (with sources-api running)
1. Follow the UI POST/PATCH request flow.
2. See successful source creation

Take sources-client offline before PATCH:
1. With all services running, do the 4 POSTs to the sources-api
2. Take sources-client down.
3. PATCH to koku...See 424 Failed Dependency

koku/sources/kafka_listener.py
```diff
@@ -501,33 +502,56 @@ def execute_koku_provider_op(msg, cost_management_type_id):
...
 
+    except SourcesProviderCoordinatorError as account_error:
+        raise SourcesIntegrationError("Koku provider error: ", str(account_error))
+    except ValidationError as account_error:
+        err_msg = (
+            f"Unable to {operation} provider for Source ID: {str(provider.source_id)}. Reason: {str(account_error)}"
+        )
+        LOG.error(err_msg)
+        sources_client.set_source_status(account_error)
     except RabbitOperationalError:
```
chambridge, line 550: Do we need this error anymore? We aren't creating a celery task, right?

dccurtis:
@chambridge 
The main point of this PR is to reduce our architectural complexity with how sources are managed.  By having the sources-worker;  account construction now takes two different paths (koku-api-> sources-worker, and sources-client-> sources-worker).  It's proven to be challenging to manage this complexity.

By having the sources-client be the one to do all source crud handling it serializes all of these event streams and makes it much more deterministic.

There are two issues that are noted in the jira description and also I'm seeing this as the contributing factor to some stuck providers while running QE tests.  For this case the celery task failing (due to a cascading delete problem while processing) is not reported back to the sources-client so there is no retry that is happening.

The problems are mainly exposed via API tests.  I'm sure we could add more code to handle these situations but I thought making it simpler and freeing up some OCP quota was the way to go.

koku/sources/kafka_listener.py
dccurtis: yup, think this can go 👍

dccurtis:
As for running multiple sources-clients.  We are not prepared for that today and I'd argue we should make the case to the platform to provide enhancements which would eliminate our requirement for getting cost-specific bits via PATCH API calls first.  Getting rid of our PATCH requirements would also eliminate RPC!  It's something I have been thinking about and am planning on making a proposal to the platform to simplify our integration.  I'm not sure we really have a need to run more than one client though.  It could take up to 24 hours before a user sees anything in our dashboard so I don't think we have a strong use case to ensure very responsive event handling.

koku/sources/kafka_listener.py
maskarb: Don't need the error. I'll remove it.

chambridge:
@dccurtis I don't disagree with you around decreasing the complexity, I would also prefer to eliminate the PATCH. 

I suppose what I'm asking is why add the XMLRPC, just because it gets us another thread and port? Could be a dumb question, but isn't koku-sources already running django with gunicorn? Is it just one worker or can it not process more than one request? Basically, if you are going to make a network call to yourself, why not just add REST entry points and keep it consistent.

I'm sure I'm missing something, but have to ask the dumb questions :smile:

dccurtis:
@chambridge 
I thought the proxy was causing too much confusion!  Maybe I'm not getting it, but it sounds like you are saying to have the koku-api just forward the PATCH request from 3scale to a REST endpoint in sources-client (this is what I mean by saying the koku-api proxy, which was the old implementation).

The main implementation problem was that the sources-client had to make the database saves so the callbacks would trigger account creation/edit/deletes.

I feel like our terminology could be missing each other :). We can discuss later if you want.  

Lets just get rid of our PATCH requirements and make all this moot!  I hope this won't be needed for long...

chambridge:
@dccurtis No, its my confusion. I had it in my head that sources was defined to talk to `koku-sources` and not via 3scale for some reason (which it would have to via the UI). I had to go back and look at https://github.com/project-koku/koku/blob/master/koku/sources/api/urls.py to clarify that `koku-sources` only supports **source-status** and what the old proxy was doing.

🤞 we can remove PATCH sooner rather than later, but I'm fine with this for now. Just needed to walk through it in my head.

PR #2212, opened 2020-06-17
Cost 103 Consolidate Refresh Materialized Views

lrlevine6:
The goal of this issue is to combine the functionality of two very similar methods in two different files: `tasks.py` and `ocp_cloud_summary_updater.py`.  To accomplish this, the `refresh_materialized_views` method from the `ocp_cloud_summary_updater` file was removed and functionality was added to the `refresh_materialized_views` method in the `tasks` file. Unit tests were added to reflect these changes in addition to adding a specification to migrations that allows the `OCP_ON_INFRASTRUCTURE_MATERIALIZED_VIEWS` to be refreshed concurrently.

**Testing:**

1. To test that the materialized views are being refreshed correctly we can use docker-compose to wipe and rebuild the local database:
`make docker-reinitdb-with-sources` or `make create-test-customer` if the database is already clean and does not need to be rebuilt.
2. Check that the database is up and is empty. We can use the query `SELECT * FROM acct10001.reporting_ocpall_cost_summary WHERE usage_start='CURRENT_DATE'`. Where `CURRENT_DATE` is the current date following the example format '2020-06-16'. When run, this should return empty as data has not been populated.
3. We can then load test data into the database using:
`make load-test-customer-data`
4. We can now query the database again using the same query. Two objects should return: one for OCP on AWS and one for OCP on Azure. Since the cost value for Azure is very small we can focus on the OCP on AWS for updates. Take note of the `unblended_cost`. 
5. Now repeat steps 3 and 4 and compare the `unblended_cost`.  These two values should be different displaying that the material views are being updated.


koku/reporting/migrations/0116_ocpall_unique_index.py
```diff
@@ -0,0 +1,26 @@
...
+    operations = [
+        migrations.RunSQL(
+            """
+            DROP INDEX IF EXISTS ocpall_cost_daily_summary;
+            CREATE UNIQUE INDEX ocpall_cost_daily_summary
+            ON reporting_ocpallcostlineitem_daily_summary (source_type, usage_start, cluster_id, namespace, node, usage_account_id, resource_id, product_code, product_family, instance_type, region, availability_zone, tags)
+            ;
+            DROP INDEX IF EXISTS ocpall_cost_project_daily_summary;
+            CREATE UNIQUE INDEX ocpall_cost_project_daily_summary
+            ON reporting_ocpallcostlineitem_project_daily_summary (source_type, usage_start, cluster_id, namespace, node, usage_account_id, resource_id, product_code, product_family, instance_type, region, availability_zone, pod_labels)
```
adberglund, line 18:
```suggestion
            ON reporting_ocpallcostlineitem_project_daily_summary (source_type, usage_start, cluster_id, data_source, namespace, node, usage_account_id, resource_id, product_code, product_family, instance_type, region, availability_zone, pod_labels)
```

PR #2214, opened 2020-06-18
Cost 156 partitioned table prototype

Red-HAP:
## Description

This is a prototype for making a partitioned table out of `reporting_ocpusagelineitem_daily_summary`. 

The main changes to the **koku** API code is adding the OCPUsageLineItemDailySummary model `_meta.managed = False`

The main change to the **masu** processing code is to add a call to a new database procedure called `create_date_partitions` to automatically create needed partitions at data processing time when the table will have new records inserted.

All of these changes rely on a set of migrations to be run. The reason this is a set of migrations is that it will be good to have a `commit` run at certain points in the table conversion.

The basis of these migrations was the work on the `scripts/pg_partable.py` script which will run a one-and-done conversion of tables. It is a starting point for the table conversion. It was created as a script in case, for production, the data needed to be migrated outside of a migration step in case of a pod timeout issue if the existing data needed to be preserved.

The method for these migrations was to leave a net result of the same object names (including indexes, sequence, constraints) for the table as were defined in the model in the hopes that django would not complain about any differences or attempt to unnecessarily create objects because there was a mismatched name. Setting the model to unmanaged helps alleviate this concern as well as give more time to determine the best strategy for primary keys and django modules that support declarative partitioning.

Declarative partitioning was decided upon because it meets our needs; was the simplest to implement; and put the control of these partitions directly on PostgreSQL so that partitions would be created properly and the engine itself would manage from which partitions data are selected and to where data are inserted. Other methods would require upkeep of trigger functions which creates another burden as the maintenance of these functions would have to be undertaken.

## Testing

1. Pull branch
3. `make load-test-customer-data start=2020-02-01 end=2020-04-01`
4. Let masu/worker finish
5. Connect to db (psql, pgAdmin, dbeaver, ...)
6. Examine the tables under `acct10001`
7. You should see: 
    * `reporting_ocplineitemusage_daily_summary` (partitioned table)
    * `reporting_ocplineitemusage_daily_summary_2020_02` (table)
    * `reporting_ocplineitemusage_daily_summary_2020_03` (table)
    * `reporting_ocplineitemusage_daily_summary_default` (table)

A record count from the `2020_02` and `2020_03` tables should show records. 
A count from the `default` table should show zero records. 
A count from `reporting_ocplineitemusage_daily_summary` should be the sum of the 2020 tables' records. 

## Developer Impact

The biggest change for this prototype is that migration for the OCPUsageLineitemDailySummary model must be made manually at this time. I have evaluated a few django modules that appear to support declarative partitioning, but neither of them would work properly. One failed to even load. So if a change to the model is needed, then a manual migration should be done. This also eliminates any worry about a composite primary key. Manually creating these migrations will give us time to either find a module that does work or find a better PK solution.

In terms of API work, for optimal performance, every request should have some sort of time window parameter to use against usage_start. This is not a strict requirement, but leaving a time parameter out will cause the scan of all partitions, so it is highly recommended to be included. Otherwise, there should not be that much of an impact on the developer experience working on koku.

## Jira Ticket

https://issues.redhat.com/browse/COST-156
https://issues.redhat.com/browse/COST-277   (testing ticket)


db_functions/record_hash.sql
```diff
@@ -0,0 +1,55 @@
+/*
+Copyright 2020 Red Hat, Inc.
```
adberglund, line 2: Is record_hash still being used?

koku/reporting/migrations/0116_partitioned_ocplineitemusage_daily_summary.py
```diff
@@ -0,0 +1,160 @@
...
+    dependencies = [("api", "0021_db_functions"), ("reporting", "0115_populate_source_uuid")]
+
+    operations = [
+        migrations.RunSQL(
+            """
+-- ===================================
+--  Create tracking table
+-- ===================================
+
+CREATE TABLE IF NOT EXISTS partitioned_tables (
```
adberglund, line 16: I thikn the partitioned_tables table could be built from a Django model since it's okay for it to manage this.

db_functions/record_hash.sql
Red-HAP: Nah, I can take it out.

koku/reporting/migrations/0116_partitioned_ocplineitemusage_daily_summary.py
Red-HAP: Yes.

Red-HAP: done

PR #2223, opened 2020-06-22
[COST-206] foundation to switch listener to DeploymentConfig

maskarb:
1. switch from aiokafka to confluent_kafka consumer and producer.
2. use ciso8601 to parse dates in ocp report processor.
3. introduce a `kafka_utils` folder. We utilize very similar code between the koku-listener and the sources-client. A util file could help use reduce the amount of repeated code and make maintenance a little easier. (this PR does not touch the sources-client. We can simplify things there in a follow up PR)

corresponding e2e PR: [https://github.com/RedHatInsights/e2e-deploy/pull/1873](https://github.com/RedHatInsights/e2e-deploy/pull/1873)

~TODO:~ Done
~the `max.poll.interval.ms` may need to be adjusted upwards, currently set to 15 minutes. An approx. 80 MB file ingestion takes approx. 12 min in the perf cluster, but the maximum file size is 100 MB. Need to test a file at the max allowable size to ensure we do not timeout between polls.~

A 99.1 MB file was ingested in the performance cluster. The interval was bumped to 18 minutes.

Testing:
- Testing was performed in the performance cluster using the corresponding e2e template config. With 3 listeners running, large ocp cost and usage reports were sent to ingress and the files were processed without issues in the listeners. 
- additional testing - Listener re-deploy during processing:
1. Send large files to listeners. Once processing is complete, make note of the line count in one of the OCP tables.
2. Delete all the data
3. Ingest the same files, but redeploy all the listeners during processing.
4. Count the lines again and note that the line count matches the first ingestion.

Local tests:
- consumer polling errors (with ingress running):
1. set the consumer timeout to 60,000 ms, bring up listener
2. in nise: `nise yaml ocp -c default -r -o large_ish.yml --ocp-cluster-id my-cluster-id`
3. add an OCP source with cluster-id my-cluster-id
4. from nise: `INSIGHTS_ACCOUNT_ID=10001 INSIGHTS_ORG_ID=123 nise report ocp --file-row-limit 1800000 --ocp-cluster-id my-ocp-cluster --static-report-file large_ish.yml --insights-upload http://localhost:8080/api/ingress/v1/upload`
5. Watch the listener logs. See a timeout error similar to this:
```
koku_listener     | [2020-06-25 16:15:40,695] ERROR None [listen_for_messages_loop] consumer.poll message: <cimpl.Message object at 0x7f41f437df48>. Error: KafkaError{code=_MAX_POLL_EXCEEDED,val=-147,str="Application maximum poll interval (60000ms) exceeded by 72ms"}
```
6. See the consumer continue consuming messages after logging the error.


koku/masu/external/kafka_msg_handler.py
```diff
@@ -721,26 +702,46 @@ async def process_messages(msg, loop=EVENT_LOOP):
...
-async def listen_for_messages_loop(event_loop):
+def listen_for_messages_loop():
     """Wrap listen_for_messages in while true."""
-    while True:
-        consumer = get_consumer(event_loop)
-        await listen_for_messages(consumer)
+    consumer = get_consumer()
+    LOG.info("Consumer is listening for messages...")
+    for _ in itertools.count():  # equivalent to while True, but mockable
+        msg = consumer.poll(timeout=1.0)
```
dccurtis, line 726: Looks like this will essentially call `consume()` internally and will block for a maximum of 1s waiting for a new message?  Do you know if it's possible to get two messages within 1 second?

koku/masu/external/kafka_msg_handler.py
```diff
@@ -721,26 +702,46 @@ async def process_messages(msg, loop=EVENT_LOOP):
...
+        if msg is None:
+            continue
+
+        if msg.error():
+            KAFKA_CONNECTION_ERRORS_COUNTER.inc()
+            if msg.error().code() == KafkaError._PARTITION_EOF:
```
dccurtis, line 732: What circumstances lead to this? From the docs this is `Broker: No more messages`?

koku/masu/external/kafka_msg_handler.py
maskarb: From what I have seen, this will only take one message at a time.

koku/masu/external/kafka_msg_handler.py
```diff
@@ -721,26 +702,46 @@ async def process_messages(msg, loop=EVENT_LOOP):
...
+                # End of partition event
+                LOG.warning(
+                    f"[listen_for_messages_loop] {msg.topic()} {msg.partition()} reached end at offset {msg.offset()}"
+                )
+                continue
+            elif msg.error():
+                raise KafkaException(msg.error())
```
dccurtis, line 739: This will be game over for the listener?  I don't see this being caught in `koku_listener_thread`

koku/masu/external/kafka_msg_handler.py
```diff
@@ -762,73 +763,56 @@ async def listen_for_messages(consumer):
...
-
-    except KafkaError as error:
-        LOG.error(f"[listen_for_messages] Kafka error encountered: {type(error).__name__}: {error}", exc_info=True)
-        KAFKA_CONNECTION_ERRORS_COUNTER.inc()
+        offset = msg.offset()
+        partition = msg.partition()
+        LOG.info(f"Processing message offset: {offset} partition: {partition}")
+        process_messages(msg)
+        LOG.debug(f"COMMITTING: message offset: {offset} partition: {partition}")
+        consumer.store_offsets(msg)
```
dccurtis, line 778: I see `‘enable.auto.offset.store’ must be set to False when using this API` in the docs https://docs.confluent.io/current/clients/confluent-kafka-python/#confluent_kafka.Consumer.store_offsets

Do you know what that's about?

koku/masu/external/kafka_msg_handler.py
```diff
@@ -58,25 +58,26 @@
...
-def backoff(interval, maximum=64):  # pragma: no cover
-    """Exponential back-off."""
-    wait = min(maximum, (2 ** interval)) + random.random()
-    LOG.info("Sleeping for %.2f seconds.", wait)
-    time.sleep(wait)
+def delivery_callback(err, msg):
+    """Acknowledge message success or failure."""
+    if err is not None:
+        LOG.error(f"Failed to deliver message: {str(msg)}: {str(err)}")
+        raise KafkaMsgHandlerError(err)
```
dccurtis, line 78: Does this callback happen in the same thread as `send_confirmation`?  normally callback mechanisms would not.. Just curious how this exception would be handled if that were the case

dccurtis: Discussed, we addressed it

PR #2233, opened 2020-06-30
Update Azure daily summary SQL for unit normaliz.

adberglund:
## Summary 
This is a SQL change to Azure's Daily Summary table to normalize usage units. For example, VM usage is recorded in 100 Hour blocks, this conversts down to Hours and multiplies the usage effectively. Same thing for 100 GB/Month for storage. IN cases where there isn't a value and unit in the `unit_of_measure` column, the original value is maintained and usage is simply multipled by `1`. This should be considered incomplete work and might actually cause smoke test failures as we have changed usage values. But this enables the API work for https://issues.redhat.com/browse/COST-241.


koku/masu/database/sql/reporting_azurecostentrylineitem_daily_summary.sql
```diff
@@ -1,29 +1,56 @@
 -- Place our query in a temporary table
 CREATE TEMPORARY TABLE reporting_azurecostentrylineitem_daily_summary_{{uuid | sqlsafe}} AS (
+    WITH cte_split_units AS (
+        SELECT li.id,
+            m.currency,
+            CASE WHEN split_part(m.unit_of_measure, ' ', 2) != '' AND m.unit_of_measure != '100 Hours'
```
adberglund, line 6:
```suggestion
            CASE WHEN split_part(m.unit_of_measure, ' ', 2) != '' AND NOT (m.unit_of_measure = '100 Hours AND m.meter_category='Virtual Machines')'
```

Feel free to ignore my use of NOT replacing the != you had, but I do think adding in the meter_category is good for now since we only know for sure that the virtual machine data is impacted

koku/api/report/azure/provider_map.py
```diff
@@ -221,15 +223,18 @@ def __init__(self, provider, report_type):
...
                             "usage": Sum("usage_quantity"),
-                            "usage_units": Coalesce(Max("unit_of_measure"), Value("Storage Type Placeholder")),
+                            "usage_units": Coalesce(Max("unit_of_measure"), Value("GB-Mo")),
                             "source_uuid": ArrayAgg(F("source_uuid"), distinct=True),
                         },
                         "delta_key": {"usage": Sum("usage_quantity")},
-                        "filter": [{"field": "service_name", "operation": "icontains", "parameter": "Storage"}],
+                        "filter": [
+                            {"field": "service_name", "operation": "icontains", "parameter": "Storage"},
+                            # {"field": "unit_of_measure", "operation": "icontains", "parameter": "-Mo"},
```
adberglund, line 232:
```suggestion
                            # {"field": "unit_of_measure", "operation": "icontains", "parameter": "GB-Mo"},
```

PR #2242, opened 2020-07-06
adjusted cache file to handle use of DummyCache 

cgoodfred:
When deleting a source and having view caching disabled a KokuCacheError was raised when trying to invalidated caches since DummyCache wasn't a caching backend that was handled. DummyCache is now on the list since that is being used to disable view caching.

To test: 
1. Disable view caching by changing "CACHED_VIEWS_DISABLED"  to True in .env
2. add a source and go to the following endpoint: [http://localhost:8000/api/cost-management/v1/sources/](http://localhost:8000/api/cost-management/v1/sources/) to see the source
3. go to the source and attempt to delete it, before a KokuCacheError was raised and now it should not be.

### DIFF
--- a/koku/koku/tests_cache.py
+++ b/koku/koku/tests_cache.py
@@ -75,11 +75,7 @@ class KokuCacheTest(IamTestCase):
         self.assertIsNone(self.cache.get(key_to_clear))
         self.assertIsNotNone(self.cache.get(remaining_key))
 
-    @override_settings(
-        CACHES={
-            "default": {"BACKEND": "django.core.cache.backends.dummy.DummyCache"}
-        }
-    )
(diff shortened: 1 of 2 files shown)


koku/koku/tests_cache.py
```diff
@@ -75,6 +75,16 @@ def test_invalidate_view_cache_for_tenant_and_cache_key(self):
         self.assertIsNone(self.cache.get(key_to_clear))
         self.assertIsNotNone(self.cache.get(remaining_key))
 
+    @override_settings(
+        CACHES={
+            "default": {"BACKEND": "django.core.cache.backends.dummy.DummyCache", "LOCATION": "worker_cache_table"}
```
maskarb, line 80:
```suggestion
            "default": {"BACKEND": "django.core.cache.backends.dummy.DummyCache"}
```
The DummyCache doesn't set a location in the settings.py, so don't include one here.

PR #2249, opened 2020-07-07
COST-309: Use ciso8601 to parse dates in azure report processor.

myersCody:
For a quick way to test this: https://gist.github.com/myersCody/c85600085fad13c146c566d9e3f27962

### DIFF
--- a/koku/masu/processor/azure/azure_report_processor.py
+++ b/koku/masu/processor/azure/azure_report_processor.py
@@ -20,6 +20,7 @@ import logging
 from datetime import datetime
 from os import remove
 
+import ciso8601
 import pytz
 import ujson as json
 from dateutil import parser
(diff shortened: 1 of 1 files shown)


koku/masu/processor/azure/azure_report_processor.py
```diff
@@ -120,9 +121,11 @@ def _create_cost_entry_bill(self, row, report_db_accessor):
 
         report_date_range = utils.month_date_range(parser.parse(row_date))
         start_date, end_date = report_date_range.split("-")
+        start_date = start_date.replace(" +0000 UTC", "+0000")
+        end_date = end_date.replace(" +0000 UTC", "+0000")
```
maskarb, line 125: I'm not sure we need these here. Azure doesn't have the weird ` +0000 UTC` in its datetime strings.

myersCody: Sweet! I can totally take it out then. I saw that it didn't have it in our nise data, but I didn't know if the nise reports were exactly like azure reports so I added it to be safe. But if we are certain its not needed ill remove it.

PR #2251, opened 2020-07-08
[COST-315] Cost model date issue

cgoodfred:
When adding a cost-model to a source, a ValidationError was occurring in the koku-worker logs and preventing the cost model from being added correctly. The following error was happening:
```
django.core.exceptions.ValidationError: ["'2020-07-01T00:00:00Z' value has an invalid date format. It must be in YYYY-MM-DD format."]
```
and with the changes to the dates in cost_model_manager that should no longer happen.

Testing:

1. Have the koku-worker logs up and running using the following command:
```
docker-compose logs -f koku-worker
```

2. Create some azure data using nise (you'll need to fill in your path to the project):
```
nise report azure -s 2020-07-01 --azure-container-name /path/to/project/koku/testing/local_providers/azure_local --azure-report-name cur
```

3. Create a source using a POST request to the sources endpoint ([http://localhost:8000/api/cost-management/v1/sources/](http://localhost:8000/api/cost-management/v1/sources/)) with some needed info, for example:
```
{
    "name": "Test Azure Source",
    "source_type": "AZURE-local",
    "authentication": {
        "credentials": {
            "subscription_id": "12345678-1234-5678-1234-567812345679",
            "tenant_id": "12345678-1234-5678-1234-567812345679",
            "client_id": "12345678-1234-5678-1234-567812345679",
            "client_secret": "12345"
        }
    }, "billing_source": {
        "data_source": {
            "resource_group": {
                "directory": "",
                "export_name": "cur"
            },
            "storage_account": {
                "local_dir": "/tmp/local_container",
                "container": ""
            }
        }
    }
}
```

4. Get the uuid for the source by going to the sources endpoint ([http://localhost:8000/api/cost-management/v1/sources/](http://localhost:8000/api/cost-management/v1/sources/)) and finding the source you just created.

5. Try adding a cost model to that source using a POST request to the cost-models endpoint ([http://localhost:8000/api/cost-management/v1/sources/](http://localhost:8000/api/cost-management/v1/sources/)) with the uuid from above, for example:
```
{
    "name": "Test Cost Model",
    "description": "Test",
    "source_type": "OCP",
    "source_uuids": "{uuid}"
    "markup": {"value": 10, "unit": "percent"}
}
```

6. After that request, you should not see the following ValidationError in the koku-worker logs:
```
django.core.exceptions.ValidationError: ["'2020-07-01T00:00:00Z' value has an invalid date format. It must be in YYYY-MM-DD format."]
```

7. Go to the following masu endpoint to ensure that everything is the same since update_cost_model_costs was changed and you do not see the Validation Error in the worker-logs (using the uuid from above): http://localhost:5000/api/cost-management/v1/update_cost_model_costs/?provider_uuid={uuid}_&schema=acct10001&start_date=2020-07-01


koku/cost_models/cost_model_manager.py
```diff
@@ -84,8 +84,8 @@ def update_provider_uuids(self, provider_uuids):
                 raise CostModelException(log_msg)
             CostModelMap.objects.create(cost_model=self._model, provider_uuid=provider_uuid)
 
-        start_date = DateHelper().this_month_start
-        end_date = DateHelper().today
+        start_date = str(DateHelper().this_month_start.date())
+        end_date = str(DateHelper().today.date())
```
blentz, line 88: While casting to string may work, I think using `.strftime()` is more readable and ensures the string formatting doesn't change unexpectedly.

cgoodfred: 👍  Good idea!

PR #2253, opened 2020-07-08
COST-331: sources logging fix

maskarb:
After the sources-worker removal, the following logs were seen in sources-client logs for every Source.update kafka event:
```
None Cost management does not allow duplicate accounts. {provider} already exists. Edit source settings to configure a new source.
```
This PR makes the Provider update method check for duplicate providers only when the billing or authentication information changes.


koku/api/provider/serializers.py
```diff
@@ -394,8 +397,9 @@ def update(self, instance, validated_data):
 
             try:
                 instance.save()
-            except IntegrityError:
-                raise serializers.ValidationError(error_obj(ProviderErrors.DUPLICATE_AUTH, message))
+            except IntegrityError as error:
+                message = f"Unknown IntegrityError: {error}"
```
dccurtis, line 401: we might want to log this error and leave `message` to be the user facing string from above.  Otherwise this IntegrityError would be pushed to the sources UI

maskarb: Will we only hit IntegrityError if there is a duplicate? If that's the case, then I would rather remove this try/except altogether since I moved the raise to where we detect the duplicate.

dccurtis: Yes, lets remove this try/except

maskarb: Done.

PR #2260, opened 2020-07-10
[COST-311] Cost Model costs remain after deleting cost model

cgoodfred:
When a cost_model was deleted from a source, it would skip updating the cost models because it was not associated with a source. Now when a model is deleted it will update the cost models and refresh the materialized views.

Some Notes:
- Changed update_cost_model_costs task to no longer check if a cost model exists and just run the update, moved the check to update_summary_tables. Moved because when a cost_model was deleted, it would not update cost_model_costs since there was no cost_model.
- Refactored update_summary_tables for more logical flow

Testing:
1. Add an azure source
2. Add a cost model to that source:
```
{
    "name": "Test Cost Model",
    "description": "Test",
    "source_type": "Azure",
    "source_uuids": ["{uuid}"],
    "markup": {"value": 10, "unit": "percent"}
}
```
3. Check the costs endpoint: [http://127.0.0.1:8000/api/cost-management/v1/reports/azure/costs/](http://127.0.0.1:8000/api/cost-management/v1/reports/azure/costs/)
4. Delete the cost model
5. Check the costs endpoint again and see that the markup is 0.


koku/masu/processor/tasks.py
```diff
@@ -352,6 +352,9 @@ def update_cost_model_costs(schema_name, provider_uuid, start_date=None, end_dat
         updater = CostModelCostUpdater(schema_name, provider_uuid)
         if updater:
             updater.update_cost_model_costs(start_date, end_date)
+            if provider_type:
+                refresh_materialized_views.delay(schema_name, provider_type)
```
dccurtis, line 356: Since this is already being called in a worker task.  I think we could remove this `delay` to run it synchronously.

koku/masu/api/update_cost_model_costs.py
```diff
@@ -53,6 +54,11 @@ def update_cost_model_costs(request):
 
     LOG.info("Calling update_cost_model_costs async task.")
 
-    async_result = cost_task.delay(schema_name, provider_uuid, start_date, end_date)
+    try:
+        provider = Provider.objects.get(uuid=provider_uuid)
+    except Provider.DoesNotExist:
+        return Response({"Error": "Provider does not exist."}, status=status.HTTP_400_BAD_REQUEST)
+    else:
+        async_result = cost_task.delay(schema_name, provider_uuid, start_date, end_date, provider.type)
```
dccurtis, line 62: I need to think on the implications of making this asynchronous.  The main concern I have is that the API is synchronous but if the work to update the report data is asynchronous, it might impact QE tests since from they will not know when the actual processing is complete.

maskarb: The only thing that was added here was the try/except. This was already an async call.

koku/masu/processor/tasks.py
cgoodfred: Removed the `delay`.

PR #2261, opened 2020-07-13
removing num_files_processed

dccurtis:
Removing the `num_files_processed` field on `CostUsageReportManifest` and instead doing a query against `CostUsageReportStatus` to determine if all report files have been processed.

This logic did not handle multiple workers and listeners well.  The file count bump is not accurate since the field was not protected by a lock.  Because of this we are seeing reports failing to summarize in smoke and vortex.

**Testing**
* Smokes passed
* Manual OCP listener tests coming next


maskarb:
Testing:
1. With ingress running, create 3 kafka partitions. Log into kafka container:
```
kafka-topics --bootstrap-server localhost:29092 --alter --topic platform.upload.hccm --partitions 3
```
2. Bring up 3 koku-listeners
3. Send data from nise to ingress:
```
INSIGHTS_ACCOUNT_ID=10001 INSIGHTS_ORG_ID=123 nise report ocp --file-row-limit 10000 --ocp-cluster-id my-cluster-id --insights-upload http://localhost:8080/api/ingress/v1/upload -s 2020-07-01
```
4. Ensure all three listeners are processing files.
5. When all is done, check that the `"files_processed": "x/x"` shows all the files as complete.

Issue #264, opened 2018-07-12
Locale support


blentz:
## User Story
As koku user, I want koku to be locale-aware so that I can see location-dependent details (currency, time zone, etc.) in my locale.

Note: This is a pre-requisite for #19 and #20

## Impacts
API, UI

## Role
HCCM Consumer

## Assumptions
- Units data is stored in a consistent manner
- 

## API Details

| API  | Detail|
 | --- | --- |
 | **Method** | GET/POST/PUT/PATCH/DELETE |
 | **Path** | /api/v1/?/ |
 | **Body** |  |
 | **Response** | |

## UI Details
- User should be able to set Locale preferences - currency, time zone

## Acceptance Criteria

- [x] Currency preference can be set 
- [x] Time zone preference can be set
- [ ] API queries return data in the user's preferred Currency or Time Zone

blentz:
After doing some investigation, I'm flagging this as needing additional discussion on what exactly we're looking to do and what the use cases for it are.

Requirements:
1. At a bare minimum, we can *not* display costs that are different than what's on the customer's invoice. If we tell somebody they should expect to spend $25k and their invoice says they owe $22k or $28k, that's going to make it very difficult to trust our tool's analysis.

Timezones:

Right now, AWS uses UTC for all datetimes. Everything we get from them is UTC.

1. Converting timezones during ingestion is a non-starter. It will cause too many problems for cost-calculations.

2. Converting timezones during reporting could cause confusion when Amazon says "Start date: 1-Sept-2018 00:00:00+0" and we tell the customer "Start date: 31-Aug-2018 22:00:00-2".   

This problem only gets worse when factoring in DST and the fact that for regions that observe DST, there are two distinct hours that are both 01:00:00-02:00:00.

Currencies:

Currency conversions are time-sensitive, as currency rates fluctuate over time. This means that if we include currency conversions in our estimates, that introduces variance and increases our margins of error. 

Just like with time zones, this can be seen as a negative by the customer and may reduce the trust they have in our application.


So, for both of these, I think we need to clarify the use cases and define the expected behaviors more precisely.

chargio:
Timezones can get complicated when more than one provider is added. As soon as we add some provider that is not aligned with UTC (inc. on-premises), we will be needing to work this out.

https://aws.amazon.com/premiumsupport/knowledge-center/non-us-currency/

In Amazon, exchange rates are calculated when the invoice is generated, so any currency conversion will vary until the final moment of the invoice.

[...]

PR #2265, opened 2020-07-14
COST-224: update OCP test data to align with Nise PR #264

blentz:
This PR lines up the report test data to align with the changes in this Nise PR: https://github.com/project-koku/nise/pull/264


koku/api/report/test/ocp_aws_static_data.yml
```diff
@@ -4,135 +4,121 @@ generators:
...
               pods:
-                - pod:
-                  pod_name: pod_name3
+                - pod_name: pod_name3
                   cpu_request: 1
                   mem_request_gig: 2
                   cpu_limit: 1
                   mem_limit_gig: 4
                   pod_seconds: 3600
-                  labels: label_environment:qe|label_app:cost|label_version:beta
```
adberglund, line 72: These had `label_` on them because that's how our real data looks like coming from OpenShift. I think in dropping this we drift farther from a real report.

koku/koku/koku_test_runner.py
```diff
@@ -46,7 +46,6 @@ class KokuTestRunner(DiscoverRunner):
 
     account = "10001"
     schema = f"acct{account}"
-    settings.HOSTNAME = "koku-worker-10-abcdef"
```
adberglund, line 49: Do we know why this was set to begin with?

koku/api/report/test/aws/tests_queries.py
```diff
@@ -691,7 +696,8 @@ def test_execute_query_curr_month_by_avail_zone(self):
             month_data = data_item.get("azs")
             self.assertEqual(month_val, cmonth_str)
             self.assertIsInstance(month_data, list)
-            self.assertEqual(len(month_data), self.availability_zone_count_cur_mo)
+            # Add 1 to the count for "no-az"
+            self.assertEqual(len(month_data), self.availability_zone_count + 1)
```
adberglund, line 700: This test was just changed in https://github.com/project-koku/koku/pull/2332 to fix an intermittent test error. Want to make sure this isn't re-introducing that.

koku/koku/koku_test_runner.py
dccurtis: this can be removed.  It was a leftover from one of the worker cache fix iterations.

koku/api/report/test/ocp_aws_static_data.yml
blentz: Ok. I can add it back, but it needs to match the labels in the AWS tag data. So, the AWS tags also need to include `label_`, which they didn't.

As far as I can tell, Masu wasn't stripping the `label_` bit off and the data-as-is wasn't lining up for the ocp-on-<cloud> stuff to work.

So, there's some behavior here that probably needs documenting, if you are able to talk through what it's supposed to be.

koku/koku/koku_test_runner.py
blentz: This might be an error in merge conflict resolution. I'll double-check.

koku/api/report/test/aws/tests_queries.py
blentz: I'll look into it. Thanks for catching it.

blentz: I've re-aligned this with the newer changes from #2332

koku/api/report/test/ocp_aws_static_data.yml
blentz: done!

koku/api/report/test/utils.py
```diff
@@ -188,18 +202,25 @@ def load_aws_data(self, customer, static_data_file, account_id=None, role_arn=No
...
             report_path = self.build_report_path(provider_type, bill_date, base_path)
-            for report in os.scandir(report_path):
-                if os.path.isdir(report):
-                    for report in [f.path for f in os.scandir(f"{report_path}/{report.name}")]:
-                        if os.path.isdir(report):
-                            continue
-                        elif "manifest" in report.lower():
-                            continue
-                        self.process_report(report, "GZIP", provider_type, provider, manifest)
+            try:
```
maskarb, line 205: I'm not sure we need to catch exceptions here. The koku-test-runner has broad exception handling, so anything that gets raised here is caught there.

maskarb (commented):
Some comments:

koku/api/report/test/utils.py
blentz: That might be true, but I don't think that's desirable behavior.

The main reason for catching the IOError here is to log the error and have something meaningful in the console output if there's a failure processing report data.

PR #2271, opened 2020-07-18
Cost 100 ocp on infra sql updates

adberglund:
## Summary
Work to make OCP on Infra SQL run faster

## Testing
Running this branch against master with a full month of data locally went from 
1 minute to run OCP on AWS to 35 seconds. Was reaaaalllly hoping for a bigger gain, and maybe there's more to do, but I think the big gains here would be on a real customer who has a lot of AWS data that isn't OCP related. We will do a much better job of filtering that out before matching now.

Smoke Tests
```
==== 3947 passed, 80 skipped, 1709 deselected in 1439.44s (0:23:59) =====
```


koku/reporting/migrations/0121_auto_20200728_2258.py
```diff
@@ -56,12 +56,15 @@ class Migration(migrations.Migration):
...
 
+                DROP INDEX IF EXISTS ocpall_cost_daily_summary;
                 DROP INDEX IF EXISTS ocpallcstdlysumm_node;
                 DROP INDEX IF EXISTS ocpallcstdlysumm_node_like;
                 DROP INDEX IF EXISTS ocpallcstdlysumm_nsp;
                 DROP INDEX IF EXISTS ocpall_product_code_ilike;
+
                 DROP MATERIALIZED VIEW IF EXISTS reporting_ocpallcostlineitem_daily_summary;
 
+                DROP INDEX IF EXISTS ocpall_cost_project_daily_summary;
```
maskarb, line 67: 🤔  Is this allowed? Modifying an old migration?

adberglund: 🙊 🤐 😯

PR #2276, opened 2020-07-22
COST-370 - Sources warning and prometheus metrics

maskarb:
[COST-370](https://issues.redhat.com/browse/COST-370)

1. log Provider issues as log.warning
2. Add prometheus metrics to increment when we retry either processing a kafka msg or a provider operation (create, update, etc.)

Corresponding e2e: [https://github.com/RedHatInsights/e2e-deploy/pull/2005](https://github.com/RedHatInsights/e2e-deploy/pull/2005)


chambridge:
Could you open a JIRA issue for this and include the issue number in the squash commit so Luke can reference it for what should be tested?

PR #2302, opened 2020-08-04
[COST-382] RBAC not respecting exact name matching

cgoodfred:
Previously, when a user would request some information the filters were adjusted to only include the information they had access to through RBAC. The changes proposed here will now return a 403 if a request is made for information the user does not have access too and will respect exact matching on RBAC restrictions instead of contains.

Example: if you only had access to account 9999999999 and hit `aws/costs/?filter[account]=589173575009&filter[account]=9999999999` the returned information would only show you results from account 9999999999 even though you asked for both. Now this will result in a 403 since you do not have access to the other account. 

The RBAC exact matching is a little different, previously if you had access to `project-1` you could also pull up the information on `project-1-2` because it contains `project-1` now that is not possible since the filter was updated to only include items in the users access.

Testing:

1. Load in some data, for example `make load-aws-org-unit-tree`

2. Set your RBAC access in your .env to something specific, such as: `"access": {"aws.account": {"read": ["*"]}, "aws.organizational_unit": {"read": ["R_001", "OU_001"]}}`

3. Attempt to load an endpoint relating to your access such as: `reports/aws/costs/?filter[org_unit_id]=OU_001&filter[org_unit_id]=OU_002` and you should see a 403 (this is based off a discussion in koku dev) whereas previously you would have seen data associated with only OU_001

4. Generate some nise data with specific node names that overlap such as `pretty-specific` and `pretty-specific-1` and set your access on one of those nodes. For example: `"access": {"openshift.node":{"read": ["pretty-specific"]}}. I used this yaml (github doesn't let you upload a yaml so remove the .txt) [ocp-data-test.yml.txt](https://github.com/project-koku/koku/files/5024507/ocp-data-test.yml.txt) 

5. Create a source for that data, *don't forget to add a cost-model to the OCP source (i forgot this step in the instructions)*,and go to `/reports/openshift/costs/` and note the total, it should be the same as the cost in `/reports/openshift/costs/?filter[node]=pretty-specific` whereas previously, the cost would include the cost from `pretty-specific-1`


koku/api/report/aws/query_handler.py
```diff
@@ -469,6 +470,22 @@ def __resolve_conditions(condition, alias="t", where=None, values=None):
...
+                filters.add(q_filter)
+        elif filt["field"] == "organizational_unit__org_unit_path":
+            filt["field"] = "organizational_unit__org_unit_id"
+            filt["operation"] = "in"
+            q_filter = QueryFilter(parameter=access, **filt)
+            filters.add(q_filter)
+        else:
+            filt["operation"] = "in"
+            q_filter = QueryFilter(parameter=access, **filt)
+            filters.add(q_filter)
```
maskarb, line 487: The only real difference here between this func and the one in queries.py is the elif. Maybe this could be a bit cleaner?
```suggestion
        if not isinstance(filt, list) and filt["field"] == "organizational_unit__org_unit_path":
            filt["field"] = "organizational_unit__org_unit_id"
       super()._set_access_filters(self, access, filt, filters)
```

koku/api/report/queries.py
```diff
@@ -211,6 +221,17 @@ def _get_search_filter(self, filters):
         LOG.debug(f"_get_search_filter: {composed_filters}")
         return composed_filters
 
+    def _set_access_filters(self, access, filt, filters):
```
maskarb, line 224: Function needs a docstring.

cgoodfred: added

koku/api/report/test/tests_queries.py
```diff
@@ -347,6 +347,47 @@ def test_set_operator_specified_tag_filters_or(self):
         self.assertIsInstance(output, QueryFilterCollection)
         assertSameQ(output.compose(), expected.compose())
 
+    def test_set_access_filter(self):
```
maskarb, line 350: Can you make this 2 tests? One for the list, the other for the non-list.

cgoodfred: Updated to be separate tests.

koku/api/organizations/queries.py
```diff
@@ -131,6 +131,13 @@ def _get_filter(self, delta=False):  # noqa: C901
                 for item in filter_value:
                     q_filter = QueryFilter(parameter=item, **filter_obj)
                     filters.add(q_filter)
+            access = self.parameters.get_access(filter_key)
+            filt = self.FILTER_MAP.get(filter_key, [])
+            if access and filt:
+                filt = self.FILTER_MAP.get(filter_key)
```
abaiken, line 137: i don't think you need to redefine `filt` right here since it is defined on line 135.

cgoodfred: Good call, removed

PR #2303, opened 2020-08-05
COST-376: Fix bug order by delta during pagination.

myersCody:
Jira Issue: https://issues.redhat.com/browse/COST-376

This pr is to fix a bug that bubbled up during my work on adding default pagination. I noticed that once I added the default pagination that the `order_by[delta]` tests were failing with a 500 error. 

#### Replicate bug on master

Hit these endpoints combinations cause a 500 fail before customer data is added and after the customer data is present. 

- http://127.0.0.1:8000/api/cost-management/v1/reports/aws/instance-types/?order_by[delta]=asc
- http://127.0.0.1:8000/api/cost-management/v1/reports/aws/instance-types/?filter[limit]=5&filter[offset]=0&order_by[delta]=asc

---

#### How to test

Pull down this branch and bring up the server. 
Before loading customer data hit these endpoints:

- http://127.0.0.1:8000/api/cost-management/v1/reports/aws/instance-types/?order_by[delta]=asc
- http://127.0.0.1:8000/api/cost-management/v1/reports/aws/instance-types/?filter[limit]=5&filter[offset]=0&order_by[delta]=asc

Notice that they no longer fail.

Then add some customer data and hit the same points again to notice that they no longer 500 & order by the delta.


koku/api/report/aws/openshift/query_handler.py
```diff
@@ -61,9 +61,20 @@ def execute_query(self):  # noqa: C901
                     account_alias=Coalesce(F(self._mapper.provider_map.get("alias")), "usage_account_id")
                 )
 
-            if self._limit:
-                rank_order = getattr(F(self.order_field), self.order_direction)()
-                rank_by_total = Window(expression=RowNumber(), partition_by=F("date"), order_by=rank_order)
+            # Set default delta if one was not passed in.
+            if self.order_field == "delta":
+                if self._delta is None:
```
adberglund, line 66: Do we know why self._delta isn't being set when we are ordering by delta? Could this be a bug in the QueryParams class?

myersCody: @adberglund So, I had not considered that this might be a bug in the QueryParameters class. The `self._delta` is set by passing it into the url. So you will notice in my tests I use two different urls to test this:
```
"?filter[limit]=5&filter[offset]=0&order_by[delta]=asc",
 "?filter[limit]=5&filter[offset]=0&order_by[delta]=asc&delta=usage",
```
When the `&delta=usage` is **not** added to the url the `self._delta` is None. However, when it is added on it it becomes the value of whatever you set there. 

My solution to the problem was for when delta is none, I check the provider map for the `delta_key`, and grab the first key in that dictionary. However, that is just one possible solution, another way I could have handled this was to change the serializer to require `&delta=x` to be passed in whenever someone does `&order_by[delta]=asc`.

blentz: @myersCody  - I think your second solution is probably the better one. This is more a validation issue than anything else. If `order_by[delta]` is used, then `delta=X` is a required parameter. So, it can't be `None` and the serializer should raise a `ValidationError` in that case.

myersCody: @blentz That works for me too! I will try to get this update in today. I will ping ya for another review on slack when I do.

koku/api/report/aws/query_handler.py
```diff
@@ -523,13 +523,19 @@ def execute_individual_query(self, org_unit_applied=False):  # noqa: C901
...
+                if self._limit and query_data:
+                    rank_orders = []
+                    if self.order_field == "delta":
+                        rank_orders.append(getattr(F(self._delta), self.order_direction)())
+                    else:
+                        rank_orders.append(getattr(F(self.order_field), self.order_direction)())
+                    rank_by_total = Window(expression=RowNumber(), partition_by=F("date"), order_by=rank_orders)
+                    query_data = query_data.annotate(rank=rank_by_total)
+                    query_order_by.insert(1, "rank")
+                    query_data = self._ranked_list(query_data)
```
blentz, line 538: I don't think you need the extra level of indentation

```suggestion
            if self._limit and query_data and not org_unit_applied:
                rank_orders = []
                if self.order_field == "delta":
                    rank_orders.append(getattr(F(self._delta), self.order_direction)())
                else:
                    rank_orders.append(getattr(F(self.order_field), self.order_direction)())
                rank_by_total = Window(expression=RowNumber(), partition_by=F("date"), order_by=rank_orders)
                query_data = query_data.annotate(rank=rank_by_total)
                query_order_by.insert(1, "rank")
                query_data = self._ranked_list(query_data)
```

PR #2306, opened 2020-08-06
Update to settings.local.yaml for iqe

adberglund:
## Summary 
Update to be in line with iqe settings.

### DIFF
--- a/testing/conf/settings.local.yaml
+++ b/testing/conf/settings.local.yaml
@@ -4,11 +4,13 @@ local:
     ocp_dir: /var/tmp/masu/insights_local
     aws_dir: /tmp/local_bucket
     azure_dir: /tmp/local_container
-    masu:
-      path: api/cost-management/v1
-      port: 5000
-      hostname: localhost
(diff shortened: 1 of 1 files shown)


testing/conf/settings.local.yaml
```diff
@@ -4,11 +4,13 @@ local:
...
-      hostname: localhost
-      scheme: http
+    service_objects:
+      masu:
+        config:
+          path: api/cost-management/v1
+          port: 5000
+          hostname: localhost
+          scheme: http
   SOURCES:
```
lcouzens, line 14: Ah I just noticed we have local sources settings in here too. Don't need to do that now but might want to catch that at some point. Probably something like this 

```SOURCES:
    service_objects:
      sources:
        config:
          path: "api/v1.0"
          hostname: localhost
          port: 3000
          scheme: http```

dccurtis: yeah, don't have to do it now.  I can revisit once we get the local source type changes put into the sources repo

PR #2309, opened 2020-08-07
Updating WorkerCache key to remove worker deployment config suffix

dccurtis:
As part of the recent change to move our worker to a deployment config, we now have to strip the OCP generated suffix hash from the workers hostname.  Previously using stateful sets, the host name was static.

**Testing**
1. Stop a worker in the middle of processing. Verify that manifest is not complete due to the file not being processed.
2. GET /download endpoint to try again 592d34d6-4416-4292-8c35-bf812daeba67-July-2020-local-bucket-1.csv.gz.  Verify that it is re-downloaded and processing is complete
3. Start worker 2 and generate a new manifest.  Verify processing is successful and worker-2 is in the mix.

**Test Results**
[worker_cache_ut.txt](https://github.com/project-koku/koku/files/5038731/worker_cache_ut.txt)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -116,7 +116,7 @@ services:
     - POSTGRES_USER=${DATABASE_USER-postgres}
     - POSTGRES_PASSWORD=${DATABASE_PASSWORD-postgres}
     ports:
-      - "15432:5432"
+      - "15433:5432"
     volumes:
       - ./pg_data:/var/lib/${DATABASE_DATA_DIR-postgresql}/data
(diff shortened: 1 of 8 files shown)


maskarb:
All the other failing tests are due to an invalid hostname. We _could_ set the HOSTNAME in the `koku_test_runner` if you think that might be a good place to do it. Just add:
```
settings.HOSTNAME = "koku-worker-10-abcdef"
```
somewhere in the constants in that file and it'll apply to all the tests. Tis just one idea. Or override the setting for the individual tests.

PR #2312, opened 2020-08-07
Cost 275 empty tenant

lrlevine6:
The goal of this issue is to remove stale tenants that have been empty for more than 2 weeks. A celery task has been added that will run once a day to remove any two week old stale tenants. This should improve lookup time by reducing the number of entries in the table. For functional testing, the Django ORM can be used to step through the process of deleting tenants that will be done by the task.

To start the ORM run `DJANGO_READ_DOT_ENV_FILE=True python koku/manage.py shell` in the koku environment. This should bring up a cli to run commands to follow. 

1. Import the Customer model using `from api.iam.models import Customer` and assign the objects in the model to a variable. For example `data = Customer.objects.all()`. To see what is currently contained in the model we can print the variable using `print(data)`. 

2. To add multiple customers to the model a few changes can be made to the settings file. By changing `DEVELOPMENT_IDENTITY = DEFAULT_IDENTITY` and changing the account number listed in the `DEFAULT_IDENTITY` we can add customers to the model by saving and hitting an endpoint. After doing so, you should see several customers in the returned query set. For example, 
`>>> print(data)`
`<QuerySet [<Customer: Customer object (2)>, <Customer: Customer object (3)>, <Customer: Customer object (5)>]>` 

3. Now update the `date_created` for at least one of the Customers given. The date created for the first object in the query set can be obtained using `print(data[0].date_created)`. This will return a date in a format similar to `2020-08-07 18:20:56.835374+00:00`. For simplicity, the month of the date can be changed to be one month prior. This can be done using:
`>>> record = Customer.objects.get(schema_name=data[0].schema_name)`
`>>> record.date_created = '2020-07-07 18:20:56.835374+00:00'`
`>>> record.save()`

4. Now we can import the cursor and run a variation of code that is used to remove the Tenants.
`>>> from django.db import connection`
`>>> from api.iam.models import Tenant`
`>>> t = Tenant.objects.all()`
`>>> print(t)`
`<TenantQueryset [<Tenant: Tenant object (2)>, <Tenant: Tenant object (3)>, <Tenant: Tenant object (4)>]>`
Here we can see that there are three tenant objects in the query set. Now set a variable with the sql 
`>>> table_sql = "SELECT schema_name FROM api_customer c LEFT JOIN api_provider p ON c.id = p.customer_id LEFT JOIN api_sources s ON p.uuid::text = s.koku_uuid WHERE s.source_id IS null AND c.date_created < now() - INTERVAL '2 weeks'"`
Then query the database:
`>>> with connection.cursor() as cursor:`
`...     cursor.execute(table_sql)`
`...     data = cursor.fetchall()`
`...     print(data)`
And remove the stale tenants:
`>>> Tenant.objects.filter(schema_name__in=[i[0] for i in data]).delete()`
`(1, {'api.Tenant': 1})`
`>>> print(t)`
`<TenantQueryset [<Tenant: Tenant object (3)>, <Tenant: Tenant object (4)>]>`
Now we can see only two tenants remain!


koku/masu/processor/tasks.py
```diff
@@ -700,3 +701,23 @@ def convert_to_parquet(
         msg = f"Failed to convert the following files to parquet:{','.join(failed_conversion)}."
         LOG.warn(log_json(request_id, msg, context))
         return
+
+
+@app.task(name="masu.processor.tasks.remove_stale_tenants", queue_name="remove_stale_tenants")
+def remove_stale_tenants(self):
```
dccurtis, line 707:
```suggestion
def remove_stale_tenants():
```

dccurtis: This change was making the celery beat schedule not fire.  After removing it it ran as expected 👍

koku/masu/processor/tasks.py
```diff
@@ -700,3 +701,23 @@ def convert_to_parquet(
...
+        ON c.id = p.customer_id
+      LEFT
+      JOIN api_sources s
+        ON p.uuid::text = s.koku_uuid
+     WHERE s.source_id IS null AND c.date_created < now() - INTERVAL '2 weeks';
+        """
+    with connection.cursor() as cursor:
+        cursor.execute(table_sql)
+        data = cursor.fetchall()
+        Tenant.objects.filter(schema_name__in=[i[0] for i in data]).delete()
```
dccurtis, line 723: Can we log that an attempt to remove tenants was done and also which ones were removed?

lrlevine6: removed self

PR #2318, opened 2020-08-12
Add documentation to Application Settings

boaz0:
closes COST 220

Adding the "Learn more" link to the end of the description.

image


koku/api/settings/utils.py
```diff
@@ -54,21 +54,34 @@ def create_subform(name, title, fields):
...
 
 
+def generate_doc_link(path):
+    """
+    Generate cost management link to a given path.
+
+    Args:
+        (String) path - path to the documentation.
+    """
+    version = "4.5"
```
adberglund, line 80: Recommendation here: if we have to hard code this consider adding a constant to settings.py and importing that in here to reference the version. That way we can modify it at the top level and have it update anywhere we reference the constant.

boaz0: Thanks! I moved it to `koku.settings` as you suggested.

adberglund (dismissed):

koku/koku/settings.py
```diff
@@ -444,3 +444,5 @@
     DEMO_ACCOUNTS = ENVIRONMENT.json("DEMO_ACCOUNTS", default={})
 except JSONDecodeError:
     pass
+
+OPENSHIFT_DOC_VERSION = "4.5"
```
maskarb, line 448: Just a suggestion, but maybe it would be good to make this an importable variable?
```suggestion
OPENSHIFT_DOC_VERSION = ENVIRONMENT.get_value("OPENSHIFT_DOC_VERSION", default="4.5")
```

PR #2328, opened 2020-08-17
Cost 273 python 3.8

Red-HAP:
## Jira Ticket

[COST-273](https://issues.redhat.com/browse/COST-273)

## Description

This will update our Python language version to 3.8.

Mostly it's just changing references, but there were a few things of note:

* `virtualenv`, `virtualenv_bin` had to be explicitly installed to work with our existing scripts
* The imagestream key was changed to be more generic so, hopefully, not as many changes need to be made for the next language update.
* py36 was removed from the tox.ini
* A script was created to handle e2e yaml changes for deploying a different branch from master for koku,, koku-flower, rdsexporter

## Deploy

The `scripts/e2e-deploy.sh` was changed to accept an optional positional parameter that should be a git branch name or ref. This will be used to set some yaml files for hccm, hccm-optional to use that ref. If you pass `current` (or `CURRENT`) to the script, the current git ref will be used. If no parameter is passed, `master` will be used. If the script aborts on error, then the yaml files will have to be reset manually.

Reset the yaml files executing `scripts/update_e2e_srr.py master`

----

Please give these changes careful scrutiny. Things _should_ build in crc, hopefully there are no outliers.

## e2e PR

Companion changes were made to the e2e-deploy repo for hccm, hccm-optional
https://github.com/RedHatInsights/e2e-deploy/pull/2147


Pipfile
```diff
@@ -44,6 +44,9 @@ kafka-python = ">=2.0.1"
 cachetools = ">=4.1.0"
 pyarrow = ">=0.17.1"
 kombu = ">=4.6.10"
+zipp = "*"
+typed-ast = "*"
+importlib-metadata = "*"
```
maskarb, line 49: Why are these now required?

Red-HAP: Builds were failing on Fedora without importlib-metadata.
The other two, I thought were already required. I can take them out and re-verify.

Red-HAP: @maskarb 
I've taken out zipp, typed-ast.

PR #2334, opened 2020-08-20
Celery task check is running

adberglund:
## Summary
* This adds a constraint on the refresh materialized viewsto only allow 1 task to run per tenant at a time. This hopefully reduces some situations where several workers are refreshing views at the same time.
* This adds a constraint on the cost model update tasks so that only 1 task for a tenant/source can be running at a time.

## Summary
Ran smoke tests successfully with no new failures. Both with single and multi-worker setups.


koku/masu/processor/tasks.py
```diff
@@ -421,9 +443,24 @@ def update_cost_model_costs(schema_name, provider_uuid, start_date=None, end_dat
...
+    bind=True,
+)
+def refresh_materialized_views(self, schema_name, provider_type, manifest_id=None):
     """Refresh the database's materialized views for reporting."""
+    task_id = None
+    if self.request:
+        task_id = self.request.id
+    if is_task_currently_running("masu.processor.tasks.refresh_materialized_views", task_id, [schema_name]):
+        msg = f"Already running refresh_materialized_views for {schema_name}."
+        raise TaskRunningError(msg)
```
blentz, line 462: Is this an error that we need to raise an exception on? Or is this just a condition where we need to not execute the materialized view refresh code?

What purpose does raising an exception here serve?

koku/api/provider/provider_manager.py
```diff
@@ -234,4 +235,7 @@ def provider_post_delete_callback(*args, **kwargs):
 
         delete_func = partial(delete_archived_data.delay, provider.customer.schema_name, provider.type, provider.uuid)
         transaction.on_commit(delete_func)
-    refresh_materialized_views(provider.customer.schema_name, provider.type)
+    try:
+        refresh_materialized_views.s(provider.customer.schema_name, provider.type).apply()
+    except TaskRunningError:
+        pass
```
adberglund, line 241:
```suggestion
        # Forcing this error will engage Celery's task retry logic
        pass
```

koku/masu/processor/tasks.py
adberglund, line 462:
```suggestion
        # Raising an exception will fail the task and trigger Celery's retry logic
        raise TaskRunningError(msg)
```

koku/masu/processor/tasks.py
```diff
@@ -407,6 +421,14 @@ def update_cost_model_costs(schema_name, provider_uuid, start_date=None, end_dat
...
 
     """
+    task_id = None
+    if self.request:
+        task_id = self.request.id
+    if is_task_currently_running(
+        "masu.processor.tasks.update_cost_model_costs", task_id, [schema_name, provider_uuid]
+    ):
+        msg = f"Already running update_cost_model_costs for {schema_name} and provider {provider_uuid}."
+        raise TaskRunningError(msg)
```
adberglund, line 431:
```suggestion
        # Raising an exception will fail the task and trigger Celery's retry logic
        raise TaskRunningError(msg)
```

koku/masu/processor/tasks.py
adberglund: Comments added, and I wish there was a better way, but celery needs the task to fail with an exception to auto-retry.

PR #2338, opened 2020-08-23
Provider data update timestamp

adberglund:
## Summary
This adds another timestamp to signify data is updated. Particularly when we trigger cost model summarization when a cost model is updated, we do not track against a particular manifest, so we don't update any timestamps. This timestamp would cover all cases where we update our data. 

Will require https://gitlab.cee.redhat.com/insights-qe/hccm-plugin/-/merge_requests/658 to merge too.

## Testing

With this PR was able to get all smoke tests passing locally.


koku/masu/processor/tasks.py
```diff
@@ -422,7 +425,7 @@ def update_cost_model_costs(schema_name, provider_uuid, start_date=None, end_dat
 
 
 @app.task(name="masu.processor.tasks.refresh_materialized_views", queue_name="reporting")
-def refresh_materialized_views(schema_name, provider_type, manifest_id=None):
+def refresh_materialized_views(schema_name, provider_type, manifest_id=None, provider_uuid=None):
```
Red-HAP, line 428: Would it make any sense at all to have the `ProviderDBAccessor` detect if `provider_uuid` is an instance of `Provider` and not have to fire another query to resolve it? There seems to be a lot of places here that will have to execute another query to resolve something that has already been resolved.

adberglund: We had an offline conversation and decided that this question is more a bigger question around our database accessors. There's also the issue of serialization and celery tasks here where passing in ORM objects might not work as expected for a task that is run on a worker.

dccurtis (approved):

PR #2339, opened 2020-08-24
[COST-272] Filter on tag key/value

cgoodfred:
Previously there was not a filter available for tag keys or values and now there is. At the tags endpoint for each provider, you can filter on a key (/aws/?filter[key]=version) and after going to a specific key you can filter the values (/aws/version/?filter[value]=prod). It supports partial matching of keys and values.

Added many to many field in the tags tables in order to make them filterable.

Testing:
1. load some data in for each provider `make create-test-customer` and `make load-test-customer-data` 
2. Go to the tags endpoint for a provider such as aws, `tags/aws/` and see all the tags and values for that provider
3. Pick one of the tags and try to filter on that tag, using the aws example: `tags/aws/?filter[key]=version` and you should see the values: prod, master, gamma and beta.
4. Go to the endpoint for that specific tag and see the same values as above, `tags/aws/version/`
5. Filter for a specific value or a partial match, `tags/aws/version/?filter[value]=prod` and see the tag or matching tags if you do a partial match
6. repeat for the other providers

Smokes results: 
```
============================================================================ short test summary info =============================================================================
FAILED iqe_cost_management/tests/rest_api/v1/test_cost_model.py::test_api_cost_model_markup_calculation_aws - assert 0.0 != 0.0
==================================================== 1 failed, 3946 passed, 81 skipped, 1729 deselected in 1341.54s (0:22:21) ====================================================
```
This test is unrelated and fails on master.


koku/api/tags/ocp/queries.py
```diff
@@ -69,3 +83,31 @@ def __init__(self, parameters):
             parameters.set_filter(**{"enabled": True})
         # super() needs to be called after _mapper is set
         super().__init__(parameters)
+
+    def _get_key_filter(self):
```
maskarb, line 87: So we don't have a lot of repetitive code, let's move this function into tags/queries.py and save the QueryFilter params somewhere. Maybe a class attribute? e.g.
```
KEY_FILTERS = [
    {"field": "ocpusagepodlabelsummary__key"},
    {"field": "ocpstoragevolumelabelsummary__key"},
]
```
Then the main function can just iterate thru this list to add the appropriate filters:
```
        filters = QueryFilterCollection()
        if self.parameters.get_filter("value"):
            for something in self.KEY_FILTERS:
                filters.add(
                    QueryFilter(
                        field=something.get("field"),
                        operation="exact",
                        parameter=self.key,
                        composition_key="tag_filter",
                    )
                )
        else:
            ...
```
With something like this in place, we can remove `_get_key_filter` from all of the individual queries files.

cgoodfred: Good idea, changed to this functionality.

koku/api/tags/queries.py
```diff
@@ -378,6 +417,17 @@ def execute_query(self):
         if self.parameters.get("key_only"):
             tag_data = self.get_tag_keys()
             query_data = sorted(tag_data, reverse=self.order_direction == "desc")
+        elif self.parameters.get_filter("value"):
+            tag_data = self.get_tag_values()
+            # This is sorted by values that start with the filter first, then values that contain the filter
+            # based on a discussion with UX
```
maskarb, line 423: 👏🏻 👏🏻 👏🏻 
I was gonna ask you to put a comment in here since this logic is so weird

PR #2344, opened 2020-08-25
[COST-479] Remove NISE_REPO_PATH

cgoodfred:
Removed NISE_REPO_PATH as a required env variable and any references to it. Required yml files were brought over into `scripts/NISE_ymls/` and references updated.

Testing:
the two make commands `make load-aws-org-unit-tree` and `make load-test-customer-data` should still function like normal and correctly load the data.

1. `make load-aws-org-unit-tree` and go to `v1/organizations/aws/` and see the information loaded
2. `make create-test-customer` and `make load-test-customer-data` and then you should be able to see the data associated with each provider


docs/source/testing.rst
```diff
@@ -126,10 +126,9 @@ Within the koku repo there are scripts to automate the creation of a test custom
     - OpenShift running on Azure
     - OpenShift on premises
 
-There is a complementary script to subsequently load test data for each source. The `nise`_ project is used to generate this test data. Nise is a dev dependency of koku, so a working copy of the nise command line tool shoudl be availble after running `pipenv install --dev`. A local clone of the nise repository is also required as it contains static YAML files used by this script. Before using this script, make sure the following variables are defined in your environment. See `.env.example`_ for example values.
+There is a complementary script to subsequently load test data for each source. The `nise`_ project is used to generate this test data. Nise is a dev dependency of koku, so a working copy of the nise command line tool shoudl be availble after running `pipenv install --dev`. See `.env.example`_ for example values.
```
maskarb, line 129:
```suggestion
There is a complementary script to subsequently load test data for each source. The `nise`_ project is used to generate this test data. Nise is a dev dependency of koku, so a working copy of the nise command line tool should be available after running `pipenv install --dev`. See `.env.example`_ for example values.
```

maskarb (dismissed):
LGTM!

PR #2351, opened 2020-08-26
Cost 418 drop unneeded temp tables

Red-HAP:
## Jira Ticket

[COST-418](https://issues.redhat.com/browse/COST-418)

## Description

This change tries to make better cooperative use of limited resources by dropping temp tables as soon as they are no longer needed. The technique used is **not** `ON COMMIT DROP`, but using explicit `TRUNCATE` statements followed by `DROP` statements. The reason for this is that, in a few of the SQL files, a CALL is made to a procedure that can, if directed, issue a commit. As a safety measure in case of changes, the use of `ON COMMIT DROP` was abandoned for the explicit calls. Also, this will free up resources immediately after their usefulness has expired.

## Testing

1. `git checkout COST-108-drop-unneeded-temp-tables && git pull`
3. `docker-compose restart` # Just in case not all services come up correctly after db restart
4. `make load-test-customer-data`
5. `docker logs koku-worker | grep -e 'TRUNCATE' -e 'DROP'`

You should see: 
`TRUNCATE TABLE <table>_<uuid>; `
followed by 
`DROP TABLE <table>_<uuid>; `

There should be no errors in processing.


Pipfile.lock
```diff
@@ -93,6 +93,7 @@
         },
         "boto3": {
             "hashes": [
+                "sha256:34ee7608afa457da7bf4ad42dc53ac967b2f28769ef4ef1a0a41e1b40f35d897",
```
adberglund, line 96: Probably want to back out a pipfile.lock change that doesn't update alllll the requirements files. Also no overall hash change, so this would probably fail to deploy in OpenShift.

Red-HAP: @adberglund 
Removed the extra hash entry.
The file should match master now.

PR #2364, opened 2020-09-01
Fix the status_probe failure

Red-HAP:
## Jira Ticket

[COST-476](https://issues.redhat.com/browse/COST-476)

## Description

Changed to use the `requests` module instead of `urllib`

Also, if the status url is not passed in, the probe will assume a file should be read. If no file given, it will read from stdin.

The `run_migrations.sh` script will try the status probe directly first. If that fails to populate `COMMIT`, then `curl` will be used with the status probe reading and parsing the result of the call.

## Testing

1. The script should not fail to run (`run_migrations.sh` or `status_probe.py`)
2. If the hash can be determined, then `COMMIT` should be set properly blocking repeat executions of the migrations.
3. This is difficult, if not impossible, to test locally. But the results can be viewed in CI, QA, and staging. _The staging environment is where I saw it fail._

### DIFF
--- a/scripts/status_probe.py
+++ b/scripts/status_probe.py
@@ -18,11 +18,15 @@
 import argparse
 import json
 import sys
-import urllib.request  # Using urllib because that's included in any python distro
+
+import requests
 
 
 parser = argparse.ArgumentParser()
-parser.add_argument("--status-url", dest="status_url", metavar="URL", required=True, help="Application status URL")
+parser.add_argument("--status-url", dest="status_url", metavar="URL", required=False, help="Application status URL")
+parser.add_argument(
+    "--file", dest="filename", metavar="FILE", required=False, help='File to read (use "-" for stdin)', default="-"
+)
 parser.add_argument(
     "--path",
     dest="path",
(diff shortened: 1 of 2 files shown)


scripts/status_probe.py
```diff
@@ -35,9 +39,24 @@
...
+else:
+    if args.filename == "-":
+        jfile = sys.stdin
+    else:
+        jfile = open(args.filename, "rt")
+
+    content = jfile.read()
+    if jfile != sys.stdin:
+        jfile.close()
+    app_status = json.loads(content)
```
blentz, line 57: `open()` and `read()` and `json.loads()` can all raise exceptions.

Do you want to catch any of them?

Red-HAP: I should at least log 'em.

Red-HAP: Very logging enabled.

scripts/status_probe.py
```diff
@@ -33,22 +38,91 @@
...
+    app_status = resp.json()
+else:
+    if args.filename == "-":
+        fname = "stdin"
+        jfile = sys.stdin
+    else:
+        fname = args.filename
+        try:
+            jfile = open(args.filename, "rt")
+        except Exception as e:
```
blentz, line 80:
```suggestion
        except (IOError, OSError) as e:
```

scripts/status_probe.py
```diff
@@ -33,22 +38,91 @@
...
+            LOG.error(f"{e.__class__.__name__}: Cannot open file {fname}: {e}")
+            sys.exit(-1)
+
+    LOG.debug(f"Reading status from {fname}")
+    try:
+        content = jfile.read()
+    except Exception as e:
```
blentz, line 87:
```suggestion
    except (IOError, OSError) as e:
```

scripts/status_probe.py
```diff
@@ -33,22 +38,91 @@
...
+        LOG.error(f"{e.__class__.__name__}: Cannot read file {fname}: {e}")
+        sys.exit(-1)
+
+    if jfile != sys.stdin:
+        jfile.close()
+
+    if content:
+        try:
+            app_status = json.loads(content)
+        except Exception as e:
```
blentz, line 97:
```suggestion
        except (IOError, OSError, json.JSONDecodeError) as e:
```

blentz (changes requested):
Let's tighten up those exception catches a little, but then I think this should be ready.

blentz (approved):

PR #2371, opened 2020-09-02
Skip rows that don't have data for required fields

adberglund:
## Summary
This will cause us to not process OpenShift rows that don't have required data present. This is a very rare occurrence and we've now only seen this happen with the very first timestamp in a new cluster's data. This would cause us to skip that first hour of data.


maskarb (dismissed):
LGTM!

I manually tested by removing the pod, node, namespace from different rows of static data, ingesting, counting rows. Then retried without modifying the data, and the difference in row count made sense.

PR #2375, opened 2020-09-02
COST-430: Add support for tag based rates in cost model API.

myersCody:
Jira Issue: 
- https://issues.redhat.com/browse/COST-430

This pr is to add support for the tag based rates.

#### How to test

I have already added an example tag rate to the `openshift_on_prem_cost_model.json`. 

I think it would be beneficial to add a couple more scenarios. Like multiple of the same tag key, multiple tag values, a rate for infrastructure and a rate for supplementary. 

After you have finished modifying the json you can then run:
`make load_test_customer_data` to populate the database and upload the cost model.  

Then you will want to check to see if the tag rates show up: 
- http://localhost:8000/api/cost-management/v1/cost-models/


blentz (dismissed):

docs/source/specs/openapi.json
```diff
@@ -3608,6 +3620,64 @@
                     }
                 }
             },
+            "TagRate": {
+                "required": [
+                    "metric"
```
boaz0, line 3625: maybe I am missing something but I do not see `metric` under `TagRate`'s properties?

docs/source/specs/openapi.json
```diff
@@ -3608,6 +3620,64 @@
...
+                ],
+                "properties": {
+                    "tag_key": {
+                        "type": "string",
+                        "enum": ["key_one"]
```
boaz0, line 3630: What does `enum` mean here?

boaz0:
Hey @myersCody well done.
I just have two questions, can you answer them?
I will run this changes on my local dev environment to give a better feedback.

Thanks.

docs/source/specs/openapi.json
myersCody: enum is short for "enumerate" and is a data type that consists of predefined values. Essentially it is just part of the openapi spec syntax.

docs/source/specs/openapi.json
myersCody: This is a good question about the metric name, I will have to look into it a bit and come back with an answer.

myersCody: Update: Good catch @boaz0 I have this positioned in the wrong spot. I am going to move this over to WIP for now, and try to get those changes in today. I will ping you on slack when its ready for another review.

koku/cost_models/serializers.py
```diff
@@ -294,6 +389,8 @@ class Meta:
 
     rates = RateSerializer(required=False, many=True)
 
+    # tag_rates = TagRateSerializer(required=False, many=True)
+
```
adberglund, line 393:
```suggestion
```

adberglund (dismissed):
Just one commented out thing that could maybe be removed?

koku/cost_models/serializers.py
myersCody: Whoops! Totally missed this line when I was doing my clean up. I am gonna remove it now.

adberglund (approved):

PR #2378, opened 2020-09-03
Use check_migrations management command for run_migrations script

adberglund:
## Summary
Switch to using our check_migrations function to determine if we should run migrations.


blentz (dismissed):
LGTM, but if there's a chance the `check_migrations` output could be null, the bash script may want to also check for an empty string.

maskarb (approved):
lgmt!

PR #2380, opened 2020-09-03
Rename rate to cost_model for RBAC

adberglund:
## Summary
This renames the "rate" RBAC permission to "cost_model"

Dependent on https://github.com/RedHatInsights/rbac-config/pull/87


koku/koku/rbac.py
```diff
@@ -41,7 +41,7 @@
     "openshift.node": ["read"],
     "openshift.project": ["read"],
     "provider": ["read", "write"],
```
chambridge, line 43: not really related, but would be nice to delete this provider line.

boaz0:
I need to update this in [the UI](https://github.com/boaz0/koku-ui/blob/master/src/store/rbac/selectors.ts#L24) as well

boaz0 (dismissed):

chambridge (approved):
LGTM

PR #2381, opened 2020-09-04
Cost 455 AWS Account api

lrlevine6:
Added endpoint for AWS Accounts API as a part of the RBAC Custom Resource APIs epic. This API returns unique AWS accounts. To check that this is properly constructed, load data into the database and visit the `/api/cost-management/v1/resource-types/aws-accounts/` endpoint.


koku/api/resource_types/test/tests_views.py
```diff
@@ -24,16 +24,16 @@
 
 
 class ResourceTypesViewTest(IamTestCase):
-    """Tests the resource types view."""
+    """Tests the report view."""
 
     ENDPOINTS_RTYPE = ["resource-types"]
-    ENDPOINTS_AWS = ["aws-accounts", "aws-organizational-units"]
+    ENDPOINTS_AWS = ["aws-accounts"]  # , "aws-organizational-units"]
```
Red-HAP, line 30: Is there a reason to leave in the commented-out part of the change? 
If so, please leave an extra comment to that effect.

lrlevine6: I had left it commented out since it would be the next api that I worked on. I can remove it for now and add it back when this API is added!

koku/api/resource_types/test/tests_views.py
```diff
@@ -56,16 +56,3 @@ def test_endpoint_view(self):
                 self.assertIsNotNone(json_result.get("data"))
                 self.assertIsInstance(json_result.get("data"), list)
                 self.assertTrue(len(json_result.get("data")) > 0)
-
-    def test_endpoints_invalid_query_param_still_returns(self):
```
Red-HAP, line 60: Why remove this test?

lrlevine6: I was unsure if it was necessary since the only additional thing it is testing is the added group by but I do not believe it makes sense to ever use a group by for this api. I am happy to return it if you think test would be worth it!

adberglund: Yes, these APIs aren't supporting a group_by or any query params, so totally good to remove this test.

adberglund (approved):

PR #2384, opened 2020-09-08
Cost 493 azure partitioned deadlock orm

Red-HAP:
**Please also review the testing method described below.**

## Jira Tickets

* [COST-322](https://issues.redhat.com/browse/COST-322)  (Azure Processing deadlock)
* [COST-493](https://issues.redhat.com/browse/COST-493)  (Create Partition deadlock)

## Description

After looking at the processing flow for Azure, it became clear that it was structured with a long-lived transaction. During this transaction, shared records in shared tables are attempting to be written at the same time. Similar to this:

```
shared_table (
    id int primary key
    ext_id text not null,
    constraint uniq_ext_id unique (ext_id)
)
```

1. Connection 1: Write record (1, 'eek')
2. Connection 2: Write record (2, 'ook')
3. Connection 2: Write record (3, 'eek')  -- Connection 2 now blocked by Connection 1
4. Connection 1: Write record (4, 'ook')  -- Connection 1 now blocked by Connection 2

Each of these connections blocking the other is the deadlock state

This is the issue with multiple workers running with the existing workflow. This is affecting shared tables like `reporting_azuremeter` (and others) as well as others. Also at risk is the `partitioned_tables` table as well as system catalog tables if the same PostgreSQL object (in this case, a table, and more speficically, a table partition) is being created by two different connections within their own transactions.

This change attempts to break a long-lived transaction into a series of shorter transactions by wrapping parts of the processing in `transaction.atomic()` contexts. To better handle basic partition management (create, drop, attach, detach) database triggers have been added to the `partitioned_tables` table as well as an `active` boolean column. The processing that will create these partitions for Azure is now being done, primarily outside of the SQL for the data processing and that is also wrapped in a `transaction.atomic()` context.

### Sidebar

The changes to the partitioning creation logic exposed some flaws in the partitioning testing that needed to be corrected. These test corrections exposed a bug I introduced with the partitioning changes. So some fixes to the `pg_partitioning.py` file had to be implemented along with updates to the partitioning tests.

## Testing Method

**Note:** Before this testing method was utilized a full `tox` run has passed.

This is the testing method that was utilized in local testing. This testing method passed twice before putting up this PR.

1. Unset the `DATABASE_DATA_DIR` environment variable
2. `make docker-down`
3. `make remove-db`
4. `docker volume prune -f`
5. `make docker-up`
6. `make docker-reinit-with-sources`
7. `docker-compose down`
8. `docker network create koku_default`
9. `docker-compose -f ./testing/compose_files/docker-compose-multiworker.yml up -d`  (You should have 10 workers now)
10. Edit `./scripts/load_test_customer_data.sh`:
    * Find this line: `nise report azure --static-report-file "scripts/nise_ymls/ocp_on_azure/azure_static_data.yml" --azure-container-name "$KOKU_PATH/testing/local_providers/azure_local" --azure-report-name azure-report --start-date "$START_DATE" --end-date "$END_DATE"` 
    * Change to:  `nise report azure --static-report-file "scripts/nise_ymls/large_azure.yml" --azure-container-name "$KOKU_PATH/testing/local_providers/azure_local" --azure-report-name azure-report`
11. `make load-test-customer-data start=2020-06-01 end=2020-10-01`
12. `git restore ./scripts/load_test_customer_data.sh`

You can watch the logs with: 
```
docker-compose -f ./testing/compose_files/docker-compose-multiworker.yml logs --tail=200 -f koku-worker-1 koku-worker-2 koku-worker-3 koku-worker-4 koku-worker-5 koku-worker-6 koku-worker-7 koku-worker-8 koku-worker-9 koku-worker-10
```

Once processing has finished grep the logs for any message about a lock with:
```
docker-compose -f ./testing/compose_files/docker-compose-multiworker.yml logs koku-worker-1 koku-worker-2 koku-worker-3 koku-worker-4 koku-worker-5 koku-worker-6 koku-worker-7 koku-worker-8 koku-worker-9 koku-worker-10 koku-server masu-server | grep -i lock
```


koku/masu/processor/azure/azure_report_processor.py
```diff
@@ -314,6 +320,16 @@ def process(self):
                 for row in reader:
                     if not self._should_process_row(row, "UsageDateTime", is_full_month):
                         continue
+                    li_usage_dt = row.get("UsageDateTime")
+                    if li_usage_dt:
+                        try:
+                            li_usage_dt = parser.parse(li_usage_dt).date().replace(day=1)
```
adberglund, line 326: Can you switch this to using ciso8601? Datetime handling in Python it turns out is quite slow. Dropping down to C with that library has actually helped speed up some bottlenecks in report processing. Should be an example of it in this file.

Red-HAP: OK, thanks!!

testing/compose_files/docker-compose-multiworker-make.yml
```diff
@@ -0,0 +1,864 @@
+version: '3'
```
adberglund, line 1: What's the difference between this and the docker-compose-multiworker.yml?

Red-HAP: This one uses an env var for the path to volumes. The other one uses a relative path so the cwd must be the same location as that file.
I wasn't sure if there was any testing or tooling already in place, so I just made a new one ending with `-make` for use with the Makefile new CLI param `compose_file`

adberglund (approved):

PR #2385, opened 2020-09-08
Cost 456 AWS Organizational Unit/ Azure Subscription Guid Resource type API

lrlevine6:
Added endpoint for AWS Organizational Units API as a part of the RBAC Custom Resource APIs epic. This API returns unique AWS Organizational Units. To check that this is properly constructed, load data into the database and visit the `/api/cost-management/v1/resource-types/aws-organizational-units/` endpoint.


koku/api/resource_types/aws_org_unit/view.py
```diff
@@ -0,0 +1,42 @@
...
+from api.common.permissions.resource_type_access import ResourceTypeAccessPermission
+from api.resource_types.serializers import ResourceTypeSerializer
+from reporting.provider.aws.models import AWSOrganizationalUnit
+
+
+class AWSOrganizationalUnitView(generics.ListAPIView):
+    """API GET list view for AWS organizational units."""
+
+    queryset = (
+        AWSOrganizationalUnit.objects.filter(deleted_timestamp__isnull=True)
```
adberglund, line 32: @myersCody @abaiken Does this query seem okay to get the list of existing active org units?

myersCody: @lrlevine6 @adberglund 

I am not quite sure if you need a `schema_context`; however, this query looks like it is correctly filtering out the old values.

```
>>> with schema_context("acct10001"):
...     AWSOrganizationalUnit.objects.filter(deleted_timestamp__isnull=True).annotate(**{"value": F("org_unit_id")}).values("value").distinct()
...
<QuerySet [{'value': 'OU_001'}, {'value': 'OU_002'}, {'value': 'OU_003'}, {'value': 'OU_005'}, {'value': 'R_001'}]>`
```

Awesome work Lauren! I know filtering out both the moved & deleted org units can be confusing.

adberglund (approved):

PR #2392, opened 2020-09-10
Use DRF browsable API in dev only

adberglund:
## Summary
Remove DRF Browsable API access outside of the dev environment.


chambridge (approved):
LGTM

PR #2394, opened 2020-09-11
COST-498: Add ability to upload aws org unit tree through masu endpoint.

myersCody:
Jira Issue: https://issues.redhat.com/browse/COST-498

This pr is to add the ability to upload an aws org unit tree through the masu endpoint.

I didn't want to duplicate the code I already written inside of: https://github.com/project-koku/koku/blob/master/scripts/insert_aws_org_tree.py

So, I converted that script to use the django orm and placed inside masu so that it could be used for both this endpoint & be used to populate the test database for the unit tests inside of the koku test runner. Then I rewrote the development workflow for inserting an aws org unit tree to use the new endpoint. 

The idea being that if the crawler ever changes how it inserts data into the db, we can change it in one place to update the unittests, development workflow,  and now the smoke tests via this new endpoint.  

**Special Note:** 

I have reached out to @lcouzens to go ahead and try to connect his smoke tests using this pr. That way if we discover we need to make changes to get the smoke test side of things working we can do it before this pr is merged. And that way the smoke tests and this pr can go in at the same time. 

----

### HOW TO TEST

**One: Test the development workflow**

You can test the development workflow by running either this command: `make load-test-customer-data` or `make load-aws-org-unit-tree`. After masu has finished processing the reports you can check out these endpoints to make sure the data was populated correctly. 

- http://127.0.0.1:8000/api/cost-management/v1/reports/aws/costs/?group_by[org_unit_id]=R_001

**Two: Test the koku runner changes using tox**

I ran into some wonkiness running the tox tests more than once. On the first run it would work, but on the second pass the tests would fail. I am pretty certain I fixed the issue, but it would be good for someone to run this back to back on their machine to make sure:

```
tox -e py36 masu.test.api.test_crawl_account_hierarchy.crawlAccountHierarchyTest
```


lcouzens (approved):
LGTM! works in my automation anyway :+1:

abaiken (approved):
LGTM 👍

PR #2400, opened 2020-09-17
fine-tuning grafana dashboards

blentz:
This is just fine-tuning a couple of details in the grafana dashboards.

- Sets the datasource to be prometheus.
- Sets some sane variable defaults to be selected
- Small tweak to one dashboard title for consistency.


Red-HAP (commented):
This is all for staging only, correct?
I see no mention of that in the comments or a reference to a Jira ticket for that information.

blentz:
@Red-HAP - This will be only in staging until an app-interface MR promotes it to the production instance of grafana. This doesn't impact the running pods at all, just grafana instances.

Red-HAP (approved):

PR #2402, opened 2020-09-18
Cost 522 empty csv multi group by

Red-HAP:
## Jira Ticket

[COST-522](https://issues.redhat.com/browse/COST-522)

## Description

This change has moved all csv post-processing until after all calls to `execute_individual_query()` have run.

Also, the CSV output has been changed to more closely match the JSON serializer output in field (column) names as well as data (such as the `type` field and setting `id`, `alias` to sub-org-id, name as the JSON output formatter does.

## Testing

1. `git checkout COST-522-empty-csv-multi-group-by`
2. `make docker-down`
3. `sudo rm -rf ./pg_data`
4. `make docker-up`
5. `make docker-reinitdb-with-sources`
6. `make load-test-customer-data`

Using Postman or curl, test output with these urls:

* `http://127.0.0.1:8000/api/cost-management/v1/reports/aws/costs/?group_by[account]=9999999999990`
* `http://127.0.0.1:8000/api/cost-management/v1/reports/aws/costs/?group_by[org_unit_id]=R_001`
* `http://127.0.0.1:8000/api/cost-management/v1/reports/aws/costs/?group_by[account]=9999999999990&group_by[org_unit_id]=R_001`

Now test with the same urls, but also setting `Accept=text/csv` in the request header. CSV output should generate for all cases.


myersCody:
So far this looks great HAP! I will say that I think it may be easy for the csv for multiple group bys for org units to be broken again accidentally. So we should add a simple check that does a text/csv on a multiple group by of org units. That way if someone makes changes in the query handler that break the csv we can catch it in the unittests.

myersCody (approved):
Awesome work @Red-HAP

PR #2404, opened 2020-09-21
Cost 541 tag summary values change

adberglund:
## Summary
This moves away from using a many to many table to two separate tag tables, one for the summary, one for tracking key:value pairs. There was some Postgres specific logic that Presto will have trouble with. This also moves the tables to using a uuid for primary key to avoid sequence overflow as we do a lot of upserts on the table.

## Testing
Manual testing of endpoints to confirm filters and results are working on the tags API endpoints + smoke tests. 

```
===== 3948 passed, 80 skipped, 1732 deselected in 1378.94s (0:22:58) =====
```


blentz (approved):
LGTM

PR #2405, opened 2020-09-22
remove safe methods check for resource type

adberglund:
## Summary
Right now resource type apis should be for org admins only. Not allowed as a "safe method"


chambridge (approved):
LGTM

chambridge (approved):

PR #2413, opened 2020-09-23
COST-521: update verbiage to more inclusive terms

blentz:
This updates the couple of places where we use non-inclusive terms in the code to more inclusive terms.


chambridge (approved):
LGTM

PR #2414, opened 2020-09-23
COST-509: Modify rbac org units to follow the ou tree hierarchy 

myersCody:
Initially this issue was a bug with RBAC not working correctly for AWS Org Unit group_bys & multiple group_bys _foreshadowing_
-  https://issues.redhat.com/browse/COST-509

**[TLDR]**
Most of the stuff below is context for @lcouzens & our future selves to help clarify what has been worked on in this pr because the issue has turned from a bug to redesign of how RBAC should work for AWS org units. 

----
### [TLDR]: Additional Context

While working on this issue it kind of transformed from a bug to a bit of redesign regarding how RBAC works for the AWS org unit feature. So, in order to understand the changes in this pr in regards to the initial issue, I am going to give a little spiel regarding the initial design as well as explain what I mean by have RBAC follow the hierarchy of the tree. 

#### Initial RBAC Implementation

The initial implementation was essentially a check to see if the org units being returned to the customer was in the rbac access list. Therefore, if the OU RBAC list was set to `[R_001, OU_002]`, then the only org unit nodes the user would have access to were those two org unit nodes. Therefore, even if `OU_003` was a child node of `OU_002` then you wouldn't have access to it unless it was in the list. Which means for a user to have full access to aws org tree, they would have to place every node in the tree in the rbac list and keep it up to date for when new nodes are added. Additionally, the excluded nodes would not show up in the total cost. So, even if `OU_003` was a child of `OU_002` it would not be reflected in the total cost unless it was added to the list. When I discovered this after I fixed the initial bug, I looped Sergio & Andrew into the discussion and we came up with the idea to have it follow the tree hierarchy.

#### New RBAC Implementation

So, the new implementation of the RBAC org units follows the hierarchal structure of the tree. By that I mean if a user has access to to a parent node through the rbac access list, then the user is granted access to all children & sub children of the parent node. For example, now if the rbac access list is `[OU_001]` then the user is granted access to both `OU_001` & `OU_002`. Which means now if a user wants to grant access to the entire tree they can simply place the root node in the rbac access list. This also provides some protection for a node being left out while calculating the total cost of a parent node. 
Screen Shot 2020-09-23 at 6 28 09 PM

----

### How to test
Populate the database with dummy data: `make load-test-customer-data`

In order to test this you will need to update the `DEVELOPMENT_IDENTITY` inside of your `.env`.
 **NOTE: DON'T FORGET TO CHANGE THIS BACK AFTER TESTING THIS PR OR THE NEXT TIME YOU RUN THE SMOKE TESTS LOCALLY YOU ARE GONNA HAVE A BAD TIME** 

**Identity One: Root node**
```
DEVELOPMENT_IDENTITY='{"identity": {"account_number": "10001", "type": "User", "user": {"username": "user_dev", "email": "user_dev@foo.com", "access": {"aws.account":{"read": ["*"]}, "aws.organizational_unit":{"read": ["R_001"]}, "azure.subscription_guid":{"read": ["*"]}, "openshift.cluster":{"read": ["*"]}, "openshift.project":{"read": ["*"]}, "openshift.node": {"read": ["pretty-specific-1"]}}}},"entitlements": {"cost_management":{"is_entitled": "True"}}}'
```
You may have to bring the docker containers down and back up to get the development identity to change. 

With access to the root node you should have access to everything in the tree.
1. Hit the organizations endpoint and make sure you can see all the nodes: 

2. Hit the report endpoint with a group by to make sure you have access to children nodes:
- http://127.0.0.1:8000/api/cost-management/v1/reports/aws/costs/?group_by[org_unit_id]=OU_002
- http://127.0.0.1:8000/api/cost-management/v1/reports/aws/costs/?group_by[or:org_unit_id]=OU_001&group_by[or:org_unit_id]=OU_002

**Identity Two: Limited Access (OU_001)**
```
DEVELOPMENT_IDENTITY='{"identity": {"account_number": "10001", "type": "User", "user": {"username": "user_dev", "email": "user_dev@foo.com", "access": {"aws.account":{"read": ["*"]}, "aws.organizational_unit":{"read": ["OU_001"]}, "azure.subscription_guid":{"read": ["*"]}, "openshift.cluster":{"read": ["*"]}, "openshift.project":{"read": ["*"]}, "openshift.node": {"read": ["pretty-specific-1"]}}}},"entitlements": {"cost_management":{"is_entitled": "True"}}}'
```
1. Hit the organizations endpoint and make sure you can see all the nodes: 
(You should only see OU_001 & OU_005)

- http://127.0.0.1:8000/api/cost-management/v1/reports/aws/costs/?group_by[org_unit_id]=OU_001
- http://127.0.0.1:8000/api/cost-management/v1/reports/aws/costs/?group_by[or:org_unit_id]=OU_001&group_by[or:org_unit_id]=OU_005

2. Negative testing (all of these should be 403):
- http://127.0.0.1:8000/api/cost-management/v1/reports/aws/costs/?group_by[or:org_unit_id]=OU_002
- http://127.0.0.1:8000/api/cost-management/v1/reports/aws/costs/?group_by[org_unit_id]=OU_002
- http://127.0.0.1:8000/api/cost-management/v1/reports/aws/costs/?group_by[or:org_unit_id]=OU_002&group_by[or:org_unit_id]=OU_001

Feel free to try some different combinations, I just wanted to provide some examples.

-----

TODO:

- [x] Reach out to spank to see if we need any doc updates.
- [x] Finish running smoke tests and see if this breaks @lcouzens rbac tests


myersCody:
Smoke Tests: 
```

======================================================================================= 3981 passed, 80 skipped, 1734 deselected in 2126.02s (0:35:26) =======================================================================================
(nise) bash-3.2$
```

dccurtis (approved):
Nice work Cody!  Great PR write up 👍

PR #2415, opened 2020-09-24
Cost-458 - 461 Openshift nodes, projects and clusters endpoints and cost model endpoint for Resource Types

lrlevine6:
This PR creates endpoints for Openshift nodes, projects and clusters as well as an endpoint for cost model resource types. The endpoints created are as follows:

- Openshift Clusters: `/api/cost-management/v1/resource-types/openshift-clusters/`
- Openshift Nodes: `/api/cost-management/v1/resource-types/openshift-nodes/`
- Openshift Projects: `/api/cost-management/v1/resource-types/openshift-projects/`
- Cost Model Resource Types: `/api/cost-management/v1/resource-types/cost-models/`

Each of these endpoints as well as previously created endpoints for resource types have been updated to include ordering. By default, the ascending order will be returned. To dictate the direction of ordering add the following query:

for ascending: `?ordering=value` 
for descending: `?ordering=-value` 

Combined the would create: 
`/api/cost-management/v1/resource-types/openshift-projects/?ordering=-value`

To confirm the creation of these endpoints please visit them and play around with the sorting using the above syntax.


dccurtis:
Looks great @lrlevine6 !  Still playing around with the queries and want to check out a situation with multiple accounts.  So far the only deviation I see from the target spec is the `rates` -> `cost-models` URL change.

Have there been any discussions on when we want to update the OpenAPI spec?  Since this is the last round of updates for this epic.  Maybe we should add it to this PR?

dccurtis (approved):
👍

PR #2416, opened 2020-09-25
COST-553 Add pvc to metrics

boaz0:
This adds "PVC" to metrics both in the API and in masu.


boaz0:
The CI fails because of

```
Warning: Your Pipfile requires python_version 3.8, but you are using 3.6.9 (/usr/bin/python3).
  $ pipenv --rm and rebuilding the virtual environment may resolve the issue.
  $ pipenv check will surely fail.
```

Now that's odd because I see it setups Python 3.8 so I am not sure why it fails.
In addition, running the tests on my machine it works fine (after upgrading to Python 3.8 of course).

cgoodfred (approved):
LGTM 👍

boaz0:
I got regression in the unit tests after updating to master

boaz0:
Never mind :grin:

dccurtis (approved):

boaz0:
Thanks!

PR #2417, opened 2020-09-25
COST-565: test jira integration

dccurtis:


dccurtis:
Link showed up in jira.  Closing to see if jira state updates.

PR #2422, opened 2020-09-28
COST-556: Remove aws org unit tree when aws source is deleted.

myersCody:
### How to test:

Add an org tree which you can do with `make load-test-customer-data`. After the tree has been added you can then delete a source to see the tree disappear. It would also help to run a real scan to make sure the provider is being populated. If you add the default aws source, then do a get request to the crawler hierarchy endpoint to watch a scan run. 

-----

#### Current Smoke Tests

- Smoke Test fixes: https://gitlab.cee.redhat.com/insights-qe/hccm-plugin/-/merge_requests/675

TODO: 
- [x] Fix smoke tests


myersCody:
I merged in master and a new migration was added to master that is conflicting with the migration I added for this pr. I will need to delete my migration and recreate it to get past the conflicting migration files.

dccurtis (approved):

PR #2424, opened 2020-09-29
Fix the container name to point to UBI8 for manifest file.

chambridge:
* Update manifest script
* Run make requirements


maskarb (approved):
lgtm!

PR #2429, opened 2020-10-01
COST-589: recreating tenant in sources client after being cleaned

dccurtis:
This should address: https://sentry.io/organizations/project-koku/issues/1865053269/?project=5183020&referrer=slack

When the tenant cleaner task removes a tenant, the sources client is failing to recreate the tenant on the OCP source creation flow.  This doesn't apply to AWS and Azure since they require an API PATCH call and the tenant is recreated successfully on that path.

**Testing**
1. Set DEVELOPMENT=False and create an OCP source without hitting an APIs (to ensure no tenant exists and sources client creates it)
2. Force tenant cleaner to run and remove tenant
3. Re-add source and verify tenant is re-created and source create is successful

**Test Results**
[sources_tenant_recreate_ut.txt](https://github.com/project-koku/koku/files/5313458/sources_tenant_recreate_ut.txt)

### DIFF
--- a/koku/api/provider/provider_builder.py
+++ b/koku/api/provider/provider_builder.py
@@ -112,7 +112,10 @@ class ProviderBuilder:
         """Call to create provider."""
         connection.set_schema_to_public()
         context, customer, _ = self._create_context()
-        tenant = Tenant.objects.get(schema_name=customer.schema_name)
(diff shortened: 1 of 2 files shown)


maskarb (approved):
lgtm!

PR #2430, opened 2020-10-02
COST-557: update to django 3.1

blentz:
This upgrades us to Django 3.1.2, which is the latest version.

Notable changes in this PR
- Fix a bug in cost_models that was logging a validation exception without re-raising it to throw a ValidationError.
- Swap from the deprecated contrib postgres JSONField to the first-class version
- Added custom `migrate` and `migrate_schemas` commands to work around `django-tenant-schemas` being a dead project and not being updated for Django 3.1.x. There will be a future PR to replace django-tenant-schemas with something else.

Testing:
- unit tests all pass
- smoke test results:


Red-HAP (approved):
Overrides are magic but it looks good to me if it's passing tests.

koku/masu/management/commands/migrate.py
```diff
@@ -0,0 +1,35 @@
...
+    """Override the migrate command from django-tenant-schemas.
+
+    This override is here to workaround a dead upstream.
+    This enables django-tenant_schemas to work with Django 3.1.x
+    """
+
+    requires_system_checks = []
+
+
+if django_is_in_test_mode():
```
adberglund, line 34: Why was this needed?

PR #2440, opened 2020-10-07
COST-609: Fix Django 3 deprecation and settings warnings

boaz0:
When running koku server the following warnings are shown:
```
WARNINGS:
api.User.is_active: (fields.W903) NullBooleanField is deprecated. Support for it (except in historical migrations) will be removed in Django 4.0.
	HINT: Use BooleanField(null=True) instead.
django.conf.settings: (tenant_schemas.W001) You should put 'tenant_schemas' before any django core applications in INSTALLED_APPS.
	HINT: This is necessary to overwrite built-in django management commands with their schema-aware implementations.
reporting.PartitionedTable.partition_parameters: (fields.W904) django.contrib.postgres.fields.JSONField is deprecated. Support for it (except in historical migrations) will be removed in Django 4.0.
	HINT: Use django.db.models.JSONField instead.
```

This patch fixes these warnings.


dccurtis (approved):
Thanks!

boaz0:
3..2..1 :bomb:

PR #2441, opened 2020-10-07
Change github-action cache key

maskarb:
Our github-actions cache seems corrupted again. One solution is to change the cache key so that it will be recreated.

This idea came from here: [https://github.com/python-poetry/poetry/pull/1928](https://github.com/python-poetry/poetry/pull/1928). It resolved their issues, so it should work for us too.


adberglund (approved):
🧶

boaz0:
LGTM

PR #2444, opened 2020-10-08
[COST-431] Tag Based Rates Calculations

cgoodfred:
This PR adds support for tag based rates by retrieving them from the database, and calculating the monthly/infrastructure/supplementary costs defined in the rates. 

A few notes:

- SQL is not my strongest point and Andrew helped a bit, however it still needs to be looked at thoroughly. I am not sure how this will operate on extremely large datasets or on something with thousands of tags for a single tag key.
- I do not know if there is a naming convention for the way we name sql files, definitely willing to change them if anybody has a better reccomendation.
- After a discussion with Natalie, Sergio, and Andrew, it was determined that the default rate applies to every tag under a tag key that does not have its own rate defined for it and that is why it is implemented like it is. 

Testing:

1. load in the test customer data with `make create-test-customer` followed by `make load-test-customer-data`

2. check out the cost models page (http://localhost:8000/api/cost-management/v1/cost-models/) and pick one that you would like to add some tag based rates to (I recommend the Test OCP on AWS one since the pod labels are static).

3. Pull up the openshift costs page (http://localhost:8000/api/cost-management/v1/reports/openshift/costs/)  to see the current costs and either keep the page open or make a note of what they are.

4. Find that cost model you would like to edit and go to its cost model page (http://localhost:8000/api/cost-management/v1/cost-models/<uuid>/) and edit the raw data to add some tag based rates, something like the following is a good example, it may also be good to test a few different metrics and tags:
`    "rates": [
        {
            "metric": {
                "name": "cluster_cost_per_month",
                "label_metric": "Cluster",
                "label_measurement": "Currency",
                "label_measurement_unit": "cluster-month"
            },
            "tag_rates": [
                {
                    "tag_key": "app",
                    "tag_values": [
                        {
                            "tag_value": "banking",
                            "unit": "USD",
                            "usage": {
                                "unit": "USD",
                                "usage_end": null,
                                "usage_start": null
                            },
                            "value": 0.2,
                            "description": "",
                            "default": false
                        },
                        {
                            "tag_value": "mobile",
                            "unit": "USD",
                            "usage": {
                                "unit": "USD",
                                "usage_end": null,
                                "usage_start": null
                            },
                            "value": 100.0,
                            "description": "",
                            "default": true
                        },
                        {
                            "tag_value": "weather",
                            "unit": "USD",
                            "usage": {
                                "unit": "USD",
                                "usage_end": null,
                                "usage_start": null
                            },
                            "value": 5.0,
                            "description": "",
                            "default": false
                        }
                    ]
                }
            ],
            "cost_type": "Infrastructure"
        },
        {
            "metric": {
                "name": "cpu_core_usage_per_hour",
                "label_metric": "CPU",
                "label_measurement": "Usage",
                "label_measurement_unit": "core-hours"
            },
            "tag_rates": [
                {
                    "tag_key": "app",
                    "tag_values": [
                        {
                            "tag_value": "banking",
                            "unit": "USD",
                            "usage": {
                                "unit": "USD",
                                "usage_end": null,
                                "usage_start": null
                            },
                            "value": 0.7,
                            "description": "",
                            "default": false
                        },
                        {
                            "tag_value": "mobile",
                            "unit": "USD",
                            "usage": {
                                "unit": "USD",
                                "usage_end": null,
                                "usage_start": null
                            },
                            "value": 3.0,
                            "description": "",
                            "default": false
                        },
                        {
                            "tag_value": "weather",
                            "unit": "USD",
                            "usage": {
                                "unit": "USD",
                                "usage_end": null,
                                "usage_start": null
                            },
                            "value": 4.0,
                            "description": "",
                            "default": true
                        }
                    ]
                }
            ],
            "cost_type": "Supplementary"
        }
    ],`

5. Go ahead and update the cost model costs for the source uuid that cost model is associated with (http://localhost:5000/api/cost-management/v1/update_cost_model_costs/?provider_uuid=<provider_uuid>&schema=acct10001)

6. Reload the Openshift Costs page (http://localhost:8000/api/cost-management/v1/reports/openshift/costs/) and see that the values have gone up! To thoroughly check that those numbers are correct, you would need to go into the database and look at the ocpusagelineitem_daily_summary table and do math based on the core hours and the rates and comparing that with the usage costs which might not be a bad thing to check for at least a few entries. 

7. Go back to your cost model and remove the rates and set it back to `"rates": []` and then check the costs again and see that the related costs have disappeared.

Smoke Results: 
`===== 3930 passed, 80 skipped, 1736 deselected, 27 errors in 1048.77s (0:17:28) =====`

Checked on master and the results were the same, all unrelated tests:
`===== 3930 passed, 80 skipped, 1736 deselected, 27 errors in 940.33s (0:15:40) =====`


dccurtis (approved):
Well done Corey!

PR #2445, opened 2020-10-08
bugfixes for grafana dashboards

blentz:
This PR fixes a couple of issues with our dashboards:

1. Each configmap object name needs to be unique, otherwise they won't deploy correctly.
2. Specifying more appropriate Y-axis units on some graphs
3. Adding an `rds_datasource` variable and updating the RDS graphs so that they're able to display data.


dccurtis (approved):
I trust that you got this 👍

dccurtis (approved):

PR #2446, opened 2020-10-08
Update Open API for Resource Types Endpoints

lrlevine6:
This change involves updating the open api spec for recently added resource type endpoints. Here two formatted out parameters and corresponding pagination schemas are added to be used for the resource types endpoint including `name`, `path`, and `count` for each of the added endpoints and for all remaining endpoints including `value` properties.


docs/source/specs/openapi.json
```diff
@@ -1726,6 +1726,278 @@
...
+                    }
+                }
+            }
+        },
+        "/resource-types/aws-accounts/": {
+            "get": {
+                "tags": [
+                    "Resource Type"
+                ],
+                "summary": "List Resource Types For RBAC",
```
dccurtis, line 1768: Can you update these summary strings for the various endpoints?

dccurtis (approved):

PR #2447, opened 2020-10-08
COST-439 azure presto summary

adberglund:
## Summary
This adds processing for Azure in Presto. It covers both report versions we know of. This includes an account that has Azure sources with either/both report types. 

## Testing

1. Set these env vars

```
ENABLE_PARQUET_PROCESSING=True
S3_BUCKET_NAME=koku-bucket
S3_ENDPOINT=http://kokuminio:9000
S3_ACCESS_KEY=kokuminioaccess
S3_SECRET=kokuminiosecret
```

2. make docker-up-min-presto
3. make create-test-customer
4.  make load-test-customer-data
5. Hit http://127.0.0.1:8000/api/cost-management/v1/reports/azure/costs/?filter[time_scope_value]=-1&group_by[subscription_guid]=* and see both v1 and v2 Azure accounts processed.


koku/masu/processor/report_processor.py
```diff
@@ -75,7 +75,7 @@ def _set_processor(self):
             (Object) : Provider-specific report processor
 
         """
-        if settings.ENABLE_PARQUET_PROCESSING:
+        if settings.ENABLE_PARQUET_PROCESSING and self.provider_type not in (Provider.PROVIDER_OCP,):
```
adberglund, line 78:
```suggestion
        if settings.ENABLE_PARQUET_PROCESSING:
```

dccurtis (approved):
👍

PR #2452, opened 2020-10-13
Cost 584 migration check upgrade

Red-HAP:
## Ticket

[COST-584](https://issues.redhat.com/browse/COST-584)

## Description

In an attempt to be more thorough regarding detecting the need to run migrations, I have created a function that will scan all tenants that exist (meaning exist in `api_teanat` and in `pg_namespace`) and will cycle through all of them checking the latest recorded migration name by app against the latest code migration leaf by app.

If the `app.leaf` migration name is greater than the `app.latest` migration name, then migrations will be run.

If an app appears in the leaf that is not recorded, then migrations should be run.

This will mean that, when all migrations are up-to-date, this check function will run longer than normal. If, however, migrations need to be run, it will stop processing at the first schema that is out-of-date.

### Setup

1. `git fetch && git checkout COST-584-migration-check-upgrade`

#### Migrations should not be run

1. `python ./koku/manage.py check_migrations`

#### Migrations should be run

1. 
```bash
$ cat <<EOF >./koku/api/migrations/0999_eek.py
from django.db import migrations
class Migration(migrations.Migration):
    dependencies = [("api", "$(basename $(ls -1 ./koku/api/migrations/0*.py | tail -1) | sed -e 's/\.py$//g')")]
    operations = [migrations.RunSQL("""select 1;""")]
EOF

```
2. `python ./koku/manage.py check_migrations`
3. `rm ./koku/api/migrations/0999_eek.py`

### Call db function directly

1. `python ./koku/manage,py shell`
2. 
```python
import json
from koku.database import 
from django.db import connection
from django.db.migrations.executor import MigrationExecutor
from koku.database import verify_migrations_dbfunc
if connection.connection.closed:
    connection.connect()
verify_migrations_dbfunc(connection)
connection.prepare_database()
executor = MigrationExecutor(connection)
targets = executor.loader.graph.leaf_nodes()
json.dumps(dict(targets))
# Copy the string (including the surrounding single-quotes) that is the output from the json.dumps() call
exit
```
3. psql postgresql://postgres@localhost:15432/postgres
4. 
```sql
select public.app_needs_migrations(<paste string here>::jsonb);
-- returns 't' if migrations need to be run else 'f'

-- if you want to see verbose output:
select public.app_needs_migrations(<paste string here>::jsonb, true);

-- if you want to play around, change an app key value in the json string or change the name of one of the migrations.
```
5. `\q`

### DIFF
--- a/db_functions/app_migration_check_func.sql
+++ b/db_functions/app_migration_check_func.sql
@@ -1,0 +1,89 @@
+/*
+Copyright 2020 Red Hat, Inc.
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU Affero General Public License as
+published by the Free Software Foundation, either version 3 of the
+License, or (at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Affero General Public License for more details.
+
+You should have received a copy of the GNU Affero General Public License
+along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+-- Function public.app_migration_check(leaf_migrations, _verbose)
+-- Returns bool: true = run migrations; false = migrations up-to-date
+-- leaf_migrations (jsonb) = leaf migration names by app from the django code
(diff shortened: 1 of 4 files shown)


adberglund (approved):
Passed my test!

PR #2453, opened 2020-10-14
adjusting resource type operationids in spec

dccurtis:
@lcouzens noticed that the API client generated had some redundant naming in the path.


lcouzens (approved):
Saweet! Just what we need!

PR #2454, opened 2020-10-14
PoC for schema cloning for new tenant creation

Red-HAP:
## Proof of Concept: Schema Cloning for Tenant Creation

These changes will do several things:
1. Create and utilize a database function to clone schema objects to a new schema.
2. Integrate the call to this function in the koku `Tenant.save()` method
3. The `public` schema will be synced as normal via the migrations
4. A new "tenant" schema will be created and used as a template (default name: `template0`)
5. When a new tenant schema is needed, the database function is called instead of the migrations. The current state of `template0` will be copied to the new customer tenant schema name.

Some base timings on Fedora 31 using the latest squashed migrations:
* Migrations to create customer schema: **~13s - 13.4s** on a newly initialized database
* Clone of template schema: **~1.8s** on a newly initialized database

Cloned objects include:
* Sequences
* Tables (including indexes and primary key, foreign key, unique constraints)
* Views and Materialized Views (including matview indexes)
* Functions
* Triggers

Data in `partitioned_tables` and `django_migrations` tables are _always_ copied.
Other table data are copied with the `copy_data => true` parameter passed to `public.clone_schema` func. 

### Pros

1. The template schema can be created ahead of time otherwise it will be created on-demand.
2. The template is _always_ created from migrations
3. The template will be kept in-sync with the app migrations
4. The clone is not only extremely fast, it should **stay** extremely fast over time.
5. Not-too-much hammering on the Tenant model to get the base integration done

### Cons

1. It's still rough around the edges and all cases may not yet be covered.
2. Initial integration was not too hard, but `tox` falls on its face out of the gate regarding AWSOrgTree setup
3. There are probably more unknown unknowns that will be need to checked during testing

### Django

1. `git fetch && git checkout COST-581-clone-template-schema`
3. `cd koku && python ./manage.py shell && cd -`
4. 
```python
import timeit
import gc
from api.iam.models import Tenant

# See the template schema name
Tenant._TEMPLATE_SCHEMA

# Create a new customer schema
Tenant(schema_name='acct11111').save()
# You'll see some log messages about verifying the tenant schema
# If the template schema does not exist, it is created via the migrations
# Afterward, the 'acct11111' customer schema is created via the clone function

# Create another customer
Tenant(schema_name='acct22222').save()
# Note that there is no attempt to create the template as it should exist. The template will be cloned.

# Want to time things?
# Remove the template
Tenant.objects.filter(schema_name=Tenant._TEMPLATE_SCHEMA).delete()

# Create the template schema with timeit. It will use migrations
timeit.Timer('Tenant(schema_name=Tenant._TEMPLATE_SCHEMA).save()', 'gc.enable()', globals=globals()).timeit(1)

# Note the time it took for 1 statement call

# Now create a new customer schema. This will use the clone func
timeit.Timer('Tenant(schema_name="acct33333").save()', 'gc.enable()', globals=globals()).timeit(1)

# Note the time difference

exit
```

### psql

1. `git fetch && git checkout COST-581-clone-template-schema`
3. `psql postgresql://postgres@localhost:15432/postgres`
4. 
```sql
-- show namespaces
\dn

-- get records in api_tenant
select * from public.api_tenant;

-- clone template0
select public.clone_schema('template0', 'acct5483496', true);

-- The func returns a boolean. So a value of 't' is success and 'f' is failure.

-- clone template0 and see messages
select public.clone_schema('template0', 'acct7847473', true, true);

-- want to time things?
\timing on
select public.clone_schema('template0', 'acct84684622', true);

-- a boolean you say? what returns a false?
select public.clone_schema('template0', 'acct84684622', true);
-- trying to clone to a schema that already exists will simply return false.

-- what if the source does not exist?
select public.clone_schema('eek', 'acct84684622', true);
-- This should also return false

\q
```

## Migrations

Migrations were tested manually and do apply to the template as designed.


koku/api/iam/models.py
```diff
@@ -62,8 +86,101 @@ class Meta:
...
+
+        # Verify name structure
+        _check_schema_name(self.schema_name)
+
+        # Make sure all of our special pieces are in play
+        ret = self._check_clone_func()
+        if not ret:
+            errmsg = "Missing clone_schema function even after re-applying the function SQL file."
+            LOG.critical(errmsg)
+            return CloneSchemaFuncMissing(errmsg)
```
blentz, line 168:
```suggestion
            raise CloneSchemaFuncMissing(errmsg)
```

koku/api/iam/models.py
```diff
@@ -62,8 +86,101 @@ class Meta:
...
+        result = None
+        with conn.cursor() as cur:
+            # This db func will clone the schema objects
+            # bypassing the time it takes to run migrations
+            sql = """
+select public.clone_schema(%s, %s, copy_data => true) as "clone_result";
+"""
+            LOG.info(f'Cloning template schema "{self._TEMPLATE_SCHEMA}" to "{self.schema_name}" with data')
+            cur.execute(sql, [self._TEMPLATE_SCHEMA, self.schema_name])
+            result = cur.fetchone()
```
blentz, line 145: This could raise a few different exceptions, if the connection or cursor are broken, or if fetchone() returns something other than one result.

Do we need to catch any of those exceptions?

Red-HAP: If the connection or cursor are broken, then there's precious little that can be done to gracefully handle the condition.
I have placed python checks around the call of this function. If it fails, then the all further DB transaction traffic will fail until it is closed. 
If any exception is thrown from the DB I'd rather it bubble up through the existing infrastructure.

koku/api/iam/models.py
Red-HAP: Oops.

koku/api/iam/models.py
Red-HAP: Also the function will return only a boolean. Anything else will be an exception from the database that will be picked up via psycopg2 and wrapped by django.

blentz (approved):
:+1:

dccurtis (approved):

PR #2462, opened 2020-10-20
update Pipfile

blentz:
Simple PR to update the Pipfile & friends. This removes an unneeded dependency.


lrlevine6 (approved):
Great catch and LGTM!

PR #2465, opened 2020-10-22
COST-641: Update sources API for GCP accounts.

myersCody:
Jira Issue: https://issues.redhat.com/browse/COST-641

**Why the change is needed**
This issue is to update the sources api for the GCP accounts. Since GCP will be using BigQuery in the future to download the billing data, we need to prepare the GCP source to collect the information we will need to use Big Query. 

The sources client will be collecting the table_id, whereas the billing_source in the sources API will be collecting the project_id & the dataset. 

**Real GCP source environment variable setup**
Some of the extra work accomplished in this pr is to setup an easy way to create both a `GCP-local` & a `GCP` source. In order to set up an actual GCP source you will need to fill in four environment variables:

1. `GOOGLE_APPLICATION_CREDENTIALS` This is the file path service account json file used to access your GCP data. Remember that the filepath is the file path inside of the container and not inside of your local machine. I recommend putting the json inside of the `koku/testing` directory since it is currently in the `.gitignore` for the project. That way you don't accidentally do a git add on it, and it will automatically be docker volumed into the container.
```
GOOGLE_APPLICATION_CREDENTIALS=/koku/testing/name.json
```
2. The other new envs:
```
GCP_DATASET
GCP_TABLE_ID
GCP_PROJECT_ID
```

#### How to create a real GCP source
```
make gcp-source gcp_name=my_gcp_source
```

#### How to test
**Test Real GCP Source**
1) Create a real GCP source and make sure it makes it into the database:
```
make gcp-source gcp_name=my_gcp_source
```
Then check the database:
```
postgres=# SELECT name, account_id, source_type, authentication, billing_source  FROM public.api_sources WHERE source_type='GCP';
      name       | account_id | source_type |                    authentication                     |                                                     billing_source
-----------------+------------+-------------+-------------------------------------------------------+------------------------------------------------------------------------------------------------------------------------
 my_gcp_source | 10001      | GCP-local   | {"credentials": {"project_id": "project_id_example"}} | {"data_source": {"dataset": "dataset_id_example", "table_id": "table_id_example", "project_id": "project_id_example"}}
```
**Test GCP-local source**
I updated the `make docker-reinitdb-with-sources` to now automatically include a GCP-local source.
```
postgres=# SELECT name, account_id, source_type, authentication, billing_source  FROM public.api_sources WHERE source_type='GCP-local';
      name       | account_id | source_type |                    authentication                     |                                                     billing_source
-----------------+------------+-------------+-------------------------------------------------------+------------------------------------------------------------------------------------------------------------------------
 Test GCP source | 10001      | GCP-local   | {"credentials": {"project_id": "project_id_example"}} | {"data_source": {"dataset": "dataset_id_example", "table_id": "table_id_example", "project_id": "project_id_example"}}
```

**Negative Testing**
You can do some negative testing by putting an incorrect project_id in your environment variable for `GCP_PROJECT_ID` and running `make gcp-source gcp_name=my_gcp_source`

TODO:

- [x] Run Smoke Tests


myersCody:
Smoke Tests Results:
```
========================================================================================================== short test summary info ===========================================================================================================
FAILED lib64/python3.6/site-packages/iqe_cost_management/tests/rest_api/v1/test_metrics.py::test_api_ocp_metrics - AssertionError: Expected metrics not all listed
FAILED lib64/python3.6/site-packages/iqe_cost_management/tests/rest_api/v1/test_source.py::test_api_ocp_on_aws_source_crud - AssertionError: Raw calculated value does not match report value
FAILED lib64/python3.6/site-packages/iqe_cost_management/tests/rest_api/v1/test_source.py::test_api_aws_source_raw_calcs - AssertionError: Raw calculated value does not match report value
===== 3 failed, 3954 passed, 80 skipped, 1736 deselected in 1804.93s (0:30:04) =====
```

I am pretty certain that these failures are unrelated to the changes that I have made here.

koku/providers/gcp/provider.py
```diff
@@ -11,6 +13,8 @@
 
 LOG = logging.getLogger(__name__)
 
+REQUIRED_IAM_PERMISSIONS = ["bigquery.jobs.create", "bigquery.tables.getData"]
```
dccurtis, line 16: We can add the `bigquery.tables.list` check when we handle the new account setup flow where the table_id isn't immediately available at source create time.

koku/providers/gcp/provider.py
```diff
@@ -24,24 +28,34 @@ def cost_usage_source_is_reachable(self, credentials, data_source):
...
+            check_permissions = {"permissions": REQUIRED_IAM_PERMISSIONS}
+            request = service.projects().testIamPermissions(resource=project, body=check_permissions)
+            response = request.execute()
+            permissions = response.get("permissions", [])
+
+            for required_permission in REQUIRED_IAM_PERMISSIONS:
+                if required_permission not in permissions:
+                    key = "authentication.project_id"
+                    err_msg = f"Improper IAM permissions: {permissions}."
+                    raise serializers.ValidationError(error_obj(key, err_msg))
```
dccurtis, line 49: FYI: There is a list of user facing error strings that we can include for GCP when we complete the platform sources integration: https://github.com/project-koku/koku/blob/master/koku/providers/provider_errors.py.  We can surface those messages here.

koku/api/provider/test/tests_serializers.py
```diff
@@ -367,7 +367,9 @@ def test_create_gcp_provider(self):
             "name": "test_provider_one",
             "type": Provider.PROVIDER_GCP.lower(),
             "authentication": {"credentials": {"project_id": "gcp_project"}},
-            "billing_source": {"data_source": {"bucket": "test_bucket", "report_prefix": "precious"}},
+            "billing_source": {
+                "data_source": {"dataset": "test_dataset", "table_id": "test_table_id", "report_prefix": "precious"}
```
dccurtis, line 371: I know the `report_prefix` was here before... But what what is the purpose for including it here?  Maybe that it's just ignored?

dccurtis (approved):
Was able to successfully add a GCP source and do the negative IAM test.  Nice work Cody!

koku/providers/gcp/provider.py
myersCody: Ohhh. Thanks for pointing this out @dccurtis I didn't know this existed.

koku/api/provider/test/tests_serializers.py
myersCody: I didn't know if the report prefix was used to help generate the path of where the csv file would go during the downloader. So, I went ahead and left it in here to make sure that it could still be used. However, while working on the downloader if we decide we don't need it then we can cut it out.

koku/providers/gcp/provider.py
myersCody: I made sure to include this in the `Check to see if the Table Id is created` issue in the design doc since it will be the issue that has to add that permission.

PR #2466, opened 2020-10-23
COST-642, COST-665: GCP Sources Integration

dccurtis:
**Overview**
This change adds preliminary GCP support for sources client.  The design for how credentials will be stored in platform sources is still a work in progress.  But Using the attached platform branch we can get preliminary GCP sources support which will allow us to continue development while the integration design unfolds with the platform.

Some benefits to having this support early is so we can workout the details for:
* User Facing error message pushing
* Polling for the table-id after source create when a new GCP account is setup.  This will happen several hours after the source is created on a new GCP account.  The Sources Client will need to be able to get the table-id from the user's account when it's available.  At which point we can complete the account onboarding into Cost Management

**Testing**
1. Create source with WIP platform sources changes for GCP
2. Create source with WIP platform sources changes and do not include table id.  Verify that it is discovered and source is created successfully.

Note: WIP Platform Sources change: https://github.com/RedHatInsights/sources-api/compare/master...dccurtis:sources_gcp

**Testing Results**
[sources_gcp_initial_support.txt](https://github.com/project-koku/koku/files/5431845/sources_gcp_initial_support.txt)
[gcp_source_table_discovery.txt](https://github.com/project-koku/koku/files/5478198/gcp_source_table_discovery.txt)


myersCody:
Hey @dccurtis I think we are going to need to update the required iam list here:  https://github.com/project-koku/koku/blob/4e54e24420bf88bd38f7f241e027752a9f955c0d/koku/providers/gcp/provider.py#L16

To include: `bigquery.tables.list` to make sure we have access to the table_id.

myersCody (approved):
Awesome work @dccurtis

PR #2482, opened 2020-10-30
COST-645: Download GCP report using Google BigQuery

myersCody:
How to test

Run `make gcp-source gcp_name=test_gcp_source` and then check to see if the csv file was created.


myersCody:
Codecov patch is being weird, this pr really only touches the `gcp_report_downloader.py`
```
koku/masu/external/downloader/gcp/gcp_report_downloader.py                     115     10      4      1    91%   101-104, 113, 256-264, 275, 274->275
```

dccurtis (approved):
LGTM! I was able to use this and download data two days in a row.  As we discussed changes could be needed once processing work begins.  Great job!

PR #2484, opened 2020-11-02
Fix date ranges in tests

adberglund:
## Summary
Fix two test that fail now with months of different lengths.


boaz0 (approved):
LGTM :+1:

PR #2486, opened 2020-11-02
COST-659: Update tag rate serializer to only allow one metric per cost type

myersCody:
How to test:

I added an example, tag rate to the on prem cost model json. I would duplicate that metric and then try running `make load-test-customer-data`.


cgoodfred (approved):
LGTM!

PR #2488, opened 2020-11-04
Cost 702 api enhanced metrics

adberglund:
## Summary
This will add the account ID to a set of Django Prometheus API metrics. We would then be able to filter/group request latency, response status codes, and requests by view per account. 

## Testing

1. Added `ACCOUNT_ENHANCED_METRICS=True` environment variable.
2. Set pdb in `AccountEnhancedMetricsAfterMiddleware`

```
(Pdb) new_labels
{'account': '10001', 'view': 'sources-list', 'transport': 'http', 'method': 'GET'}
(Pdb) metric._name
'django_http_requests_total_by_view_transport_method'
```

### DIFF
--- a/koku/koku/middleware.py
+++ b/koku/koku/middleware.py
@@ -380,7 +380,7 @@ class DisableCSRF(MiddlewareMixin):
 class AccountEnhancedMetrics(Metrics):
     def register_metric(self, metric_cls, name, documentation, labelnames=(), **kwargs):
         if name in EXTENDED_METRICS:
-            labelnames += ("account",)
+            labelnames.extend(("account",))
         return super().register_metric(metric_cls, name, documentation, labelnames=labelnames, **kwargs)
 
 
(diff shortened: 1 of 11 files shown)


koku/koku/middleware.py
```diff
@@ -366,3 +375,32 @@ def process_request(self, request):
...
+
+class AccountEnhancedMetricsBeforeMiddleware(PrometheusBeforeMiddleware):
+    metrics_cls = AccountEnhancedMetrics
+
+
+class AccountEnhancedMetricsAfterMiddleware(PrometheusAfterMiddleware):
+    metrics_cls = AccountEnhancedMetrics
+
+    def label_metric(self, metric, request, response=None, **labels):
+        new_labels = labels
```
blentz, line 395: All of these should have docstrings for the classes and methods. The docstrings need to document how/where the classes are used. If it makes sense, they can also reference upstream docs.

blentz (changes requested):
The code itself looks okay. But it needs documentation.

blentz (approved):

PR #2490, opened 2020-11-05
Allow list in rbac attribute filter value

adberglund:
## Summary
The new custom resource work in RBAC has changed a comma separated string to a list of values. We were not handling this and this should allow either string or list. 

## Testing
Currently just the unit test which mirrors what is returned by the RBAC API, but real testing will require RBAC up. Will test immediately in CI.


chambridge (approved):
LGTM

PR #2492, opened 2020-11-05
COST-623: checking filter source type settings when updating

dccurtis:
Updating settings API to ensure that only source types listed in POST data are updated.

**Testing**
1.  Enable tags for OCP and AWS
2. Clear AWS enabled tags and verify OCP remains

**Test Results**
[settings_source_type_ut.txt](https://github.com/project-koku/koku/files/5496546/settings_source_type_ut.txt)


boaz0 (approved):
:+1:

PR #2493, opened 2020-11-06
AWS tag enablement settings out of DEV

dccurtis:
Once @lcouzens wraps up testing AWS tags enablement and the corresponding settings update to support multiple source types, this change will turn on AWS tag enablement.


dccurtis:
@lcouzens wrapped up the API testing so this should be good to go.  @boaz0 can you take a look and sanity check it from the DDF perspective?  We discussed offline but just one last check before enabling AWS in CI/production

boaz0 (approved):
:+1:

PR #2467, opened 2020-10-26
[COST-451] AWS Tag Enablement

cgoodfred:
This PR is targeted at allowing AWS tags that are enabled/disabled to reflect those changes in the daily summary. It also puts new tags into the enabledtagkeys table to make them enabled by default. 

The data migration was done to fill the `aws_enabledtagkeys` table in production since the sql checks to see if a tag exists in the `aws_tagsummary` table before adding it to the enabled tag keys table. It does this to prevent adding keys that already exist and have been disabled when a report file is processing with those keys in it.

Testing:

1. Generate some nise data, (here is an example yml but you will need to remove the .txt, github doesn't let you upload .yml's):
[aws_tags_enabled_testing.yml.txt](https://github.com/project-koku/koku/files/5440638/aws_tags_enabled_testing.yml.txt)
```nise report aws --static-report-file aws_tags_enabled_testing.yml --aws-s3-bucket-name testing/local_providers/aws_local_0 --aws-s3-report-name testingAwsTags```

2. create a source for the data
3. Go check out the database and check the `reporting_awsenabledtagkeys` table and make sure all the tags are there, there should be 5:
 ```
id |     key     
----+-------------
  1 | crazy
  2 | environment
  3 | show
  4 | title
  5 | trash
(5 rows)
```
4. Check out the tags on the `reporting_awscostentrylineitem_daily_summary` table and make sure you see the appropriate distinct tags for the data you use, here are the ones from the example yml:
```
                  tags                   
-----------------------------------------
 {"show": "house", "environment": "dev"}
 {"trash": "wednesday"}
 {"crazy": "dogs"}
 {"title": "green", "trash": "tuesday"}
(4 rows)
```

5. Note the total costs for aws:  `http://127.0.0.1:8000/api/cost-management/v1/reports/aws/costs/`

6. Disable (delete it from the `reporting_awsenabledtagkeys` table) some tags, I have removed the trash tag from my example

7. Resummarize the data

8. Check the distinct tags from the `reporting_awscostentrylineitem_daily_summary` table again and make sure the tags do not include the disabled tag. The example now looks like the following:
```
                  tags                   
-----------------------------------------
 {"title": "green"}
 
 {"show": "house", "environment": "dev"}
 {"crazy": "dogs"}
(4 rows)
```
9. Check the aws cost and ensure that they have not changed.

10. Also test OCP on AWS with a similar method.

PR #2501, opened 2020-11-12
[COST-630] azure tag enablement

cgoodfred:
This PR is targeted at allowing Azure tags that are enabled/disabled to reflect those changes in the daily summary. It also puts new tags into the enabledtagkeys table to make them enabled by default. This PR is similar to the AWS PR here: https://github.com/project-koku/koku/pull/2467 with one difference, the `azureenabledtagkeys` table is populated at the beginning of the daily_summary sql since there is not a sql file for daily processing on azure.

The data migration was done to fill the `azureenabledtagkeys` table in production since the sql checks to see if a tag exists in the `azure_tagsummary` table before adding it to the enabled tag keys table. It does this to prevent adding keys that already exist and have been disabled when a report file is processing with those keys in it.

Testing:

1. you can use the make commands, or your own nise file to generate some azure data, `make create-test-customer` and `make load-test-customer-data`

2. After that is done you should check the `reporting_azureenabledtagkeys` table and make sure it was filled with the tags from the data correctly. If using the make commands it should be similar to this:
```
 id |     key      
----+--------------
  3 | app
  4 | environment
  5 | storageclass
  6 | version
(4 rows)
```

3. Check out the tags on the `reporting_azurecostentrylineitem_daily_summary` table and make sure you see the appropriate distinct tags for the data you use

4. Note the total costs for azure: http://127.0.0.1:8000/api/cost-management/v1/reports/azure/costs/

5. Disable (delete it from the `reporting_awsenabledtagkeys` table) a tag or two

6. Resummarize the data

7. Check the distinct tags from the `reporting_azurecostentrylineitem_daily_summary` table again and make sure the tags do not include the disabled tag. 

8. Check the azure cost and ensure that they have not changed.

9. Also test OCP on Azure with a similar method.

Smoke Results:
on master: `========================================= 1 failed, 2532 passed, 47 skipped, 761 deselected, 6 errors in 1504.56s (0:25:04) ==========================================`

on this branch: `========================================= 2 failed, 2531 passed, 47 skipped, 761 deselected, 6 errors in 1533.60s (0:25:33) ==========================================` 

The difference was this test which doesn't seem like it should be impacted by the changes to azure tags so maybe it was a random failure?: 
```FAILED tests/rest_api/v1/test_aws_org_units.py::test_api_aws_org_units_returned_data_validation[org_unit_with_region] - IndexError: list index out of range```


dccurtis (approved):
👍

PR #2506, opened 2020-11-16
COST-669: Initial update to process gcp reports.

myersCody:
Jira Issue: https://issues.redhat.com/browse/COST-669

The initial changes needed to get the line item daily populated with data from the csv report. 

#### How to test:
I recommend during testing, to run with a fresh db. So, `make docker-reinitdb`.

After you have a fresh db, you can then add the real gcp source with:
```
make gcp-source gcp_name=test_gcp_source
```

After you the gcp source is created, it will automatically kick of the download, and start the processing. Don't be concerned if you do see an error message in the docker logs, because the summarization of the gcp report is not setup. So, the `update_summary_tables` celery task will fail.

However, after the report has been processed, you will then want to check the `report_gcp*` tables to make sure that data has been populated inside of those tables. You can either use pgadmin your psql queries, whichever is easier for you. 

After the initial upload of data it now because time to check the `_delete_line_items_in_range` 3 day range section. So, unlike the other downloaders we currently have in koku, the gcp downloader works a little different. Since we can't control when services will push the cost data to the gcp dataset you could end up with "partial" data being collected for a given day.

**TLDR;** Example about gcp downloader

> So, say 11/18/2020 we ran the gcp scan at 9am and only got information about Service A. However, service B pushed it's cost data to the dataset at 2pm. Then the next day we ran the scan again and we got data for `Service A` & `Service B`. However in the processing side of things we got to make sure we don't calculate the Service A cost twice, but still need to calculate the cost for Service B.  The simple solution was to delete all the cost item daily rows that have usage starts on 11/18/2020 and reenter the data to make sure don't duplicate and we have the freshest information possible in our 3 day range.

So, we need to recreate the scenario in the example above in our manual testing, but obviously I can't make yall wait a day to run the scan again 😂 

Therefore, with a little sql voodoo magic, we can change the usage start data.
```
UPDATE public.reporting_common_costusagereportmanifest SET assembly_id='1234:1234:1' WHERE id=1;
UPDATE acct10001.reporting_gcpcostentrylineitemdaily SET start_time = start_time - INTERVAL '1 DAY';
UPDATE acct10001.reporting_gcpcostentrylineitemdaily SET end_time = end_time - INTERVAL '1 DAY';
SELECT COUNT(*) FROM acct10001.reporting_gcpcostentrylineitemdaily WHERE start_time <= NOW() - INTERVAL '3 DAY';
SELECT COUNT(*) FROM acct10001.reporting_gcpcostentrylineitemdaily WHERE start_time >= NOW() - INTERVAL '3 DAY';
```
**TLDR;** what this sql is doing

> The first thing I do is change the assembly_id in the manifest so that the code won't think the report is the same as the previous run. Then I am setting the start_time to 1 day back & the end_time to one day back for every single row. 

This isn't exactly a perfect solution, but it basically allows a little wiggle room for getting partial data outside of our 3 day range. Then we can hit the download endpoint and rerun the downloader/processor and check to make sure we didn't exactly duplicate our data again.


koku/masu/database/gcp_report_db_accessor.py
```diff
@@ -16,3 +24,54 @@ def __init__(self, schema):
...
+            bills = self._get_db_obj_query(table_name).values(*columns)
+            return {(bill["billing_period_start"], bill["provider_id"]): bill["id"] for bill in bills}
+
+    def get_cost_entry_bills_query_by_provider(self, provider_uuid):
+        """Return all cost entry bills for the specified provider."""
+        table_name = GCPCostEntryBill
+        with schema_context(self.schema):
+            return self._get_db_obj_query(table_name).filter(provider_id=provider_uuid)
+
+    def mark_bill_as_finalized(self, bill_id):
```
dccurtis, line 44: Do you know if GCP has billing finalization?

koku/masu/processor/gcp/gcp_report_processor.py
```diff
@@ -60,17 +95,120 @@ def __init__(self, schema_name, report_path, compression, provider_uuid, manifes
...
+                f" bill ID: {bill_id}\n"
+                f" on or after {scan_start}."
+            )
+            LOG.info(log_statement)
+            line_item_query.delete()
+
+        return True
+
+    def _process_tags(self, row):
+        """Return a JSON string of AWS resource tags.
```
dccurtis, line 158:
```suggestion
        """Return a JSON string of GCP tags.
```

koku/masu/processor/gcp/gcp_report_processor.py
```diff
@@ -60,17 +95,120 @@ def __init__(self, schema_name, report_path, compression, provider_uuid, manifes
...
+
+        Args:
+            row (dict): A dictionary representation of a CSV file row
+
+        Returns:
+            (str): A JSON string of AWS resource tags
```
dccurtis, line 164:
```suggestion
            (str): A JSON string of GCP tags
```

koku/masu/processor/gcp/gcp_report_processor.py
```diff
@@ -60,17 +95,120 @@ def __init__(self, schema_name, report_path, compression, provider_uuid, manifes
...
+        if isinstance(labels, str):
+            labels = json.loads(labels.replace("'", '"'))
+        for label in labels:
+            key = label["key"]
+            value = label["value"]
+            tags_dict[key] = value
+        return json.dumps(tags_dict)
+
+    def _get_usage_type(self, row):
+        """Return instance_type if there is one.
```
dccurtis, line 178: More comment updates :)

myersCody: @dccurtis Starting to see how bad I am at copying and pasting 😂 

I will push up these changes soon, I am going to do it as one commit since I am going to remove the finalize report method. That was a good catch.

dccurtis (approved):

PR #2512, opened 2020-11-17
scripting bugfixes

blentz:
This is a small batch of bugfixes to some of the dev/test scripts.

### DIFF
--- a/scripts/create_test_customer.py
+++ b/scripts/create_test_customer.py
@@ -47,6 +47,7 @@ koku:
 import argparse
 import os
 import sys
+import time
 from base64 import b64encode
 from json import dumps as json_dumps
 from urllib.parse import quote
@@ -99,6 +100,8 @@ class KokuCustomerOnboarder:
         print("\nAdding customer...")
         response = requests.get(self.endpoint_base + "reports/aws/costs/", headers=get_headers(self.auth_token))
         print(f"Response: [{response.status_code}] {response.text}")
+        if response.status_code not in [200, 201]:
+            time.sleep(60)
 
     def create_source_api(self):
         """Create a Koku Source using the Koku API."""
@@ -118,6 +121,8 @@ class KokuCustomerOnboarder:
 
             response = requests.post(self.endpoint_base + "sources/", headers=get_headers(self.auth_token), json=data)
             print(f"Response: [{response.status_code}] {response.reason}")
+            if response.status_code not in [200, 201]:
+                time.sleep(60)
 
     def create_provider_source(self, source_type):
         """Create a single provider, auth, and billing source in the DB."""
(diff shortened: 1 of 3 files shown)


scripts/create_test_customer.py
```diff
@@ -99,6 +100,8 @@ def create_customer(self):
         print("\nAdding customer...")
         response = requests.get(self.endpoint_base + "reports/aws/costs/", headers=get_headers(self.auth_token))
         print(f"Response: [{response.status_code}] {response.text}")
+        if response.status_code not in [200, 201]:
+            time.sleep(60)
```
blentz, line 104: One thing I've noticed in my local environment is that creating the very first new customer takes so long that the API endpoint times out and returns a 504.

On the backend, the migrations are still running and will finish successfully, if given more time.

So, I've added some sleeps to work around this issue until a better solution is found.

scripts/e2e-deploy.sh
```diff
@@ -98,6 +98,7 @@ function check_var() {
 }
 
 check_var "REGISTRY_REDHAT_IO_SECRETS"
+check_var "QUAY_IO_SECRETS"
```
blentz, line 101: This is lining up our deploy script with recent changes to the e2e deploy process.

scripts/load_test_customer_data.sh
```diff
@@ -30,6 +30,9 @@ KOKU_PATH=$1
 START_DATE=$2
 END_DATE=$3
 
+# this is the default that's in koku.masu.config
+PVC_DIR=/var/tmp/masu
```
blentz, line 34: This fixes an issue where `oc rsync` wasn't putting OCP data into the directory that the worker was expecting. So, this fixes it so the script now copies the files into the correct location.

scripts/load_test_customer_data.sh
```diff
@@ -92,54 +95,69 @@ nise report ocp --ocp-cluster-id my-ocp-cluster-3 --insights-upload "$KOKU_PATH/
...
         oc rsync --delete $SOURCEDIR $DESTDIR
     done
-    oc rsync --delete $KOKU_PATH/testing/local_providers/azure_local/* ${WORKER_POD}:/tmp/local_container
+    for SOURCEDIR in $(ls -1d $KOKU_PATH/testing/local_providers/azure_local*)
+    do
+        DESTDIR="$(echo $SOURCEDIR | sed s#$KOKU_PATH/testing/local_providers/azure_local#/tmp/local_container#)"
+        echo "uploading nise data from $SOURCEDIR to $DESTDIR"
+        oc rsh -t $WORKER_POD /usr/bin/mkdir -vp $DESTDIR
+        oc rsync --delete $SOURCEDIR/ ${WORKER_POD}:$DESTDIR
+    done
```
blentz, line 160: This is a pathing fix for Azure-local data similar to the fix for OCP data.

dccurtis (approved):

PR #2515, opened 2020-11-18
Add env var to optionally run gunicorn in docker

adberglund:
## Summary
This is adds the option to run the koku server in docker-compose using gunicorn.


dccurtis (approved):
🦄

PR #2518, opened 2020-11-19
More header normalization for Azure

adberglund:
## Summary
We have now seen yet another Azure report header format. This time they were feeling a mix of style from column name to column name. some are `columnName` and others `ColumnName`. Why not?


dccurtis (approved):
🐫 YAARHF

PR #2520, opened 2020-11-19
Update to use koku-presto

Red-HAP:
## Ticket

[COST-689](https://issues.redhat.com/browse/COST-689)

## Description

The presto server image is no longer maintained and is pinned at a very old presto release. The  repository for koku-presto can be found [here](https://github.com/project-koku/koku-presto).

The changes here are to properly configure presto and the JVM on which it runs.

There is also a fix for a recent grafana image change that removes the grafana group.


testing/compose_files/docker-compose-presto.yml
```diff
@@ -47,13 +47,13 @@ services:
       - ./../../testing/metastore/db-data:/var/lib/hive
       - ./../../testing/hadoop/hadoop-dfs-name:/hadoop/dfs/name
       - ./../../testing/hadoop/hadoop-dfs-data:/hadoop/dfs/data
-      - ./../../testing/hadoop/hadoop-logs:/opt/hadoop/logs
+#      - ./../../testing/hadoop/hadoop-logs:/opt/hadoop/logs
```
adberglund, line 50: Why is this commented?

Red-HAP: I had a lot of problems getting that to function properly and consistently. Please check out the [ubi-hadoop Dockerfile](https://github.com/RedHatInsights/ubi-hadoop/blob/main/Dockerfile) and [ubi-hive Dockerfile](https://github.com/RedHatInsights/ubi-hive/blob/main/Dockerfile). If there's some secret sauce for this, I'd definitely like to know.

The issue I had was that the mount would change from being owned by `hadoop` to `root`. Java could not write to the dir and the Java VM would not start and the container would exit on error.

adberglund (approved):

PR #2526, opened 2020-11-24
Add a masu endpoint to return a list of active tasks in masu for QE.

myersCody:
After talking with Luke about the aws org unit automation tests he is working on, we discovered that part of the issue he was running into was a timing problem. Where he was hitting the organizations endpoint, before the crawler was finished running. Therefore, I created the `running_celery_tasks` internal masu endpoint to return a list of running celery tasks. That way he can check the uuid returned from the crawl_account_hierarchy endpoint is in that list.


lcouzens (approved):
LGTM!

PR #2527, opened 2020-11-30
[COST-644] namespace label processing

cgoodfred:
This PR adds processing for namespace labels in a similar fashion to how node labels are processed.

Testing: Create nise data with namespace labels using a static yml that includes namespace labels and create a source for the data so it loads in. Check the OCP namespace label table in the db for the relevant data from the nise data.


dccurtis (approved):
LGTM

PR #2530, opened 2020-12-01
COST-760 Forecast cost model costs

adberglund:
## Summary
- Use the provider map definitions to determine what fields to use for total cost in the forecast. 
- Added back ability to control dates for nise data in load test customer script.

## Testing
Loaded data for each source type. Checked that forecast included all cost terms for each source type.


dccurtis (approved):
👍

PR #2533, opened 2020-12-02
COST-670: Summarize GCP Reports.

myersCody:
Creating a draft pr for early feedback.


koku/masu/database/sql/reporting_gcpcostentrylineitem_daily_summary.sql
```diff
@@ -0,0 +1,128 @@
+-- TODO:
+-- 1) Figure out what to do with: usage_to_pricing_units
```
myersCody, line 2: Do I need to capture the `usage_to_pricing units` in the daily_summary?

koku/reporting/provider/gcp/models.py
```diff
@@ -81,3 +101,137 @@ class Meta:
...
+    markup_cost = models.DecimalField(max_digits=24, decimal_places=9, null=True)
+    tags = JSONField(null=True)
+    source_uuid = models.UUIDField(unique=False, null=True)
+
+    # TODO: I am not sure if these are needed or just aws specific. I went ahead
+    # commented them out in case we need them.
+    # unblended_rate = models.DecimalField(max_digits=24, decimal_places=9, null=True)
+    # blended_rate = models.DecimalField(max_digits=24, decimal_places=9, null=True)
+    # blended_cost = models.DecimalField(max_digits=24, decimal_places=9, null=True)
+
```
myersCody, line 194: I am still unsure about the comment above ^^^

adberglund: You can safely remove these.

koku/masu/database/__init__.py
```diff
@@ -60,3 +60,13 @@
     "ocp_on_azure_project_daily_summary": "reporting_ocpazurecostlineitem_project_daily_summary",
     "ocp_on_azure_tags_summary": "reporting_ocpazuretags_summary",
 }
+
+GCP_REPORT_TABLE_MAP = {
+    "line_item": "reporting_gcpcostentrylineitem",
+    "line_item_daily": "reporting_gcpcostentrylineitem_daily",
+    "line_item_daily_summary": "reporting_gcpcostentrylineitem_daily_summary",
+    "tags_summary": "reporting_gcpenabledtagkeys",
```
adberglund, line 68:
```suggestion
    "tags_summary": "reporting_gcptags_summary",
```

koku/masu/database/sql/reporting_gcpcostentrylineitem_daily_summary.sql
adberglund: I think so, or we need to do the unit conversion ourselves. Since they already did it might as well spare the computation. Example: In the API we don't want to show byte-seconds, so if we just use the usage in the pricing unit in the API we're probably good to go.

adberglund (commented):
I'll pull and run this in the morning, but it looks great!

koku/reporting/provider/gcp/models.py
```diff
@@ -81,3 +101,137 @@ class Meta:
...
+    project_name = models.CharField(max_length=256)
+    service_id = models.CharField(max_length=256, null=True)
+    service_alias = models.CharField(max_length=256, null=True, blank=True)
+    sku_id = models.CharField(max_length=256, null=True)
+    sku_alias = models.CharField(max_length=256, null=True)
+    usage_start = models.DateField(null=False)
+    usage_end = models.DateField(null=True)
+    region = models.CharField(max_length=50, null=True)
+    instance_type = models.CharField(max_length=50, null=True)
+    usage_to_pricing_units = models.DecimalField(max_digits=24, decimal_places=9, null=True)
```
adberglund, line 183: Hate to come back with more suggestions, but after thinking about this a bit more. Let's remove `usage_to_pricing_units` in the daily summary table and just populate `usage_amount` with that value from the daily table.

koku/reporting/provider/gcp/models.py
```diff
@@ -81,3 +101,137 @@ class Meta:
...
+    usage_end = models.DateField(null=True)
+    tags = JSONField(null=True)
+    usage_type = models.CharField(max_length=50, null=True)
+    region = models.CharField(max_length=256, null=True, blank=True)
+    cost = models.DecimalField(max_digits=24, decimal_places=9, null=True, blank=True)
+    currency = models.CharField(max_length=256, null=True, blank=True)
+    conversion_rate = models.CharField(max_length=256, null=True, blank=True)
+    usage_amount = models.DecimalField(max_digits=24, decimal_places=9, null=True)
+    usage_unit = models.CharField(max_length=256, null=True, blank=True)
+    usage_to_pricing_units = models.DecimalField(max_digits=24, decimal_places=9, null=True)
```
adberglund, line 135: I think `usage_to_pricing_units` would make more sense as `usage_in_pricing_units` the to makes me think the column is a conversion factor and not the actual amount.

dccurtis (approved):
Excellent work Cody!

PR #2536, opened 2020-12-02
partition gcp tables

Red-HAP:
Changes for partitioning **all** of the gcp tables in PostgreSQL

Please review and, if you have any tests, run the tests.


myersCody (approved):

koku/reporting/migrations/0154_gcp_partitioned.py
```diff
@@ -0,0 +1,92 @@
...
+        target_schema=target_schema,
+        source_schema=target_schema,
+    )
+
+    # Push the button, Frank.
+    p_converter.convert_to_partition()
+
+
+def convert_gcpcostentrylineitem_to_partitioned(apps, schema_editor):
+    convert_to_partitioned("reporting_gcpcostentrylineitem")
```
adberglund, line 43: Let's just stick with line item daily summary for now. I don't think we'll be using these long enough to worry about them being partitioned as well.

adberglund (changes requested):

koku/reporting/migrations/0154_gcp_partitioned.py
Red-HAP: No prob. I thought it would be an interesting experiment since it was to be short-lived anyway.

Red-HAP: Cody will make the changes as he's under the hood now.

PR #2539, opened 2020-12-04
COST-798 Forecast supp and infra costs too

adberglund:
## Summary
This provides independent forecasts for total cost, supplementary cost, and infrastructure cost. Because each is predicted on it's own, we do end up with predictions where supp + infra cost predictions do not sum to the total cost prediction for that day. 

## Testing 

Loaded data sets for AWS, Azure, and OpenShift
- AWS: Confirm infrastructure and cost terms give same values
- Azure: Confirm infrastructure and cost terms give same values
- OpenShift on AWS: Confirm infrastructure and cost terms give same values
- OpenShift on Azure: Confirm infrastructure and cost terms give same values
- OpenShift on All: Confirm infrastructure and cost terms give same values
- OpenShift: Confirm non-zero different forecasts for infrastructure, supplementary, and cost terms with an OpenShift cluster with corresponding cost model with infra and supp terms.


koku/forecast/test/tests_forecast.py
```diff
@@ -52,23 +49,23 @@
 LOG = logging.getLogger(__name__)
 
 
-class MockDateHelper(DateHelper):
-    """Testing class."""
+class MockQuerySet:
+    def __init__(self, data):
+        """Set the queryset data as a instance variable."""
+        self.data = data
+        # super().__init()
```
adberglund, line 56:
```suggestion
```

dccurtis (approved):

PR #2543, opened 2020-12-07
COST-796: Allow for $0 rate

dccurtis:
@chargio described a use case where he wanted to prevent the default tag rate from being applied.  Comment is linked in the jira issue.

Keeping this as a Draft PR until I have a chance to discuss more with Sergio to ensure this will work.


adberglund (approved):

boaz0:
@dccurtis I will open a follow-up ticket for UI.

PR #2545, opened 2020-12-08
COST-748: Create cost materialized views for gcp.

myersCody:
I created the cost materialized views for GCP.

# How to test

Bring up the django server, and run `make gcp-source` make target. After the data is downloaded, process, and summarized check the materialized view tables for GCP and make sure they were populated correctly.


koku/reporting/migrations/0156_gcp_cost_summary.py
```diff
@@ -0,0 +1,135 @@
...
+                DROP INDEX IF EXISTS gcp_cost_summary_project;
+                DROP MATERIALIZED VIEW IF EXISTS reporting_gcp_cost_summary_by_project;
+
+                DROP INDEX IF EXISTS gcp_cost_summary_region;
+                DROP MATERIALIZED VIEW IF EXISTS reporting_gcp_cost_summary_by_region;
+
+                DROP INDEX IF EXISTS gcp_cost_summary_service;
+                DROP MATERIALIZED VIEW IF EXISTS reporting_gcp_cost_summary_by_service;
+
+                CREATE MATERIALIZED VIEW reporting_gcp_cost_summary AS(
```
adberglund, line 26: Instead of manually creating these here with SQL, you can instead make a function to load them from the files you've created. See https://github.com/project-koku/koku/blob/master/koku/reporting/migrations/0146_auto_20200917_1448.py#L9-L16.

This way we don't end up with different versions of the views in migration and the file becomes the single source of truth.

koku/reporting/provider/gcp/sql/views/reporting_gcp_cost_summary.sql
```diff
@@ -0,0 +1,22 @@
+DROP INDEX IF EXISTS gcp_cost_summary;
+DROP MATERIALIZED VIEW IF EXISTS reporting_gcp_cost_summary;
+
+CREATE MATERIALIZED VIEW reporting_gcp_cost_summary AS(
+    SELECT row_number() OVER(ORDER BY usage_start, source_uuid) as id,
+        usage_start,
+        usage_start as usage_end,
+        sum(unblended_cost) as unblended_cost,
+        sum(markup_cost) as markup_cost,
+        currency
```
adberglund, line 10:
```suggestion
        max(currency) as currency
```

koku/reporting/provider/gcp/sql/views/reporting_gcp_cost_summary.sql
```diff
@@ -0,0 +1,22 @@
...
+        source_uuid
+    FROM reporting_gcpcostentrylineitem_daily_summary
+    -- Get data for this month or last month
+    WHERE usage_start >= DATE_TRUNC('month', NOW() - '1 month'::interval)::date
+    GROUP BY usage_start, source_uuid, currency
```
adberglund, line 15:
```suggestion
    GROUP BY usage_start, source_uuid
```

adberglund: You can duplicate this change across the board so you don't have to group by the currency term.

adberglund (changes requested):

koku/reporting/migrations/0156_gcp_cost_summary.py
myersCody: Oohhhh. That is new, I like that a lot better. I will make that change now.

adberglund (approved):
🎅

PR #2547, opened 2020-12-10
Working on funcs to move default partition data to new partitions

Red-HAP:
# Ticket

[COST-812](https://issues.redhat.com/browse/COST-812)

# Description

Functions that will move partitioned data from the default partition by month to newly created month partitions.

It's broken up into a few components:
* class `PartitionDefaultData` : This is the meat of the processing. The flow works like this: Given a schema, partitioned table, and default table partition
    1. Find all partition bounds in the default partition's data
    2. For each of these bounds, create a new partition with a date range for 100 years from the bounds (This is to let PG create all of the internal bits and bobs that a table partition will utilize)
    3. Detach this new partition (This will disable the partition checks)
    4. Use a CTE delete/returning with an insert to move the data for the specific partition range bounds from the default partition directly into the new, detached partition
    5. Attach the new partition with the correct bounds.
    6. Goto second step until no more bounds
* function `repartition_default_data` : This is a driver function that will get default partition information from **all** tenant schemata and use `PartitionDefaultData` to fix 'em

Designing the code this way was necessary for two reasons:
1. PostgreSQL will not allow a new partition to be created with a range overlapping any data from the default partition
2. The partition management triggers are locked into specific workflows at the moment.

Utilization of the code can be done as a task to crawl the entire database or to simply execute once processing has been completed, whichever is more relevant. 

This code is **not** meant to bypass the most efficient means of managing partitions which is to create them as needed during data processing. 

Currently, this branch contains a migration that will apply the functionality on a one-time basis.

# Review Request

Please have a close look at the tests to see if there is something I missed in testing. This must be bulletproof as it is **moving** customer data (delete + insert).

# Testing

This is easiest to check with the `reporting_ocpusagelineitem_daily_summary` table (`OCPUsageLineItemDailySummary` model)

## Hard

* Insert new records with a `usage_start` value 10 years in the future to get data into the default partition. 
* Then either use the class or the driver func to operate on the table.
* Then check with psql to examine that the default partition is now empty and new partitions were created.

## Easy

* Checkout master
* Insert new records with a `usage_start` value 10 years in the future to get data into the default partition. 
* Checkout `move_default_partition_data` branch
* `make run-migrations`
* Check that the default partition table is empty
* Check that there is a new partition for the record(s) you entered
* Check that **only** these record(s) are in the new and correct partition(s).


db_functions/clone_schema.sql
```diff
@@ -183,9 +183,10 @@ BEGIN
             EXECUTE 'CREATE TABLE IF NOT EXISTS ' || dest_obj || ' (LIKE ' || source_obj || ' INCLUDING ALL) ;';
         END IF;
 
-        IF copy_data OR
-           (object_rec.table_name ~ 'partitioned_tables') OR
-           (object_rec.table_name ~ 'django_migrations')
+        IF (copy_data OR
+            (object_rec.table_name ~ 'partitioned_tables') OR
+            (object_rec.table_name ~ 'django_migrations')) AND
+           (object_rec.table_kind = 'r')
```
Red-HAP, line 189: This is to fix a bug copying partitioned table data.

db_functions/partitioned_tables_manage_trigger_function.sql
```diff
@@ -26,40 +26,52 @@ DECLARE
 BEGIN
     IF ( TG_OP = 'DELETE' )
     THEN
-        alter_stmt = 'ALTER TABLE ' ||
-                     quote_ident(OLD.schema_name) || '.' || quote_ident(OLD.partition_of_table_name) ||
-                     ' DETACH PARTITION ' ||
-                     quote_ident(OLD.schema_name) || '.' || quote_ident(OLD.table_name) ||
-                     ' ;';
+        IF ( OLD.active )
```
Red-HAP, line 29: The multiple `active` checks here prevent unnecessary (and unwanted) action when updating the partitiion_parameters when the record is inactive.

db_functions/partitioned_tables_manage_trigger_function.sql
```diff
@@ -26,40 +26,52 @@ DECLARE
...
-            action_msg = 'FOR VALUES FROM ( ' ||
-                         quote_literal(NEW.partition_parameters->>'from') || '::date ) TO (' ||
-                         quote_literal(NEW.partition_parameters->>'to') || '::date )';
+            action_stmt = 'ALTER TABLE ' ||
+                            quote_ident(OLD.schema_name) || '.' || quote_ident(OLD.partition_of_table_name) ||
+                            ' ATTACH PARTITION ' ||
+                            quote_ident(OLD.schema_name) || '.' || quote_ident(OLD.table_name) || ' ';
+            IF ( (NEW.partition_parameters->>'default') = 'true' )
+            THEN
+                action_stmt = action_stmt || 'DEFAULT ;';
```
Red-HAP, line 61: Fixes malformed statement build

grafana/Dockerfile-grafana
```diff
@@ -2,5 +2,5 @@ FROM grafana/grafana:latest
 USER root
 RUN apk add sqlite
 USER grafana
-COPY --chown=grafana:grafana grafana.db.sql /var/lib/grafana/grafana.db.sql
+COPY --chown=grafana:0 grafana.db.sql /var/lib/grafana/grafana.db.sql
```
Red-HAP, line 5: Fix stupid grafana image build issue from latest image.

koku/koku/pg_partition.py
```diff
@@ -1066,3 +1070,281 @@ def drop_original_objects(self):
...
+            self.partition_name,
+            self.partitioned_table,
+            self.partition_type,
+            self.partition_key,
+            json.dumps(partition_parameters),
+        )
+        cur = conn_execute(part_track_insert_sql, vals, _conn=self.conn)
+        self.tracking_rec = fetchone(cur)
+
+    def _create_partition(self):
```
Red-HAP, line 1190: `_create_partition`, `_attach_partition`, and `_detach_partition` were written to take advantage of the trigger code if the triggers are present or do it manually if not. I don't know if the triggers and trigger funcs made it into prod.

koku/koku/pg_partition.py
```diff
@@ -1066,3 +1070,281 @@ def drop_original_objects(self):
...
+        LOG.info(f'Getting partition bounds for data in "{self.schema_name}"."{self.default_partition}"')
+        cur = conn_execute(partition_start_sql)
+        res = [(r["partition_start"], r["partition_end"]) for r in fetchall(cur)]
+
+        if len(res) == 0:
+            LOG.info("Nothing to do")
+
+        return res
+
+    def repartition_default_data(self):
```
Red-HAP, line 1256: This is the main method to operate on a partitioned table's default partition.

koku/koku/pg_partition.py
```diff
@@ -1066,3 +1070,281 @@ def drop_original_objects(self):
...
+                    conn_execute(mv_recs_sql.format(full_partition_name), (p_from, p_to), _conn=self.conn)
+
+                    # Re-attach partition with actual bounds
+                    self.partition_parameters["from"] = p_from
+                    self.partition_parameters["to"] = p_to
+                    self._attach_partition()
+
+
+# This is a crawler interface to the PartitionDefaultData class
+def repartition_default_data(schema_name=None, partitioned_table_name=None):
```
Red-HAP, line 1306: This is a crawler that will gather default partition info from across all of our tenants for processing all at once. It can also be constrained to a single schema, a single partitioned table or both.

adberglund:
@Red-HAP can you include the bugfix in the OCP processor here so we do start creating partitions 

See: https://github.com/project-koku/koku/blob/master/koku/masu/processor/ocp/ocp_report_processor.py#L143 
where we have to do a string replace to get a format that the parser accepts. 

Here is the bug: https://github.com/project-koku/koku/blob/master/koku/masu/processor/ocp/ocp_report_processor.py#L241

on that line we are not doing the replace, and the try actually always fails on the parser, but we fail silently and no partition is created.

adberglund (approved):
1. Checked out master 
2. Ran `make load-test-customer-data start=2020-11-01 end=2021-01-25` 
3. Checked and only saw default partition
4. Checked out move_default_partition_data
5. Ran run-migratons
6. Checked partitions and confirmed all data is where it should be.

PR #2558, opened 2020-12-18
COST-842 Check for term first

adberglund:
## Summary
We remove outliers prior to re-sorting this data, so it's possible for a date to not exist in the dictionary and we were hitting a KeyError. This adds an if check first.


abaiken (approved):
LGTM

PR #2560, opened 2020-12-20
Change measurement label to Cost for Node, Cluster & PVC metrics

boaz0:
closes COST 775

//cc @lcouzens


adberglund (commented):
I think what we're actually measuring here is the `Count` of each resource. How do you feel about switching this to `Count`?

boaz0:
@lcouzens what do you think? `Count` seems more accurate here than `Cost`.

lcouzens:
Yeh I think Andrew is correct, count sounds good to me! :+1:

blentz (approved):

boaz0:
Thank you @adberglund & @blentz !

PR #2575, opened 2021-01-08
COST-855, COST-848 - fix incorrect x-axis specification

blentz:
This fixes COST-855 and COST-848. 

This corrects a distortion issue by using integers for the X-axis. Also, this adds a constant to the regression to avoid forcing the regresion through the origin (0,0).

So, you can see with the attached graphs that the regression is a better fit to the data.


koku/forecast/forecast.py
```diff
@@ -178,12 +183,19 @@ def _predict(self, data):
...
 
-        # arrange the data into a form that statsmodels will accept.
         dates, costs = zip(*data)
-        X = [int(d.strftime("%Y%m%d")) for d in dates]
+
+        # format the independent variable so that all points are equidistant.
+        # in this case, every point is 86400.0 seconds apart.
+        X = []
+        for d in dates:
+            X.append(datetime(d.year, d.month, d.day, 0, 0, 0, 0).timestamp())
```
dccurtis, line 192: @esebesto made a suggestion in Jira that this could be just simple small integers.  Have you been able to give that a try?  It would be easier to work with while testing/debugging.

The comment also had some great use cases to try regarding whether or not this PR would address the missing constant.

blentz: We're discussing that stuff outside of github. :slightly_smiling_face:

blentz:
Figure_7

Here is an updated test graph.

Pipfile
```diff
@@ -49,6 +49,7 @@ kombu = ">=4.6.10,<5.0"
 importlib-metadata = "*"
 presto-python-client = ">=0.7.0"
 statsmodels = ">=0.12"
+matplotlib = ">=3.3"
```
dccurtis, line 52: Do you think this should go in the `dev-packages` section since it's being used only in the script?

blentz: Fixed!

dccurtis (approved):

PR #2580, opened 2021-01-11
COST-791: GCP RBAC

dccurtis:
Adding RBAC (account and project) for GCP accounts.

**Testing**
1. Use the resource-types GCP API and verify that the account list is shown after ingesting data.
2. Use the ENV DEVELOPMENT_IDENTITY to set "gcp.account": {"read": ["real-account-number"]}.  Verify the GCP Cost API has data
3. Update ENV to use a different gcp account. Veirfy that the cost API does not show data.
4. Update the RBAC ENV to read all gcp accounts (*) and verify cost API shows data.
5. Get a list of projects with http://localhost:8000/api/cost-management/v1/resource-types/gcp-projects/
6. Ensure all COST API is visible for RBAC "gcp.project": {"read": ["*"]}
7. Ensure COST API data is visible with "gcp.project": {"read": ["test-project"]}

[gcp_rbac_ut.txt](https://github.com/project-koku/koku/files/5798284/gcp_rbac_ut.txt)


adberglund (changes requested):
I think `project` is probably the main thing we want RBAC around for GCP. We can also offer `billing account`

dccurtis:
Yeah I thought about project as well.  Let's talk about it.

adberglund (approved):

PR #2583, opened 2021-01-13
COST-865: GCP Tags API

dccurtis:
GCP Tag API with filters for account_id, project_id, and project_name.

**Testing**
1. Ingest GCP data (with tags) and get a list of accounts and projects with the resource-type API:
2. Verify that the GCP tags API contains tag data.
3. Verify that the key_only=True parameter works.
4. Verify that filtering on an invalid account id shows no data:
5. Verify that filtering on an invalid project shows no data
6. Verify that filtering on a valid account shows data.
7. Verify that filtering on a valid project shows data
8. Verify that filtering on a valid project name also shows the same data as in step 7.

**Test Results**
[gcp_tag_api_ut.txt](https://github.com/project-koku/koku/files/5815756/gcp_tag_api_ut.txt)


adberglund (approved):
🌞

PR #2586, opened 2021-01-14
provider model char to text

Red-HAP:
## Jira Ticket

[COST-836](https://issues.redhat.com/browse/COST-836)

## Description

The initial change was to fix the identified issue from Sentry. Buuuut, it affects the Provider model/table as well.

This change will alter the provider table and model columns name and type to text type from varchar(256) and varchar(50) type

## Testing

Try adding a source or provider with a name length > 256 chars


koku/api/provider/serializers.py
```diff
@@ -202,7 +202,7 @@ class ProviderSerializer(serializers.ModelSerializer):
     """Serializer for the Provider model."""
 
     uuid = serializers.UUIDField(allow_null=True, required=False)
-    name = serializers.CharField(max_length=256, required=True, allow_null=False, allow_blank=False)
+    name = serializers.CharField(required=True, allow_null=False, allow_blank=False)
```
boaz0, line 205: `max_length` is required to avoid DDoS attack (according to a report we got from the security team in Red Hat).

koku/api/provider/models.py
```diff
@@ -109,8 +109,8 @@ class Meta:
     CLOUD_PROVIDER_LIST = [choice[0] for choice in CLOUD_PROVIDER_CHOICES]
 
     uuid = models.UUIDField(default=uuid4, primary_key=True)
-    name = models.CharField(max_length=256, null=False)
-    type = models.CharField(max_length=50, null=False, choices=PROVIDER_CHOICES, default=PROVIDER_AWS)
+    name = models.TextField(null=False)
```
boaz0, line 112: `max_length` is required to avoid DDoS attack (according to a report we got from the security team in Red Hat).

koku/api/provider/models.py
```diff
@@ -109,8 +109,8 @@ class Meta:
...
+    type = models.TextField(null=False, choices=PROVIDER_CHOICES, default=PROVIDER_AWS)
```
boaz0, line 113: `max_length` is required to avoid DDoS attack (according to a report we got from the security team in Red Hat).

boaz0 (changes requested):
Hey, I am just letting you know that the `max_length` was meant to avoid a case when an attacker send a huge amount of data and causes DDoS attack.

Nonetheless, if this PR doesn't compromise this, then ignore my review.

Thanks.

boaz0:
So it seems the issue is that people can create a source with a large name. We can 1) change the value 2) Add the source name max-length restriction in the sources-UI/backend

PR #2587, opened 2021-01-14
COST-747 & COST-750: Create mat views & instance type api for GCP.

myersCody:
What I have tackled in this pr:
1) Initially the cost api was completed before the nise work. Therefore, we used a tmp csv file to populate the data for the api unit tests. However, now that the nise work is finished we have replaced that temporary csv file work with the built in nise feature. 
2) I have created the materialize views for the gcp instance type api
3) I have created the gcp instance type api. 

## How to test

Run `make load-test-customer-data` then hit the gcp instance-type endpoints to see if they come up.


koku/reporting/migrations/0160_gcp_compute_summary.py
```diff
@@ -0,0 +1,24 @@
+# Generated by Django 3.1.5 on 2021-01-14 21:13
```
myersCody, line 1: I want to make sure this is the right way to do the migration I think something weird happened last time so I wanted to leave a note to double check.

koku/api/report/gcp/provider_map.py
```diff
@@ -61,8 +66,9 @@ def __init__(self, provider, report_type):
                         {"field": "project_name", "operation": "icontains", "composition_key": "project_filter"},
                         {"field": "project_id", "operation": "icontains", "composition_key": "project_filter"},
                     ],
+                    "instance_type": {"field": "instance_type", "operation": "icontains"},
```
adberglund, line 69: Other source tpyes don't have a instance_type filter, so I think maybe we don't need this?

koku/api/report/gcp/provider_map.py
```diff
@@ -61,8 +66,9 @@ def __init__(self, provider, report_type):
...
                 },
-                "group_by_options": ["account", "region", "service", "project"],
+                "group_by_options": ["account", "region", "service", "project", "instance_type"],
```
adberglund, line 71: Similarly, are we okay without this as a group by option? It isn't technically something we support grouping on, the instance type API just inherently groups by it.

koku/api/report/gcp/provider_map.py
```diff
@@ -121,6 +127,67 @@ def __init__(self, provider, report_type):
...
+                            "cost_markup": Sum(Coalesce(F("markup_cost"), Value(0, output_field=DecimalField()))),
+                            "cost_units": Coalesce(Max("currency"), Value("USD")),
+                            "usage": Sum("usage_amount"),
+                            "usage_units": Coalesce(Max("unit"), Value("hour")),
+                            "source_uuid": ArrayAgg(F("source_uuid"), distinct=True),
+                        },
+                        "delta_key": {"usage": Sum("usage_amount")},
+                        "filter": [
+                            {"field": "instance_type", "operation": "isnull", "parameter": False},
+                            # I assume we only want the core hours
```
adberglund, line 180: ✅

koku/api/report/gcp/serializers.py
```diff
@@ -34,12 +34,13 @@
 class GCPGroupBySerializer(GroupSerializer):
     """Serializer for handling query parameter group_by."""
 
-    _opfields = ("account", "region", "service", "project")
+    _opfields = ("account", "region", "service", "project", "instance_type")
```
adberglund, line 37: Shouldn't need to be a group by option for GCP, it's just implicit in the instance-type API

myersCody: @adberglund So, I think this is the only place I need to keep the instance_type group at. When I remove it I get:
```
{
    "group_by": {
        "instance_type": [
        ]
    }
```

adberglund (approved):

PR #2592, opened 2021-01-18
COST-854: pass exog vars to wls_prediction_std

blentz:
This fixes the bug described in the fourth list item in COST-854. This ensures that both the forecast and the confidence intervals are using the same exogenous variables, including the added constant.


dlabrecq (approved):
LGTM

PR #2593, opened 2021-01-18
COST-848: fix use of DateHandler

blentz:
This fixes the use of DateHelper in a couple of classes. Instead of using a static class variable (initialized once at application startup), I've moved these to instance variables (initialized each time a new instance is created). 

This will ensure that the dates being referenced are relative to the current API request, rather than the application's start date. 

Additionally, I've moved setting the instance variable into the base `QueryHandler` class to DRY up the code and eliminate the need for each sub-class to implement its own version.


dlabrecq (approved):
LGTM

PR #2597, opened 2021-01-19
COST-891 aws report processor memory bugfix

adberglund:
## Summary
This fixes some string parsing around memory for the AWS report processor. 

## Testing
- Successfully processed report file with previously failing values. 
- Successfully processed previously passing report.


blentz (approved):
This is fine as-is, or you could migrate this to using `pint` and simplify the code slightly.

PR #2601, opened 2021-01-20
COST-741: User Access API 

dccurtis:
@dlabrecq is working on enhancements to our empty state pages which will enable hiding Insights navigation links.  This API will also facilitate hiding in-progress source-types (GCP) in production.

Platform has told us that a simple `data:true` response is sufficient for being called from cloud-services-config.  I made the response with no query parameter to be something useful to view but won't be used by cloud-services-config.

**Testing**
1. With a development identity with no rbac set for aws, ocp, azure, gcp or cost model and org-admin=false.  Verify all access is false
2. With a development identity with rbac set for aws, azure, gcp and cost model and org-admin=false.  Verify all access is true except for ocp.
3. With a development header for an org-admin with no explicit rbac access.  Verify all access is true
4. Query with an invalid type and verify API responds with a 400

**Test Results**
[cost_741_ut.txt](https://github.com/project-koku/koku/files/5844830/cost_741_ut.txt)


dlabrecq (approved):
LGTM

dlabrecq (approved):

PR #2603, opened 2021-01-20
COST-797: Add processing for start and end dates in manifest 

abaiken:
### Description

* This PR allows start and end dates that are specified in the manifest to be passed down the processing pipeline so that we can process reports from air gapped clusters where the data might fall outside of our previous 3 day processing window
---
### Manual Testing: 
1. Get Ingress up and working in conjunction with the Koku services so that you can upload files to the listener - feel free to reach out to me for info on how I did this. 
2. Create an OCP source with the same cluster ID as what is in the payloads we are using for testing: 
```
curl -d '{"name": "aaiken", "source_type": "OCP", "authentication": {"credentials": {"cluster_id": "5997a261-f23e-45d1-8e01-ee3c765f3aec"}}}' -H "Content-Type: application/json" -X POST http://0.0.0.0:8000/api/cost-management/v1/sources/
```
3. Download the following file and then upload it using the `make local-upload-data file=costreport1.tar.gz` command
[costreport1.tar.gz](https://github.com/project-koku/koku/files/5845351/costreport1.tar.gz)
4. Connect to the DB and look at the reporting_ocpusagelineitem_daily_summary table. I used the copy command to write the table contents to a CSV file so that I could compare them for before/after: 
```
postgres=# \copy (SELECT * FROM acct10001.reporting_ocpusagelineitem_daily_summary) TO '/Users/ashleybrookeaiken/Development/koku/ocpreporting.csv' DELIMITER ',' CSV HEADER;
```
5. Download the following file (start and end dates are 1/10 - 1/12) and upload it using the `make local-upload-data file=costreport1.tar.gz` command. 
[costreport2.tar.gz](https://github.com/project-koku/koku/files/5845353/costreport2.tar.gz)

6. You should now repeat step 4 but copy the contents to a different file so that you can compare the data. You should see that the `pod_usage_cpu_core_hours` doubles from the first report to the second for 1/10 and 1/11.


Makefile
```diff
@@ -539,6 +540,12 @@ backup-local-db-dir:
 	@cd - >/dev/null
 	$(DOCKER_COMPOSE) start db
 
+local-upload-data:
```
dccurtis, line 543: nice!

dccurtis (approved):
Looks great!

PR #2605, opened 2021-01-20
include a repo secret in action cache key

maskarb:
This PR will allow us to reset the github-action cache key without needing a PR. In the repo settings, I defined the secret: `ACTIONS_CACHE_KEY_UUID`. If our cache ever gives us issues again, we can just give this secret a new uuid value and restart the failing actions.


dccurtis (approved):
Great idea!

PR #2608, opened 2021-01-21
Fixing user access to expect all rbac types to be present

dccurtis:
I was working on a bad assumption that rbac-types would only be present if they were configured.


dlabrecq (approved):
LGTM

PR #2609, opened 2021-01-22
COST-716: Finalizing GCP sources and supporting app-specific settings for AWS and Azure

dccurtis:
GCP integration changes are coded against: https://github.com/RedHatInsights/sources-api/pull/299 which should be merging today. 

*UPDATE:*
* Change merged and it works great against this PR

The sources team agreed to use the new `Applications.extra` value to hold the dataset (was previously using PATCH).  This change also brings along support for AWS and Azure to also use the new `extra` field while at the same time maintaining backwards compatibility with PATCH for AWS and Azure.  Data migrations in all environments must happen before we decommission the PATCH path for AWS and Azure.

**Testing**
1. Regression test AWS source creation with PATCH
2. Regression test Azure source creation with PATCH
3. Add AWS source with Application-extra flow
4. Add Azure source with Application-extra flow
5. Update AWS bucket for source_id =3 (Application.extra path) to an incorrect bucket and verify status is updated accordingly
6. Add GCP source (using WIP sources branch) with an invalid project name (verify status is correct).  Fix project name and verify source is created successfully.

**Test Results**
[cost_716_ut.txt](https://github.com/project-koku/koku/files/5856621/cost_716_ut.txt)


dccurtis:
E2E Deploy update for GCP:
https://github.com/RedHatInsights/e2e-deploy/pull/2678

maskarb (approved):
LGTM!

PR #2610, opened 2021-01-22
Rc 2021.01.22

dccurtis:
Bug fix release


lcouzens (approved):
LGTM, after fun with smoke testing we are down to a single Azure failure which passes for me locally and against CI

PR #2611, opened 2021-01-25
Change the start date to the beginning of the month for gcp local download

myersCody:
We need the start date to be the beginning of the month for Luke's automation.


cgoodfred (approved):
LGTM

PR #2616, opened 2021-01-26
Sources ARN move from password to username filed

dccurtis:
Platform sources is moving the AWS ARN value from the Authentications password field to the username filed so that it is visible in the UI during source editing:

Platform Issues:
https://issues.redhat.com/browse/RHCLOUD-11834
https://issues.redhat.com/browse/RHCLOUD-11835

This change will attempt to get the ARN from the username field and then fall back to the password field until the platform change is running in production.  At that point in time we can remove the password ARN logic.

**Testing**
1. Add an AWS source with ARN in the Authentication.password field.
2. Add an AWS source with ARN in the Authentication.username field.

**Test Results**
[sources_arn_move_ut.txt](https://github.com/project-koku/koku/files/5873850/sources_arn_move_ut.txt)

### DIFF
--- a/koku/sources/sources_http_client.py
+++ b/koku/sources/sources_http_client.py
@@ -235,6 +235,14 @@ class SourcesHTTPClient:
         if not authentications_response.get("data"):
             raise SourcesHTTPClientError(f"Unable to get AWS roleARN for Source: {self._source_id}")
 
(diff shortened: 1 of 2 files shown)


maskarb (approved):
lgtm!

PR #2619, opened 2021-01-27
Log warning on invalid CSV report files before being processed

boaz0:
- [x] ~Remove~ Warn invalid CSV reports
- [x] Add unit test


koku/masu/external/kafka_msg_handler.py
```diff
@@ -287,6 +302,7 @@ def extract_payload(url, request_id, context={}):  # noqa: C901
 
     """
     temp_dir, temp_file_path, temp_file = download_payload(request_id, url, context)
+    clean_invalid_csv(temp_dir)
```
dccurtis, line 305: I think your going to have to call this after `extract_payload_contents` since `download_payload` will result in the tarball being in the `temp_dir`

koku/masu/external/kafka_msg_handler.py
```diff
@@ -237,6 +238,22 @@ def construct_parquet_reports(request_id, context, report_meta, payload_destinat
     return daily_parquet_files
 
 
+def remove_invalid_csv(dir_path):
+    for (root, _, filenames) in os.walk(dir_path):
+        csv_reports = [fl for fl in filenames if fl.endswith(".csv")]
+        for csv_report in csv_reports:
+            file_path = f"{root}/{csv_report}"
+            try:
+                pd.read_csv(file_path)
```
dccurtis, line 247: One other option is to call this here instead:
https://github.com/project-koku/koku/blob/262bd28754c6ea57f07c8229924dcc7c01db4d56/koku/masu/util/ocp/common.py#L285

That should not only verify that it's a CSV file but we can also verify that the file is a valid OCP report

koku/masu/external/kafka_msg_handler.py
```diff
@@ -237,6 +238,22 @@ def construct_parquet_reports(request_id, context, report_meta, payload_destinat
...
+            except Exception as error:
+                LOG.warning(f"File {file_path} is not a valid CSV file. Reason: {str(error)}")
+                LOG.debug(f"Removing file {file_path} because it is invalid CSV")
+                try:
+                    os.remove(file_path)
```
dccurtis, line 252: Instead of removing the file can we just log enough information so that we know what account the payload came from.  Then `extract_payload` would return `None` so that no further processing is done.

We might have to examine what's in the Kafka message to figure out if we can determine the account the payload came from.

boaz0:
CI failed because of unrelated unit tests failures.

koku/masu/external/kafka_msg_handler.py
```diff
@@ -237,6 +237,19 @@ def construct_parquet_reports(request_id, context, report_meta, payload_destinat
     return daily_parquet_files
 
 
+def check_invalid_csv(dir_path, context):
```
maskarb, line 240: What is the purpose of this function? From what I can tell, all it does is log when we have an invalid report type. Does it need to remove the invalid files?

koku/masu/external/kafka_msg_handler.py
maskarb: I thought we needed to remove the invalid files so that processing of valid files could continue?

koku/masu/external/kafka_msg_handler.py
boaz0: I forgot to renamed the function. It's just going to log a warning message if the csv file is not a valid OCP report.

boaz0:
@dccurtis any other important data you want to add to the warning message? Should we also include `cluster-id` too?

boaz0:
@dccurtis - So this is what we were talking about.
I made sure the returned value in such case is "None" and that we are logging the files and account name.
Although I thought it would be better to log each bad report in a different log line and in the end logging that invalid reports were found with the context data to trace the account, cluster id, etc later.

What do you think?

koku/masu/external/kafka_msg_handler.py
```diff
@@ -313,6 +328,15 @@ def extract_payload(url, request_id, context={}):  # noqa: C901
     report_meta["account"] = schema_name[4:]
     report_meta["request_id"] = request_id
 
+    invalid_csvs = find_invalid_csvs(temp_dir)
```
dccurtis, line 331: I think we might need to catch the Exception here for the case that the report type is invalid.  That way we can remove the temp_dir and not continue to process these invalid files.

koku/masu/external/kafka_msg_handler.py
dccurtis: @maskarb what do you think about processing partially correct payloads?  I'm sort of on the fence.  On hand hand, if we have some reports (pod, storage, etc) that are good it would be good to just process it.  On the other if one is bad, failing the entire payload would bring it to the user and therefore our attention must faster.

PR #2629, opened 2021-02-02
reapply application-side text field size check

Red-HAP:
## Jira Ticket

[COST-836](https://issues.redhat.com/browse/COST-836)

## Description

Checking into expanding the data type change to the provider table was going to require that the change be propagated to many other `reporting_*` tables as well as requiring manual migrations against at least one partitioned table. Also some of the materialized views would have to be dropped and re-created.

After speaking with Doug, we decided to re-implement the 256-char limit on the `api_sources.name` column. This will be done in the api by forcing field validations to run against `Sources` model instances on a call to `save()`. This change will run the default and specified validators only without calling the `clean()` functionality.

This was done as a mixin that will override `Model.save()` and can be applied to any Cost Management Django model.

## Testing

Creating a source with a name > 256 will throw a django ValidationError exception.

### DIFF
--- a/koku/api/model_utils.py
+++ b/koku/api/model_utils.py
@@ -1,0 +1,32 @@
+#
+# Copyright 2021 Red Hat, Inc.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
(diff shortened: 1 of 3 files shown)


blentz (approved):
LGTM

PR #2630, opened 2021-02-03
COST-908: bugfix for summary table lookup

blentz:
This fixes the bug described in COST-908. A conditional branch was referencing a part of the ProviderMap that doesn't exist.

### DIFF
--- a/koku/forecast/test/tests_forecast.py
+++ b/koku/forecast/test/tests_forecast.py
@@ -46,6 +46,7 @@ from forecast import OCPAWSForecast
 from forecast import OCPAzureForecast
 from forecast import OCPForecast
 from forecast.forecast import LinearForecastResult
+from reporting.provider.aws.models import AWSCostEntryLineItemDailySummary
 from reporting.provider.gcp.models import GCPCostSummary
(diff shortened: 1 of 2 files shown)


Red-HAP (approved):

koku/forecast/test/tests_forecast.py
```diff
@@ -428,6 +429,13 @@ def test_enumerate_dates(self):
                 out = instance._enumerate_dates(scenario["dates"])
                 self.assertEqual(out, scenario["expected"])
 
+    def test_summary_table(self):
+        """COST-908: Test that the expected summary table is used."""
+        mock_access = {"aws.organizational_unit": {"read": ["1234", "5678"]}}
+        params = self.mocked_query_params("?", AWSCostForecastView, access=mock_access)
+        instance = AWSForecast(params)
+        self.assertEqual(instance.cost_summary_table, AWSCostEntryLineItemDailySummary)
```
adberglund, line 437: @blentz In the case that we have RBAC restriction only on org unit, we should actually be getting the `reporting_aws_cost_summary_by_account` materialized view, so I think we actually want to update the views attribute in the AWS provider map.

PR #2633, opened 2021-02-03
process big deletes as a loop of smaller transactions

Red-HAP:
## Jira Ticket

[COST-821](https://issues.redhat.com/browse/COST-821)

## Description

This is to allow potentially two long-running processes to interact with the same table. The suspicion is that two processes are acquiring locks for a large set of data within the table. This change will break up potentially long-running delete transactions into loops of smaller transactions. The statements are written such that row locks will be acquired and locked rows will be skipped.

The original query will be kept and used as a count check to make sure all records were deleted. If not, the processing will be tried for a fixed number of iterations. If all iterations fail to remove all of the records, an error log message will be written.

## Testing

This is the difficult one as it depends on a race condition. But the base-level test is that data processing and Expired data should appear to work as normal. Otherwise, two processes that expire data can be run simultaneously (against the same table) and neither should deadlock the other. Similarly with two file processing processes that happen to have overlapping date rnages should process successfully.

## Risk Exposure

If I am incorrect and this is related to deletes and inserts/updates happening at the same time, the deletes may win due to the retry strategy.


koku/masu/database/koku_database_access.py
```diff
@@ -113,3 +115,54 @@ def delete(self, obj=None):
...
+def mini_transaction_delete(base_select_query):
+    """
+    Uses a select query to make a mini transactinal delete loop.
+    Schema should be set before calling this function.
+    Args:
+        base_select_query (QuerySet) : Single-table django select query
+    Returns:
+        tuple : (deleted_record_total, records_remaining)
+    """
+    del_record_limit = int(os.getenv("DELETE_CYCLE_RECORD_LIMIT", 5000))
```
adberglund, line 133: To be extra Django-y, we could move these to settings.py and `from django.conf import settings` and reference them from settings like `settings. DELETE_CYCLE_RECORD_LIMIT` but I suspect we'll be changing this again soon to only remove expired partitions, so optional.

adberglund (approved):

PR #2634, opened 2021-02-04
Update settings API to support new dual list

boaz0:
Don't merge before https://github.com/RedHatInsights/settings-frontend/pull/357
This is to catch up with the changes of the schema the new dual list supports

This doesn't have the expanded dual list work yet because DDF doesn't support it for now.

//cc @nlcwong


dccurtis:
@boaz0 Can you check with me before merging this one.  Just want to make sure the timing of the merge works with our current effort to get the staging environment healthy.

boaz0:
@dccurtis no problem. I will touch base before merging just to make sure we are sync between settings UI envs (prod/ci/etc)

PR #2638, opened 2021-02-08
reduce the potential of constraint violation

Red-HAP:
## Jira Ticket

[COST-820](https://issues.redhat.com/browse/COST-820)

## Description

Change `x.objects.create` to `x.objects.get_or_create` to reduce the window that could introduce a constraint violation.

## Testing

Attempt to introduce a race condition via masu processing.


blentz (approved):

adberglund (approved):
🚀 🚀 🚀

PR #2644, opened 2021-02-09
COST-721 Use redis for celery backend

adberglund:
## Summary
* Switch our celery broker backend to use redis. 
* This is step one, further cleanup of templates will follow. 

## Testing
Loaded test customer data locally with rabbit container removed to make sure we weren't somehow still using it. Everything ran successfully. APIs returned expected data.


maskarb (approved):
LGTM!

Could consolidate `CELERY_BROKER_URL` and `CELERY_RESULTS_URL`, but that can certainly be done in a follow up to remove all the references to RabbitMQ.

PR #2646, opened 2021-02-09
Sources GCP BigQuery Table ID should be maintained while updating app-settings

dccurtis:
A miss from yesterday.  We need to ensure that the BigQuery Table ID stays in tact when we update GCP application settings.  Table ID is not tracked by platform sources since it is detected by Cost Management and not something the user enters.

**Testing**
GCP crud testing in QA environment was successful


lcouzens (approved):
Sweet!

adberglund (approved):

PR #2648, opened 2021-02-09
COST-1002: GCP Numeric overflow error during csv processing

myersCody:
I want to check in with someone with more db knowledge than myself to see if we should be doing the decimal bump for every usage_amount field in the gcp models or will just bumping the line item table fix the problem for us.


Red-HAP (approved):

adberglund (commented):
I think this is likely too big a change. By the time we get to the daily summary table we are using the much smaller `usage_in_pricing_units` for usage_amount, so I think the problem is really in the line item and line item daily table which wouldn't require updated daily summary or any of the views. https://github.com/project-koku/koku/blob/master/koku/masu/database/sql/reporting_gcpcostentrylineitem_daily_summary.sql#L43

adberglund (approved):

dccurtis (approved):

PR #2652, opened 2021-02-11
[COST-997] Bugfix: Remove sources API caching

adberglund:
## Summary
Our sources API caching is causing some problems, and doesn't account for the changes made to sources via the non-API pathed sources changes. Removing caching implementation as the benefit was marginal. We make significantly fewer calls to this API in the UI than we used to.


lcouzens (approved):
LGTM

PR #2671, opened 2021-02-17
scripting updates - improve resilience, DRY up a few things

blentz:
This PR adds some updates to our dev env scripting.

Change Summary:
- Update `Makefile` use of `docker-compose` to open the door for using `podman-compose`. This isn't 100% working, in that there's some additional details to work through with podman. But, adding almost-working support for podman doesn't adversely impact docker-compose users.

- `create_test_customer.py` got a few updates.
  - add `wallclock()` function to help expose ingestion timing issues.
  - define lists of source type constants that work with the API vs. the DB. DRY up those strings.
  - moved all sleeps to use a single variable.
  - fixed the `--bypass-api` flow so that it creates sources in the DB.  
  
- `load_test_customer.sh` also got several updates.
  - added a bunch of comments to document the script better
  - added `DEBUG` variable and function to print extra info only when `DEBUG=1`
  - fixed pathing logic so that we no longer need to pass a directory into the script; this script can now be executed from anywhere on your filesystem.
  - swapped out some `if/then` blocks for setting defaults at variable assignment time.
  - DRYed up the code for creating cost models
  - DRYed up the code for rendering nise static files

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,9 +37,9 @@ OC_TEMPLATES = $(wildcard $(OC_TEMPLATE_DIR))
 
 # Docker compose specific file
 ifdef compose_file
-    DOCKER_COMPOSE = docker-compose -f $(compose_file)
+    DOCKER_COMPOSE = $(DOCKER)-compose -f $(compose_file)
 else
-	DOCKER_COMPOSE = docker-compose
+	DOCKER_COMPOSE = $(DOCKER)-compose
 endif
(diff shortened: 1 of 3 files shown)


Red-HAP:
@blentz 
Please see suggested changes on https://github.com/project-koku/koku/pull/2672

Red-HAP (approved):

PR #2680, opened 2021-02-19
COST-1067: Extending start_date query parameter to cover materialized view window

dccurtis:
The `start_date` query parameter should be functional for the same number of billing months that we are storing in summary-materialized views.

This change will use the environment variable `RETAIN_NUM_MONTHS` to keep the summary data and API query validation in sync.

**Testing**
1. Set RETAIN_NUM_MONTHS=3 and verify 3 billing periods are accessible in a report API.
2. Verify that a date prior to the RETAIN_NUM_MONTHS=3 window throws a 400 error.
3. Change RETAIN_NUM_MONTHS=2 and repeat steps 1 and 2.

**Test Results**
[cost_1067_ut.txt](https://github.com/project-koku/koku/files/6010720/cost_1067_ut.txt)


dlabrecq (approved):

koku/api/utils.py
```diff
@@ -124,6 +127,12 @@ def last_month_start(self):
         last_month = self.this_month_start - self.one_day
         return last_month.replace(day=1)
 
+    @property
+    def materialized_view_month_start(self):
+        """Datetime of midnight on the first of the month where materialized summary starts."""
+        summary_month = self.this_month_start - relativedelta(months=Config.MASU_RETAIN_NUM_MONTHS - 1)
+        return summary_month
+
```
blentz, line 135: I'm not sure this is the right place for this. DateHelper is supposed to only care about helping us get our date math correct. It doesn't really need to know about the app in any way. I think this is a case of yes we _can_ do it this way, but I'm not sure we _should_ do it this way. :slightly_smiling_face:  

I also think there's a simpler solution.

In the validator, we make this change:
```
-         if value >= dh.last_month_start.date() and value <= dh.today.date():
+        MAX_QUERY_DAYS=60
+        if value >= dh.n_days_ago(dh.today, MAX_QUERY_DAYS).date() and value <= dh.today.date():
```

I think I'd prefer to not use MASU_RETAIN_NUM_MONTHS. I think we're coalescing around the day being the basic unit we calculate around. For that reason, I think the MASU_RETAIN_NUM_MONTHS feature probably needs a refactor and maybe a small redesign to update it to our current direction. It might be good to open a new issue to maybe do a spike to investigate if that's something we'd want to do.

blentz (changes requested):

blentz (approved):

PR #2682, opened 2021-02-19
Localdev db ssl

Red-HAP:
## Description

Add a script to easily generate a self-signed cert for local development use of PostgreSQL with SSL support turned on.
Added options to `docker-compose.yml` to use the certificate and enable SSL support.

This is opt-in.


blentz (approved):
LGTM.

PR #2684, opened 2021-02-19
COST-1071: Re-enable GCP status when BigQuery table is discovered

dccurtis:
**Testing**
Verified functionality in QA environment.  Will add API tests next


lcouzens (approved):
LGTM!

PR #2686, opened 2021-02-22
update requirements; library security updates

blentz:
This updates the Pipfile requirements to pull in django 3.1.7, among other things.

### DIFF
--- a/docs/rtd_requirements.txt
+++ b/docs/rtd_requirements.txt
@@ -11,8 +11,8 @@ azure-mgmt-storage==11.2.0
 azure-storage-blob==12.7.1
 beautifulsoup4==4.9.3
 billiard==3.6.3.0
-boto3==1.17.8
-botocore==1.20.8; python_version >= '2.7' and python_version not in '3.0, 3.1, 3.2, 3.3, 3.4, 3.5'
+boto3==1.17.12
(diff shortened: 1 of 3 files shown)


docs/rtd_requirements.txt
```diff
@@ -1,4 +1,5 @@
--i https://pypi.python.org/simple
+-i http://localhost:3141/root/pypi/+simple/
+--trusted-host localhost:3141
```
maskarb, line 2: What is this?

maskarb (commented):
Everything looks good, but a question about the `rtd_requirements.txt`:

docs/rtd_requirements.txt
blentz: t's because I have a local devpi instance deployed.

AFAIK, it's not a fatal issue. It just says where those things were pulled from.

maskarb: Let's change this back. When I run it locally, i get:
```
╰─$ pip install -r docs/rtd_requirements.txt 
adding trusted host: 'localhost:3141' (from line 2 of docs/rtd_requirements.txt)
Looking in indexes: http://localhost:3141/root/pypi/+simple/
```
which fails to find anything.

blentz: ok. i'll fix that. :slightly_smiling_face:

blentz: fixed!

maskarb (approved):
lgtm!

PR #2695, opened 2021-03-01
COST-1045: Sources dataset update for downloader

dccurtis:
Follow up to COST-1045.  The worker needs to adjust the dataset format.  I didn't want to change what was stored in the Sources model because it's better to stay in sync with what the platform stores.

**Testing**
Added GCP source with both formats and verified ingest was successful.


lcouzens (approved):
LGTM!

PR #2696, opened 2021-03-01
COST-1095: Return false for unauthorized user-access

dccurtis:
This change will prevent the PrermissionDenied exception from being raised from middleware during the RBAC check only when the user-access API is requested.  The user-access API will return a `false` indicating that the user does not have access.  This change is needed because the platform UI can not catch exceptions and is expecting a boolean response.  

FYI: @dlabrecq 

**Testing**
Tested in the QA environment with a non-org admin user that has no cost management RBAC access

User does not have access to Cost Management
image

Correct empty state page is shown:
image


adberglund (approved):

dlabrecq (commented):
LGTM

PR #2670, opened 2021-02-17
Add IBM Cloud Provider

boaz0:
This is essential for starting with the IBM Cloud Downloader.

# How to test it
What I personally did to make sure it works is adding IBM source(s) to `./scripts/test_customer.yaml`.
This file contains all the sources that are created using `make create-test-customer`.

Add under `sources`:
```
---
customer:
  account_id: 10001
  email: test@example.com
  user: test_customer
  sources:
    - ibm_source:
      source_name: Test IBM Source
      source_type: 'IBM'
      authentication:
        credentials:
          iam_token: 'yadayadayoow'
      billing_source:
        data_source:
          enterprise_id: 'blablabla'
    ......
```

You should see in `koku_server` that this is failing due to bad API token.
Ping me so I can give you a valid API Token and Enterprise ID to verify that it works when the data is right.

PR #2703, opened 2021-03-04
Add IBM Cloud report downloader

boaz0:
~DO NOT MERGE - waiting on https://github.com/project-koku/koku/pull/2670~


koku/masu/util/ibm/download.py
```diff
@@ -0,0 +1,94 @@
...
+from ibm_cloud_sdk_core.authenticators import IAMAuthenticator
+from ibm_platform_services.common import get_sdk_headers
+
+from api.provider.models import Provider
+from masu.config import Config
+from masu.util.aws.common import copy_local_report_file_to_s3_bucket
+from masu.util.common import get_path_prefix
+
+LOG = logging.getLogger(__name__)
+
```
myersCody, line 32: I am not quite sure if this needs to be in its own utl since these methods would probably only be used in the `ibm_report_downloader.py` file. It may it easier to just have them methods in the `IBMReportDownloader` class, but that is more of a preference comment really.

koku/masu/util/ibm/download.py
```diff
@@ -0,0 +1,94 @@
...
+
+class NoMoreReportPages(Exception):
+    pass
+
+
+def page_downloader_factory(credentials, data_source, month):
+    service_version = "v1"
+    service_name = "resource-instance-usage-reports"
+    authenticator = IAMAuthenticator(credentials.get("iam_token", ""))
```
myersCody, line 41: We probably want to try to wrap some of this in a try except statement to catch potential errors.

koku/masu/util/ibm/download.py
boaz0: You know. In a second thought we can do it. I just don't want to make the class bloated and just do one thing and that's preparing the csv report files for the processor.

That's why I created `util/ibm/download` so it is easy to separate between the responsibility of the class and the utilities it uses. I guess I can move these to the ibm_report_downloader.py file but not inside the class but as functions that are used by that class. Will that work?

If you still think it should  be inside the class, let me know. Would like to hear more.

koku/masu/util/ibm/download.py
boaz0: Yes, that's a good idea. At least log an error message so it will be easy to debug!

koku/masu/external/downloader/ibm/ibm_report_downloader.py
```diff
@@ -0,0 +1,242 @@
...
+class NoMoreReportPages(Exception):
+    pass
+
+
+def page_downloader_factory(credentials, data_source, month):
+    service_version = "v1"
+    service_name = "resource-instance-usage-reports"
+    try:
+        authenticator = IAMAuthenticator(credentials.get("iam_token", ""))
+        service = BaseService(service_url="https://enterprise.cloud.ibm.com", authenticator=authenticator)
```
dccurtis, line 55: We don't have to do this now, but we might want to consider making this be an environment variable.

koku/masu/external/downloader/ibm/ibm_report_downloader.py
```diff
@@ -0,0 +1,242 @@
...
+        except NoMoreReportPages:
+            return
+        writer(response.get_result().text)
+        page += 1
+
+
+def create_daily_archives(
+    request_id, account, provider_uuid, local_filename, full_file_path, manifest_id, start_date, context
+):
+    if settings.ENABLE_S3_ARCHIVING or settings.ENABLE_PARQUET_PROCESSING:
```
dccurtis, line 121: Can you replace `settings.ENABLE_PARQUET_PROCESSING` with `enable_trino_processing()`. Here's an example: https://github.com/project-koku/koku/blob/9d893857401c996990e8fedfcc4b6cb6156e1301/koku/masu/external/downloader/gcp/gcp_report_downloader.py#L95

dccurtis (approved):

boaz0:
Thanks @dccurtis :bow: & @myersCody :rose:

PR #2708, opened 2021-03-05
COST-1075: implement improvements to ranking logic.

blentz:
This PR fixes a bug that is present for queries that use `resolution=daily`, as described in COST-1075.

Previously, the behavior of `filter[limit]=LIMIT&group_by[GROUP]=*` did not constrain the number of groups in the whole response. There would be `LIMIT` number of elements in each per-day list, but the total number of groups in the response could exceed the `LIMIT`.

With this change, the response now ensures that only `LIMIT` number of groups are present in the whole response being returned.

### Azure

Example query: `LIMIT=5; curl -s -g "http://localhost:8000/api/cost-management/v1/reports/azure/costs/?filter[limit]=$LIMIT&filter[offset]=0&group_by[subscription_guid]=*&order_by[cost]=desc&end_date=2021-02-28&start_date=2021-02-01&filter[resolution]=daily" | jq '.'`

Example output: 
[azure_query.txt](https://github.com/project-koku/koku/files/6091570/azure_query.txt)

### AWS
Example query: `LIMIT=5; curl -s -g "http://localhost:8000/api/cost-management/v1/reports/azure/costs/?filter[limit]=$LIMIT&filter[offset]=0&group_by[region]=*&order_by[cost]=desc&end_date=2021-02-28&start_date=2021-02-01&filter[resolution]=daily" | jq '.'`

Example output: 
[aws_query.txt](https://github.com/project-koku/koku/files/6091573/aws_query.txt)

### OCP + AWS
Example query: `LIMIT=5; curl -s -g "http://localhost:8000/api/cost-management/v1/reports/openshift/infrastructures/aws/costs/?filter[limit]=$LIMIT&filter[offset]=0&group_by[project]=*&order_by[cost]=desc&end_date=2021-02-28&start_date=2021-02-01&filter[resolution]=daily" | jq '.'`

Example output: 
[ocp_aws_query.txt](https://github.com/project-koku/koku/files/6091581/ocp_aws_query.txt)

### OCP
Test Data: 9 projects

Functional testing:
```
$ for LIMIT in $(seq 1 10); do curl -s -g "http://localhost:8000/api/cost-management/v1/reports/openshift/costs/?filter[limit]=$LIMIT&filter[offset]=0&group_by[project]=*&order_by[cost]=desc&end_date=2021-02-28&start_date=2021-02-01&filter[resolution]=daily" | jq '[.data[].projects[].project] | unique | length'; done
1
2
3
4
5
6
7
8
9
9
```

Example query: `$ LIMIT=5; curl -s -g "http://localhost:8000/api/cost-management/v1/reports/openshift/costs/?filter[limit]=$LIMIT&filter[offset]=0&group_by[project]=*&order_by[cost]=desc&end_date=2021-02-28&start_date=2021-02-01&filter[resolution]=daily" | jq '.'`

Example response:
[query_json.txt](https://github.com/project-koku/koku/files/6091290/query_json.txt)


dlabrecq (dismissed):
LGTM

koku/api/report/queries.py
```diff
@@ -682,37 +690,92 @@ def _percent_delta(self, a, b):
...
         ranked_list = []
         others_list = []
         other_sums = {column: 0 for column in self._mapper.sum_columns}
         for data in entry:
             if other is None:
                 other = copy.deepcopy(data)
-            rank = data.get("rank")
+
+            if ranks:
+                rank = ranks.index(str(data.get(self._get_group_by()[0])))
```
adberglund, line 774:
```suggestion
                rank = ranks.index(str(data.get(self._get_group_by()[0]))) + 1
```

adberglund: We either want to start at 1 here or change the if statement below

`if rank > self._offset and rank <= self._limit + self._offset:`

to

`if rank >= self._offset and rank < self._limit + self._offset:`

adberglund (commented):
We have a handful of smoke tests failing on this. Essentially when we're returning daily data, and a particular day doesn't have an entry for one of the ranks, we need to add in a zeroed out or blank entry for it. This will be important for the table in the cost explorer so we always have an entry, even if it is 0. Happy to chat about this too.

koku/api/report/queries.py
blentz: I've applied the if-statement change. Thank you for catching this. ^_^

dccurtis:
If everything is set for this one.  Lets run smokes locally one more time before merging.

dlabrecq (approved):

dccurtis (approved):
Thanks for slogging through this one Brett!

PR #2711, opened 2021-03-08
Ocp query optimizations

adberglund:
## Summary
A few SQL optimizations for OCP report processing. 
- When getting the node relationship for volumes, add a WHERE clause for the time period and cluster of interest. Instead of scanning ALLLL the data. 
- Run tag summary SQL for time range of interest. Add a CTE to make sure we don't overwrite tag values not in the current time frame, but merge in the new values.


koku/masu/database/sql/reporting_ocpstoragevolumelabel_summary.sql
```diff
@@ -24,16 +26,37 @@ cte_values_agg AS (
...
+),
+cte_distinct_values_agg AS (
+    SELECT v.key,
+        array_agg(DISTINCT v.values) as values,
+        v.report_period_id,
+        v.namespace,
+        v.node
+    FROM (
+        SELECT va.key,
+            unnest(va.values || ls.values) as values,
```
Red-HAP, line 38: If a null in the array is not OK, then I'd suggest this change:

```sql
unnest(va.values || '{}'::text[]) as values
```

Also, I'd suggest putting `values` in dbl-quotes `"values"`

(same for all occurrences in this PR)

Red-HAP (approved):

PR #2715, opened 2021-03-08
COST-1133 - Sources delete task

maskarb:
This PR moves the source deletion into a celery task.

Testing:

---
1. create source
2. delete source. observe deletion taking place within a worker pod
---
1. create source
2. ingest large amt of data (large enough so that deletion takes some time)
3. delete the source.
4. kill the sources-client and restart
5. observe that the sources-client adds the source delete to the task queue again.
6. observe that the worker does not fail when the it tries to delete the same source again.


koku/koku/celery.py
```diff
@@ -119,6 +119,11 @@ def task(self, *args, **kwargs):
     "args": [],
 }
 
+# task to clean up sources with `pending_delete=t`
+app.conf.beat_schedule["delete_source_beat"] = {
+    "task": "sources.tasks.delete_source_beat",
+    "schedule": crontab(hour="4"),
```
dccurtis, line 125: Can we make this an environment variable?

dccurtis (approved):

koku/koku/celery.py
maskarb: Discussed offline. Decided to keep this value hardcoded. Task will run once per day at 4 am.

PR #2722, opened 2021-03-11
Add IBM to user-access view

boaz0:
w/o this patch, requesting "/user-access/?type=IBM" will return "Unknown source type: IBM". (reported by @dlabrecq).
I added IBM to the sources type with the "ibm.account" access_key.
This is only for making everything works with the current IBM details page with no errors whatsoever in the console log.

Related to COST-1088


dccurtis (dismissed):

dccurtis (approved):

boaz0:
Thanks again @dccurtis

PR #2723, opened 2021-03-15
[COST-1138] Large Delta Percent

cgoodfred:
This change handles what is probably an edge case where a previous cost can be extremely small resulting in an exceptionally large delta percent. If the previous cost value rounds to 0.00 at 2 decimal places, it will be treated as 0 for the purposes of the percent delta calculation. 

Smoke results: `===================== 2986 passed, 10 skipped, 521 deselected, 2 warnings in 2010.92s (0:33:30) =====================`


blentz (dismissed):
LGTM

maskarb (approved):

PR #2734, opened 2021-03-18
OCP log statement for summary start and end dates

dccurtis:
This log statement will tell us which operator the OCP cluster is using.  The start and end dates are present only in the manifests generated in the new operator.


adberglund (approved):
📆

PR #2743, opened 2021-03-22
COST-1161: Source async delete follow ups 

dccurtis:
This change stops ingest of pending delete sources and also allows for the user to re-add the source back before the provider is completely removed.

**Testing**
1. Create a source shutdown the worker then delete the source.  Verify that the provider is not active
2. POST /source-status for the source should return a 404 since we do not want to re-activate a provider if it's pending deletion.
3. Add another source with duplicate source name, billing source, and authenticaiton.  Verify creation is successful.  Even though the first source is still pending delete.
4. Start worker and verify that first source is remvoed. And second completes ingest.

**Test Results**
[cost_1161_ut.txt](https://github.com/project-koku/koku/files/6183169/cost_1161_ut.txt)


koku/sources/storage.py
```diff
@@ -278,6 +290,7 @@ def enqueue_source_delete(source_id, offset, allow_out_of_order=False):
         if not source.pending_delete and not source.out_of_order_delete:
             source.pending_delete = True
             source.save()
+            mark_provider_as_inactive(source.koku_uuid)
```
maskarb, line 293: This raises `Provider.DoesNotExist`. Is this exception caught anywhere?

dccurtis: Right here: https://github.com/project-koku/koku/pull/2743/files#diff-c25b68ba7a2e6f35a566bf4844d96b05c08ae2b4a72ffb29e99b07118f02bdacR272

adberglund (approved):

PR #2747, opened 2021-03-23
Add ibm.account to RESOURCE_TYPES in RBAC

boaz0:
This is a missing piece in IBM RBAC.

When a request is sent, the request contains the X_RH_IDENTITY header which is used by the middleware to fetch the user permissions. Then we check if cost management resource types are in the permissions and transform it to an object that user_access view can understand and returns whether that specific user has permission or not.

Unfortunately, I forgot to add "ibm.account" to the list of cost management resource types which caused to return `access denied` to users regardless if they have the "ibm.account" permission or not.

This patch is supposed to fix this problem.

//cc @ddonahue007 @dlabrecq @dccurtis

### DIFF
--- a/koku/koku/tests_rbac.py
+++ b/koku/koku/tests_rbac.py
@@ -96,6 +96,13 @@ def mocked_requests_get_200_no_next(*args, **kwargs):
     return MockResponse(json_response, status.HTTP_200_OK)
 
 
+def mocked_requests_get_200_no_next_ibm(*args, **kwargs):
+    """Mock valid status response that has no next."""
+    IBM = {"permission": "cost-management:ibm.account:*", "resourceDefinitions": []}
+    json_response = {"links": {"next": None}, "data": [IBM]}
+    return MockResponse(json_response, status.HTTP_200_OK)
+
+
 def mocked_requests_get_200_next(*args, **kwargs):
     """Mock valid status response that has no next."""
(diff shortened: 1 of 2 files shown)


ddonahue007 (dismissed):
LGTM

dlabrecq (dismissed):
LGTM

koku/koku/tests_rbac.py
```diff
@@ -458,6 +472,29 @@ def test_get_access_for_user_data_limited(self, mock_get):
             "openshift.cluster": {"read": []},
             "openshift.node": {"read": []},
             "openshift.project": {"read": []},
+            "ibm.account": {"read": []},
+        }
+        self.assertEqual(access, expected)
+        mock_get.assert_called()
+
+    @patch("koku.rbac.requests.get", side_effect=mocked_requests_get_200_no_next_ibm)
+    def test_ninja(self, mock_get):
```
adberglund, line 481: ⚔️

dccurtis, line 481: Test NINJA!

boaz0: *me trying to pickup a good name for a test*
!()[https://media.giphy.com/media/vVO14Ew5PYPLy/giphy.gif]

ddonahue007 (approved):
LGTM

dccurtis (approved):

PR #2750, opened 2021-03-23
COST-1110: forecast should not include negative numbers

blentz:
This resolves COST-1110. This PR ensures that any negative numbers in the forecast results are substituted with zero. I've also added a unit test to ensure this doesn't regress.


.pre-commit-config.yaml
```diff
@@ -1,5 +1,5 @@
 default_language_version:
-  python: python3.8
+  python: python3
```
blentz, line 2: This change is to ensure we can run pre-commit across any Python3 environment. Fedora 33 has python 3.9 as the default.

dlabrecq (approved):
LGTM

PR #2753, opened 2021-03-24
COST-680: Move gcp tag enablement out of development.

myersCody:
Did I make it a draft, because I have not run test yet, but I figured I could have travis do it while I am at lunch? Maybe.


boaz0:
@myersCody you forgot to update the unit tests (a small change though) :smiley: . Maybe you can create a variable that will hold the number of providers we have and use it in all tests while you're doing that. :man_shrugging:

boaz0 (approved):

boaz0 (commented):
Maybe we need to add GCP to "test_post_settings_ocp_tag_enabled" and "test_post_settings_ocp_tag_disabled"?

PR #2762, opened 2021-03-26
[COST-1237] Only call account labeler when reports processing

adberglund:
## Summary
We currently run the account labeler every time we check for a manifest, even when there is no new information. It's very slow and could easily be done just when there are new reports to process. 

## Testing

1. Ingest our AWS account
2. See `Running AccountLabel to get account aliases.` logged.
3. Run /download endpoint again. 
4. No logged message for the account labeler.

### DIFF
--- a/koku/masu/processor/orchestrator.py
+++ b/koku/masu/processor/orchestrator.py
@@ -244,7 +244,7 @@ class Orchestrator:
                     if account_number:
                         LOG.info("Account: %s Label: %s updated.", account_number, label)
 
-        return async_result
+        return
 
     def remove_expired_report_data(self, simulate=False, line_items_only=False):
         """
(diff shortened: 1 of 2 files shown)


myersCody (dismissed):

koku/masu/processor/orchestrator.py
```diff
@@ -230,15 +233,18 @@ def prepare(self):
...
+                    labeler = AccountLabel(
+                        auth=account.get("credentials"),
+                        schema=account.get("schema_name"),
+                        provider_type=account.get("provider_type"),
+                    )
+                    account_number, label = labeler.get_label_details()
+                    if account_number:
+                        LOG.info("Account: %s Label: %s updated.", account_number, label)
+
+        return async_result
```
adberglund, line 247:
```suggestion
        return
```

myersCody (approved):

PR #2769, opened 2021-03-29
[COST-1238] targeted download api

adberglund:
## Summary 
Allow the masu download API to work for a given source and billing period. 

## Testing
1. Hit http://127.0.0.1:5000/api/cost-management/v1/download/ and see normal behavior
2. http://127.0.0.1:5000/api/cost-management/v1/download/?provider_uuid={uuid} and see 2 month download check for just that source
3. http://127.0.0.1:5000/api/cost-management/v1/download/?provider_uuid={uuid}&bill_date=2021-03-01 and see 1 month download check for just that source and specified month.


myersCody (approved):
I was able to re-download a single source for a period of time.  👍

PR #2777, opened 2021-03-30
[COST-1394] Clowderize Koku

maskarb:
- last few code changes
- add Dockerfile / convert docker-compose to this new Dockerfile
- add build_deploy.sh and pr_check.sh

related PR: https://github.com/RedHatInsights/e2e-deploy/pull/3088


koku/koku/database.py
```diff
@@ -145,9 +146,9 @@ def check_migrations():
         with transaction.atomic():
             verify_migrations_dbfunc(connection)
             res = check_migrations_dbfunc(connection, targets)
-
         return res
-    except OperationalError:
+    except OperationalError as err:
+        LOG.info(f"what's going on here: {err}")
```
adberglund, line 151: 😄

scripts/run_migrations.sh
```diff
@@ -1,9 +1,9 @@
 #!/bin/bash
 
-RESULT=$(scl enable rh-python38 "${APP_HOME}/manage.py check_migrations | tail -1")
+RESULT=$(python3.8 ${APP_HOME}/manage.py check_migrations | tail -1)
```
adberglund, line 3: Why this change?

adberglund (commented):
Top level question. Because we have the platform logging set in the Dockerfile now, it's totally okay to remove the cloudwatch logger now right? We'll still get logs independent of being a ClowdApp?

scripts/run_migrations.sh
maskarb: In the base ubi8/python-38 image,`python3.8` is enabled for us, so we don't need the extra `scl enable` syntax:
```
RUN source scl_source enable rh-python38 && \
    python3.8 -m venv ${APP_ROOT} && \
    chown -R 1001:0 ${APP_ROOT} && \
    fix-permissions ${APP_ROOT} -P && \
    rpm-file-permissions
```
Is this change _necessary_? Not likely, but I do think it simplifies the script just a tiny bit.

koku/koku/database.py
maskarb: ha! whoops. Removed.

maskarb:
@adberglund to answer the Cloudwatch logging question, once we start using the Dockerfile this PR adds, then yes, we can drop the cloudwatch config. But, I'm thinking for the interim, I should add it back.

koku/sources/management/commands/sources_listener.py
```diff
@@ -0,0 +1,42 @@
+#
```
maskarb, line 1: This file enables us to separate the sources-api and the sources-listener into separate pods. This separation will likely happen with Clowder, not e2e.

adberglund (approved):

PR #2781, opened 2021-04-02
WIP:updating masu openapi.json

lcouzens:
This is changes to the masu openapi.json spec, there is a couple of issues I noticed while trying to call /status/ in my automation.

### DIFF
--- a/koku/masu/openapi.json
+++ b/koku/masu/openapi.json
@@ -372,39 +372,49 @@
         },
         "parameters": []
       },
-      "/status/": {
-        "get": {
-          "operationId": "status_list",
-          "description": "Packages response for class-based view.",
-          "parameters": [
-            {
-              "name": "celery",
-              "in": "query",
-              "description": "Whether to return celery queue/task data",
-              "required": false,
-              "schema": {
-                "type": "boolean",
-                "example": true
-              }
-            }
-          ],
-          "responses": {
-            "200": {
-              "description": "",
-              "content": {
-                "application/json": {
-                  "schema": {
-                    "$ref": "#/components/schemas/StatusResponse"
-                  }
+    "/status/": {
+      "get": {
+        "operationId": "getStatus",
+        "description": "Packages response for class-based view.",
+        "responses": {
+          "200": {
+            "description": "A JSON object containing pet information",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/StatusResponse"
                 }
               }
             }
-          },
-          "tags": [
-            "Status"
-          ]
+          }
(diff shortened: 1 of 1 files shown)


koku/masu/openapi.json
```diff
@@ -372,39 +372,49 @@
...
-                  }
+          }
+        },
+        "tags": [
+          "Status"
+        ]
+      },
+      "parameters": []
+    },
+    "/celery-status/": {
```
lcouzens, line 397: Wondering whether we should split this out since as far as I can tell you can't have multiple responses returned based from passed in parameters. The response from /status/?celery=true is completely different to /status/

`
{
    "scheduled_count": 0,
    "reserved_count": 0,
    "active_count": 0
}`

koku/masu/openapi.json
```diff
@@ -763,9 +773,8 @@
           "type": "object",
           "properties": {
             "api_version": {
-              "type": "integer",
-              "format": "int64",
-              "example": 1
+              "type": "string",
```
lcouzens, line 776: This is actually returning a string instead of an integer, I changed this just to get the spec returning the data correctly but this probably needs fixing in masu to return the integer instead.

koku/masu/openapi.json
```diff
@@ -847,10 +856,29 @@
           "type": "object",
           "properties": {
             "database": {
-              "type": "string"
+              "type": "string",
+              "nullable": "true"
```
lcouzens, line 860: Again here I added this just to get the spec working with the returned data, I think we are returning an unexpected DB here and probably needs fixing. `"database_status": [
        {
            "database": null,
            "database_connections": 0
        },
        {
            "database": "template1",
            "database_connections": 0
        },
        {
            "database": "template0",
            "database_connections": 0
        },
        {
            "database": "postgres",
            "database_connections": 6
        },
        {
            "database": "hive",
            "database_connections": 0
        }`

koku/masu/openapi.json
```diff
@@ -847,10 +856,29 @@
...
             },
             "database_connections": {
-              "type": "string"
+              "type": "integer",
+              "format": "int64"
+            }
+          }
+        },
+        "CeleryResponse": {
```
lcouzens, line 868: This is the response content that is actually missing from https://issues.redhat.com/browse/COST-1174 but like I mentioned, I'm not sure its accessible without having it as a new endpoint rather than a param on /status/ . I could be wrong but if not then we would need to change masu to add this endpoint for this to work.

lcouzens (commented):
@adberglund Please take a look and either I can update this with what we want to do or just close it out and create something new.

PR #2794, opened 2021-04-13
Cost 1272 trino table creation update

adberglund:
## Summary
Updates to reduce the number of times we try to create hive tables and clear data from S3 while processing parquet files. 
- Only create Trino/Hive tables one time per report processing (instead of for every file)
- Check for schema/table existence before creating
- Only delete previous CSVs/Parquet from S3 one time per manifest
- Wrap Trino errors in our own error to make it sleepable in sentry
- Don't load entire file into memory with `.read()` when sending data to S3, instead use upload_fileobj.


docker-compose.yml
```diff
@@ -226,7 +226,7 @@ services:
           - "${RABBITMQ_PORT-5674}:5672"
 
   koku-worker:
-      container_name: koku-worker
+      # container_name: koku-worker
```
Red-HAP, line 229: You've got this commented out. Is this intentional to be left in or is this a hanger-on during development?

Red-HAP (approved):

PR #2797, opened 2021-04-14
Pre process enabled tag keys

Red-HAP:
## Jira Ticket

[COST-1295](https://issues.redhat.com/browse/COST-1295)

## Description

Change structure `reporting_awsenabledtagkeys` 
* Delete primary key column `id`
* Drop unique index on `key` column
* Make `key` column the new primary key
* Add boolean column `enabled` default = True (database default of true also set)

Update business logic to not delete from the table, but to "soft" delete by setting enabled = False

Update tag management API to utilize the soft delete logic

Update parquet file processing to update/add enabled tags with data from the file transform to parquet

Disable the original parquet processing calls to create, update the AWS enabledtagkeys table

Update, create unit tests


koku/masu/processor/parquet/parquet_report_processor.py
```diff
@@ -35,16 +35,18 @@
 from masu.processor.azure.azure_report_parquet_processor import AzureReportParquetProcessor
 from masu.processor.gcp.gcp_report_parquet_processor import GCPReportParquetProcessor
 from masu.processor.ocp.ocp_report_parquet_processor import OCPReportParquetProcessor
+from masu.util import common as mutil
 from masu.util.aws.common import aws_post_processor
 from masu.util.aws.common import copy_data_to_s3_bucket
 from masu.util.aws.common import get_s3_resource
 from masu.util.aws.common import remove_files_not_in_set_from_s3_bucket
 from masu.util.azure.common import azure_post_processor
-from masu.util.common import get_column_converters
```
adberglund, line 43: I liked the explicit imports of the things we needed. Feels like this just added more work to go change tests and all that, but can live with it.

koku/masu/processor/aws/aws_report_parquet_summary_updater.py
```diff
@@ -121,9 +121,9 @@ def update_summary_tables(self, start_date, end_date):
                 accessor.populate_line_item_daily_summary_table_presto(
                     start, end, self._provider.uuid, current_bill_id, markup_value
                 )
-                accessor.populate_enabled_tag_keys(start, end, bill_ids)
+                # accessor.populate_enabled_tag_keys(start, end, bill_ids)
             accessor.populate_tags_summary_table(bill_ids, start_date, end_date)
-            accessor.update_line_item_daily_summary_with_enabled_tags(start_date, end_date, bill_ids)
+            # accessor.update_line_item_daily_summary_with_enabled_tags(start_date, end_date, bill_ids)
```
adberglund, line 126: We'll probably still want to figure out how to only populate the daily summary table with enabled tags since this is no longer being done. Can we run smoke tests on this to make sure this and the tag enablement for AWS still passes?

koku/reporting/migrations/0176_update_aws_enabled_keys.py
```diff
@@ -0,0 +1,26 @@
...
+            model_name="awsenabledtagkeys",
+            name="key",
+            field=models.CharField(max_length=253, primary_key=True, serialize=False),
+        ),
+        migrations.AddIndex(
+            model_name="awsenabledtagkeys", index=models.Index(fields=["key", "enabled"], name="aws_enabled_key_index")
+        ),
+        migrations.RunSQL(
+            """
+alter table reporting_awsenabledtagkeys alter column enabled set default true;
```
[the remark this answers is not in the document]
Red-HAP: Yep

koku/masu/processor/parquet/parquet_report_processor.py
Red-HAP: Was chasing my tail on this one since the actual error was masked. I thought I needed to qualify with a ns because of a wonky import or something.

koku/masu/processor/aws/aws_report_parquet_summary_updater.py
Red-HAP: Altered trino sql to handle tag filtering up-front for AWS.

adberglund (approved):

PR #2798, opened 2021-04-15
Add scale variable to docker make

adberglund:
## Summary
This removes the container name for the worker in docker compose, and adds the optional scale arg to the docker make commands. 

Ex. `make docker-up-min-no-build scale=4` would start the minimum set of containers, without rebuilding, and supply 4 worker containers. 

To get a single feed for those workers `docker-compose logs -f koku-worker` which is really a feed of 4 containers, so it can be confusing. Each container can still be docker log'd on its own.


Makefile
```diff
@@ -322,16 +323,16 @@ _koku-wait:
...
-	$(DOCKER_COMPOSE) up -d
+	$(DOCKER_COMPOSE) up -d --scale koku-worker=$(scale)
 
 docker-up-min:
-	$(DOCKER_COMPOSE) up --build -d db redis koku-server masu-server koku-worker
+	$(DOCKER_COMPOSE) up --build -d --scale koku-worker=$(scale) db redis koku-server masu-server koku-worker
 
 docker-up-min-no-build:
-	$(DOCKER_COMPOSE) up -d db redis koku-server masu-server koku-worker koku-listener
+	$(DOCKER_COMPOSE) up -d --scale koku-worker=$(scale) db redis koku-server masu-server koku-worker koku-listener
```
maskarb, line 335: can we pull the `db` part out and make it like the `docker-up-no-build`?

```
docker-up-min-no-build: docker-up-db
$(DOCKER_COMPOSE) up -d --scale koku-worker=$(scale) redis koku-server masu-server koku-worker koku-listener
```

this ensures the db is running before starting the other services.

maskarb (approved):

PR #2801, opened 2021-04-15
COST-1289: Handling OCP on Cloud exceptions so infrastructure completes processing.

dccurtis:
OCP to Cloud cost correlation summary failures should not prevent the individual cloud and OCP sources from completing processing.  This change will introduce a new exception type that will be caught by the update_summary_tables celery task.

**Testing**
1. Add an OCP source and ingest data.
2. Add a test Exception to force error during OCP correlation Cloud processing.  Add a cooresponding AWS source such that OCP is running on AWS.  Verify exception is caught and AWS processing completes.

**Test Results**
[cost_1289_ut.txt](https://github.com/project-koku/koku/files/6320202/cost_1289_ut.txt)

### DIFF
--- a/koku/masu/processor/report_summary_updater.py
+++ b/koku/masu/processor/report_summary_updater.py
@@ -44,6 +44,12 @@ class ReportSummaryUpdaterError(Exception):
     pass
 
 
+class ReportSummaryUpdaterCloudError(Exception):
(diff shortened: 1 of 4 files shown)


adberglund (approved):
🙆‍♂️

PR #2814, opened 2021-04-21
COST-1228: "no-project" ignores limit

myersCody:
Fix issue with no-project ignoring limit.

Eva created some smoke tests for us to check to see if this is fixed: `test_api_ocp_filtered_top_no_project`

Smoke Results:
```
lib64/python3.6/site-packages/iqe_cost_management/tests/rest_api/v1/test_ocp_cost_reports.py::test_api_ocp_filtered_top_no_project[1]
=== 4 passed, 6561 deselected in 28.93s ====
```

### DIFF
--- a/koku/api/report/queries.py
+++ b/koku/api/report/queries.py
@@ -48,6 +48,12 @@ def strip_tag_prefix(tag):
     return tag.replace("tag:", "").replace("and:", "").replace("or:", "")
 
 
+def is_grouped_by_tag(parameters):
+    """Determine if grouped by tag."""
+    group_by = list(parameters.parameters.get("group_by", {}).keys())
+    return [key for key in group_by if "tag" in key]
+
+
 def is_grouped_by_project(parameters):
     """Determine if grouped or filtered by project."""
     group_by = list(parameters.parameters.get("group_by", {}).keys())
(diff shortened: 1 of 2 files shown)


koku/api/report/queries.py
```diff
@@ -48,6 +48,12 @@ def strip_tag_prefix(tag):
     return tag.replace("tag:", "").replace("and:", "").replace("or:", "")
 
 
+def is_grouped_by_tag(parameters):
+    """Determine if grouped by tag."""
+    group_by = list(parameters.parameters.get("group_by", {}).keys())
+    return [key for key in group_by if "tag" in key]
```
adberglund, line 54:
```suggestion
    return any(["tag" in key for key in group_by])
```

adberglund: Totally optional, as its use will be the same, but this will return a single boolean if we have any tag group bys.

adberglund (approved):

dccurtis (approved):
Nice work!

koku/api/report/queries.py
myersCody: I actually use the return later on to get what the `tag_key` is for when the tag doesn't exist to create the `no-{tag_key}` value.

PR #2821, opened 2021-04-22
COST-1197: removing time scope deprecated comments

dccurtis:
time scope parameters will be used for the foreseeable future in all non cost exporter UI


adberglund (approved):
💬

dlabrecq (commented):
LGTM

PR #2822, opened 2021-04-22
add enabled col to model

Red-HAP:
Alter enabled tag keys models for Azure, GCP, OCP to simply add the `enabled` column.


myersCody:
I believe the fix is here @Red-HAP https://github.com/project-koku/koku/pull/2848

PR #2824, opened 2021-04-23
Tenant creation temporary fix

adberglund:
## Summary
This is an option to temporarily switch to using a SQL file to create a new tenant instead of the stored proc, while we sort out stored proc slow down.


maskarb (dismissed):
lgtm

maskarb (approved):

PR #2828, opened 2021-04-24
remove patch; apply tuned-up clone func

Red-HAP:
Does what it says on the tin.

Testing:

Add 30 tenants. schema clone should continue to be quick even as you get towards that limit.


maskarb (approved):
lgtm!

PR #2830, opened 2021-04-26
add tags to group by

Red-HAP:


adberglund (approved):
On latest master I ran

```
make create-test-customer
make load-test-customer-data
```

Then pulled this branch and copied `0174_update_ocpall_matviews.py` into `0177_update_ocpall_matviews.py` as a temporary migration to force the rebuild of the views with data already present in the DB. 

Everything worked!

PR #2840, opened 2021-04-28
[COST-1348] Option for timing request/response time in django

adberglund:
## Summary
This would add time to the logged message when we produce a response for an API call. 

Ex. 

```
[2021-04-28 16:37:54,290] INFO None {'method': 'GET', 'path': '/api/cost-management/v1/sources/', 'status': 200, 'request_id': None, 'account': '10001', 'username': 'user_dev', 'is_admin': 'True', 'response_time': '217.31 ms'}
```

### DIFF
--- a/koku/koku/middleware.py
+++ b/koku/koku/middleware.py
@@ -18,6 +18,7 @@
 import binascii
 import logging
 import threading
+import time
 from http import HTTPStatus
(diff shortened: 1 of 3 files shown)


boaz0 (approved):

maskarb (approved):
lgtm!

PR #2842, opened 2021-04-28
COST-1239: GCP project name is not being updated when changed.

myersCody:
Changes:
- Added the project name to the key so that way it wouldn't return the old value if name changes
- Whenever a project conflict happens now it will update instead of doing nothing

### DIFF
--- a/koku/masu/processor/gcp/gcp_report_processor.py
+++ b/koku/masu/processor/gcp/gcp_report_processor.py
@@ -248,7 +248,7 @@ class GCPReportProcessor(ReportProcessorBase):
         data = self._get_data_for_table(row, table_name._meta.db_table)
         data = report_db_accessor.clean_data(data, table_name._meta.db_table)
 
-        key = (data["account_id"], data["project_id"])
+        key = (data["account_id"], data["project_id"], data["project_name"])
         if key in self.processed_report.projects:
             return self.processed_report.projects[key]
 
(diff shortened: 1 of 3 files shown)


koku/masu/processor/gcp/gcp_report_processor.py
```diff
@@ -248,16 +248,16 @@ def _get_or_create_gcp_project(self, row, report_db_accessor):
...
         if key in self.processed_report.projects:
             return self.processed_report.projects[key]
 
         if key in self.existing_projects_map:
             return self.existing_projects_map[key]
 
         with transaction.atomic():
-            project_id = report_db_accessor.insert_on_conflict_do_nothing(
-                table_name, data, conflict_columns=["project_id"]
+            project_id = report_db_accessor.insert_on_conflict_do_update(
```
dccurtis, line 259: @adberglund does anything need to be done for trino?

myersCody: @dccurtis I actually had the same concerns when I was working on this. I chatted with @adberglund to make sure we don't need to do the same change in the parquet processor. From my understanding, the parquet processor doesn't use the `reporting_gcpproject` table, it just populates the data needed in the summary table. So, instead of updating the row in the `reporting_gcpproject` table in the parquet processor, the new row would just be inserted into the daily summary table with the correct value.

dccurtis: ok great!  That was what I suspected but I'm glad you confirmed it. 👍

dccurtis (approved):

PR #2843, opened 2021-04-29
Sources refactor

maskarb:
This is essentially a re-write of our sources integration.

- remove all RPC code
- rework the kafka listener so logic flows better
- use newer sources-api endpoints so that we make fewer api calls


dccurtis (approved):
Awesome work here Michael! 🙇

PR #2845, opened 2021-04-30
COST-1308: Re-enable source status pushes

dccurtis:
This changes moves the regular source status re-validation and status pushes to source-api to a celery beat schedule.  The sources wizard makes POST source-status calls which has proven to be unreliable because of race conditions of when the source-status API pushes status and when the sources client pushes status as a result of a source edit.

The change will re-evaluate sources and push status to the sources backend on a regular cadence.  This frequency is configurable with the `SOURCE_STATUS_FREQUENCY_MINUTES` environment variable and is defaulted to 30 minutes.

**Testing**
1. Create a GCP dataset with no billing_export.  Add source and verify no source status is pushed because Billing table is not present.
2. Set SOURCE_STATUS_FREQUENCY_MINUTES=1 to accelerate the status beat task.  Create a billing BigQuery table and verify the GCP provider is created and status is pushed to sources-api.
3. Add an AWS source and verify source status is pushed for both the AWS and GCP source.

**Test Results**
[cost_1308_ut.txt](https://github.com/project-koku/koku/files/6402449/cost_1308_ut.txt)

e2e-deploy: https://github.com/RedHatInsights/e2e-deploy/pull/3036


maskarb (approved):
lgtm!

PR #2852, opened 2021-05-04
COST-1362 - API empty state when no schema exists

maskarb:
- disable auto_schema_creation in middleware
- return`{}` if a request is associated with a tenant that has not created a schema
- create schema during source creation

### DIFF
--- a/koku/api/iam/models.py
+++ b/koku/api/iam/models.py
@@ -20,7 +20,6 @@ import os
 import pkgutil
 from uuid import uuid4
 
-from django.conf import settings
 from django.core.exceptions import ValidationError
 from django.db import connection as conn
 from django.db import models
@@ -32,6 +31,8 @@ from tenant_schemas.utils import schema_exists
 from koku.database import dbfunc_exists
 from koku.migration_sql_helpers import apply_sql_file
 from koku.migration_sql_helpers import find_db_functions_dir
+
+# from django.conf import settings
 
 
 LOG = logging.getLogger(__name__)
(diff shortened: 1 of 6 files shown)


koku/koku/middleware.py
```diff
@@ -113,6 +114,18 @@ def __init__(self, dikt):
         super().__init__(data)
 
 
+class KokuTenantSchemaExistsMiddleware(MiddlewareMixin):
+    """A middleware to check if schema exists for Tenant."""
+
+    def process_request(self, request):
+        tenant_username = request.user.username
+        user = User.objects.get(username=tenant_username)
```
Red-HAP, line 122:
```suggestion
        user = User.objects.select_related('customer').get(username=tenant_username)
```

koku/api/iam/models.py
```diff
@@ -32,6 +32,8 @@
 from koku.migration_sql_helpers import apply_sql_file
 from koku.migration_sql_helpers import find_db_functions_dir
 
+# from django.conf import settings
```
Red-HAP, line 35:
```suggestion
from django.conf import settings
```

koku/api/iam/models.py
```diff
@@ -111,6 +113,9 @@ class Tenant(TenantMixin):
     # Delete all schemas when a tenant is removed
     auto_drop_schema = True
 
+    # Schema creation is triggered by source creation
+    # auto_create_schema = settings.DEVELOPMENT or False
```
Red-HAP, line 117:
```suggestion
    auto_create_schema = bool(settings.DEVELOPMENT)
```

Red-HAP (changes requested):
These changes will produce a better DB query for user-> customer and keep merge conflicts with my work to a minimum

koku/koku/middleware.py
maskarb: I removed the user lookup altogether because the request contains the tenant info. So there is no need to look anything up.

PR #2856, opened 2021-05-04
COST-1362 API empty state when no schema exists

maskarb:
- return `{}` for a request whose tenant does not have a schema.
- return appropriate response for `sources-list` and `user-access` endpoints.


maskarb:
incorporated into https://github.com/project-koku/koku/pull/2859

PR #2857, opened 2021-05-05
[COST-1363] Api no tenant option two

adberglund:


adberglund:
Ended up combining branches for this.

PR #2870, opened 2021-05-11
use pg catalog for schema create

Red-HAP:
Use the template schema data in the PostgreSQL catalog as the source for creating the schema.

Tighten up checking for errors so that they are logged.

Adding some messaging around using the catalog as the clone data source


adberglund (dismissed):

adberglund (approved):

boaz0:
@adberglund - when this merged, I just need to pull the master and continue with my work (in addition to run `make make-migration` again for IBM databases), right?

adberglund:
@boaz0 That should be right. I'd pull in and rebuild everything (start a fresh DB) but this makes it so we can stop updating that static SQL file.

boaz0:
Sweet!
Good job @Red-HAP & @adberglund

PR #2871, opened 2021-05-12
Rc 2021.05.12

myersCody:


myersCody:
The smoke tests are failing to deploy here because sources is failing, we confirmed that we can run the smoke tests locally. After talking with Andrew & Doug we decided to move the release into staging so that we can re-run the smoke job against the stable branch tomorrow.

PR #2882, opened 2021-05-14
Cost 1410 matview index def

Red-HAP:
Remove project_costs from materialized view and unique index.
Added hash of tags and namespace. Using hash values in unique index. Hoping that will be unique enough (sha256).

Ticket: https://issues.redhat.com/browse/COST-1410


koku/reporting/migrations/0179_matview_tags_hash.py
```diff
@@ -0,0 +1,33 @@
...
+        view_sql = pkgutil.get_data("reporting.provider.all.openshift", f"sql/views/{view}.sql")
+        view_sql = view_sql.decode("utf-8")
+        LOG.info(f"""Creating materialized view "{view}"...""")
+        with connection.cursor() as cursor:
+            cursor.execute(view_sql)
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [("reporting", "0178_auto_20210511_1851")]
```
adberglund, line 31: I think the only other thing we need is to update 0178 to make it a no-op so this is the migration that does work.

Red-HAP: @adberglund Done. See above.

adberglund (approved):

PR #2888, opened 2021-05-17
COST-1404 add api to return supported currencies 

abaiken:
#### Testing 

Visit http://127.0.0.1:8000/api/cost-management/v1/currency/?limit=15 and see the list of currencies returned


dccurtis:
@abaiken can you update the openapi spec?  Thanks!

dccurtis (approved):

PR #2905, opened 2021-05-31
Update LICENSE to Apache

chargio:


maskarb:
What do we need to do about all the python source files that contain the following header:
```python
#
# Copyright YYYY Red Hat, Inc.
#
# This program is free software: you can redistribute it and/or modify
# it under the terms of the GNU Affero General Public License as
# published by the Free Software Foundation, either version 3 of the
# License, or (at your option) any later version.
#
# This program is distributed in the hope that it will be useful,
# but WITHOUT ANY WARRANTY; without even the implied warranty of
# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
# GNU Affero General Public License for more details.
#
# You should have received a copy of the GNU Affero General Public License
#
```

chargio:
We must delete it.

Sergio

Sergio Ocón-Cárdenas

He / Him / His

Senior Principal Product Manager, Hybrid Cloud Business Services

Red Hat <https://www.redhat.com/>

***@***.***
M: +34-692058130
<https://www.redhat.com/>

kholdaway:
Let's do that in this PR. It doesn't make sense to only change the license and leave the headers.

PR #2948, opened 2021-06-08
COST-1341: Fix no attribute bug in provider db accessor.

myersCody:
Jira Issue: https://issues.redhat.com/browse/COST-1341

The error we were seeing:
```
  File "/opt/app-root/src/koku/masu/util/ocp/common.py", line 213, in get_cluster_id_from_provider
    cluster_id = provider_accessor.get_credentials().get("cluster_id")
  File "/opt/app-root/src/koku/masu/database/provider_db_accessor.py", line 134, in get_credentials
    return self.provider.authentication.credentials
AttributeError: 'NoneType' object has no attribute 'credentials'
```

I added some conditionals, then wrote up a unittest that forces a none  in the authentication to confirm.

### DIFF
--- a/koku/masu/database/provider_db_accessor.py
+++ b/koku/masu/database/provider_db_accessor.py
@@ -135,7 +135,10 @@ class ProviderDBAccessor(KokuDBAccess):
                     example: {"bucket": "my-s3-cur-bucket"}
 
         """
-        return self.provider.billing_source.data_source
+        data_source = None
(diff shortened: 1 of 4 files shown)


maskarb:
I'm not saying that this is a problem, but if `authentication.credentials` gives us issues, then I would assume `billing_source.data_source` could do the same thing. Should we also add conditionals for data_source too?

dccurtis (approved):

PR #2950, opened 2021-06-11
Cost 1037 expire partitions

Red-HAP:
## Jira Ticket

[COST-1037](https://issues.redhat.com/browse/COST-1037)
[COST-1036](https://issues.redhat.com/browse/COST-1036)

## Description

Don't forget the **Note** at the bottom.

1. When expiring data, drop whole partitions of partitioned tables.
2. New Partitioned Tables:
    * `reporting_ocpawscostlineitem_daily_summary`
    * `reporting_ocpawscostlineitem_project_daily_summary`
    * `reporting_ocpazurecostlineitem_daily_summary`
    * `reporting_ocpazurecostlineitem_project_daily_summary`
3. Fix to partition manager trigger (reduced 2 trigger functions + 2 triggers to 1 of each)
4. While under the hood, add support for sub-partitions (multi-level partitioning) including in clone_schema
5. Bumped masu retention to 6 months
6. Fixed some tests by replacing hard-coded values with ones that will adapt as config changes are made

## Testing

Verify that data in the partitioned tables is not only deleted, but the containing table partition is dropped. Table partitions are currently splitting data by month ranges These ranges are added to the table name in YYYY_MM format. The `reporting_ocpusagelineitem_daily_summary` table is partitioned; so a table partition for data from **2021-06-01** to **2021-07-01** would have the table name `reporting_ocpusagelineitem_daily_summary_2021_06` because that table will only hold data for the month of June, 2021.

Verify that this holds true for the newly partitioned tables (bullet points under **2** above

Partitions are still successfully auto-created during download + processing for the database or for Trino

Item 4 is for potential future use. There should be no errors regarding creating or dropping partitions. Clone schema should continue to run successfully.

Verify that data can be retained for 6 months in the summary tables.

## Note

This should be **very-reviewed** and the tires kicked before merge into master.


db_functions/partitioned_manage_trigger.sql
```diff
@@ -0,0 +1,24 @@
+/*
```
dccurtis, line 1: Please update new files with the Apache text:  Example: https://github.com/project-koku/koku/blob/ed866cb9c0f45e7aeb5ee0caab8e41fb8fe1ebdb/db_functions/clone_schema.sql#L3

koku/masu/processor/expired_data_remover.py
```diff
@@ -130,18 +130,13 @@ def remove(self, simulate=False, provider_uuid=None, line_items_only=False):
             ([{}]) List of dictionaries containing 'account_payer_id' and 'billing_period_start'
 
         """
-        removed_data = None
+        removed_data = []
         disable_purge_line_item = (Provider.PROVIDER_AZURE, Provider.PROVIDER_AZURE_LOCAL, Provider.PROVIDER_OCP)
```
dccurtis, line 134: I think we can remove this since line item data removal is now gone.

koku/masu/processor/expired_data_remover.py
```diff
@@ -130,18 +130,13 @@ def remove(self, simulate=False, provider_uuid=None, line_items_only=False):
...
         no_data_msg = "%s has no line item data to be be removed."
+        line_items_only = False
```
dccurtis, line 136: We can remove the `line_items_only` argument from `ExpiredDataRemover.remove` and the other places throughout the code.

This if block can also be removed since we no longer have to remove line item data as part of the summary processing:
https://github.com/project-koku/koku/blob/2087dcce7761656bd04f02c191f0f3ea474aa671/koku/masu/processor/tasks.py#L429

When trying your branch locally it seems to be removing all summary data because of this.

dccurtis: Doing that test I could definitely see data being removed from the partition tables but I still see the tables with psql (`reporting_awscostentrylineitem_daily_summary_2021_06`). Should these be removed after the table drop?

koku/masu/processor/expired_data_remover.py
Red-HAP: There are still data returned as a list even though line-item data are not deleted. This is for account info.

koku/masu/processor/expired_data_remover.py
Red-HAP: The table drop should eliminate the table partition. I will double-check the code here.

dccurtis (dismissed):
looks good!  As discussed lets wait for https://github.com/project-koku/koku/pull/3022 to be merged before merging this.

dccurtis (approved):

PR #2953, opened 2021-06-11
Use raw sql to execute deletes

Red-HAP:
## Jira Ticket

[COST-1490](https://issues.redhat.com/browse/COST-1490)

## Description

Use raw sql vs ORM to do deletes. This change takes the ORM query and uses django's own SQLDeleteCompiler to get the text of the SQL statement plus the parameters. This is then processed directly via `connection` and `cursor` objects without using django ORM to attempt to verify and handle the foreign keys.

`Provider().delete()` is now tuned up as well to call a func named `cascade_delete` that will walk ORM relations and execute SQL to delete related records bottom-up.

## Testing

The resulting delete action should be **no different** from usual. But the memory usage should not expand to OOM state during delete operations for expired data or anything else that a `db_cleaner` would be called for.

If you run `make docker-reinitdb-with-sources` then run a django shell, you can watch the cascade delete work.
How-To-Do-It Lessons:
1. Navigate to top-of-repo
3. `make load-test-customer-data`
4. You can check the relations from Provider to the line-item tables, if you want.
5. `cd koku`
6. `export KOKU_LOG_LEVEL=DEBUG`
6. `python ./manage.py shell`
7. `from api.provider.models import Provider`
8. `p = Provider.objects.first()` (or `p = Provider.objects.get(pk=<UUID>)` if you have identified a provider to use.)
9. `p.delete()`
10. You should see a flurry of log messages with `Level` and `Cascade` in them. With logging set to debug, you'll also see the formatted SQL statements that are being executed.
11. `ctrl+d` to exit the shell
12. ``unset KOKU_LOG_LEVEL` to clear debug logging


dccurtis:
@Red-HAP Can this be extended to cover the flow for removing a provider here:
https://github.com/project-koku/koku/blob/13b91d62beadf459d5bfbd91b7326e39b60f48fc/koku/api/provider/provider_manager.py#L208

This is what is called when a source is destroyed and is the path in which we are suspecting the Django ORM being inefficient at managing the foreign key relationships while deleting.

dccurtis (approved):
Awesome work HAP!

PR #2959, opened 2021-06-17
[Cost 1501] Resource Types need RBAC for type-ahead use

Fozzywozzybear:
Adding Rbac to Resource type endpoints for GCP/AWS/AZURE/OCP. This gives users access to data that only they have access too. 

Steps to reproduce
1.Create and load test customer data.
2. Edit this line in your local .env file `DEVELOPMENT_IDENTITY='{"identity": {"account_number": "10001", "type": "User", "user": {"username": "user_dev", "email": "user_dev@foo.com", "is_org_admin": "True", "access": {}}},"entitlements": {"cost_management": {"is_entitled": "True"}}}' 
`
Making sure to change it, example 
`DEVELOPMENT_IDENTITY='{"identity": {"account_number": "10001", "type": "User", "user": {"username": "user_dev", "email": "user_dev@foo.com", "is_org_admin": "False", "access": {"aws.account": {"read": ["9999999992"]}}}},"entitlements": {"cost_management": {"is_entitled": "True"}}}'`
This gives the user access to all AWS data that the account 9999999992 has access too. This can be changed to other RBAC preset types all that can be used are listed below 

- aws.account aws.organizatoinal_unit For All Aws Endpoints, Regions and services use aws.account
- gcp.account and gcp.project For all gcp endpoints, regions and services use gcp.account
- azure.subsription_guid For all Azure endpoints 
- openshift.project openshift.cluster and openshift.node For all Openshift endpoints.

3. Load up the end point http://localhost:8000/api/cost-management/v1/resource-types/aws-accounts/ . If you are Admin you will have all access to see all accounts, if you have the example set you should see only accounts that your set account has access to. 
4. Repeat steps 2-3 for all endpoints under gcp/aws/azure/ocp
5. On local testing with how our middleware works if you want to test the pages results where you are not the org admin, you need to change in the env file "is_org_admin": "True" to "is_org_admin": "" . Since the middle ware checks if there is a string in that spot it will always return true even when set too false. An empty string will return false as org admin.

Smoke tests 
```
FAILED lib64/python3.6/site-packages/iqe_cost_management/tests/rest_api/v1/test_azure_cost_reports.py::test_api_azure_cost_tag_validate_keys_visible - AssertionError: Failing from https://iss...
FAILED lib64/python3.6/site-packages/iqe_cost_management/tests/rest_api/v1/test_source.py::test_api_ocp_source_crud_upload_service - AssertionError: Raw calculated value does not match report...
FAILED lib64/python3.6/site-packages/iqe_cost_management/tests/rest_api/v1/test_source.py::test_api_multi_ingest_ocp_source - AssertionError: assert {'cluster': 'test_cost_ocp_static_EiZqhEuh...
============================================================ 3 failed, 5778 passed, 200 skipped, 703 deselected in 3163.39s (0:52:43) ============================================================
```


cgoodfred (dismissed):
LGTM

cgoodfred (approved):

PR #2960, opened 2021-06-18
Trino aws bugfix empty tags

adberglund:
## Summary
Our trino SQL fails for AWS when no user defined tags are sent through the cost and usage report. This changes a `NaN` value to an empty dict for the empty tags and allows us to fully process and summarize.


chambridge (approved):
LGTM

PR #2968, opened 2021-06-25
COST-1077: Editing azure secret error message to cover expired case.

dccurtis:
Azure secret credentials have an expiration date.  If a user's credentials expire the secret that they are using for cost management becomes invalid and we push the failed source status to platform sources.  This changes edits our message to tell the user that their credentials may have expired.

The Azure UI makes it very clear what credential expired and they can go and create a new one and edit the source to continue using the same source with cost management.  A subsequent docs update will be completed to explain this.

Wizard error message
image

Sources UI pop over error message
image

Source edit error message
image


dccurtis:
Discussed with @xJustin and @nlcwong.  Going to revise the wording to be more generic that the secret isn't working.  Docs updates will contain a troubleshooting table where the user can cross-reference this error string to corrective actions steps in the azure console

koku/providers/provider_errors.py
```diff
@@ -78,7 +78,7 @@ class ProviderErrors:
         "Edit your Azure source to include these details."
     )
     AZURE_INCORRECT_CLIENT_SECRET_MESSAGE = (
-        "The client secret was entered incorrectly for this source. "
+        "The client secret was entered incorrectly or has expired for this source. "
```
xJustin, line 81: @dccurtis @nlcwong Maybe something like this:
A problem has been detected with the Azure client secret for this source. Refer to the Microsoft Azure troubleshooting guide in the cost management documentation for details.   
??? too long?

dccurtis: I think that would be ok.  I can update the code and give an example in the UI next.  If we reference the docs we should wait until the docs are updated before making the error message update.

dccurtis: @nlcwong I updated the screenshots with @xJustin wording suggestion.  How's it look to you?

nlcwong:
too bad we can't tell if it is expired vs. incorrectly used, but it LGTM.

adberglund (approved):

PR #2900, opened 2021-05-25
COST-1446: Update Cost Model API for Cost Distribution

myersCody:
Criteria:
- Only allow the distribution choice when doing a monthly node or a monthly cluster cost
- Make sure the distribution type is either memory or cpu
- Update openapi.json

You can run `make load-test-customer-data` to test. 
Then you will need to delete the default cost model and put in the new one:
```
{
    "name": "Cost Management OpenShift Cost Model",
    "description": "A cost model of on-premises OpenShift clusters.",
    "source_type": "OCP",
    "source_uuids": ["e7308aa0-961b-46f5-8a3a-b31c926b387e"],
    "rates": [{
        "metric": {"name": "cpu_core_usage_per_hour"},
        "tiered_rates": [{
            "unit": "USD",
            "value": 0.0070000000,
            "usage_start": null,
            "usage_end": null
        }],
        "cost_type": "Supplementary"
    },
    {
        "metric": {"name": "cpu_core_request_per_hour"},
        "tiered_rates": [{
            "unit": "USD",
            "value": 0.2000000000,
            "usage_start": null,
            "usage_end": null
        }],
        "cost_type": "Supplementary"
    },
    {
        "metric": {"name": "memory_gb_usage_per_hour"},
        "tiered_rates": [{
            "unit": "USD",
            "value": 0.0090000000,
            "usage_start": null,
            "usage_end": null
        }],
        "cost_type": "Supplementary"
    },
    {
        "metric": {"name": "memory_gb_request_per_hour"},
        "tiered_rates": [{
            "unit": "USD",
            "value": 0.0500000000,
            "usage_start": null,
            "usage_end": null
        }],
        "cost_type": "Supplementary"
    },
    {
        "metric": {"name": "storage_gb_usage_per_month"},
        "tiered_rates": [{
            "unit": "USD",
            "value": 0.01,
            "usage_start": null,
            "usage_end": null
        }],
        "cost_type": "Supplementary"
    },
    {
        "metric": {"name": "storage_gb_request_per_month"},
        "tiered_rates": [{
            "unit": "USD",
            "value": 0.01,
            "usage_start": null,
            "usage_end": null
        }],
        "cost_type": "Supplementary"
    },
    {
        "metric": {"name": "node_cost_per_month"},
        "tiered_rates": [{
            "unit": "USD",
            "value": 1000.0,
            "usage_start": null,
            "usage_end": null
        }],
        "cost_type": "Infrastructure"
    },
    {
[...]

PR #2977, opened 2021-06-29
COST-1447, COST-1448: Cluster to node distribution (cpu & memory).

myersCody:
This pr follows along pretty closely to the design laid out in 
- https://issues.redhat.com/browse/COST-1447
- https://issues.redhat.com/browse/COST-1448

# Background
The overall biggest change is the transition from the `supplementary_monthly_cost` and `infrastructure_monthly_cost` from decimals to a jsonb object. We needed to convert it to a json object so that we could distinguish between distribution type `cpu`, `memory`, and `pvc`. For additional context I would check out the comment in the issue to the cost distribution breakdown google doc. 

# Testing
How to test, you can now run `make load-test-customer-data` to populate openshift on premise cluster with a distribution cost. The default distribution is `cpu`, but you can change it to `memory` using the cost model example in this pr description: https://github.com/project-koku/koku/pull/2900

After you have successfully loaded the data you can now check out the ocp endpoints. You will notice that for the ocp endpoints we now have a new `distributed_cost` column:
```
"infrastructure": {
                "raw": {
                    "value": 1333.3458948441,
                    "units": "USD"
                },
                "markup": {
                    "value": 126.45735272621002,
                    "units": "USD"
                },
                "usage": {
                    "value": 0.0,
                    "units": "USD"
                },
                "distributed": {
                    "value": 17000.0,
                    "units": "USD"
                },
                "total": {
                    "value": 18459.80324757031,
                    "units": "USD"
                }
            },
```
The first thing to check is these two endpoints: 
- http://localhost:8000/api/cost-management/v1/reports/openshift/costs/?filter[time_scope_units]=month&filter[time_scope_value]=-1&filter[resolution]=daily
-  http://localhost:8000/api/cost-management/v1/reports/openshift/costs/?filter[time_scope_units]=month&filter[time_scope_value]=-1&group_by[node]=*&filter[resolution]=daily

So before this change the cluster cost was not being distributed at the node level, so you would get a different cost when you group by node. However, that is now fixed. 

Additionally, for however you decided to distribute the cost, the distributed cost will only show up on that endpoint. 
CPU: http://localhost:8000/api/cost-management/v1/reports/openshift/compute/?filter[time_scope_value]=-30&filter[time_scope_units]=day&filter[resolution]=daily
Memory: http://localhost:8000/api/cost-management/v1/reports/openshift/memory/?filter[time_scope_value]=-30&filter[time_scope_units]=day&filter[resolution]=daily

I also had to do a few changes to all of the materialize views to handle the new jsonb field without causing a regression. Its also important to add up all the individual cluster costs per node to make sure they add back up to the original cost.

 Feel free to ask me any questions! 😄 
 
 # Local Development Changes
 
 Also, I added a static file for the ocp on prem nise generation for when you run `make load-test-customer-data` because the node & cluster capacities get all wonky without the static file making the distribution seem really off.


dchorvat1 (dismissed):
I tested the branch, tried out cost distribution by both CPU/memory, ran our existing tests to check for regressions and started working on new automation dedicated for cost distribution. Cody's fix also solved the cost model update issue he found, LGTM!

koku/masu/database/ocp_report_db_accessor.py
```diff
@@ -1163,58 +1189,125 @@ def tag_upsert_monthly_default_pvc_cost_line_item(
...
+        node_column = "node_capacity_cpu_core_hours"
+        cluster_column = "cluster_capacity_cpu_core_hours"
+        if "memory" in distribution:
+            node_column = "node_capacity_memory_gigabyte_hours"
+            cluster_column = "cluster_capacity_memory_gigabyte_hours"
+
+        with schema_context(self.schema):
+            distributed_node_list = (
+                OCPUsageLineItemDailySummary.objects.filter(
+                    usage_start__gte=start_date, usage_start__lt=end_date, cluster_id=cluster_id
```
myersCody, line 1235: Should I be filtering by `report_id`, and `cluster_alias` here as well?

dccurtis (approved):
LGTM!

PR #2980, opened 2021-06-30
[COST-1579] trino ocp datesplit update

adberglund:
## Summary
When an OpenShift manifest has the version field, we record the version in the manifest table. We also do not split the files into files by date now for this case, since we only receive new data with the new operator.

## Testing
Run make create-test-customer and make load-test-customer-data in Trino mode. Check manifest in DB to see that the operator version is added. See that OpenShift is not date split and only a single file is processed per report type.


koku/masu/external/downloader/ocp/ocp_report_downloader.py
```diff
@@ -93,6 +96,7 @@ def create_daily_archives(request_id, account, provider_uuid, filename, filepath
                 context,
             )
             daily_file_names.append(daily_file.get("filepath"))
+    LOG.info(f"\t\tReturning {daily_file_names}")
```
adberglund, line 99:
```suggestion
```

abaiken (approved):
LGTM

PR #2982, opened 2021-06-30
COST-1446: Fix updating distribution choice in the cost model.

myersCody:
During the qa progress for my initial COST-1446 pr, @dchorvat1  noticed that we did not have the ability to update the distribution choice after initial creation. Awesome catch @dchorvat1! 

This pr is to allow you to update the distribution choice. 

How to test:
1) run `make load-test-customer-data` this will create cost models for you with a default distribution choice of cpu. Update the distribution choice to be `memory` and then check the cost models endpoint to make sure the change happened.


abaiken (approved):
LGTM

PR #3005, opened 2021-07-09
Set constraints to immediate

Red-HAP:
Deletes are now part of an atomic transaction that will set constraints to immediate.

I currently suspect that the deferred constraints are causing a problem with accessing the provider table during these long-running deletes. If that is the case, then setting the constraints from deferred to immediate should help.


maskarb (approved):
lgtm!

PR #3007, opened 2021-07-09
Temporarily disabling provider deletes

dccurtis:


maskarb (approved):

maskarb (commented):
I think this change is ok, but I do wonder if instead of blocking the `coordinator.destroy_account`, if it makes sense to block `self._provider_builder.destroy_provider(koku_uuid)` instead and let the code delete the source? This way, we could mark the provider as inactive and delete the source. It could make cleaning up the providers later a little more straightforward.

Either way, I still approved this PR.

PR #3019, opened 2021-07-15
Cost 1599 aws tag bugs2

Red-HAP:
## Jira Ticket

[COST-1599](https://issues.redhat.com/browse/COST-1599)

## Description

Addresses an issue where the provider tag query handlers were "locked" into always returning enabled tags even if filters were set for `enabled=False`

## Testing

I'm relying on Eva's smoke tests as the gold standard here.


dccurtis:
Looks like there may be an issue with Azure.  Here's some more detail on the unit test failure:
```
{'handler': <class 'api.tags.azure.queries.AzureTagQueryHandler'>, 'view': <class 'api.tags.azure.view.AzureTagView'>, 'name': 'azure'}
(Pdb) query_output
{'filter': OrderedDict([('resolution', 'monthly'), ('time_scope_value', '-1'), ('time_scope_units', 'month'), ('enabled', False)]), 'key_only': True, 'group_by': OrderedDict(), 'order_by': OrderedDict(), 'access': OrderedDict(), 'data': []}
(Pdb) 
```

`'data': []`

dccurtis (approved):

PR #3022, opened 2021-07-19
Process smaller dataframes for OCP on cloud

adberglund:
## Summary
Instead of concatenating the smaller daily data frames, this keeps a list of them and processes each smaller frame.


maskarb (approved):
lgtm!

PR #3039, opened 2021-07-28
[Cost1567] Prevent query params besides those that are allowed on resource type endpoints

Fozzywozzybear:
On all resource type endpoints any query parameters where allowed, this update only allows the correct parameters on each endpoint. Example 
Allowed Parameters
/resource-types/openshift-projects/?limit=1000
/resource-types/openshift-projects/?search=foo
/resource-types/aws-accounts/?openshift=true
No Allowed
/resource-types/openshift-projects/?foo=
limit search and openshift are the three allowed query parameters at the moment on resource type endpoints.  When trying to add a parameter that is not allowed the user will receive this error 
http://localhost:8000/api/cost-management/v1/resource-types/openshift-projects/?foo
`{
    "foo": [
        [
            "Unsupported parameter"
        ]
    ]
}`
This error changes dynamically to the parameter that is entered. This also returns a 400 bad request error.
Smoke tests 
`FAILED tests/rest_api/v1/test__ocp_cost_reports.py::test_api_ocp_cost_order_by_name_pagination
FAILED tests/rest_api/v1/test_azure_cost_reports.py::test_api_azure_cost_tag_validate_keys_visible
FAILED tests/rest_api/v1/test_gcp_cost_reports.py::test_api_gcp_cost_order_by_name_pagination
FAILED tests/rest_api/v1/test_source.py::test_api_ocp_on_aws_source_crud
FAILED tests/rest_api/v1/test_source.py::test_api_aws_source_raw_calc - As...
`


dlabrecq (approved):

lcouzens:
@Fozzywozzybear Looks good for the most part but your missing this change on `resources-types/cost-models/`

dccurtis (dismissed):

dccurtis (approved):

PR #3053, opened 2021-08-04
Avoid adding an empty dictionary to the operator_errors field for the manifest

abaiken:
Previously since we default errors to an empty dictionary, the empty dictionary was added to the field for operator errors. In the koku-daily reports I'm going to be counting the errors based on whether or not this field is Null, so an empty dictionary was problematic. This PR cleans up the empty dictionary and only adds errors if there are errors. 

This PR also adds the cluster_id to the manifest table

Related PR: 
* https://github.com/project-koku/koku-daily/pull/43


koku/masu/external/downloader/ocp/ocp_report_downloader.py
```diff
@@ -143,7 +143,8 @@ def process_cr(report_meta):
             case_info = cr_status.get(case, {})
             if case_info.get("error"):
                 errors[case + "_error"] = case_info.get("error")
-        manifest_info["operator_errors"] = errors
+        if errors != {}:
+            manifest_info["operator_errors"] = errors
```
maskarb, line 147: just to prevent any future silly `keyErrors` could we do:
```suggestion
        manifest_info["operator_errors"] = errors or None
```

```python
>>> d1 = {}
>>> d2 = {'k1':'v1'}
>>> d2['errors'] = d1 or None
>>> print(d2)
{'k1': 'v1', 'errors': None}
```

maskarb (approved):
lgtm!

PR #3059, opened 2021-08-06
Cost 1486 initial migration

Red-HAP:
## Jira Ticket

[COST-1486](https://issues.redhat.com/browse/COST-1486)

## Description

This PR is for a migration file that will add partitioned tables for the materialized views `reporting_ocpallcostlineitem_daily_summary` and `reporting_ocpallcostlineitem_project_daily_summary`. It will also create any requisite partitions for these tables including a default partition. Any data in the source tables for AWS and Azure will be copied to these tables.

There are reverse steps in this migration if needed.

There are two optional parameters now for the `run-migrations` make target: `applabel` and 'migration`. **Specify both, or neither.** These parameters can control the target migration when running a downgrade.

This is an additive migration that has no API code changes. 

This PR does not close the ticket.

## Testing

1. Checkout main
3. `make load-test-customer-data` _Use start, end params to get 2 months of data in_
3. Checkout this branch
4. `make run-migrations`
5. `psql 'postgresql://postgres:postgres@localhost:15432/postgres?application_name=tester&sslmode=prefer' -c '\dt acct10001.*' | grep -E '_p[_ ]'`
6. `psql 'postgresql://postgres:postgres@localhost:15432/postgres?application_name=tester&sslmode=prefer' -c '\dm acct10001.*' | grep '_p '`
7. Look for object names suffixed with `_p` or `_p_YYYY_MM` or `_p_default` in the output
8. `psql 'postgresql://postgres:postgres@localhost:15432/postgres?application_name=tester&sslmode=prefer' -c 'select count(*) from acct10001.reporting_ocpallcostlineitem_daily_summary_p;'`
9. `psql 'postgresql://postgres:postgres@localhost:15432/postgres?application_name=tester&sslmode=prefer' -c 'select count(*) from acct10001.reporting_ocpallcostlineitem_project_daily_summary_p;'`
10. You should see a count number for each of these queries
11. `make run-migrations applabel=reporting migration=0189_auto_20210803_2056`
12. Repeat steps 5 and 6
13. The objects to check for in step 7 should not be present. (There should be no output)
14. Repeat steps 8 and 9
15. You should get "does not exist" errors for both.


Red-HAP:
Finally fixed testing step in comment.

PR #3064, opened 2021-08-11
New squash method

Red-HAP:
New squash to factor-out `pg_partition` which began to fail in the prior squashed reporting initial migration.

This squash uses a dump from the template schema and includes the state up to reporting **0189** migration.

This dump is only for the `partitioned_tables` and `reporting_` tables. All other tables will be initialized/created by existing migrations.

Application to an existing database at state 0189 _should_ be a no-op.

Application to a new database should succeed.

**All databases** should be at state of reporting 0189 before this is applied.


maskarb:
My tagging logic doesn't seem to work when it's not tagged _just_ right (we'll need to fix this). Gonna remove some tags and retest.

adberglund:
@Red-HAP can this be closed?

PR #3070, opened 2021-08-12
MVP Currency

boaz0:


koku/currency/models.py
```diff
@@ -0,0 +1,30 @@
+from django.db import models
```
dccurtis, line 1: FYI: We're going to want to make the migration additions in a separate PR to help simplify release rollbacks

boaz0: So should the migration PR be merged first or last?

PR #3078, opened 2021-08-17
[COST-1784] Delete ocp on cloud parquet too

adberglund:
## Summary
* When we process AWS or Azure data, we delete previous parquet files for the month. We were not doing this for ocp on Cloud parquet files, causing duplication. This will delete these files before reprocessing the month. 
* Fixes OCP on Azure resource ID matching using string contains instead of isin

## Testing
Ran make-load-test-customer-data several times before making this change and had a parquet file from each run in the daily/openshift parquet path for AWS. Made this change and re-ran resulting in a single file covering the entire month's costs.


myersCody (approved):
LGTM

Red-HAP (approved):

PR #3082, opened 2021-08-19
automate pre-release

maskarb:
The goal of the pre-release action is to write the release notes once the tag has been created.

The idea would be to create an empty release for a commit, and then this action will populate release body.


adberglund (approved):
🥨

dccurtis (approved):

PR #3097, opened 2021-08-27
[Cost-1810]Can't access nodes/projects resource-types when rbac is restricted to a cluster

Fozzywozzybear:
This fix allows users that have set cluster restrictions to now see the proper data when they hit the resource-types/openshift-projects/ and openshift-nodes/.
steps to reproduce
1. Make a load customer data
2. Set rbac restrictions for openshift-cluster.
3. Hit the resource-type/openshift-projects/ or /resource-type/openshift-nodes/ 
4. The data returned should be related to that cluster that is set in RBAC

Please reach out to me or QA for any more details 
Jira ticket 
https://issues.redhat.com/browse/COST-1810
--- All Smokes pass ---


koku/api/resource_types/openshift_projects/view.py
```diff
@@ -47,8 +47,14 @@ def list(self, request):
...
             return super().list(request)
         elif request.user.access:
-            user_access = request.user.access.get("openshift.project", {}).get("read", [])
+            if request.user.access.get("openshift.cluster", {}).get("read", []):
+                user_access = request.user.access.get("openshift.cluster", {}).get("read", [])
+                # We hold a copy of the filtered queryset just incase the user has a wildcard for user access
+                query_holder = self.queryset.filter(cluster_id__in=user_access)
+            elif request.user.access.get("openshift.project", {}).get("read", []):
+                user_access = request.user.access.get("openshift.project", {}).get("read", [])
+                query_holder = self.queryset.filter(cluster_id__in=user_access)
```
maskarb, line 56: Is this correct? Shouldn't it be:
```suggestion
                query_holder = self.queryset.filter(namespace__in=user_access)
```

maskarb (changes requested):

koku/api/resource_types/test/tests_views.py
```diff
@@ -132,18 +132,6 @@ def test_openshift_endpoints_view(self):
...
-        """Test endpoint runs with a customer owner."""
-        for endpoint in self.ENDPOINTS_AZURE:
-            with self.subTest(endpoint=endpoint):
-                url = reverse(endpoint)
-                response = self.client.get(url, **self.headers)
-                self.assertEqual(response.status_code, status.HTTP_200_OK)
-                json_result = response.json()
-                self.assertIsNotNone(json_result.get("data"))
-                self.assertIsInstance(json_result.get("data"), list)
-
```
maskarb, line 146: Why was this test removed?

Fozzywozzybear: It was an accident, good catch !

Fozzywozzybear: It has been added back

lcouzens (approved):
lgtm

PR #3100, opened 2021-08-31
enable finer logging control for the individual services

maskarb:
- Remove `LOG_LEVEL` variable: It was unclear what this variable was for.
- Add `CELERY_LOG_LEVEL`: this is the log level set for `celery` during worker initialization
- Add `GUNICORN_LOG_LEVEL`: log level for gunicorn. Only used in koku-api, masu, sources-client
- Create variables to set `KOKU_LOG_LEVEL` independently in koku-api, masu, listener, sources-client, and sources-listener. All of the workers will still use `KOKU_LOG_LEVEL`.
- Leave `DJANGO_LOG_LEVEL` alone. All services will use the same log level.


pr_check.sh
```diff
@@ -14,7 +14,7 @@ ARTIFACTS_DIR="$WORKSPACE/artifacts"
 export IQE_PLUGINS="cost_management"
 export IQE_MARKER_EXPRESSION="cost_smoke"
 export IQE_FILTER_EXPRESSION="test_api"
-export IQE_CJI_TIMEOUT="2h"
+export IQE_CJI_TIMEOUT="75m"
```
adberglund, line 17: Time liberator!

adberglund (approved):

PR #3105, opened 2021-09-01
[COST-1821] - upgrade Celery and Kombu

maskarb:
- drop celery-prometheus-exporter [[COST-520](https://issues.redhat.com/browse/COST-520)] -  as far as I can tell, we weren't using celery-prometheus-exporter anyway, so dropping it as a dependency will not affect anything.
- Remove some coveralls dependencies too
- unpin Kombu and Celery

~On hold pending a new Kombu release:~
~- https://github.com/celery/kombu/issues/1063~
~- This PR may fix our issues directly related to [COST-1821](https://issues.redhat.com/browse/COST-1821): https://github.com/celery/kombu/pull/1378~

~Once the above PR is released, we should try upgrading Kombu.~


koku/masu/test/database/helpers.py
```diff
@@ -141,105 +138,92 @@ def create_ocp_report_period(self, provider_uuid, period_date=None, cluster_id=N
...
             data["node"] = node
-        with OCPReportDBAccessor(self.schema) as accessor:
-            return accessor.create_db_object(table_name, data)
+        with schema_context(self.schema):
+            return baker.make(
+                model, report_period_id=report_period.id, report_id=report.id, **data, _fill_optional=True
+            )
 
-    def create_columns_for_table(self, table):
+    def create_columns_for_table_with_bakery(self, table, data={}):
```
maskarb, line 220: The whole reason I am updating this file is because of this `create_columns_for_table` function. I _think_ the update of `faker` was causing a lot of test failures (something related to Django and Decimals).

So I re-wrote this particular function and converted most functions in this file to use model_bakery. The bakery is a better solution for generating fake data for a Django model because it complete fills the model fields (e.g. a Charfield with a max length of 256 characters is going to get an entry with 256 characters.)

cgoodfred (approved):
LGTM! Great Work Buddy

PR #3106, opened 2021-09-01
replace prometheus multiprocess_mode `livesum` with `liveall`

maskarb:


maskarb:
Our alerts were fixed by updating the prom-rule to ignore the masu pod. We can close this PR and revisit later if we see weird results.

PR #3108, opened 2021-09-02
[COST-1725, COST-1795, COST-1664] Switch to warning log for common trino issues

adberglund:
## Summary
For some of the common, intermittent Trino issues, we can log a warning instead of an error.


Red-HAP (approved):
I think this is fine for the moment. What I'm worried about is potentially masking a larger issue that these exceptions could be a sign of. 

Can we agree to review warning logs in future to determine if they are occurring with alarming frequency?

PR #3110, opened 2021-09-02
Use the new AWSEnabledTagKeys model to enabled, all keys

Red-HAP:
## Jira Ticket

[COST-1800](https://issues.redhat.com/browse/COST-1800)

## Description

The old method was to use the QueryHandler class to get available and the model to get enabled.

I switched this to use the model to get available and enabled as they should all be in the enabled tag keys table.

This should pass smokes for tests this time. :crossed_fingers:


maskarb (approved):
lgtm!

PR #3113, opened 2021-09-02
fix order-by-date services test checks

Red-HAP:
Pretty much does what it says on the tin.

### DIFF
--- a/koku/api/report/test/aws/openshift/test_ocp_aws_query_handler.py
+++ b/koku/api/report/test/aws/openshift/test_ocp_aws_query_handler.py
@@ -6,6 +6,7 @@
 import copy
 import logging
 from datetime import timedelta
+from unittest import skip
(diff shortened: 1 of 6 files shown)


myersCody (dismissed):

maskarb:
I'm having a hard time understanding how this change actually fixes the test. We are taking data from the endpoint, and then comparing it to itself. Will this test ever actually assert anything?

Red-HAP:
> I'm having a hard time understanding how this change actually fixes the test. We are taking data from the endpoint, and then comparing it to itself. Will this test ever actually assert anything?


@maskarb 
The test will check the content of the lists only when both lists contain content. If one or the other is empty, the check is skipped.

maskarb (approved):

myersCody (approved):

PR #3117, opened 2021-09-07
Cost 1811 gcp topology table

adberglund:
## Summary
This establishes a topology table for GCP account, project, service, and region. With this we have a single look up table for our resource type APIs, and all fields for RBAC are available here.


adberglund:
@Fozzywozzybear opened a new PR with this content + some

PR #3124, opened 2021-09-10
move currencies from a json file into python code

maskarb:
This PR moves the currencies into a Python dictionary instead of a json file. This drops the dependency of a static file and makes the api a little simpler.

```
GET http://localhost:8000/api/cost-management/v1/currency/

{
meta: {
count: 15
},
links: {
first: "/api/cost-management/v1/currency/?limit=10&offset=0",
next: "/api/cost-management/v1/currency/?limit=10&offset=10",
previous: null,
last: "/api/cost-management/v1/currency/?limit=10&offset=5"
},
data: [
{
code: "AUD",
name: "Australian Dollar",
symbol: "AU$",
description: "AUD (AU$) - Australian Dollar"
},
```

Smoke test:
```
$ iqe tests plugin cost_management -k test_api_pre_tenant_create_check

25 passed, 6626 deselected, 25 warnings in 5.90s
```

### DIFF
--- a/koku/api/currency/currencies.py
+++ b/koku/api/currency/currencies.py
@@ -1,4 +1,11 @@
-[
+#
+# Copyright 2021 Red Hat Inc.
+# SPDX-License-Identifier: Apache-2.0
+#
+"""List of currencies."""
+# turn off black formatting
+# fmt: off
+CURRENCIES = [
     {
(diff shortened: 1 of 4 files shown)


abaiken (approved):
🥇

PR #3129, opened 2021-09-10
[COST-1629] Could not obtain last modified date for BigQuery table

cgoodfred:
After talking with Doug, this is an error that we wouldn't really be able to do much about since it would be a user side problem so we are just logging it as a warning and moving on. The processing still needs to exit smoothly so an error is still raised but handled differently than our broad exception check in `get_report_files` by logging it as a warning and moving on.


maskarb (approved):

maskarb:
Only 2 smokes failures, which are unrelated to this change. LGTM!

PR #3140, opened 2021-09-13
create `check-migrations` workflow

maskarb:
1. Update workflow names
2. add migrations check -> check if the migrations are in sync with the code defined models

Testing:
- [missing migration](https://github.com/project-koku/koku/runs/3593124558)
- [migration was created, but model updates are missing](https://github.com/project-koku/koku/runs/3593143416)


myersCody (approved):
LGTM!
This will be a great way to double check ourselves. The only thing that might slip by is the tables that have `managed=False`.

PR #3143, opened 2021-09-16
Cost 1811 gcp resource type migration

adberglund:
## Summary

This adds a new GCPTopology table for use in the resource type APIs.


maskarb (approved):
LGTM! Let's merge when unit tests pass.

PR #3150, opened 2021-09-21
Cost 1885 tags in csvs

adberglund:
## Summary
This will add an implicit group by for tags to our reports APIs when exporting CSV data. This will force tags to be included in the CSV. For now it is not optional.


lcouzens:
Smoke tests are updated

PR #3153, opened 2021-09-22
Cost 1486 fix partition creation

Red-HAP:
Like the other COST-1486, but adds support for non-trino (yikes) processing and fixes some bugs with the call to the partition handler.

This should be tested with and without Trino processing.

If everything is working, then non-default partitions should be created for `reporting_ocpallcostlineitem_daily_summary_p` and `reporting_ocpallcostlineitem_project_daily_summary_p` and the matviews should have data as well.

Just load ~2months of data and you should see results in those tables.

## Smokes Tests

[Output](https://ci.ext.devshift.net/job/project-koku-koku-pr-check/1622/console)

```
 1 failed, 5087 passed, 171 skipped, 862 deselected, 5 warnings in 2133.89s (0:35:33)
```

### DIFF
--- a/koku/reporting/migrations/0195_ocpall_add_source_fk.py
+++ b/koku/reporting/migrations/0195_ocpall_add_source_fk.py
@@ -6,7 +6,10 @@ from django.db import models
 
 class Migration(migrations.Migration):
 
-    dependencies = [("api", "0049_auto_20210818_2208"), ("reporting", "0193_gcptopology")]
+    dependencies = [
+        ("api", "0049_auto_20210818_2208"),
+        ("reporting", "0194_awscostentrylineitemdailysummary_savingsplan_effective_cost"),
+    ]
 
     operations = [
         migrations.AlterField(


koku/reporting/provider/all/openshift/models.py
```diff
@@ -577,7 +577,7 @@ class Meta:
     # It is used to divide cost evenly among projects
     shared_projects = models.IntegerField(null=False, default=1)
 
-    source_uuid = models.UUIDField(unique=False, null=True)
+    source_uuid = models.ForeignKey("api.Provider", on_delete=models.CASCADE, db_column="source_uuid", null=True)
```
adberglund, line 580: Why the foreign key here?

Red-HAP: @adberglund 
To get the table data to delete when the source is deleted. Else we run the risk of stale data inflating totals. (in testing, at least)

koku/masu/database/ocp_report_db_accessor.py
```diff
@@ -2282,6 +2283,13 @@ def _execute_processing_script(self, script_file_path, sql_params):
             if sql_stmt:
                 sql_stmt, params = self.jinja_sql.prepare_query(sql_stmt, sql_params)
                 with connection.cursor() as cur:
+                    try:
+                        LOG.debug(cur.mogrify(sql_stmt, params).decode("utf-8"))
+                    except ProgrammingError:
+                        LOG.error(f"STATEMENT: {sql_stmt}")
+                        LOG.error(f"PARAMS: {params}")
+                        LOG.error(f"INPUT_PARAMS: {sql_params}")
```
maskarb, line 2291: Let's consolidate these into a single Log.error with a meaningful error message. Otherwise we will get 3 different sentry alerts with unclear information.
```
LOG.error(f"something happened! STATEMENT: {sql_stmt} | PARAMS: {params} | INPUT_PARAMS: {sql_params}")
```

koku/reporting/provider/all/openshift/models.py
Red-HAP: Will link with bill id instead.

adberglund (approved):

adberglund (dismissed):

adberglund (approved):

PR #3154, opened 2021-09-22
Cot 1405 made celery task to call api

Fozzywozzybear:
-wip

### DIFF
--- a/koku/masu/celery/tasks.py
+++ b/koku/masu/celery/tasks.py
@@ -9,6 +9,7 @@ import os
 from datetime import datetime
 from datetime import timedelta
 
+import requests
 from botocore.exceptions import ClientError
 from celery.exceptions import MaxRetriesExceededError
 from django.conf import settings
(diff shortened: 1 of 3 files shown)


koku/api/currency/models.py
```diff
@@ -0,0 +1,45 @@
...
+        (HKD, "HKD"),
+        (JPY, "JPY"),
+        (NOK, "NOK"),
+        (NZD, "NZD"),
+        (SEK, "SEK"),
+        (SGD, "SGD"),
+        (USD, "USD"),
+        (ZAR, "ZAR"),
+    )
+    baseCurrency = models.CharField(max_length=100, choices=SUPPORTED_CURRENCIES, unique=True)
```
adberglund, line 44: Real quick, just for Python-y consistency, you can name these like `base_currency`. 

This table probably needs three columns. The base currency, the currency you are converting to, and the exchange rate.

koku/masu/celery/tasks.py
```diff
@@ -277,6 +279,79 @@ def clean_volume():
...
+    #     "GBP",
+    #     "HKD",
+    #     "JPY",
+    #     "NOK",
+    #     "NZD",
+    #     "SEK",
+    #     "SGD",
+    #     "ZAR",
+    # ]
+    endpoint = "http://api.exchangeratesapi.io/v1/latest?access_key=76b80adfa0400a63c7a8c8e5bc843b89"
```
boaz0, line 304: Should we store this access key as a secret?

koku/api/currency/models.py
```diff
@@ -0,0 +1,45 @@
...
+    GBP = "gbp"
+    HKD = "hkd"
+    JPY = "jpy"
+    NOK = "nok"
+    NZD = "nzd"
+    SEK = "sek"
+    SGD = "sgd"
+    USD = "usd"
+    ZAR = "zar"
+    SUPPORTED_CURRENCIES = (
```
boaz0, line 27: Maybe you can use https://github.com/project-koku/koku/blob/main/koku/api/currency/currencies.py#L8 and map them into tuples of tuples?

Something like this.

```
SUPPORTED_CURRENCIES = tuple([(curr.code.toLower(), curr.code)for curr in CURRENCIES])
```

koku/api/currency/models.py
Fozzywozzybear: Yep! These have been changed to the naming convention. There is now a third column for target_currency.

koku/masu/celery/tasks.py
Fozzywozzybear: This has been done, now imports as     endpoint = settings.CURRENCY_ENDPOINT

PR #3155, opened 2021-09-23
COST-1013: Add export time to manifest for GCP bigquery

myersCody:
## Description of Problem
I believe this is the source of the doubling problem we are seeing in our day to day workflow for GCP. GCP is different from the other providers we support because we have to create those csv files ourselves. The tricky part is that if we download at 9 am and more data comes in at 1 pm of the same day if we don't download that day again we will have gaps in our data. The initial design was to create a "rolling window" where we delete all the data in the line item table for the past 3 days and re-download those three days to ensure we have all the data.

However, I believe the heart of the doubling daily problem is that rolling window design. I believe the parquet files for the rolling window are being double causing the data in the gcp_line_items table to become doubled. We could try to figure out how to delete the duplicate parquet files; however, the approach in this pull request it to prevent the duplicate parquet files from being created. By adding export time to our bigquery we can ensure that we never pull data that we have already pulled. 

## How to Test
- Still trying to figure this part out to be honest. The day to day workflow is tricky because the `DATE_OVERRIDE=` development feature won't really help us here. Since we are pulling from the past we can never force a scenario where we only grab part of a day's data. In other words, we are pull at 9am, but the complete data is already available cause its from yesterday 😅 

## example of day to day workflow
```
koku-worker_1     |             FROM cloud-billing-292519.billing_datset.gcp_billing_export_v1_018984_D0AAA4_940B88
koku-worker_1     |             WHERE DATE(_PARTITIONTIME) >= '2021-09-18'
koku-worker_1     |             AND DATE(_PARTITIONTIME) < '2021-09-23'
koku-worker_1     |             AND export_time >= '2021-09-21 23:59:31.855000+00:00'
```


myersCody:
Smoke Test Results: https://ci.ext.devshift.net/job/project-koku-koku-pr-check/1492/
I still need to look at the results, but I am gonna look at them in the morning.

adberglund (approved):

PR #3159, opened 2021-09-27
add horizontalpodautoscaling for the worker pods

maskarb:
These changes will require adding `- HorizontalPodAutoscaler` to app-interface under [`managedResourceTypes`](https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/data/services/insights/hccm/deploy-clowder.yml#L22).

Autoscaling Rules:
1. max pods set to 5.
2. min pods (needs to be updated in app-interface) set to 1.
3. For all queues except download, add 1 pod every 5 minutes until the desiredReplaces is met (we don't want to scale up so fast that we overwhelm the db)
4. for download, scale up as fast as possible (we currently run 4 pods in Prod already and things seem happy, so scaling from 1 to 5 immediately shouldn't be a problem)
5. Use default scaleDown which scales to desiredReplicas immediately after 5 min stabilization window


maskarb:
We cannot implement HPAs on custom metrics. This feature was [removed in OCP 4.8](https://docs.openshift.com/container-platform/4.8/release_notes/ocp-4-8-release-notes.html#ocp-4-8-deprecated-removed-features).

PR #3160, opened 2021-09-28
disable re-summarization for all providers with a single api query

maskarb:
If all providers are summarized thru masu, it's possible to cause our summary queue to get backed up for a lengthy amount of time (e.g. for days). This PR removes that ability for environments where `DEVELOPMENT=False`.

[Passing smokes](https://ci.ext.devshift.net/job/project-koku-koku-pr-check/1613/)


myersCody (approved):
LGTM! 

This is small change that will have a lot of impact!

PR #3161, opened 2021-09-28
convert data to stringData in koku-aws secret so that we do not need to b64 encode the values

maskarb:
When deploying into the ephemeral cluster, this PR enables us to take our AWS credentials and put them directly into our bonfire configuration without needing to b64 encode the values first.

```
    AWS_ACCESS_KEY_ID_EPH: <access-key-id>
    AWS_SECRET_ACCESS_KEY_EPH:<access-secret-key>
```


abaiken (approved):
Tested 👍 Nice change haha

PR #3164, opened 2021-09-29
Add currency support in cost models

boaz0:
This adds the currency fields in the `/cost-models/` response.

For now, if currency is not given in POST/PATCH we will use the default currency (`USD` unless set differently in settings.KOKU_CURRENCY_DEFAULT)


adberglund (commented):
How does the default value get populated here? 

Also are we storing the currency value on the cost model model?

koku/cost_models/serializers.py
```diff
@@ -394,6 +395,12 @@ class Meta:
         choices=metric_constants.DISTRIBUTION_CHOICES, required=False, allow_blank=True
     )
 
+    currency = serializers.ChoiceField(
+        choices=tuple([(currency.get("code"), currency.get("code")) for currency in CURRENCIES]),
+        required=False,
+        allow_blank=True,
+    )
```
dlabrecq, line 402: Is this a list of currencies?

In the UI, we just need the API to return a single currency -- the currency used to create the cost model.

boaz0:
@adberglund yes we store the currency value in the cost model. I also fixed the populating default value here by checking what is the currency and if it's not set we get the default currency that was set in the api/settings

PR #3166, opened 2021-09-30
[Cost -1861] - adding a cost_type parameter to AWS report api

shannon-donahue:
Adding a cost type parameter to AWS report api to add in the capability of AWS amortized saving plan costs in the provider map. 

TO DO:
Adding Testing

This PR will be on hold till:
 COST 1802-1806 which adds the savings plan fields to the DB  is merged  
 potential COST-1850 is merged so we can incorporate a default value set from the user-settings table rather than a string


shannon-donahue:
Thank you for testing it out Ashley I appreciate it!

PR #3171, opened 2021-10-05
Fix SRE CI by adding GIT_COMMIT

boaz0:
The SRE CI job fails due to using previous git commit when updating a PR.
This causes problems because the git commit is used to get labels and according to them to run tests and build image.

To fix this I added `GIT_COMMIT` variable that is used in the `CI` script to fetch labels from Github.
I set the `GIT_COMMIT` to the current `HEAD` of a PR.


maskarb (approved):
lgmt!

boaz0:
Thanks @maskarb

PR #3172, opened 2021-10-05
fix currency issue not saving

ddonahue007:
closes: https://issues.redhat.com/browse/COST-1928

- update AWS tags only if they have changed
- added logging to any tag or currency changes
- fixed logic on calling handler functions

### DIFF
--- a/koku/api/settings/settings.py
+++ b/koku/api/settings/settings.py
@@ -242,7 +242,13 @@ class Settings:
                 raise ValidationError(error_obj(key, message))
 
             if "aws" in provider_name:
-                updated[ix] = update_enabled_keys(self.schema, enabled_tag_keys, enabled_tags_no_abbr)
+                existing_enabled_tags = []
+                for key in enabled_tag_keys.objects.all():
+                    if key.enabled:
+                        existing_enabled_tags.append(key.key)
+
+                if enabled_tags_no_abbr != existing_enabled_tags:
+                    updated[ix] = update_enabled_keys(self.schema, enabled_tag_keys, enabled_tags_no_abbr)
 
             else:
                 remove_tags = []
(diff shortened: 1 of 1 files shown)


koku/api/settings/settings.py
```diff
@@ -257,9 +263,12 @@ def _tag_key_handler(self, settings):
                             updated[ix] = True
 
                     for rm_tag in remove_tags:
+                        LOG.info(f"Updating " + provider_name + " tag '" + rm_tag.key + "': DISABLED")
```
dlabrecq, line 266: I understand wanting to log an exception, but don't know that we want to log each and every change? Did someone request we do that?

ddonahue007: None of these logs are going to show in the UI console, this is server side logging. It is extremely helpful to have logging information when API endpoints are called. We were already logging all this for AWS I just added logging so we are now tracking for the other providers.

dlabrecq: Perhaps if this worked off of a debug flag, but otherwise I have to disagree. 

We don't want to potentially fill the log with each and every change. Just moving items from one list to another will generate messages.

ddonahue007: These are only being logged during a **POST** call to the endpoint, not when just moving items around in the Dual list selector.  You have to hit the Save button and all changes that are made are recorded. This same logging has been happening for AWS I just made sure we are capturing the data for the other providers as well. When calling the end point is it is not a bad idea to log the the changes that are happening, this allows you to verify that state being changed is actually being saved properly in the backend(databases, etc...).

 I can certainly understand if you thought they were being logged during the process of just moving items from list to list in the component. This would just file the logs with erroneous information...

ddonahue007: This is true if the number of tags is a very large amount it would flood the logs. Will track total s instead of specifics

koku/api/settings/settings.py
```diff
@@ -242,7 +242,13 @@ def _tag_key_handler(self, settings):
                 raise ValidationError(error_obj(key, message))
 
             if "aws" in provider_name:
-                updated[ix] = update_enabled_keys(self.schema, enabled_tag_keys, enabled_tags_no_abbr)
+                existing_enabled_tags = []
+                for key in enabled_tag_keys.objects.all():
+                    if key.enabled:
+                        existing_enabled_tags.append(key.key)
```
maskarb, line 248: I worry about the performance of this if there are a lot of tags to deal with. We should be able to lean on Django more here to filter on the enabled tags instead of looping thru them.

This query should work here. It also makes the DB do all the work instead of iterating thru the list in Python.
```suggestion
                existing_enabled_tags = list(enabled_tag_keys.objects.filter(enabled=True).values_list('key', flat=True))

```

PR #3176, opened 2021-10-05
[COST-1802] update CJI name

abaiken:
If the CJI has a static name it will not be ran again so I need to bump the name to kick it off


myersCody (approved):
✊ 📈

maskarb (approved):

PR #3184, opened 2021-10-07
do not run _currency_handler when settings are None

ddonahue007:
closes: https://issues.redhat.com/browse/COST-1935

possible fix for the HTTP Status code 400

I am unable to get this failure to happen locally so, I am fishing here. I put in logic so we are no longer calling the handler for currency unless currency is the payload being posted.

### DIFF
--- a/koku/api/settings/settings.py
+++ b/koku/api/settings/settings.py
@@ -270,14 +270,18 @@ class Settings:
         return any(updated)
 
     def _currency_handler(self, settings):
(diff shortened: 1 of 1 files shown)


ddonahue007:
The fix for this was in the way tag name labels are being handled. 

https://github.com/project-koku/koku/pull/3185

This PR is still valid for the cleanup of how the handler is being called using DEVELOPMENT=True

PR #3208, opened 2021-10-18
[Cost-1811] BugFix returns empty queryset properly when user does not have access

Fozzywozzybear:
This PR fixes the issue where if a user had user access not related to the GCP resource endpoints, they could still access the endpoint. This would return all the data on that endpoint, this has now been fixed. The fix now has a proper check to set the queryset to empty when the user does not have access.

Testing
1. Ingest Data
2. Set RBAC access
3. Hit http://localhost:8000/api/cost-management/v1/resource-types/gcp-accounts/ and make sure that the data returned in null.

The change 

```
if gcp_project_access and gcp_project_access[0] != "*":
               query_holder = query_holder.filter(project_id__in=gcp_project_access)
if gcp_account_access and gcp_account_access[0] != "*":
               query_holder = query_holder.filter(account_id__in=gcp_account_access)
```
Over 
```
            if (
                gcp_account_access
                and gcp_account_access[0] == "*"
                or gcp_project_access
                and gcp_project_access[0] == "*"
            ):
                return super().list(request)
```
Because not filtering if the user has wildcard access is the name as returning the list unfiltered, this also will filter to the lowest level access.


koku/api/resource_types/test/gcp_accounts/tests_views.py
```diff
@@ -0,0 +1,116 @@
...
+                "US West",
+            ),
+        ]
+        mock_get_topo.return_value = mock_topo_record
+        dh = DateHelper()
+        start_date = dh.this_month_start
+        end_date = dh.this_month_end
+        self.accessor.populate_gcp_topology_information_tables(self.gcp_provider, start_date, end_date)
+        with schema_context(self.schema_name):
+            expected = GCPTopology.objects.annotate(**{"value": F("account_id")}).values("value").distinct().count()
```
maskarb, line 65: Based on the defined RBAC rules, is this really the correct expected value? Why are we not filtering `gcp-project` on `example-project-id`?

koku/api/resource_types/test/gcp_accounts/tests_views.py
```diff
@@ -0,0 +1,116 @@
...
+        # check that the expected is not zero
+        self.assertTrue(expected)
+        url = reverse("gcp-accounts")
+        response = self.client.get(url, **self.headers)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        json_result = response.json()
+        self.assertIsNotNone(json_result.get("data"))
+        self.assertIsInstance(json_result.get("data"), list)
+        self.assertNotEqual(len(json_result.get("data")), expected)
```
maskarb, line 74: The length of the data _should_ match the expected.

maskarb (changes requested):
These tests are close, but the `expected` query is not taking into account the defined RBAC rules.

koku/api/resource_types/test/gcp_accounts/tests_views.py
Fozzywozzybear: This has been updated to check the length in the same.

koku/api/resource_types/test/gcp_accounts/tests_views.py
Fozzywozzybear: This has been changed to filter on the lowest level of access.

koku/api/resource_types/test/gcp_regions/tests_views.py
```diff
@@ -0,0 +1,128 @@
...
+        dh = DateHelper()
+        start_date = dh.this_month_start
+        end_date = dh.this_month_end
+        self.accessor.populate_gcp_topology_information_tables(self.gcp_provider, start_date, end_date)
+        with schema_context(self.schema_name):
+            expected = (
+                GCPTopology.objects.annotate(**{"value": F("region")})
+                .values("value")
+                .distinct()
+                .filter(region="US East")
```
maskarb, line 69: Question: Why is this filter being applied? The endpoint code does not add a filter on `region`...

koku/api/resource_types/test/gcp_services/tests_views.py
```diff
@@ -0,0 +1,128 @@
...
+        dh = DateHelper()
+        start_date = dh.this_month_start
+        end_date = dh.this_month_end
+        self.accessor.populate_gcp_topology_information_tables(self.gcp_provider, start_date, end_date)
+        with schema_context(self.schema_name):
+            expected = (
+                GCPTopology.objects.annotate(**{"value": F("service_alias")})
+                .values("value")
+                .distinct()
+                .filter(service_alias="The Best Service")
```
maskarb, line 69: Question: Why is this filter being applied? The endpoint code does not add a filter on service_alias...

koku/api/resource_types/test/gcp_accounts/tests_views.py
```diff
@@ -0,0 +1,128 @@
...
+        dh = DateHelper()
+        start_date = dh.this_month_start
+        end_date = dh.this_month_end
+        self.accessor.populate_gcp_topology_information_tables(self.gcp_provider, start_date, end_date)
+        with schema_context(self.schema_name):
+            expected = (
+                GCPTopology.objects.annotate(**{"value": F("project_id")})
+                .values("value")
+                .distinct()
+                .filter(project_id="example-project-id")
```
maskarb, line 69: praise: very good!

maskarb (changes requested):
These tests are getting better! But some of them are still not quite right.

maskarb (approved):
👏🏻 👏🏻  lgtm!

PR #3217, opened 2021-10-20
	COST-1957	 Add masu endpoint for sources and providers

Fozzywozzybear:
As the developer I want to have access to a Masu endpoint that displays providers and source information, that is filterable by account id.
Steps to reproduce 
1. Spin up docker.
2. Make and load test customer data
3. Access the new endpoints at http://127.0.0.1:5000/api/cost-management/v1/providers/ and http://127.0.0.1:5000/api/cost-management/v1/sources/
4. The endpoint are filterable by adding the account id to the end of the endpoint. EX http://127.0.0.1:5000/api/cost-management/v1/providers/10001/ will return data that only belongs to the account 10001


lcouzens:
Minimum test run results ========= 29 passed, 6750 deselected, 2 warnings in 435.43s (0:07:15) ==========

koku/masu/api/sourceproviders/providers/views.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def get_providers_by_account_id(self, request, *args, **kwargs):
+        """Get Providers By Account Id"""
+        accountIdParam = kwargs
+        try:
+            queryset = self.queryset.filter(customer__account_id=accountIdParam["customer"])
+            if not queryset:
+                raise ProviderException("Invalid account id.")
+        except Exception:
+            raise ProviderException("Invalid arguments.")
```
maskarb, line 80: Question: Can you explain the purpose of this try/except?

In general, a broad `except Exception` is not good practice. What exceptions are we expecting to handle here?

koku/masu/api/sourceproviders/sources/views.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def get_sources_by_account_id(self, request, *args, **kwargs):
+        """Get source by schema"""
+        sourceuuidParam = kwargs
+        try:
+            queryset = self.queryset.filter(account_id=sourceuuidParam["account_id"])
+            if not queryset:
+                raise SourceException("Invalid account id")
+        except Exception:
+            raise SourceException("Invalid source uuid.")
```
maskarb, line 80: Question: same as above:
Can you explain the purpose of this try/except?

In general, a broad except Exception is not good practice. What exceptions are we expecting to handle here?

koku/masu/api/sourceproviders/sources/serializers.py
```diff
@@ -0,0 +1,31 @@
...
+
+    class Meta:
+        model = Sources
+
+    source_id = serializers.IntegerField()
+    source_uuid = serializers.UUIDField()
+    name = serializers.CharField()
+    auth_header = serializers.CharField()
+    offset = serializers.IntegerField()
+    account_id = serializers.IntegerField()
```
maskarb, line 22: In the Sources model, what is the type of `account_id`? Why is this field an `IntegerField`?

maskarb (changes requested):

koku/masu/api/sourceproviders/providers/views.py
Fozzywozzybear: We are trying to handle if the account ID is not in the queryset when we filter, should this be narrowed down or should a try and except be removed completely.

koku/masu/api/sourceproviders/sources/serializers.py
Fozzywozzybear: This should be a char field this has been changed.

koku/masu/api/sourceproviders/sources/views.py
Fozzywozzybear: This has also been removed from both providers and sources

adberglund:
Closing in favor of https://github.com/project-koku/koku/pull/3375

PR #3257, opened 2021-11-04
fix gcp-local ingest issue

maskarb:
This fixes a `not enough values to unpack (expected 2, got 1)` error during ingest for a GCP-Local provider type. I think we start with a file directory with, e.g.:
```
202110_ehocGhby_2021-10-01:2021-11-01.csv
202111_ehocGhby_2021-11-01:2021-11-04.csv
```
We start splitting these file names on the `_` into 3 pieces, the last of which is the date_range which contains a `:`.
We have a first report downloader that unpacks, say, November data into daily csvs, resulting in a file structure such as:
```
202110_ehocGhby_2021-10-01:2021-11-01.csv
202111_2021-11-01_3ee3375f5d8c1604b3db1a2531160e00.csv
202111_2021-11-02_3ee3375f5d8c1604b3db1a2531160e00.csv
202111_2021-11-03_3ee3375f5d8c1604b3db1a2531160e00.csv
202111_2021-11-04_3ee3375f5d8c1604b3db1a2531160e00.csv
202111_ehocGhby_2021-11-01:2021-11-04.csv
```

The next month's (Oct) downloader now iterates through all of these files, trying to split on the `_`. But now we have these new files which do not contain a `:` in the last part. We then try to split this string on a `:`, which the string does not contain, resulting in an error. This prevent ingest for the second months data.

Since the files we need to process will contain a date_range after the last `_`, we should only try to split strings that contain that `:`, otherwise it's not a file we are trying to process here.

### DIFF
--- a/koku/masu/external/downloader/gcp_local/gcp_local_report_downloader.py
+++ b/koku/masu/external/downloader/gcp_local/gcp_local_report_downloader.py
@@ -70,6 +70,9 @@ class GCPLocalReportDownloader(ReportDownloaderBase, DownloaderInterface):
                 if file.endswith(".csv"):
                     report_name = os.path.splitext(file)[0]
                     invoice_month, etag, date_range = report_name.split("_")
+                    if ":" not in date_range:
+                        # if date range does not contain the `:`, it is not a file to process
+                        continue
                     scan_start, scan_end = date_range.split(":")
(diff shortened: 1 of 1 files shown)


dlabrecq (approved):

myersCody (approved):

myersCody:
Thanks for catching this @maskarb!

PR #3259, opened 2021-11-04
COST-1981: Fix tag filter and update resource-types.

myersCody:
Fix a few things I missed/broke with the deprecation of `group_by[project]` in favor of `group_by[gcp_project]` for our gcp endpoints. 

Fixes:
- http://localhost:8000/api/cost-management/v1/tags/gcp/?filter[gcp_project]=example
```
Cannot resolve keyword 'project_ids' into field. Choices are: account_id, cost_entry_bill, cost_entry_bill_id, enabled, key, project_id, project_name, uuid, values
```

Adds:
- ~http://localhost:8000/api/cost-management/v1/resource-types/gcp-gcp_projects/~
- http://localhost:8000/api/cost-management/v1/resource-types/gcp-gcp-projects/


maskarb:
Question: what is the necessity of the new `resource-types/gcp-gcp_projects/` endpoint? Mixing dashes and underscores does not seem like a good idea.

myersCody:
> Question: what is the necessity of the new `resource-types/gcp-gcp_projects/` endpoint? Mixing dashes and underscores does not seem like a good idea.

@maskarb 

So, I was just trying to follow the naming convention already set in the resource-types which urls which is: 
```
/resource-types/{source_type}-{group_by}s/
```
I am personally okay with changing it, let me check with @dlabrecq to see if it messes up anything he may have templated for the UI.

maskarb (approved):

PR #3262, opened 2021-11-05
Adding enhanced labelling to pr_Check

lcouzens:
Hopefully this will give as some improved labelling methods for running pr_check jobs.

This adds 7 new labels plus a new _default_ bare minimal test suite.

List of labels and usage:
- `aws-smoke-tests` This will run all aws and ocp on aws tests
- `azure-smoke-tests` This will run all aws and ocp on azure tests
- `gcp-smoke-tests` This will run all gcp and ocp on gcp tests
- `cost-model-smoke-tests` This will run all cost-model tests
- `full-run-smoke-tests` This runs all smoke tests (basically what we used to have by default)
- `hot-fix-smoke-tests` This runs that absolute minimal tests to get an urgent bug-fix to production
- `smoke-tests` Original label, this will now run new cost_required marker, this is going to run the bare minimal amount of tests across the board that _should_ be run on all pr's as minimum.

One thing to note here is these labels are not really going to be stackable, we could change this in the future but it might be a pain to manage. That said this breakout _should_ cover most if not all cases. If you are changing something that touches multiple source types then there's a good chance it will affect all of them so full-test should be used. 

#### Note ####
I also had to update the unit_test.sh after some changes in bonfire that caught me out while testing this.

### DIFF
--- a/pr_check.sh
+++ b/pr_check.sh
@@ -14,7 +14,6 @@ ARTIFACTS_DIR="$WORKSPACE/artifacts"
 
 export IQE_PLUGINS="cost_management"
 export IQE_MARKER_EXPRESSION="cost_smoke"
-export IQE_FILTER_EXPRESSION="test_api"
 export IQE_CJI_TIMEOUT="90m"
 
 set -ex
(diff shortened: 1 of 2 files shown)


ddonahue007 (approved):
LGTM

abaiken (approved):
Great work Luke!

ddonahue007 (commented):
LGTM

PR #3267, opened 2021-11-09
[COST-2051] update make tasks

ddonahue007:
closes: https://issues.redhat.com/browse/COST-2051

update the 'docker-up-min' & 'docker-up-min-no-build' constant with each other

added two targets that mirror the tasks above and also start 'koku-listener' for running local Sources and Kafka testing.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -378,11 +378,20 @@ docker-up:
 docker-up-no-build: docker-up-db
 	$(DOCKER_COMPOSE) up -d --scale koku-worker=$(scale)
 
-docker-up-min:
+# basic dev environment targets
+docker-up-min:  docker-up-db
 	$(DOCKER_COMPOSE) build koku-base
-	$(DOCKER_COMPOSE) up -d --scale koku-worker=$(scale) db redis koku-server masu-server koku-worker
+	$(DOCKER_COMPOSE) up -d --scale koku-worker=$(scale) redis koku-server masu-server koku-worker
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -378,12 +378,20 @@ docker-up:
...
 	$(DOCKER_COMPOSE) build koku-base
-	$(DOCKER_COMPOSE) up -d --scale koku-worker=$(scale) db redis koku-server masu-server koku-worker
+	$(DOCKER_COMPOSE) up -d --scale koku-worker=$(scale) redis koku-server masu-server koku-worker
 
 docker-up-min-no-build: docker-up-db
-	$(DOCKER_COMPOSE) up -d --scale koku-worker=$(scale) redis koku-server masu-server koku-worker koku-listener
+	$(DOCKER_COMPOSE) up -d --scale koku-worker=$(scale) redis koku-server masu-server koku-worker
+
+# basic dev environment targets with koku-listener for local Sources Kafka testing
+docker-up-min-with-kafka: docker-up-min docker-up-db
```
maskarb, line 390: suggestion (blocking): let's change these to `with-listener` since these commands do not start kafka.

maskarb (changes requested):

Makefile
ddonahue007: Good point
Done

maskarb (approved):
Thanks for updating these! 👏🏻

PR #3272, opened 2021-11-10
Update Masu port

shannon-donahue:
In the new apple update for Monterey the apple control center now utilizes port 5000 so it was decided to move internal endpoints from port 5000 to 5042. This should update the endpoints to port 5042 as well as update the docs.

Testing Instructions:
```
1. Bring any currently running docker containers down
2. Pull and checkout branch
3. Bring docker containers back up 
4. test internal endpoints:
http://127.0.0.1:5000/api/cost-management/v1/download/ 
    -should no longer work
http://127.0.0.1:5042/api/cost-management/v1/download/
    -should be the new endpoint that works
```


adberglund (approved):

kholdaway:
👏

PR #3276, opened 2021-11-11
COST-2013: Migrations for gcp mat view conversion.

myersCody:


maskarb (approved):

koku/reporting/provider/gcp/models.py
```diff
@@ -949,3 +969,853 @@ class Meta:
...
+
+    class Meta:
+        """Meta for GCPComputeSummaryByServiceP."""
+
+        db_table = "reporting_gcp_compute_summary_by_service_p"
+        indexes = [
+            models.Index(fields=["usage_start"], name="gcpcompsumm_ser_usage_start"),
+            models.Index(fields=["instance_type"], name="gcpcompsumm_ser_insttyp"),
+            models.Index(fields=["invoice_month"], name="gcpcompsumm_ser_invmonth"),
+        ]
```
cgoodfred, line 1328: For compute summary by service, should we have an index on the `service_id`?

koku/reporting/provider/gcp/models.py
```diff
@@ -949,3 +969,853 @@ class Meta:
...
+    class Meta:
+        """Meta for GCPComputeSummaryByRegionP."""
+
+        db_table = "reporting_gcp_compute_summary_by_region_p"
+        indexes = [
+            models.Index(fields=["account_id"], name="gcpcompsumm_reg_account_id"),
+            models.Index(fields=["usage_start"], name="gcpcompsumm_reg_usage_start"),
+            models.Index(fields=["instance_type"], name="gcpcompsumm_reg_insttyp"),
+            models.Index(fields=["invoice_month"], name="gcpcompsumm_reg_invmonth"),
+        ]
```
cgoodfred, line 1434: For the ComputeServiceByRegion should we have an index on region?

koku/reporting/provider/gcp/models.py
```diff
@@ -949,3 +969,853 @@ class Meta:
...
+    class Meta:
+        """Meta for GCPStorageSummaryByRegionP."""
+
+        db_table = "reporting_gcp_storage_summary_by_region_p"
+        indexes = [
+            models.Index(fields=["usage_start"], name="gcpstorsumm_reg_usage_start"),
+            models.Index(fields=["instance_type"], name="gcpstorsumm_reg_insttyp"),
+            models.Index(fields=["account_id"], name="gcpstorsumm_reg_account_id"),
+            models.Index(fields=["invoice_month"], name="gcpstorsumm_reg_invmonth"),
+        ]
```
cgoodfred, line 1689: For the GCPStorageSummaryByRegion should we have an index on `region`?

koku/reporting/provider/gcp/models.py
myersCody: So, I agree with you, but I think we are gonna end up getting rid of these tables later on: https://issues.redhat.com/browse/COST-2106

koku/reporting/provider/gcp/models.py
myersCody: So, I agree with you, but I think we are gonna end up getting rid of these tables later on: https://issues.redhat.com/browse/COST-2106

koku/reporting/provider/gcp/models.py
myersCody: So, I agree with you, but I think we are gonna end up getting rid of these tables later on: https://issues.redhat.com/browse/COST-2106

PR #3277, opened 2021-11-11
[COST-2012] Azure CJI to copy data

cgoodfred:
This PR is to create the CJI to copy the azure data from the existing mat view to the new partitioned tables.

Testing: Deploy to ephemeral and post an azure source. Go to the db pod and check the azure matviews and partitioned tables, there should only be data in the matviews. Run the CJI in ephemeral and you should see it running in the cji pod logs. After go back to the db pod and check the partitioned tables and they should be filled with the data from the matviews now.


maskarb (approved):
lgtm!

PR #3281, opened 2021-11-12
Fixing QE nise initial ingest issue

lcouzens:
This is a problem for QE tests that upload nise data since it will always use todays date as the partition time.


myersCody (approved):
LGTM

PR #3282, opened 2021-11-12
[COST-2062] - Move savingsplan column in INSERT

adberglund:
## Summary
This should put the column order in the INSERT to match the column order in the SELECT. As a result we should be using the correct columns to populate the postgres tables.


cgoodfred (approved):
LGTM

PR #3287, opened 2021-11-15
OCP-on-All: drop NOT NULL constraint on currency-code

maskarb:


lcouzens (approved):
LGTM Looks like just what we needed to fix that OCP/Azure markup test

PR #3289, opened 2021-11-16
Cost 2063 standardized calc datatype

Red-HAP:
## Jira Ticket

[COST-2063](https://issues.redhat.com/browse/COST-2063)

## Description

Standardize numeric types on `max_digits=33` and `decimal_places=15`

This will affect all models with numeric fields not matching these parameters.

This also includes a new deploy YAML file that will run a new migration execution script. This script supports upgrade and downgrade. This new system is designed to be deployed via a new CJI stanza in app-interface.


deploy/koku-migrations-cji.yaml
```diff
@@ -0,0 +1,65 @@
...
+  metadata:
+    name: koku-migrations-cji
+  spec:
+    database:
+      sharedDbAppName: koku
+    dependencies:
+      - koku
+    envName: ${ENV_NAME}
+    jobs:
+    - name: script-runner-${IMAGE_TAG}
```
maskarb, line 18: This `job` name should be different than the `koku-cji` job.

Perhaps we can call this `migration-runner-${IMAGE_TAG}`?

deploy/koku-migrations-cji.yaml
```diff
@@ -0,0 +1,65 @@
...
+          value: ${PRESTO_PORT}
+        - name: PROMETHEUS_MULTIPROC_DIR
+          value: ${PROMETHEUS_DIR}
+        - name: CLOWDER_ENABLED
+          value: ${CLOWDER_ENABLED}
+        image: ${IMAGE}:${IMAGE_TAG}
+- apiVersion: cloud.redhat.com/v1alpha1
+  kind: ClowdJobInvocation
+  metadata:
+    name: koku-migrations-cji-v${IMAGE_TAG}
```
maskarb, line 37: Let's remove that `v` from this name.

deploy/koku-migrations-cji.yaml
```diff
@@ -0,0 +1,65 @@
...
+  spec:
+    appName: koku-migrations-cji
+    jobs:
+      - script-runner-${IMAGE_TAG}
```
maskarb, line 41: This will need to match the name of the `job` given above in the Clowdapp.

deploy/koku-migrations-cji.yaml
```diff
@@ -0,0 +1,65 @@
...
+    dependencies:
+      - koku
+    envName: ${ENV_NAME}
+    jobs:
+    - name: script-runner-${IMAGE_TAG}
+      podSpec:
+        args:
+        - /bin/bash
+        - -c
+        - $HOME/scripts/run_django_migrations.sh
```
maskarb, line 23: Our koku initContainer runs a different script:
https://github.com/project-koku/koku/blob/main/deploy/clowdapp.yaml#L115-L121

I think we should use one or the other in both places.

(We _do_ still need the initContainer in the koku clowdapp so that ephemeral env will run migrations.)

maskarb (changes requested):
Some initial comments:

PR #3290, opened 2021-11-16
do not update packages in Dockerfile

maskarb:
PR checks are failing because we cannot update our packages in the Dockerfile. This PR removes the update step so builds build again.


lcouzens (approved):
LGTM

PR #3304, opened 2021-11-19
Fix skipped unit tests

Fozzywozzybear:
This change removes the Skipped flag for the order by date tests on the cost reports


lcouzens (approved):
lgtm

PR #3307, opened 2021-11-26
re-pin celery to 4.4 and kombu to < 5.0

maskarb:
This PR reverts part of #3105. After the celery upgrade, the workers sometimes stop processing new messages. There is no obvious reason making troubleshooting next to impossible.

This PR re-pins celery and kombu to the versions we were on previously.


cgoodfred (approved):
Assuming the smokes run goes fine, 👍

PR #3311, opened 2021-11-29
adding-preferred-cost

shannon-donahue:
Adding the users set default cost-type value to the meta data of the cost-type endpoint

Testing Instructions:
1. pull branch
2. bring up docker and populate with test data
3. Hit the following endpoint, the response should contain "unblended_cost" in the meta data for cost-type
http://127.0.0.1:8000/api/cost-management/v1/cost-type/
```
"meta": {
        "count": 3,
        "cost-type": "unblended_cost"
    },
```
4. Front end Testing: Bring up local testing from front end repository
5. Load the settings page and select and cost type other than "unblended_cost", click apply changes
6. Now that a user defined cost_type has been defined Hit the following endpoint, and there should be the chosen cost-type set in the meta data
http://127.0.0.1:8000/api/cost-management/v1/cost-type/
```
"meta": {
        "count": 3,
        "cost-type": user defined choice from settings page
    },
```


abaiken (approved):
Great job 🎉

PR #3312, opened 2021-11-30
Change GCP error log to warning

lcouzens:
Fix GCP logging error which is constantly firing sentry alerts from QE tests.


abaiken (approved):
LGTM!

PR #3313, opened 2021-11-30
Adding account fields to compute summary table

lcouzens:
Fix for COST-1908


lcouzens:
This is not the right solution here :laughing:

PR #3321, opened 2021-12-02
add schema to partable sql

maskarb:
This PR adds `{{schema | sqlsafe}}` to the INSERT and SELECT statements for the Azure and GCP partable sql. We currently do this for AWS and OCP, so we should do the same thing for these providers.


cgoodfred (approved):
LGTM

PR #3322, opened 2021-12-02
[COST-2015] OCP on Azure partable processing

cgoodfred:
Switches azure processing to use the partitioned tables.


koku/reporting/migrations/0211_ocpaz_partables.py
```diff
@@ -0,0 +1,342 @@
+# Generated by Django 3.1.13 on 2021-12-01 13:00
```
Red-HAP, line 1: You don't need this file, do you? This should have been removed in favor of that other PR.
Or am I missing something?

cgoodfred: Yeah, the model changes/migration are included here just so the PR _can_ operate without the other one being merged in, once the migration PR is merged this won't show up as new to the PR anymore.

maskarb (approved):

PR #3326, opened 2021-12-03
[COST-2064] add instancename and change meterid to textfield

cgoodfred:
Our dev account was having some trouble with a report format, this allows the csv header `instancename` to be renamed to `instanceid` like we do with resourceid already. The meterid of `2core` was causing issues with the UUIDField so this changes it to a text field.

Smoke test that failed when running azure smokes: ```iqe_cost_management.tests.rest_api.v1.test_user.test_api_azure_settings_tag_enablement```


lcouzens (approved):
Lgtm

PR #3328, opened 2021-12-06
do _ocp_cloud_updater.update_summary_tables for cloud providers only

maskarb:
Change the cloud updater so that ocp-on-cloud daily summary tables are not updated when we receive reports from OCP clusters. 

These tables only need to be updated when the cloud provider sends us a new report, otherwise the OCP data is ahead of the cloud provider data.


adberglund (approved):

cgoodfred (approved):
LGTM

PR #3329, opened 2021-12-06
Worker cache timeout variable

adberglund:
## Summary
Allow the timeout on the worker cache to be variable.


ddonahue007 (approved):
LGTM

maskarb (approved):
Looks good to me. I checked the worker pods during the smoke run, and all the pods that should have this env variable do have it!

PR #3334, opened 2021-12-07
update positions of unit tests for resource types

Fozzywozzybear:
WIP


koku/api/resource_types/test/Aws/aws_accounts/tests_views.py
```diff
@@ -0,0 +1,47 @@
+#
```
maskarb, line 1: The structure of the test directory should match the structure of the `/resources_types` directory. We don't have provider specific folders (i.e. Aws, Azure, Gcp) in the `/resources_types` dir, so we shouldn't have them here either.

Fozzywozzybear: This has been fixed!

maskarb:
We can do this a different time.

PR #3343, opened 2021-12-13
Cost 2131

ddonahue007:
fix to scheduler on HCS pod by giving start_date default value
only bring up one worker
changed schedule on task so it is not running once a minute


abaiken (approved):
LGTM!

PR #3344, opened 2021-12-14
add HCS queue to prometheus

ddonahue007:
Closes: https://issues.redhat.com/browse/COST-2132

In the original PR I missed a commit that made sure the queue was added to prometheus.

```
I tried it in stage env and the response for the "api/cost-management/v1/celery_queue_lengths/" endpoint request doesn't contain the "hcs" item

{
    "download": 0,
    "summary": 0,
    "priority": 0,
    "refresh": 0,
    "cost_model": 0,
    "celery": 0,
    "ocp": 0
} 
```
This is fixed in this PR

```
{
    "download": 0,
    "summary": 0,
    "priority": 0,
    "refresh": 0,
    "cost_model": 0,
    "celery": 0,
    "ocp": 0,
    "hcs": 0
}
```


abaiken (approved):
Looks good! We will also want to update our app interface alert rules to alert on this queue at some point!

PR #3351, opened 2022-01-04
[COST-2178] - minimal docker image

maskarb:
Switch to `ub8/ubi-minimal` base image to...🤞🏻 ... hopefully reduce the number of CVEs for packages installed in the docker image.
This PR does:
1. Update the Dockerfile to change our base image to `ubi8/ubi-minimal`
2. Simplify the status endpoints to return a `GIT_COMMIT` that is injected into the build process through a build-arg.
3. Update all the clowdapp commands

Testing:
- Ensure the application spins up like normal, i.e. `make docker-up`


koku/koku/settings.py
```diff
@@ -34,6 +34,8 @@
 # Build paths inside the project like this: os.path.join(BASE_DIR, ...)
 BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 
+# Version info
+GIT_COMMIT = ENVIRONMENT.get_value("GIT_COMMIT", default="local-dev")
```
maskarb, line 38: This environment variable is set during Docker build in `build_deploy.sh`. The status endpoints were simplified to use this `settings.GIT_COMMIT` instead of using a subprocess to run `git rev-parse --short HEAD`. Doing this allows us to avoid copying the `.git` folder into our final image build.

Dockerfile
```diff
@@ -28,33 +36,64 @@ LABEL summary="$SUMMARY" \
...
 
+# create the koku user
+RUN \
+    adduser koku -u ${USER_ID} -g 0 && \
+    chmod ug+rw ${HOME} ${HOME}/koku ${HOME}/koku/static /tmp
 USER koku
 
-RUN umask u=rwx,g=rwx,o=rx
+# create the static files
+RUN python koku/manage.py collectstatic --noinput
```
Red-HAP, line 89: If there's a chance that an existing `app.log` will be copied into the image via line 80, I suggest this:
```suggestion
RUN rm -f ${HOME}/koku/app.log && python koku/manage.py collectstatic --noinput && rm -f ${HOME}/koku/app.log
```

Dockerfile
```diff
@@ -28,33 +36,64 @@ LABEL summary="$SUMMARY" \
...
 
-EXPOSE 8080
+# This `app.log` file is created during the `collectstatic` step. We need to
+# remove it else the random OCP user will not be able to access it. This file
+# will be recreated by the Pod when the application starts.
+RUN rm ${HOME}/koku/app.log
```
Red-HAP, line 94: This would be no longer needed if you do the above.
```suggestion
```

Dockerfile
```diff
@@ -28,33 +36,64 @@ LABEL summary="$SUMMARY" \
...
+# shadow-utils to make useradd available
+RUN INSTALL_PKGS="python38 python38-devel glibc-langpack-en gcc shadow-utils" && \
+    microdnf --nodocs -y upgrade && \
+    microdnf -y --setopt=tsflags=nodocs --setopt=install_weak_deps=0 install $INSTALL_PKGS && \
+    rpm -V $INSTALL_PKGS && \
+    microdnf -y clean all --enablerepo='*'
 
-# Get latest packages
-# RUN dnf -y upgrade
+WORKDIR ${HOME}
```
Red-HAP, line 51: Are we breaking with convention here or is `/opt/app-root` no longer relevant?

docker-compose.yml
```diff
@@ -58,12 +57,13 @@ services:
       - UNLEASH_PORT=${UNLEASH_PORT-4242}
       - UNLEASH_LOG_LEVEL=${UNLEASH_LOG_LEVEL-WARNING}
       - CURRENCY_URL=${CURRENCY_URL}
-    privileged: true
     ports:
       - 8000:8000
       - 8001:9000
     volumes:
-      - '.:/koku/'
+      - './koku:/koku/koku'
```
Red-HAP, line 64: Wouldn't this make `${KOKU_PATH}/scripts` unavailable? I feel like I'm missing something here.

Red-HAP (changes requested):

Dockerfile
maskarb: I'll add a .dockerignore entry for `.log` files.

Dockerfile
maskarb: Honestly, I'm not sure what convention is. `/opt/app-root` seemed more related to the old s2i images, but I could be wrong.

docker-compose.yml
maskarb: Do the docker-compose services need access to the scripts dir?

Red-HAP: If the scripts dir is unavailable, how is it executing the `run_migrations.sh` script?

Red-HAP: Well, it would be an issue if you were making changes to the run_migrations.sh script, that is. You'd have to rebuild the base image every time.

maskarb: I can add it back, but locally we don't rely on the `run_migrations.sh`. The koku-server is started via `run_server.sh` which runs the migrations for us.

Dockerfile
maskarb: I'll move things into `/opt/koku`.

Red-HAP (approved):

Red-HAP (approved):

PR #3353, opened 2022-01-04
update sql token parser

Red-HAP:
This updates the tokens parsed from `sqlparse.parse()` to better get tables involved from more complex queries that are frequently used in the application.

## Testing

From `${KOKU_PATH}`:

1. `cd koku`
2. `python ./manage.py shell`
3. `from koku.database_exc import *`
4. `from psycopg2.errors import DivisionByZero`
5. `exc = ExtendedDBException(DivisionByZero("Oh Noes!"))`
6. `exc.query_tables = []`
7. `tokens = sql_parse("select * from table_a a join table_b b on b.a_id = a.id;")`
8. `exc.get_tables_from_tokens(tokens, len(tokens))`
9. `exc.query_tables == ["table_a", "table_b"]`
10. ctrl+d

The `tokens` variable will be a list of token objects as returned by sql_parse. So you can try it with a zero-length list if you'd like.

### DIFF
--- a/koku/koku/database_exc.py
+++ b/koku/koku/database_exc.py
@@ -10,13 +10,19 @@ import os
 import re
 import traceback
 
-import psycopg2
 from django.db.utils import DatabaseError as DJDatabaseError
 from psycopg2.errors import DatabaseError
 from psycopg2.errors import DeadlockDetected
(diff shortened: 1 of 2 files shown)


maskarb (approved):
lgtm!

PR #3356, opened 2022-01-04
Update check_migration return value processing

Red-HAP:
## Jira Ticket

[COST-2182](https://issues.redhat.com/browse/COST-2182)

## Description

The old way of checking to see if migrations were up-to-date or not was a loop like this:

```python
while not check_migrations():
    sleep(5)
```

This does not work properly now that `check_migrations()` is no longer a function that only returns a bool. Since the `True` value is the only value indicating that migrations are up-to-date, the loops have been changed to:

```python
while check_migrations() != True:
    sleep(5)
```


maskarb (approved):
lgtm!

PR #3358, opened 2022-01-05
Localdev base image command

Red-HAP:
Local koku-base image changed to have a benign command that exits successfully. The base image should not run in local dev.


docker-compose.yml
```diff
@@ -8,6 +8,7 @@ services:
       args:
         GIT_COMMIT: ${GIT_COMMIT-local-dev}
         PIPENV_DEV: "True"
+    command: "echo 'Base Image Should Not Run'"
```
maskarb, line 11: Is this command really necessary? For me, this image never runs locally.

maskarb (commented):
A question:

docker-compose.yml
Red-HAP: On fedora, it stays running as a second koku-server.

myersCody (approved):
LGTM!

PR #3362, opened 2022-01-05
Remove zero padding from this month reference

adberglund:
## Summary
The managed tables that Hive/Trino create don't use zero padded values for month/day and we assumed they did which means we would never find data for any month/day from 1-9. This replace call will strip the padded zero from month. The replace(ltrim(replace())) will put back a zero on the end or middle of a string so we can still have 10 for October.

## Testing

[Smoke Tests](https://ci.ext.devshift.net/job/project-koku-koku-pr-check/3240/)

1. Run load test customer data on main with trino and see no data in postgres for OCP or OCP on AWS.
2. Switch to this branch 
3. Resummarize AWS + OCP source using masu endpoint. 
4. Enjoy data.


cgoodfred (approved):
LGTM

PR #3363, opened 2022-01-05
Using strpos seems to speed up query

adberglund:
## Summary
Using regular string matching instead of the JSON match functions seems to improve this join significantly. It still seems a little slow, though.

## Testing

Ran this query and the prior against larger dataset and saw significant speed up. 

[Smoke Tests](https://ci.ext.devshift.net/job/project-koku-koku-pr-check/3200/)


ddonahue007 (approved):
LGTM

PR #3373, opened 2022-01-06
Add foreign key to provider in api_sources table

adberglund:
## Summary
This adds a foreign key on the api_sources table to directly link to the provider table.


adberglund:
This does not play nice with sources deletes and we can get around needing it.

PR #3382, opened 2022-01-10
User settings endpoints

shannon-donahue:
Additional settings endpoints that will return the users preferred settings in a simple format without data driven forms

Testing Instructions:
1. pull and checkout branch
2. Test the following 3 endpoints:

Users settings endpoint which returns all user preferred settings without tags
```
http://127.0.0.1:8000/api/cost-management/v1/account-settings/

Feedback should look as follows:
"meta": {
        "count": 1
    },
    "links": {
        "first": "/api/cost-management/v1/user-settings/?limit=10&offset=0",
        "next": null,
        "previous": null,
        "last": "/api/cost-management/v1/user-settings/?limit=10&offset=0"
    },
    "data": {
            "currency": "GBP",
            "cost_type": "savingsplan_effective_cost"
    }
```

user settings cost-type endpoint which returns the users preferred cost-type only

```
http://127.0.0.1:8000/api/cost-management/v1/account-settings/cost-type/

Feedback should look as follows:
"meta": {
        "count": 1
    },
    "links": {
        "first": "/api/cost-management/v1/user-settings/cost-type/?limit=10&offset=0",
        "next": null,
        "previous": null,
        "last": "/api/cost-management/v1/user-settings/cost-type/?limit=10&offset=0"
    },
    "data": {
        "cost_type": "savingsplan_effective_cost"
    }

```

user settings currency endpoint which returns the users preferred currency only

```
http://127.0.0.1:8000/api/cost-management/v1/account-settings/currency/

Feedback should look as follows:
"meta": {
        "count": 1
    },
    "links": {
        "first": "/api/cost-management/v1/user-settings/currency/?limit=10&offset=0",
        "next": null,
        "previous": null,
        "last": "/api/cost-management/v1/user-settings/currency/?limit=10&offset=0"
    },
    "data": {
            "currency": "GBP"
    }
```


koku/api/utils.py
```diff
@@ -35,6 +37,29 @@ def get_cost_type(request):
...
+
+def get_account_settings(request):
+    """Returns users settings from the schema or the default settings"""
+    with schema_context(request.user.customer.schema_name):
+        query_settings = UserSettings.objects.all().first()
+    if not query_settings:
+        user_settings = USER_SETTINGS
+    else:
+        user_settings = query_settings
+    return user_settings
```
abaiken, line 60: To make this a tiny bit cleaner you can do the following: 
```suggestion
   if query_settings:
        return query_settings
    return USER_SETTINGS
```

koku/api/utils.py
```diff
@@ -35,6 +37,29 @@ def get_cost_type(request):
...
+def get_currency(request):
+    """get currency from the DB user settings table or sets currency to default if table is empty."""
+
+    with schema_context(request.user.customer.schema_name):
+        query_settings = UserSettings.objects.all().first()
+        if not query_settings:
+            currency = KOKU_DEFAULT_CURRENCY
+        else:
+            currency = query_settings.settings["currency"]
+    return currency
```
abaiken, line 49:
```suggestion
        currency = KOKU_DEFAULT_CURRENCY
        if query_settings:
            currency = query_settings.settings["currency"]
    return currency
```

koku/api/user_settings/views.py
```diff
@@ -4,13 +4,65 @@
...
+
+    permission_classes = [permissions.AllowAny]
+
+    def get(self, request, *args, **kwargs):
+        """Gets a list of users current settings."""
+        param = kwargs
+        if not param:
+            user_settings = get_account_settings(request)
+        else:
+            setting = param["setting"]
```
abaiken, line 44:
```suggestion
        if not kwargs:
            user_settings = get_account_settings(request)
        else:
            setting = kwargs["setting"]
```

docs/source/specs/openapi.json
```diff
@@ -3018,6 +3018,79 @@
                 }
             }
         },
+        "/user-settings/": {
```
abaiken, line 3021: We need to update this name to account-settings for swagger

abaiken (commented):
This looks amazing!! I left one comment (the name change for swagger) the other comments are suggestions that you can decide whether or not to implement because what you have works. If we could get the count in the pagination to be right, that would be amazing but I know its tricky since its a settings object. Great work 🎉

shannon-donahue:
> This looks amazing!! I left one comment (the name change for swagger) the other comments are suggestions that you can decide whether or not to implement because what you have works. If we could get the count in the pagination to be right, that would be amazing but I know its tricky since its a settings object. Great work 🎉

shannon-donahue:
Thank you Ashley for the feedback! I addressed each of those comments definitely some great spots to make the code cleaner and easier to read

abaiken (approved):
🎉 🥇

PR #3387, opened 2022-01-10
Copy ocp-on-aws, ocp-on-azure data into partitioned tables.

Red-HAP:
## Description

This is a multi-process job that will fire off 4 worker processes. Each process will connect to the database read from a queue to get the schema and table information to perform the copy. So there should be up to 4 tables from up to 2 schemata being copied concurrently.

## Testing

The script can be run directly as a test. 
1. Re-init your database making sure that `reporting_ocpawscostentrylineitem_*daily_summary_p` tables have been created
2. Load some customer data that will populate those base tables (same table names, but not ending in `_p`)
3. Execute `./scripts/copy_ocp_aws_azure_data.py <num_processes>` (I'd suggest at least `2` for this value on localdev)
4. You will see logs for the main process as well as each worker created. These logs will show the tables being copied and the number of records copied once complete.
5. Once all worker processes have completed, the main process will exit.


scripts/copy_ocp_aws_azure_data.py
```diff
@@ -0,0 +1,359 @@
...
+        # Combine for the individual job work
+        target_tables = [{"schema": schema, "table_info": table_info} for schema in schemata for table_info in tables]
+        t_tot = len(target_tables)
+        del schemata
+        del tables
+
+        # start worker processes
+        max_workers = int(sys.argv[1]) if len(sys.argv) > 1 else int(os.environ.get("NUM_JOB_WORKERS", 1))
+        if max_workers > t_tot:
+            max_workers = t_tot
```
maskarb, line 325: Question: Do we _need_ to run this script with multiple workers?

Red-HAP: Why not? It's an attempt to make such a potentially long job faster. Each table copy is discrete. Everything is logged by the process.

maskarb:
If this script is supposed to run via the CJI, then the `.dockerignore` still needs to be updated to ensure this file lands in the final build.

cgoodfred (approved):
Ran in ephemeral and it went well, 👍  LGTM

maskarb (changes requested):
Let's hold off on merging until we can get https://github.com/project-koku/koku/pull/3427 in.

cgoodfred (approved):

PR #3389, opened 2022-01-11
[COST-2190] Upgrade Django to 3.2

maskarb:
Fixes CVE-2021-45115, CVE-2021-45116, CVE-2021-45452.

* [COST-2190](https://issues.redhat.com/browse/COST-2190)
* [COST-2227](https://issues.redhat.com/browse/COST-2227)
* [COST-2228](https://issues.redhat.com/browse/COST-2228)
* [COST-2229](https://issues.redhat.com/browse/COST-2229)


koku/koku/settings.py
```diff
@@ -291,6 +291,8 @@
 
 INTERNAL_IPS = ["127.0.0.1"]
 
+DEFAULT_AUTO_FIELD = "django.db.models.AutoField"
```
maskarb, line 294: For discussion of this setting, see: https://docs.djangoproject.com/en/4.0/releases/3.2/#customizing-type-of-auto-created-primary-keys

I tried setting this to `BigAutoField` for the reporting app, but the migration didn't go well, so I think we're better off sticking with `AutoField` everywhere.

koku/reporting/migrations/0214_auto_20220112_0209.py
```diff
@@ -0,0 +1,135 @@
...
+from django.db import migrations
+from django.db import models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [("reporting", "0213_delete_mat_views")]
+
+    operations = [
+        migrations.AlterField(
```
Red-HAP, line 11: Getting rid of all of the TextFields was a requirement for DJ3.2?

maskarb: I started typing a big response here, but I just thought of something. Lemme try this without a migration. Originally, I couldn't figure out a way, but I might know one now.

maskarb: As to why I did change this field, I kept seeing these errors on Azure tests:
```django.core.exceptions.FieldError: Expression contains mixed types: TextField, CharField. You must set output_field.```

This error is the result of this ([copied from the 3.2 release notes](https://docs.djangoproject.com/en/4.0/releases/3.2/#models)):

> Value() expression now automatically resolves its output_field to the appropriate Field subclass based on the type of its provided value for bool, bytes, float, int, str, datetime.date, datetime.datetime, datetime.time, datetime.timedelta, decimal.Decimal, and uuid.UUID instances. As a consequence, resolving an output_field for database functions and combined expressions may now crash with mixed types when using Value(). You will need to explicitly set the output_field in such cases.

maskarb: I think I figured out how to fix everything by updating the provider_maps instead of requiring a migration. We'll see how the testing goes...

adberglund (approved):
😑 🍬

PR #3393, opened 2022-01-11
ATTEMPT: 002 -- Cost 2186 partition ocp on aws azure summary tables

Red-HAP:
Create partitioned tables for ocpaws, ocpazure daily summary and project daily summary tables.

This attempt changes all of the index names so there should be no index object collision.


maskarb (approved):
lgtm!

PR #3395, opened 2022-01-12
[COST-1026] small migration for ocp on gcp partables

cgoodfred:
After we got along more with processing work, we needed one other small migration for the partables, this splits it out into its own PR.


myersCody (approved):
LGTM

PR #3397, opened 2022-01-13
[COST-2199] Hive partition delete retry

adberglund:
## Summary 
This wraps our partition delete SQL for Trino/Hive in a for loop with 5 retries for each partition day. This should hopefully help with the common but intermittent deletion exception being raised.


maskarb (approved):
lgtm!

PR #3398, opened 2022-01-13
Tweaking provider map to include volume labels for costs group_by tag

lcouzens:
When doing group_by[tag:volume]=* on the openshift/costs endpoint any data with a volume tag would never be returned, all you would see is tag no-volume entries. This change adds volume labels with pod labels to the costs provider_map.


lcouzens:
Last smoke run had a single failure which looked unrelated to these changes. Also tried it locally and it passed fine.

lcouzens:
This wont actually achieve what we need, going to have to rethink this and come up with something new.

PR #3403, opened 2022-01-17
Adding Trino flag to status endpoint

lcouzens:
Adding a flag for QE to tell if Trino is enabled by default or not.


abaiken (approved):
LGTM

PR #3404, opened 2022-01-17
Cost 2209 query tuning 1

Red-HAP:
## Jira Ticket

[COST-2209](https://issues.redhat.com/browse/COST-2209)

## Description

Change the AWS, GCP, OCP, OCP-on-All Provider Maps to use the `usage_start` column for the `end_date` value when building queries. This is to utilize the usage_start column only for date range filtering since it is indexed.

## Testing

[Smoke Tests](https://ci.ext.devshift.net/job/project-koku-koku-pr-check/3562/)


maskarb:
Question: Do we need to update the ocp-on-aws and ocp-on-azure provider maps as well?

koku/api/report/azure/openshift/provider_map.py
```diff
@@ -38,7 +38,7 @@ def __init__(self, provider, report_type):
                 "provider": Provider.OCP_AZURE,
                 "alias": "subscription_guid",
                 "annotations": {"cluster": "cluster_id"},
-                "end_date": "costentrybill__billing_period_end",
+                "end_date": "costentrybill__billing_period_start",
```
maskarb, line 41: I don't quite understand why we aren't changing this one to `usage_start` as well. From what I understand, we are changing these to fields on which we index, but this is not an indexed field...is it?

koku/api/report/azure/provider_map.py
```diff
@@ -36,7 +36,7 @@ def __init__(self, provider, report_type):
                 "provider": Provider.PROVIDER_AZURE,
                 "alias": "subscription_guid",  # FIXME: probably wrong
                 "annotations": {},
-                "end_date": "costentrybill__billing_period_end",
+                "end_date": "costentrybill__billing_period_start",
```
maskarb, line 39: Same question as above.

maskarb (commented):
A question:

koku/api/report/azure/openshift/provider_map.py
Red-HAP: Yes. It is part of a multi-column index. It is the first entry in the index tuple, so I think this will work.

koku/api/report/azure/provider_map.py
Red-HAP: Same answer as above

adberglund:
It's not abundently clear to me (yet, still digging) what the end_date part of the provider map actually does. 

Looking at https://github.com/project-koku/koku/blob/d184151004d2fd5396f14e62efcf2fef63e1772a/koku/api/query_handler.py#L276

that seems to be the thing driving using `usage_end`

Red-HAP:
> It's not abundantly clear to me (yet, still digging) what the end_date part of the provider map actually does.
> 
> Looking at
> 
> https://github.com/project-koku/koku/blob/d184151004d2fd5396f14e62efcf2fef63e1772a/koku/api/query_handler.py#L276
> 
> that seems to be the thing driving using `usage_end`

Will look into it.

Red-HAP:
@adberglund 
I have updated the other relevant references from usage_end to usage_start for query filters. Please re-review.
Thanks for the eagle-eyes!!

adberglund:
I think the new changes are what we want. Since the ocp report db accessor is now being modified, I am running the OCP smoke tests now. I'm also pretty sure the "end_date" in the provider map does exactly nothing. We could probably just remove that line entirely.

adberglund (approved):

PR #3406, opened 2022-01-18
Draft: Adding effective cost to cost model metrics

lcouzens:
Initial work to add Effective cost to the list of metrics for creating cost models.

Testing
- Check new metrics are visible at http://localhost:8000/api/cost-management/v1/metrics/
- Create a new cost model with effective cost metrics 

NOTE:
This only adds the metrics to cost models it does not have any backend cost calculations.


lcouzens:
Just going to close this in favour of https://github.com/project-koku/koku/pull/3409

PR #3409, opened 2022-01-20
Cost 2216/2217/2218 - Adding affective cost metric to cost model calculations

lcouzens:
This PR add's the ability to store and calculate based from the new effective cost metric.

Testing:
- Check new metrics are visible at http://localhost:8000/api/cost-management/v1/metrics/
- Add ocp source/data
- Create a cost model using effective cost metrics
- Verify calculations are correctly applied to source data


adberglund (commented):
I thought a little more about this, and at this point we're already at daily data, and we might want to catch the `greatest()` at the hourly level, so let's talk next week about getting it in the summary sql maybe.

lcouzens:
Requires 3412

lcouzens:
Smoke run failures which are fully expected until the QE MR's are merged.
test_api_ocp_metrics 
test_api_cost_model_ocp_crud

ddonahue007 (dismissed):
lgtm

adberglund (approved):

PR #3414, opened 2022-01-24
Pin unleash version < 5

maskarb:
The unleash client is currently unable to reach the server. Pinning to an old version fixes the issue.

Changes proposed in this PR:
* Pin Unleash client < 5

Testing Instructions:
1. deploy with bonfire, and do _not_ see `Unable to get feature flag toggles, using cached provisioning`


abaiken (approved):
LGTM!

PR #3419, opened 2022-01-25
fix 500 in resource type endpoints

shannon-donahue:
Fixes [COST-2018](https://issues.redhat.com/browse/COST-2018)

Changes proposed in this PR:
* Updated search field in resource-type endpoints that used iregex to not but rather use our regular filter method
* This resolves the 500 error of * not being recognized as a regular expression

Testing Instructions:
1. Pull branch
2. bring up docker
3. create and load test costumer data
3. Expect to see this result:
  ```
GET http://127.0.0.1:8000/api/cost-management/v1/resource-types/openshift-projects/?search=*


   {
    "meta": {
        "count": 0
    },
    "links": {
        "first": "/api/cost-management/v1/resource-types/openshift-projects/?limit=10&offset=0&search=%2A",
        "next": null,
        "previous": null,
        "last": "/api/cost-management/v1/resource-types/openshift-projects/?limit=10&offset=0&search=%2A"
    },
    "data": []
}
```
4. Additionally a few other resource endpoints that have been updated and should get the same results
```
GET http://127.0.0.1:8000/api/cost-management/v1/resource-types/openshift-clusters/?search=*
GET http://127.0.0.1:8000/api/cost-management/v1/resource-types/azure-services/?search=*
GET http://127.0.0.1:8000/api/cost-management/v1/resource-types/gcp-projects/?search=*
```
5. You should still also be able to use the same link and filter by strings contained in the "values" 

---
Additional Context:


lcouzens:
One note here, we have resource type endpoint tests not sure hot-fix covers those. I'm sure this is fine but for future we should really only be using hot-fix for hot-fixes :smile: We should really be using smoke-tests.

abaiken (approved):

PR #3422, opened 2022-01-26
[COST-2249] break gcp generate daily data out from ocpgcp PR

cgoodfred:
Fixes [COST-2249](https://issues.redhat.com/browse/COST-2249)

Changes proposed in this PR:
* This breaks out the daily data generator from OCP on GCP since it was causing trouble when we sent ocp on gcp to prod. The issue we were seeing was actually caused by an empty data frame coming through and trying to be processed and the group by on an empty dataframe was causing pandas to throw an error. We now handle the empty dataframe scenario as well. This is only reproducible using the bigquery processing flow since we can get an empty csv from bigquery.

Testing Instructions (there are two things to test here, one is gcp daily parquet functionality, another is a bugfix that affected the gcp daily parquet functionality):

Locally this wasn't an issue, but through bigquery we occasionally process an empty dataframe, this is covered in the gcp smokes tests in ephemeral. Specifically the `test_api_gcp_source_raw_calc` test.

1. Run gcp smokes in ephemeral and they should all pass (or in the PR check with the gcp-smoke-tests label), `iqe tests plugin cost_management -k test_api_gcp`
2. You can run them locally as well. The suggested tests to run are the tests related to gcp processing and the command locally would be: `ENV_FOR_DYNACONF=local iqe tests plugin cost_management -k test_api_gcp`
3. After the tests finish running, you should see them all pass and no processing errors.

THIS IS A SMALL CHANGE BUT ONLY VISIBLE IN THIS PR BECAUSE IT ONLY SEEMS TO IMPACT GCP PROCESSING
https://github.com/project-koku/koku/pull/3422/files#diff-f9c85bc4f2f7812a9afad872e5a4a3a49ff37e64b0aeeec5d1422fe3fae9d827R359 This change is because in gcp we process individual files for the days, which leads to us creating a _silly_ amount of daily parquet files since we keep rewriting the list of files. This was just a bug that didn't affect other providers it seems since they process one file.  In order to test this:
1. Create the test customer and load the test customer data: `make create-test-customer` followed by `load-test-customer-data`.
2. If you do that with this change reverted, gcp will create the same file multiple times as visible in [the minio UI](http://localhost:9090/buckets/koku-bucket/browse/) by following the parquet daily route to gcp and seeing the numerous files.
3. If you do that with this change in place, you will see a much more reasonable one file per day of the month created.

---
Additional Context:
GCP SMOKES: https://ci.ext.devshift.net/job/project-koku-koku-pr-check/3712/
TRINO SMOKES: https://ci.ext.devshift.net/job/project-koku-koku-pr-check/3713/


myersCody (approved):
LGTM

PR #3428, opened 2022-01-27
move GIT_COMMIT injection to bottom of file

maskarb:
This moves the `GIT_COMMIT` build arg and env variable to the bottom of the file so that we can leverage docker build caching. If this variable is that the top of the file, all the steps afterwards will always run again.


ddonahue007 (approved):
Nice catch, lgtm

shannon-donahue (approved):
LGTM

PR #3431, opened 2022-01-28
Cost 2276 drop ocp aws azure base tables

Red-HAP:
Fixes [COST-2276](https://issues.redhat.com/browse/COST-2276)

Changes proposed in this PR:
* Drops the following tables
    * `reporting_ocpawscostlineitem_daily_summary`
    * `reporting_ocpawscostlineitem_project_daily_summary`
    * `reporting_ocpazurecostlineitem_daily_summary`
    * `reporting_ocpazurecostlineitem_project_daily_summary`

This change removes 4 Django model classes and has 1 migration.

Testing Instructions:
1. Checkout this branch
2. Run migrations `make run-migrations`
3. Logon to `psql` and execute: 
```sql
select count(*)
  from information_schema.tables
 where table_name in ('reporting_ocpawscostlineitem_daily_summary',
                                    'reporting_ocpawscostlineitem_project_daily_summary',
                                    'reporting_ocpazurecostlineitem_daily_summary',
                                    'reporting_ocpawscostlineitem_project_daily_summary');

-- You should see zero.
```

Additional Context:

This migration could run for a long time. It would be best to **ensure** that it is executed as an independent CJI run.


ddonahue007 (approved):
LGTM

PR #3432, opened 2022-01-28
[COST-2275] - Fix OCP node capacity CTE in OCP Trino SQL

adberglund:
Fixes [COST-2275](https://issues.redhat.com/browse/COST-2275)

Changes proposed in this PR:
* Adjust the node capacity common table expression to sum the results of the already aggregated daily values. This CTE was written around the hourly raw data and was not updated successfully when we switched to using the daily table.
*

Testing Instructions:
1. Ask SRE to run both forms of the SQL against a known cluster in prod.

[Smoke Tests](https://ci.ext.devshift.net/job/project-koku-koku-pr-check/3738/)

---
Additional Context:


ddonahue007 (approved):
lgtm

PR #3435, opened 2022-02-01
fix first of month test failure

maskarb:
- fix first of the month test failure
- mock out the sleeps in the HCS functions (to speed up testing)
- fix a bug in `hcs_report_data.py`

### DIFF
--- a/koku/masu/api/hcs_report_data.py
+++ b/koku/masu/api/hcs_report_data.py
@@ -21,7 +21,6 @@ from hcs.tasks import collect_hcs_report_data
 from hcs.tasks import HCS_QUEUE
 from masu.database.provider_db_accessor import ProviderDBAccessor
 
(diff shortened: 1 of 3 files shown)


koku/masu/api/hcs_report_data.py
```diff
@@ -89,7 +88,7 @@ def hcs_report_data(request):
             )
 
         for month in months:
-            async_result = collect_hcs_report_data.s(start_date, end_date).apply_async(queue=HCS_QUEUE)
+            async_result = collect_hcs_report_data.s(month[0], month[1]).apply_async(queue=HCS_QUEUE)
```
maskarb, line 91: This is actually a bug fix.

abaiken (approved):
LGTM - thanks for fixing this

PR #3436, opened 2022-02-01
Convert markup float values to Decimal

maskarb:
While working on COST-444, markup tests were failing by a few cents. Converting the markup value to a Decimal before multiplying by usage and saving in the DB resolved the test failures.

Changes proposed in this PR:
* Convert markup to Decimal in each `populate_markup_cost` function.

Testing Instructions:
1.


abaiken (approved):
LGTM!

PR #3440, opened 2022-02-02
Fix cost distribution by grouping by data source.

myersCody:
**Fixes:**

This pull request is to fix an issue we identified with how cost distribution was being handled for ocp on azure & ocp on aws. The problem is that inside of the sql we were doing a `max(data_source)` which was overwriting all data sources as either Storage or Pod depending on what the max was in the date range which was really throwing off the case statements we use for Cost Distribution.   
```
CASE WHEN resource_id_matched = TRUE AND data_source = 'Pod'
        THEN ({{pod_column | sqlsafe}} / {{cluster_column | sqlsafe}}) * pretax_cost * cast({{markup}} as decimal(24,9)) / project_rank / data_source_rank
        ELSE pretax_cost / project_rank / data_source_rank * cast({{markup}} as decimal(24,9))
    END as project_markup_cost,
```

Essentially, either the if was always getting hit, or the else was always getting hit.

**Changes proposed in this PR:**
this pr is to change the max into a group by option so that we can distribute the cost correctly for when the data source is a Pod, or ignore it probably when the data source is Storage for each row. 

Testing Instructions:
1. Smoke tests will be updated to verify this fix so that we can test it through the smoke test passing.

### DIFF
--- a/koku/masu/database/presto_sql/reporting_ocpawscostlineitem_daily_summary.sql
+++ b/koku/masu/database/presto_sql/reporting_ocpawscostlineitem_daily_summary.sql
@@ -398,7 +398,7 @@ FROM (
     SELECT pds.aws_uuid,
         max(pds.cluster_id) as cluster_id,
         max(pds.cluster_alias) as cluster_alias,
-        max(pds.data_source) as data_source,
+        pds.data_source as data_source,
         pds.namespace,
         max(pds.node) as node,
         max(pds.persistentvolumeclaim) as persistentvolumeclaim,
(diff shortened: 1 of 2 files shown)


myersCody:
On Hold: Waiting for smoke tests to be updated

myersCody:
Current State:
So, we ran into this issue while developing on OCP on GCP, and we thought the issue would be problematic in the other providers. However, we expected these changes to break the smoke tests, and they had no affect on them whats so ever. 

```
It's possible the yaml file for nise needs to have to be configured right to make this happen. As it stands for AWS we match all the pod data to the nodes on resource id, and no volumes get matched there at all, so no possibility of having two different data sources. Then in the tag matching it depends what AWS stuff matches to pods and volumes I think. So if you made an RDS instance with a tag in the AWS yaml, and then matched that tag to both a pod and a volume hopefully you'd see the difference then?
```
I need to investigate the smoke test yamls to see what is up with them.

myersCody:
This issue was resolved in: https://github.com/project-koku/koku/pull/3467

PR #3446, opened 2022-02-07
Update the version map with the latest releases

abaiken:
Follow on to [COST-2009](https://issues.redhat.com/browse/COST-2009) and other operator releases

Changes proposed in this PR:
* Update the operator versions map with the new commits mapped to which release it was 

Testing Instructions:
1. You could test this by uploading a manifest locally and adding print statements for the version/ - this is mainly to update our koku-daily reports/ tableau boards. 

---
Additional Context:
Today during the program call I realized our tableau boards for the operator version were replaced with a different color and it made me remember that this map exists and we need to update it after operator releases.


myersCody (approved):
LGTM!

PR #3453, opened 2022-02-09
Revert "Cost 2312 always log db pid"

myersCody:
We ran into an issue during the last stage release where: 
```
  File "/opt/koku/koku/koku/settings.py", line 34, in koku_log
    if connection.connection and not connection.connection.closed:
```

We need to revert this pr so we can continue to debug the issue while still being able to do stage releases.


abaiken (approved):
LGTM but can we update the description to explain the revert

PR #3454, opened 2022-02-09
[COST-2317] fix ocp on cloud pod cost

adberglund:
Fixes [COST-2317](https://issues.redhat.com/browse/COST-2317)

Changes proposed in this PR:
* When distributing resource-matched costs out to projects in OCP on Cloud, we need to use the node capacity instead of cluster capacity as the denominator in our distribution function. 
* This is because each cost from the cloud source here corresponds with running a specific node

Testing Instructions:
1. Run `this command` ...
3. Expect to see this result:
   ```

   {json: response}
   ```
4.
5.

---
Additional Context:


shannon-donahue (approved):
LGTM

PR #3468, opened 2022-02-15
Fixing openshift query set for ocp/gcp resource-type

lcouzens:
Fixes [COST-2327](https://issues.redhat.com/browse/COST-2327)

Changes proposed in this PR:
* Switching confusing resource type naming convention from gcp-gcp-projects to gcp-projects
* Removing unneeded ocpgcp namespaces resource type
* Adding openshift=True option to gcp-projects queryset

Testing Instructions:
1. Ingest ocp on gcp data
2.  GET /api/cost-management/v1/resource-types/gcp-projects/?openshift=true

---
Additional Context:

### DIFF
--- a/koku/api/resource_types/gcp_projects/view.py
+++ b/koku/api/resource_types/gcp_projects/view.py
@@ -4,6 +4,7 @@
 #
 """View for GCP Projects."""
 from django.db.models import F
+from django.db.models.functions import Coalesce
 from django.utils.decorators import method_decorator
 from django.views.decorators.vary import vary_on_headers
 from rest_framework import filters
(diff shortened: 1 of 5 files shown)


myersCody (approved):
When I hit: 
http://localhost:8000/api/cost-management/v1/resource-types/gcp-projects/?openshift=true

I see:
```
{
    "meta": {
        "count": 1
    },
    "links": {
        "first": "/api/cost-management/v1/resource-types/gcp-projects/?limit=10&offset=0&openshift=true",
        "next": null,
        "previous": null,
        "last": "/api/cost-management/v1/resource-types/gcp-projects/?limit=10&offset=0&openshift=true"
    },
    "data": [
        {
            "value": "todays-raw-calc-id"
        }
    ]
}
```

PR #3471, opened 2022-02-16
[COST-2338] Use BigQuery's TO_JSON_STRING for gcp labels, changed log message to debug

cgoodfred:
Fixes [COST-2338](https://issues.redhat.com/browse/COST-2338)

Changes proposed in this PR:
* Use BigQuery's TO_JSON_STRING function when we download the `labels, system_labels and project.labels` from the dataset so we don't have to do any string replacing and we can just try to load the labels and parse the key/values out
* Change `unable to process gcp labels` log message to a debug

Testing Instructions:
1. Spin everything up in trino `make docker-up-min-trino` for example
2. Create a nise gcp source that has some labels and system labels and stuff with the bigquery flow, if you need help with this, ask corey
3. Ingest your nise source and see that processing still goes fine

---
Additional Context:
https://cloud.google.com/billing/docs/how-to/bq-examples#group_by_label_map_as_a_json_string


adberglund (dismissed):

myersCody:
I didn't realize big query had a `TO_JSON_STRING`, this is an awesome addition @cgoodfred! 🎉

adberglund (approved):

PR #3474, opened 2022-02-16
COST-2324: Handle AWS access denied errors downloading from S3

myersCody:
Fixes [COST-2324](https://issues.redhat.com/browse/COST-2324)

Changes proposed in this PR:
* This pull request is to handle access denied errors when trying to download the csv from the S3 bucket.

Testing Instructions:
1. Add the same source that we used for the previous AWS permissions issues. (If you don't know the source details, feel free to reach out to me.)

I updated the policy permissions on that source to trigger the Access Denied when attempting to download the file from the S3 bucket. 

- Check the logs:
```
koku-worker_1     | [2022-02-16 22:12:43,460] INFO f68277fc-57cf-4a39-9541-1dc5d526fddb {'message': 'Unable to access S3 Bucket bucket-name-here: (AccessDenied)', 'tracing_id': 'no_tracing_id', 'provider_uuid': UUID('10424d1f-ccab-4add-ab22-f35211913126'), 'provider_type': 'AWS', 'account': '10001'}
```

---
Additional Context:

### DIFF
--- a/koku/masu/external/downloader/aws/aws_report_downloader.py
+++ b/koku/masu/external/downloader/aws/aws_report_downloader.py
@@ -129,8 +129,17 @@ class AWSReportDownloader(ReportDownloaderBase, DownloaderInterface):
         """
         size_ok = False
 
-        s3fileobj = self.s3_client.get_object(Bucket=self.report.get("S3Bucket"), Key=s3key)
-        size = int(s3fileobj.get("ContentLength", -1))
+        try:
+            s3fileobj = self.s3_client.get_object(Bucket=self.report.get("S3Bucket"), Key=s3key)
+            size = int(s3fileobj.get("ContentLength", -1))
+        except ClientError as ex:
+            if ex.response["Error"]["Code"] == "AccessDenied":
+                msg = "Unable to access S3 Bucket {}: (AccessDenied)".format(self.report.get("S3Bucket"))
+                LOG.info(log_json(self.tracing_id, msg, self.context))
+                raise AWSReportDownloaderAccessDenied(msg)
+            msg = f"Error downloading file: Error: {str(ex)}"
+            LOG.error(log_json(self.tracing_id, msg, self.context))
+            raise AWSReportDownloaderError(str(ex))
 
         if size < 0:
             raise AWSReportDownloaderError(f"Invalid size for S3 object: {s3fileobj}")
--- a/koku/masu/external/downloader/aws/aws_report_downloader.py
+++ b/koku/masu/external/downloader/aws/aws_report_downloader.py
@@ -35,6 +35,10 @@ class AWSReportDownloaderNoFileError(Exception):
     """AWS Report Downloader error for missing file."""
 
 
+class AWSReportDownloaderAccessDenied(Exception):
+    """AWS Report Downloader error for access denied."""
+
+
 class AWSReportDownloader(ReportDownloaderBase, DownloaderInterface):
     """
     AWS Cost and Usage Report Downloader.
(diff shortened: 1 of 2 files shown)


koku/masu/external/downloader/aws/aws_report_downloader.py
```diff
@@ -237,7 +241,10 @@ def download_file(self, key, stored_etag=None, manifest_id=None, start_date=None
                 msg = "Unable to find {} in S3 Bucket: {}".format(s3_filename, self.report.get("S3Bucket"))
                 LOG.info(log_json(self.tracing_id, msg, self.context))
                 raise AWSReportDownloaderNoFileError(msg)
-
+            if ex.response["Error"]["Code"] == "AccessDenied":
+                msg = "Unable to access S3 Bucket {}: (AccessDenied)".format(self.report.get("S3Bucket"))
+                LOG.info(log_json(self.tracing_id, msg, self.context))
+                raise AWSReportDownloaderAccessDenied(msg)
```
maskarb, line 247: question: Do we need to catch this new `AWSReportDownloaderAccessDenied` exception anywhere?

maskarb (commented):
a question:

koku/masu/external/downloader/aws/aws_report_downloader.py
myersCody: I went over it again with a fine tooth comb, and found one other place we could use it.

koku/masu/external/downloader/aws/aws_report_downloader.py
```diff
@@ -125,8 +129,17 @@ def _check_size(self, s3key, check_inflate=False):
...
+            s3fileobj = self.s3_client.get_object(Bucket=self.report.get("S3Bucket"), Key=s3key)
+            size = int(s3fileobj.get("ContentLength", -1))
+        except ClientError as ex:
+            if ex.response["Error"]["Code"] == "AccessDenied":
+                msg = "Unable to access S3 Bucket {}: (AccessDenied)".format(self.report.get("S3Bucket"))
+                LOG.info(log_json(self.tracing_id, msg, self.context))
+                raise AWSReportDownloaderAccessDenied(msg)
+            msg = f"Error downloading file: Error: {str(ex)}"
+            LOG.error(log_json(self.tracing_id, msg, self.context))
+            raise AWSReportDownloaderError(str(ex))
```
maskarb, line 142: Question: Could you explain why this error handling was added? The only time we use `_check_size` is _after_ we have the object. I cannot imagine us downloading the report first and then suddenly we can't download it again...but also, side tangent, and unrelated to this PR, I don't even know why we download it again. We should get the size when we initially download the report and pass it into here.

koku/masu/external/downloader/aws/aws_report_downloader.py
maskarb: What I mean is, where do we catch the `AWSReportDownloaderAccessDenied` exception? I assume it would need to go [here](https://github.com/project-koku/koku/blob/3feca123c54d5de8aeab5944ba2e4306245b7ad4/koku/masu/external/downloader/aws/aws_report_downloader.py#L171) or [here](https://github.com/project-koku/koku/blob/3feca123c54d5de8aeab5944ba2e4306245b7ad4/koku/masu/external/report_downloader.py#L250) or somewhere, right?

maskarb (commented):
A couple other questions:

koku/masu/external/downloader/aws/aws_report_downloader.py
myersCody: So, I interpreted your original comment to mean: "Is there anywhere else in this file where we could potentially run into an access denied error that is not being covered." 

With that context in mind, my changes here probably make more sense now. 😂 

The changes of someone loosing their permissions between calls is low, but there was a possibility so I wrapped it in a try catch. 

However, after seeing what you linked in your other comment I am now picking up what you mean. You bring up a great point about us only catching the `AWSReportDownloaderNoFileError` in those two places you mentioned, and in hindsight, I think I should just call `AWSReportDownloaderNoFileError` instead of creating a new exception. I personally think access denied would be a perfectly good reason for us to have no file.

maskarb (approved):

PR #3475, opened 2022-02-16
fix model bakery tags update

maskarb:
The ocp-on-cloud tags summary updater was not resetting the schema to public which caused the additional test Tenant creation to fail. This PR uses a context manager to update the tags which resets the schema when done.


myersCody (approved):
LGTM

PR #3486, opened 2022-02-23
Add CJI script to truncate Azure lid tables on trino accounts

Red-HAP:
## Jira Ticket

[COST-2355](https://issues.redhat.com/browse/COST-2355)

## Description

* This is a CJI to TRUNCATE the `reporting_azurecostlineitem_daily` tables for account types `Azure`, `Azure-local`

## Testing

1. Enable trino processing in unleash for an Azure account in localdev
1. Pull branch
2. `docker exec -it koku_server /bin/bash`
3. `./scripts/truncate_old_tables`
4. You should see that your account table `reporting_azurecostlineitem_daily` was truncated. It's even cooler if you have data in it first to compare before and after.

### DIFF
--- a/scripts/truncate_old_tables.py
+++ b/scripts/truncate_old_tables.py
@@ -1,0 +1,249 @@
+#! /usr/bin/env python3.8
+import datetime
+import logging
+import os
+import sys
+from decimal import Decimal
+
+import psycopg2
+from app_common_python import LoadedConfig
+from psycopg2.extras import NamedTupleCursor
+
+
+logging.basicConfig(
+    format="truncate_old_tables (%(process)d) :: %(asctime)s: %(message)s",
+    datefmt="%m/%d/%Y %I:%M:%S %p",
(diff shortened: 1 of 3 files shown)


scripts/truncate_old_tables.py
```diff
@@ -0,0 +1,249 @@
...
+
+def _execute(conn, sql, params=None):
+    cur = conn.cursor()
+    LOG.debug(cur.mogrify(sql, params).decode("utf-8"))
+    cur.execute(sql, params)
+    return cur
+
+
+def validate_tables(conn):
+    accounts = sorted(
```
maskarb, line 53: Instead of hard-coding these account numbers, can we make use of this function instead?
https://github.com/project-koku/koku/blob/1240bb93c0acee51dd9b255c73ddf1d91b4de449/koku/masu/processor/__init__.py#L19

Red-HAP: Done. That was a good catch. I forgot enablement went to the source-level.

ddonahue007 (approved):
LGTM

PR #3504, opened 2022-03-02
[COST-2063] - Switch to RunSQL for single ALTER TABLE

adberglund:
Fixes [COST-2063](https://issues.redhat.com/browse/COST-2063)

Changes proposed in this PR:
* This modifies our migration to alter the decimal precision on columns to use a single ALTER TABLE statement instead of a distinct statement for each column

Testing Instructions:
1. Run `this command` ...
3. Expect to see this result:
   ```

   {json: response}
   ```
4.
5.

---
Additional Context:


maskarb (approved):
lgtm!

PR #3508, opened 2022-03-04
Cost 2276 drop ocp aws azure base tables v2

Red-HAP:
Fixes [COST-2276](https://issues.redhat.com/browse/COST-2276)

Changes proposed in this PR:
* Final migration to drop the ocp on aws/azure regular tables leaving only the partitioned tables.


ddonahue007 (approved):
LGTM

PR #3510, opened 2022-03-05
load script refactor

ddonahue007:
Refactor on load-test-customer-data script

Changes proposed in this PR:
* better logging output (debug flag is right)
* options to allow loading individual source types

Testing Instructions:
1. Setup as normal
2. run create customer data
`make create-test-customer`
3. Test loading just Azure data
`make test_source=azure load-test-customer-data`
4. Test loading AWS with debug flag set to true
`export DEBUG=true`
`make test_source=aws load-test-customer-data`
5. Test default which will load all data
`make load-test-customer-data`

---
Additional Context:


Makefile
```diff
@@ -187,8 +187,9 @@ delete-cost-models:
 
 delete-test-customer-data: delete-test-sources delete-cost-models
 
+TEST_SOURCE=all
 load-test-customer-data:
-	$(SCRIPTDIR)/load_test_customer_data.sh $(start) $(end)
+	$(SCRIPTDIR)/load_test_customer_data.sh $(TEST_SOURCE) $(start) $(end)
```
abaiken, line 192: This is a nit (so not a blocker) but it would be nice to update the make help for this command to show that you can pass in `TEST_SOURCE` if you want to just create aws/azure/etc

ddonahue007: Good point and nice catch

abaiken (approved):
Tested the normal flow and it worked 👍 Nice enhancements

PR #3521, opened 2022-03-09
Cost 2454 db performance panel

Red-HAP:
## Jira Ticket
[COST-2454](https://issues.redhat.com/browse/COST-2454)

## Description

This is a new module that will perform live queries against the database for research regarding database and query performance.

The internal (`masu`) endpoints are:
* `api/cost-management/v1/db-performance/db-settings/` Return current database settings
* `api/cost-management/v1/db-performance/db-version/` Return the PostgreSQL engine version
* `api/cost-management/v1/db-performance/lock-info/` Return any blocking processes and those processes that are blocked
* `api/cost-management/v1/db-performance/stat-activity/` Return the connection activity grouped by database and transaction state
* `api/cost-management/v1/db-performance/stat-statements/` Return query statistics ordered by mean execution time descending

Endpoint output is HTML to make the data easier to analyze.

This PR documentation has also been saved in a Google doc [here](https://docs.google.com/document/d/1C2_gu0pT8u3UWyugmYNHDhA-tvwsbOy2epplq4B9M2k/edit?usp=sharing).

## Documentation

In order to utilize the localdev links, please check out this branch and restart your local koku.

### PostgreSQL Version

This will display the version of PostgreSQL software. Simply navigate to the `masu` endpoint: `api/cost-management/v1/db-performance/db-version/`. You should see a result page with a table showing the PostgreSQL version as `<release>.<major>.<minor>`. PostgreSQL has changed their release schedule and numbering. You will probably see **00** for the `<major>` value.

#### Localdev Example

[http://localhost:5042/api/cost-management/v1/db-performance/db-version/](http://localhost:5042/api/cost-management/v1/db-performance/db-version/)

### PostgreSQL Settings

This will display all of the current settings for PostgreSQL. This may be useful for determining configurations like shared memory, number of workers, and understanding which settings values require a reboot.

#### Localdev Example

[http://localhost:5042/api/cost-management/v1/db-performance/db-settings/](http://localhost:5042/api/cost-management/v1/db-performance/db-settings/)

### Statement Statistics

This page will display the top 100 queries and their performance ranked in descending order by the mean execution time. Note that this view is cumulative. High time and memory numbers need to be evaluated with the number of calls recorded.

This data can help track down queries that are always running long or accessing memory a lot (or not). This can help diagnose the general performance of different queries sent to the database. These queries are recorded in a parameterized form.

he definition of the returned data table is as follows:

* database - Database name
* role - Database username
* calls - Total number of times this query was executed.
* rows - Total number of rows retrieved or affected by the statement
* min_exec_time - Minimum statement execution time in milliseconds
* mean_exec_time - Mean statement execution time in milliseconds
* max_exec_time - Maximum statement execution time in milliseconds
* shared_blks_hit - Total number of shared block cache hits by the statement
* shared_blks_read - Total number of shared blocks read by the statement
* local_blks_hit - Total number of local block cache hits by the statement
* local_blks_read - Total number of local blocks read by the statement
* temp-blks-read - Total number of temp blocks read by the statement
* temp-blks-written - Total number of temp blocks written by the statement
* query - The executed query. This query will be parameterized (variables will be represented by `$n` placeholders).

See the [PostgreSQL documentation](https://www.postgresql.org/docs/13/pgstatstatements.html#id-1.11.7.38.6) for a more complete description.

The time columns are color-coded as follows:

* Less than 4ms is green
* Greater than 4ms but less than 5ms is yellow
* 5ms or more is red

#### Localdev Example

[http://localhost:5042/api/cost-management/v1/db-performance/stat-statements/](http://localhost:5042/api/cost-management/v1/db-performance/stat-statements/)

### Connection Activity

This page will display the top 250 queries ordered by database, connection state, and active query time descending. This page will have buttons that will show the cancel or terminate SQL to use to take action on the backend pid. If the backend pid is in an `idle` state (`idle`, `idle-in-transaction`) a cancel may not be sufficient. But, as these are live queries, trying a second time may, in fact, cancel the connection activity and free the locks.

Bear in mind that the queries here are fully rendered. Be careful with potentially sensitive data in the queries.

Once you have dismissed the alert showing the SQL, the page will reload.

The definition of the returned data table is as follows:

* database - Database name
* role - Database username
* backend_pid - Process ID of the server process handling the connection
* app_name - Name of the application used in the connection
* client_ip - IP Address of the connected client
* backend_start - Time when the connection to the database was established
* xact_start - The time of the current transaction start
* query_start - The time that the current active query was started
* state_change - The time of the last state change
* wait_type_event - The wait event type and the wait event itself. The type will be enclosed in parenthesis.
* state - Name of the current state of the connection.
* active_time - When an **active** query, the length of time the query has been running (up to the request) expressed as an interval
* query - The text of the current or last executed query.

See the [PostgreSQL documentation](https://www.postgresql.org/docs/13/monitoring-stats.html#MONITORING-PG-STAT-ACTIVITY-VIEW) for a more complete definition.

#### Localdev Example

[http://localhost:5042/api/cost-management/v1/db-performance/stat-activity/](http://localhost:5042/api/cost-management/v1/db-performance/stat-activity/)

### Lock Information

If there are any connections with lock contention, this page will show the blocked process as well as the blocking process. This page will have buttons that will show the cancel or terminate SQL to use to take action on the blocking or blocked backend pid. If your target pid is in an `idle` state (`idle`, `idle-in-transaction`) a cancel may not be sufficient. But, as these are live queries, trying a second time may, in fact, cancel the connection activity and free the locks.

Once you have dismissed the alert showing the SQL, the page will reload.

Bear in mind that the queries here are fully rendered. Be careful with potentially sensitive data in the queries.

#### Localdev Example

If you wish to see actual lock information, you will need to run two psql sessions at the same time. In session 1, execute

```sql
BEGIN;
UPDATE public.api_provider SET paused = false;
```

Then run the same thing in session 2. Session 2 should appear to hang.

With this setup done, you can click the link.

[http://localhost:5042/api/cost-management/v1/db-performance/lock-info/](http://localhost:5042/api/cost-management/v1/db-performance/lock-info/)

To cancel, go to session 2 and press `<ctrl>+c` then execute `ROLLBACK;`
Finally, go to session 1 and execute `ROLLBACK;` Then you can exit both sessions.


koku/db_performance/test_db_performane.py
```diff
@@ -0,0 +1,110 @@
+#
```
cgoodfred, line 1: This filename should be `test_db_performance.py` instead of `test_db_performane.py`

cgoodfred:
A few thoughts/suggestions that are definitely open for discussion:

- **Suggestion:** Since these are masu api endpoints, can we move the `db_performance` directory into the `koku/masu/api` directory? 
- **Suggestion:** Can we group the endpoints in masu under a `db-performance` endpoint or something along that nature? so for example we would have `api/cost-management/v1/db-performance/stat_activity/`
- **nitpick (non-blocking)** We use hyphens in all of our other endpoints, but for `/stat_statements/` and `/stat_activity/` you used underscores, consider switching these to hyphens to stay more consistent
- **Question (non-blocking)** I'm not too familiar with turnpike, but do we know it will render the HTML templates properly? I don't see why it wouldn't, but something to consider

koku/db_performance/test_db_performane.py
Red-HAP: This should be addressed.

Red-HAP:
> * **Suggestion:** Since these are masu api endpoints, can we move the `db_performance` directory into the `koku/masu/api` directory?
Done.
> * **Suggestion:** Can we group the endpoints in masu under a `db-performance` endpoint or something along that nature? so for example we would have `api/cost-management/v1/db-performance/stat_activity/`
Done
> * **nitpick (non-blocking):** We use hyphens in all of our other endpoints, but for `/stat_statements/` and `/stat_activity/` you used underscores, consider switching these to hyphens to stay more consistent
Done
> * **Question (non-blocking):** I'm not too familiar with turnpike, but do we know it will render the HTML templates properly? I don't see why it wouldn't, but something to consider
I don't know. Maybe @abaiken or @maskarb or @adberglund can help out with that and my other questions above.

cgoodfred (changes requested):
**issue(blocking):** I think this endpoint should be purely for information gathering and we shouldn't allow canceling/terminating processes through this endpoint. 
**suggestion(blocking):** I think these endpoints should be removed along with the functions/code that goes with them, but I am open to discussion about this, maybe with more of the team on slack as well.

(I messed up the first time adding this and put it with the relevant pieces below)

koku/masu/api/urls.py
```diff
@@ -60,6 +68,14 @@
         name="get_one_manifest_file",
     ),
     path("gcp_invoice_monthly_cost/", gcp_invoice_monthly_cost, name="gcp_invoice_monthly_cost"),
+    path("db-performance/db-settings/", dbsettings, name="db_settings"),
+    path("db-performance/lock-info/", lockinfo, name="lock_info"),
+    path("db-performance/stat-activity/", stat_activity, name="conn_activity"),
+    path("db-performance/stat-statements/", stat_statements, name="stmt_stats"),
+    path("db-performance/db-version/", pg_engine_version, name="db_version"),
+    path("db-performance/cancel-backend-pid/", pg_cancel_backend, name="db_cancel_connection"),
```
cgoodfred, line 76: **issue(blocking): ** I think this endpoint should be purely for information gathering and we shouldn't allow canceling/terminating processes through this endpoint.
**suggestion(blocking): ** I think these should be removed, but I am open to discussion about this, maybe with more of the team on slack as well.

koku/masu/openapi.json
```diff
@@ -812,7 +812,92 @@
...
-    "/gcp_invoice_monthly_cost/": {
+    "/db-performance/db-version/": {
+        "get": {
+          "summary": "Returns a table showing the PostgreSQL database software version.",
+          "operationId": "dbPerformanceDbVersion",
+          "description": "Returns a HTML document showing the database version.",
+          "responses": {
+            "200": {
+              "description": "Returns a HTML document showing the database version.",
+              "content": "text/html"
```
cgoodfred, line 823: **Issue(blocking):** This does not meet OpenAPI specs: `Structural error at paths./db-performance/db-version/.get.responses.200.content should be object`
**Suggestion:** Update the content field to meet OpenAPI specs

koku/masu/openapi.json
```diff
@@ -812,7 +812,92 @@
...
+      },
+      "/db-performance/stat-activity/": {
+        "get": {
+          "summary": "Returns a table showing current connection activity.",
+          "operationId": "dbPerformanceStatActivity",
+          "description": "Returns a HTML document showing current connection activity.",
+          "responses": {
+            "200": {
+              "description": "Returns a HTML document showing current connection activity.",
+              "content": "text/html"
```
cgoodfred, line 874: **Issue(blocking):** This does not meet OpenAPI specs: `Structural error at paths./db-performance/stat-activity/.get.responses.200.content should be object`
**Suggestion:** Update the content field to meet OpenAPI specs

koku/masu/openapi.json
```diff
@@ -812,7 +812,92 @@
...
+      },
+      "/db-performance/lock-info/": {
+        "get": {
+          "summary": "Returns a table showing any detected blocking locks.",
+          "operationId": "dbPerformanceLockInfo",
+          "description": "Returns a HTML document showing any blocking locks and processes that are blocked by them.",
+          "responses": {
+            "200": {
+              "description": "Returns a HTML document showing any blocking locks and processes that are blocked by them.",
+              "content": "text/html"
```
cgoodfred, line 891: **Issue(blocking):** This does not meet OpenAPI specs: `Structural error at paths./db-performance/lock-info/.get.responses.200.content should be object`
**Suggestion:** Update the content field to meet OpenAPI specs

cgoodfred (changes requested):
Just a few tweaks related to the OpenAPI spec and we should be good.

koku/masu/openapi.json
Red-HAP: I added this because people kept asking about that file.
But: If these endpoints are returning HTML instead of JSON, do they _really_ belong in openapi.json?

koku/masu/openapi.json
```diff
@@ -812,7 +812,92 @@
...
+      },
+      "/db-performance/db-settings/": {
+        "get": {
+          "summary": "Returns a table showing the PostgreSQL database software version.",
+          "operationId": "dbPerformanceDbSettings",
+          "description": "Returns a HTML document showing the database software settings.",
+          "responses": {
+            "200": {
+              "description": "Returns a HTML document showing the database software settings.",
+              "content": "text/html"
```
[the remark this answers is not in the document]
Red-HAP: I added this because people kept asking about that file.
But: If these endpoints are returning HTML instead of JSON, do they really belong in openapi.json?

koku/masu/openapi.json
Red-HAP: I added this because people kept asking about that file.
But: If these endpoints are returning HTML instead of JSON, do they really belong in openapi.json?

koku/masu/openapi.json
Red-HAP: I added this because people kept asking about that file.
But: If these endpoints are returning HTML instead of JSON, do they really belong in openapi.json?

cgoodfred (approved):
Nice Job!

PR #3525, opened 2022-03-10
Fixing pod/volume distribution

lcouzens:
Fixes [COST-2455](https://issues.redhat.com/browse/COST-2455)

Changes proposed in this PR:
* Added an and statement to also check if we have a kubernetes-io-cluster-id match for correct cost distribution.

Testing Instructions:
1. Ingest ocp on gcp data
2. hit http://localhost:8000/api/cost-management/v1/reports/openshift/infrastructures/gcp/costs/?group_by[project]=*
3. check distribution is what we expect

OR run this QE test test_api_ocp_on_gcp_source_raw_calc
Before this change that test would fail.


maskarb (approved):
LGTM!

PR #3537, opened 2022-03-21
Trino logging update

adberglund:

Changes proposed in this PR:
* Change some of the details logged for Trino SQL
* Remove logging full SQL except when there is an error

Testing Instructions:
1. Run `this command` ...
3. Expect to see this result:
   ```

   {json: response}
   ```
4.
5.

---
Additional Context:


shannon-donahue (approved):
lgtm!

PR #3553, opened 2022-04-01
[COST-1880] Only update cost models if provider active

adberglund:
Fixes [COST-1880](https://issues.redhat.com/browse/COST-1880)

Changes proposed in this PR:
* Conditionally recalculate cost model costs only if the source is marked active.

Testing Instructions:
1. Run `this command` ...
3. Expect to see this result:
   ```

   {json: response}
   ```
4.
5.

---
Additional Context:
This puts a condition on relaunching cost model updates when we remove a provider from a cost model. When we delete a source, we first mark it inactive, then do cleanup and then finally remove it. Our smoke tests also delete the cost model during this time, which triggers a cost model recalculation on the dying provider. It’s gone before we finish and we get the IntegrityError in the alert channel

### DIFF
--- a/koku/cost_models/test/test_cost_model_manager.py
+++ b/koku/cost_models/test/test_cost_model_manager.py
@@ -251,7 +251,7 @@ class CostModelManagerTest(IamTestCase):
     @patch("cost_models.cost_model_manager.refresh_materialized_views")
     @patch("cost_models.cost_model_manager.update_cost_model_costs")
     @patch("cost_models.cost_model_manager.chain")
-    def test_deleting_cost_model_refreshes_materialized_views(self, mock_chain, mock_update, mock_refresh):
+    def test_deleting_cost_model_triggers_tasks(self, mock_chain, mock_update, mock_refresh):
         """Test deleting a cost model refreshes the materialized views."""
         provider_name = "sample_provider"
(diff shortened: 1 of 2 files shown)


koku/cost_models/test/test_cost_model_manager.py
```diff
@@ -286,6 +286,40 @@ def test_deleting_cost_model_refreshes_materialized_views(self, mock_chain, mock
                 mock_refresh.si(self.schema_name, provider.type, provider_uuid=provider_uuid).set(),
             )
 
+    @patch("cost_models.cost_model_manager.refresh_materialized_views")
+    @patch("cost_models.cost_model_manager.update_cost_model_costs")
+    @patch("cost_models.cost_model_manager.chain")
+    def test_deleting_cost_model_not_triggers_tasks(self, mock_chain, mock_update, mock_refresh):
+        """Test deleting a cost model refreshes the materialized views."""
```
djnakabaale, line 293: is this supposed to be  "does not refresh" ?

adberglund, line 293:
```suggestion
        """Test deleting a cost model with an inactive provider does not trigger tasks."""
```

PR #3554, opened 2022-04-01
Cost 2527 default lock info limit

Red-HAP:
## Jira Ticket

[COST-2527](https://issues.redhat.com/browse/COST-2527)

## Description

* Sets a default limit of 500 records returned from the `get_lock_info` method
* Processes `limit` and `offset` from the request if supplied
* Exposes parameters in masu `openapi.json`

## Testing

First, checkout this code and restart koku

### Limit & Offset

This is best done with the Connection Activity page, but this will also work on the Statement Statistics and Lock Info pages.

1. Navigate to the Connection Activity page
2. In the url bar, append `?limit=4`. You should see only 4 records
3. Change `?limit=4` to `?limit=-2`. You should see all records as the limit value was invalid.
    1. If you change the integer to a string, the limit will also not be applied

You can also repeat the same test for `offset` by changing `limit` to `offset`
You can also combine them: `?limit=5&offset=5`

#### include_self

This parameter will include the connection that is running these check queries.

1. Navigate to the Connection Activity page
2. Add `?include_self=true` to the end of the url. You should now see a process with the application name `database_performance_stats`

#### pid & state

1. Navigate to the Connection Activity page
2. Find a single backend_pid value.
3. Append to the url: `?pid=<the pid you picked>` and reload
4. Only that pid will display in the output.
5. Remove the parameter and reload.
6. Pick multiple pids.
7. Append to the url: `?pid=<pid1>&pid=<pid2>...` and reload
    1. Altenately, you can specify: `?pid=<pid1>,<pid2>...`
8. Output will be constrained to only those pids.

These testing steps can be repeated with the `state` parameter and use of one or more of the state values.

Enter a bizarre value for `pid` (like 999999) You will get a message that no records match your criteria.
The same is true for a garbage value for state.

#### records_per_db

This is another way to limit records per db instead of the total records returned.

1. Navigate to the Connection Activity page
2. Append `?records_per_db=2` and reload
3. You should see a maximum of 2 records for each database reporting.

### DIFF
--- a/koku/masu/api/db_performance/db_performance.py
+++ b/koku/masu/api/db_performance/db_performance.py
@@ -294,8 +294,7 @@ SELECT blocking_locks.pid::int     AS blocking_pid,
 """
         LOG.info(self._prep_log_message("requsting blocked process information"))
         res = self._execute(sql, params).fetchall()
-        if not res:
-            res = [{"Result": "No blocking locks"}]
+
         return res
 
     def get_activity(self, pid=[], state=[], include_self=False, limit=500, offset=None, records_per_db=100):
--- a/koku/masu/api/db_performance/db_performance.py
+++ b/koku/masu/api/db_performance/db_performance.py
@@ -4,6 +4,7 @@
 #
 import logging
 import os
+from decimal import Decimal
 
 import psycopg2
 from psycopg2.errors import ProgrammingError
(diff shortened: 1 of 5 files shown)


ddonahue007 (approved):
LGTM

ddonahue007 (approved):
LGTM

PR #3555, opened 2022-04-05
update worker cache to accurately reflect running tasks

maskarb:
The property `worker_cache_keys` is a list of all of the running celery workers saved in the cache key called `keys`. When we instantiate the `WorkerCache`, we update this key (`keys`) in the db so it is always current with the running workers.

What this PR changes for the single task flow is to:
1. make the cache key's value the hostname.
2. when we check to see if `single_task_is_running`, we fetch the value (which is now the hostname) and check to see if it is in the `keys` list
- if the hostname is in the list, then the task is current and running
- if the hostname is not in the list, then the task is _not_ currently running.


myersCody (approved):
LGTM

PR #3563, opened 2022-04-12
Fix skip reason for test.

myersCody:
I thought currency support would fix the test, but the problem exists in main so I created an issue for it.


abaiken (approved):
LGTM

PR #3567, opened 2022-04-12
Cost 2528 clone schema bug

Red-HAP:
Fixes [COST-2528](https://issues.redhat.com/browse/COST-2528)

Changes proposed in this PR:
* Add migration to reapply the clone_schema function that was manually applied to production to all environments.
* Fix bug in handling an empty result from git ls-tree

Testing Instructions:
1. Run: `./dev/scripts/show_migrations -n a0f9e2630e917c31c3afbf9912dd75f9cdefc1e8 -o f8830a5c94e540a341cd2fd1083af866dc5c7640`

You should not see any errors like `fatal: not tree-like object`
You should see `No migrations to be run`


maskarb (approved):
LGTM!

PR #3568, opened 2022-04-12
Don't overwrite start,end dates in task

adberglund:
Fixes bug introduced in [COST-2505](https://issues.redhat.com/browse/COST-2505)


lcouzens (approved):
Such a small change with such mighty differences :D

PR #3573, opened 2022-04-14
Update Sources API Client to use only one filtering operation type

lindgrenj6:

Changes proposed in this PR:
* Updating the sources-api client code to use the "new" (it already existed, but we chose to only support this type of filtering going forward with the rewrite) filtering operations. 

Testing Instructions:
1. Smoke tests should catch any errors. It should work with the current rails implementation as well as the new go rewrite. 

---
Additional Context:
This was found during integration of the applications with the new rewrite. Thank you all for the help!


maskarb (approved):
lgtm!

PR #3586, opened 2022-04-25
Fix arg error affecting Fedora users when default date range is calcu…

Red-HAP:
Fixes [COST-2579](https://issues.redhat.com/browse/COST-2579)

Changes proposed in this PR:
* Fixes argument error for Fedora users

Testing Instructions FEDORA OS ONLY:
1. Checkout branch
2. Run `DEBUG=true ./dev/scripts/load_test_customer_data.sh AWS`
3. When you see 2 distinct sets of log lines in blue, hit `ctrl+c`
4. Expect to see no errors and a valid start date and end date in `YYYY-MM-DD` format


lcouzens (approved):
LGTM

PR #3588, opened 2022-04-25
convert errors to warnings to prevent double alerting

maskarb:
Changes proposed in this PR:
* Convert excessive `log.error` to `log.warning`
* Failed celery tasks are logged with a full traceback all the way from the source of the exception. But we currently have some places where we catch an exception, log an error, and re-raise. This `catch -> log -> raise` clutters Sentry with duplicate errors and these logged errors do not contain the full traceback.

Here are examples:
* [This](https://sentry.io/organizations/project-koku/issues/3002713571/events/9ad32b6f5cf24600a1991ce514f4616c/?project=1822753) alert is raised by Celery. The full traceback can be seen.
* [This](https://sentry.io/organizations/project-koku/issues/3223748873/events/c9acbbc6d02742b39a6f1476db22d9df/?project=1822753) is one log that was generated by the above error. This error does not contain the same level of information that the previous alert contains.

This PR aims to not alert on this second error.

### DIFF
--- a/koku/koku/koku_test_runner.py
+++ b/koku/koku/koku_test_runner.py
@@ -7,7 +7,6 @@
 import logging
 import os
 import sys
-from unittest.mock import patch
 
 from dev.scripts.insert_org_tree import UploadAwsTree
 from django.conf import settings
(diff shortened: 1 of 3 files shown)


ddonahue007 (approved):
LGTM

PR #3592, opened 2022-04-26
update aws tag query unit tests

maskarb:
Changes proposed in this PR:
* Update the AWS tags unit tests to use the `app` key and retrieve values from the database (Changes from previous COST-444 PRs made `app` the only enabled tag key)


Red-HAP:
Why lock it to "app"?
If that ever changes, then this needs to be revisited again. 
Shouldn't the tests just use the first enabled tag key?

PR #3596, opened 2022-04-26
force header override in `RbacPermissions`

maskarb:
`DevelopmentIdentityHeaderMiddleware` used to force all requests to use either the env defined `DEVELOPMENT_IDENTITY` or the built in header. This mechanism would overwrite headers that were sent with the request. The middleware was modified so that headers sent with the request would not be overwritten. This update required a change to the `RbacPermissions` wrapper so that the header used by tests was correct.

Today, we have random test failures that seem related to test headers. This PR updates the wrapper to add an additional setting called `FORCE_HEADER_OVERRIDE`. When true, this variable forces the middleware to explicitly overwrite the request header. I think the previous logic for overwriting the header was fragile leading to our random failures.

### DIFF
--- a/koku/api/iam/test/iam_test_case.py
+++ b/koku/api/iam/test/iam_test_case.py
@@ -210,7 +210,6 @@ class RbacPermissions:
 
         @functools.wraps(function)
         def wrapper(*args, **kwargs):
-            original_id = args[0].headers[RH_IDENTITY_HEADER]
             user = self.user
             user["access"] = self.access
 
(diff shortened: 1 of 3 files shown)


InsightsDroid:
732bdea8-ffea-4822-9d86-2d9689df8eb7
f99991d4-f36d-4f48-ab5a-96a7a7f946cb

ddonahue007 (approved):
LGTM

PR #3612, opened 2022-05-01
[COST-2593] in refresh_materialized_views convert default arg to empty string

maskarb:
Fixes [COST-2593](https://issues.redhat.com/browse/COST-2593)

Changes proposed in this PR:
* convert the default `provider_uuid=None` to `provider_uuid=""`

We do not pass a provider uuid into refresh materialized views from the data expiration flow. This results in a TypeError when creating the cache key from the cache_args. By setting the default to an empty string instead of a NoneType, we should not raise a TypeError when creating the cache key.


lcouzens (approved):
LGTM

PR #3631, opened 2022-05-10
Adding exception for authentication error

lcouzens:
Fixes [COST-2384]
* Adds a new exception to catch authentication errors when our cost mgmt account is not subscribed to the customers region.

Testing Instructions:
1. Try adding an OCI source with a region our dev account is not subscribed to.
2. Should see ServiceError in logs instead of a 500


myersCody (approved):
LGTM

PR #3636, opened 2022-05-11
[COST-2562] Use resource level GCP table by default

adberglund:
Fixes [COST-2562](https://issues.redhat.com/browse/COST-2562)

Changes proposed in this PR:
* This will default to looking for and using the resource-level GCP BigQuery table
* We will fall back to the non-resource level table

Testing Instructions:
1. Run `this command` ...
3. Expect to see this result:
   ```

   {json: response}
   ```
4.
5.

---
Additional Context:


koku/providers/gcp/provider.py
```diff
@@ -52,6 +61,15 @@ def update_source_data_source(self, credentials, data_source):
                 if source.billing_source.get("data_source", {}).get("dataset") == data_source.get("dataset"):
                     source_filter = Sources.objects.filter(source_id=source.source_id)
                     source_filter.update(billing_source={"data_source": data_source})
+
+                    provider_uuid = source_filter.first().koku_uuid
```
maskarb, line 65: I think I am confused by all of this logic. Why do we filter the sources instead of getting the sources entry? What happens here if the `source_filter` queryset is empty? I presume that `source_filter.first().koku_uuid` will result in an error, won't it?

maskarb (commented):
A question:

InsightsDroid:
0ee3b33a-ccf9-4db8-ac14-abc593d4676d
46c6103b-c9ae-456b-a1c7-235cec3eeb8d

koku/providers/gcp/provider.py
adberglund: Updated slightly.

InsightsDroid:
6762003f-3b5a-4483-9ff6-30bb30d0ee3c
761fd8ca-c30f-4695-aa76-95d451428cce

koku/providers/gcp/provider.py
maskarb: Since we are using `.filter`, do we even need the try/except?

maskarb (approved):

PR #3640, opened 2022-05-12
Adding datetime to script to reduce intial ingest

lcouzens:
Fixes [COST-2384](https://issues.redhat.com/browse/COST-2384)

Changes proposed in this PR:
* Adding a small datetime improvement to OCI replication script template, this should prevent us from having hundreds of files for initial ingest.

### DIFF
--- a/dev/scripts/oci_replication_func.py
+++ b/dev/scripts/oci_replication_func.py
@@ -1,3 +1,7 @@
+#
+# Copyright 2022 Red Hat Inc.
+# SPDX-License-Identifier: Apache-2.0
+#
 ##########################################################################################
 # Script to collect cost/usage reports from OCI and replicate them to another bucket
 #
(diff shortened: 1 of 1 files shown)


dev/scripts/oci_replication_func.py
```diff
@@ -75,23 +80,29 @@ def copy_reports_to_bucket(
     last_reports,
 ):
     # Iterate through cost reports list and copy them to new bucket
+    # Get todays datetime
+    today = datetime.datetime.now()
+    # Start from current month
+    start_from = today.replace(day=1)
```
Red-HAP, line 86: suggestion: This is the only block where you use `today`. Perhaps this would be better:
```suggestion
    # Start from current month
    start_from = datetime.date.today().replace(day=1)
```

Red-HAP (changes requested):

Red-HAP (approved):

PR #3657, opened 2022-05-19
Cost 2639 targeted provider delete cascade

Red-HAP:
Fixes [COST-2639](https://issues.redhat.com/browse/COST-2639)

Changes proposed in this PR:
* Create new targeted-SQL cascade_delete function
    * This is using catalog SQL queries to get linked table information
    * The code also uses name matching to get top-level tables and specific `api_*` tables
    * The code will also follow 2nd-level linking as well

Testing Instructions:
1. Checkout this branch
3. `make load-test-customer-data`
4. `python ./koku/manage.py shell`
5. Execute the following comands
    ```
    from koku.database import get_model
    Provider = get_model("Provider")
    paws = Provider.objects.filter(type__contains="AWS").first()
    paws is not None  # Verify that we got a record
    paws.delete()  # You will see logs for each action taken and number of recs deleted
    Provider.objects.filter(uuid=paws.pk).count()  # Should return zero
    #
    # Try it with other provider types as well
    #
    ```


InsightsDroid:
01108e41-a818-4433-bfb8-d43038adddde
675af430-1ce6-4074-a2d0-73432709198d

koku/api/provider/models.py
```diff
@@ -197,12 +197,187 @@ def delete(self, *args, **kwargs):
...
+                "ui_table_sregex": f"^reporting_{normalized_type}_",
+                "ui_table_sval": 1,
+                "ocp_type_sregex": f"^reporting_ocp({normalized_type}|all)",
+                "ocp_type_sval": 2,
+                "daily_summ_sregex": "_daily_summary",
+                "daily_summ_sval": 3,
+                "api_sregex": "^api_",
+                "api_sval": 10,
+                "default_sval": 5,
+            }
```
maskarb, line 356: Many of these are constants. Can we put those directly into the sql instead of parameterizing them?
example: instead of:
```
and ft.relkind = any(%(relkind)s)
```
can we do:
```
and ft.relkind = any(["p", "r"])
```

koku/api/provider/models.py
```diff
@@ -197,12 +197,187 @@ def delete(self, *args, **kwargs):
             using = router.db_for_write(self.__class__, isinstance=self)
             with schema_context(self.customer.schema_name):
                 LOG.info(f"PROVIDER {self.name} ({self.pk}) CASCADE DELETE -- SCHEMA {self.customer.schema_name}")
-                cascade_delete(self.__class__, self.__class__.objects.filter(pk=self.pk))
+                with transaction.atomic():
+                    self._cascade_delete()
+                    self._delete_from_target(
```
maskarb, line 202: At this point, the only thing left is the Provider, right? Can we just call `super().delete()`?

koku/api/provider/models.py
```diff
@@ -197,12 +197,187 @@ def delete(self, *args, **kwargs):
...
+            LOG.warning("Customer link cannot be found! Using ORM delete!")
             super().delete()
 
+    def _get_normalized_type(self):
+        """Normalize the provider type: Return lowercase type name w/o '-local'"""
+        _normalized_type = getattr(self, "_normalized_type", None)
+        if _normalized_type is None:
+            _type = self.type.lower()
+            _normalized_type = self._normalized_type = _type if "-local" not in _type else _type[: _type.index("-")]
+        return _normalized_type
```
maskarb, line 218: This function seems a bit unreadable (I don't understand why there are so many underscores). Also, instead of calling this in `cascade_delete`, can we set the `self._normalized_type` in the `delete()` func? That way, we don't need to call this function more than once and we needn't lookup whether this attribute exists or not.

```suggestion
suffix = "-local"
type = self.type.lower()
if type.endswith(suffix):
    type = type[:-len(suffix)]
return type
```

koku/api/provider/models.py
```diff
@@ -197,12 +197,187 @@ def delete(self, *args, **kwargs):
...
+            seen.add(_target_vals)
+            if target_info["table_name"] == "api_providerinfrastructuremap":
+                self._set_infrastructure_id_null()
+
+            if target_info["table_name"] in _cascade_branch_tables:
+                # Keep the slice up-tp-date with the number of public schema table names in _cascade_branch_tables
+                public_schema = (
+                    self.customer.schema_name if target_info["table_name"] in _cascade_branch_tables[1:] else "public"
+                )
+                LOG.info(f"DELETE CASCADE BRANCH TO {target_info['table_name']}")
```
maskarb, line 262: Let's change these to DEBUG statements. Otherwise there is a _lot_ of stuff logged. I think all log statements in `_cascade_delete` should be debug, and the ones in `delete` should be info.

maskarb (commented):
Some comments/question/feedback:

koku/api/provider/models.py
Red-HAP: It just seemed safer this way if, in the future, we need to change something.
Besides, there be f-strings here for parameters.

koku/api/provider/models.py
Red-HAP: The point here is to not call `super.delete()` unless we _have_ to.
Django ORM `delete()` calls will walk down all of the relations and try to get the record into an object or create a new query and then call the `delete()` method there. It's a mess!

koku/api/provider/models.py
Red-HAP: This was done so as to not _have_ to calculate the value every time since it's static (at this point) once the record is retrieved.

maskarb: Sure, but if we just set this value in the `delete` method, this entire function would only be called once, no?

koku/api/provider/models.py
maskarb: After we've called `self._cascade_delete()`, what is left? Is it not just the provider at this point?

Fozzywozzybear (dismissed):
Tested with OCP AZURE GCP AWS, all work LGTM

shannon-donahue (approved):
lgtm

PR #3668, opened 2022-05-24
[COST-2654] - do not build a list of tags to compare against

maskarb:
Enhancement for [COST-2654](https://issues.redhat.com/browse/COST-2654)

Changes proposed in this PR:
* instead of adding prefixes to all of the tags, this PR strips the various `tag:` prefixes from the query param for comparison against the list of tags.
*


adberglund (approved):

InsightsDroid:
b86bad65-44e9-4039-a113-b41cd6c7c163
ed205578-122e-4b91-b0d5-df0e1355d7c4

PR #3676, opened 2022-05-26
remove filter limit/offset for tags

maskarb:
Changes proposed in this PR:
* Remove `limit` and `offset` from FilterSerializer for tags endpoints. These are not valid queries.

Testing Instructions:
1. Go to `http://localhost:8000/api/cost-management/v1/tags/azure/?filter[limit]=101&filter[offset]=2`
2. Expect to see this result:
```
{
    "filter": {
        "offset": [
        ],
        "limit": [
        ]
    }
}
```

---
Additional Context:
Previous, we would see this error response, which doesn't make sense for this endpoint:
```
{
    "error": [
        "filter[limit] and filter[offset] requires a valid group_by param."
    ]
}
```


djnakabaale (approved):
lgtm

InsightsDroid:
896c2555-f522-4a41-a937-ba2d8f738cd2
e97857d9-7ef0-4375-98b3-907885f96a71

PR #3689, opened 2022-06-03
added gauges to grafana dash for celery

ddonahue007:
[COST-2696](https://issues.redhat.com/browse/COST-2696) 

add gauges to grafana dash for celery

Screen Shot 2022-06-03 at 9 52 02 AM


abaiken (approved):
LGTM

PR #3695, opened 2022-06-07
raise exception after logging in execute_delete_sql

Red-HAP:
Fixes [COST-2710](https://issues.redhat.com/browse/COST-2710)

Changes proposed in this PR:
* re-raise exception after logging in `execute_delete_sql`

Testing Instructions:
1. checkout branch
3. `python ./koku/manage.py shell`
4. Execute:
    ```
    from koku.database import get_model
    from koku.database import execute_delete_sql
    Customer = get_model("Customer")
    Provider = get_model("Provider")
    cq = Customer.objects.all()
    pq = Provider.objects.filter(uuid=Provider.objects.first().uuid)
    execute_delete_sql(pq)  # no exception
    execute_delete_sql(cq)  # exception thrown after error log
    from django.db import connection
    from koku.database import cascade_delete
    with connection.cursor() as cur:
        cur.execute("drop table api_provider cascade;")

    cascade_delete(Customer, cq)  # Should see warning logs saying Provider table does not exist; but NO EXCEPTIONS
    ```


koku/koku/database.py
```diff
@@ -270,8 +270,8 @@ def execute_delete_sql(query):
     try:
         return execute_compiled_sql(sql, params=params)
     except Exception as e:
-        LOG.debug(f"The following exception occurred {e}")
-        return 0
+        LOG.error(f"execute_delete_sql :: The following exception occurred: {e}")
+        raise e
```
maskarb, line 274: The purpose of this try/except is to prevent an exception from stopping Provider deletes. Is there a way to fix this without re-raising an exception?

Also, can we log this as a warning instead of error?

maskarb: This was the original issue where we added the try/except:
[COST-1846](https://issues.redhat.com/browse/COST-1846)

The original exception was a failure to delete a table that did not exist. Something like that should not prevent us from deleting a provider.

Red-HAP: Changed to only re-raise on DB exceptions to cancel (sub)transactions otherwise handle "regular" exceptions more gracefully.

Red-HAP: > This was the original issue where we added the try/except: [COST-1846](https://issues.redhat.com/browse/COST-1846)
> 
> The original exception was a failure to delete a table that did not exist. Something like that should not prevent us from deleting a provider.

If that was the case (I actually thought I put in the try/except block myself -- that's how long it had been since I looked at that code), that was the wrong approach to fix the initial problem.
If you try to delete from a table that does not exist, then that is a database error that **will** invalidate a transaction.
To get around this, we need to make sure that the table exists beforehand. (ideally, there will be no significant differences in the schemata that this would hit).

I have added a check in `cascade_delete` to detect this (assuming that's where the error came from) but the sentry link had expired and I could not see the traceback.

In any case there is no such thing as 
```sql
delete from some_table if exists where ...
```

InsightsDroid:
5e296acd-c0d4-44d5-b561-6a3b6f5a2765
c9497e02-7cd3-406e-9f85-0ad7046d9368

maskarb (approved):
lgtm

PR #3708, opened 2022-06-14
[COST-1408] Adding currency to report APIs utilizing pandas 

abaiken:

Changes proposed in this PR:
*
*

Testing Instructions:
1. Run `this command` ...
3. Expect to see this result:
   ```

   {json: response}
   ```
4.
5.

---
Additional Context:


InsightsDroid:
dd170daf-e2de-4eaf-b2d5-a04c8deecc47

koku/api/report/all/openshift/provider_map.py
```diff
@@ -395,7 +395,7 @@ def __init__(self, provider, report_type):
                             "cost_usage": Sum(Value(0, output_field=DecimalField())),
                             "cost_markup": Sum(Coalesce(F("markup_cost"), Value(0, output_field=DecimalField()))),
                             "cost_units": Coalesce(Max("currency_code"), Value("USD")),
-                            "count": Count("resource_id", distinct=True),
+                            "count": ArrayAgg("resource_id", distinct=True),
```
adberglund, line 398: ArrayAgg here will return a list of resource ids. Is that what we want?

koku/api/report/azure/openshift/query_handler.py
```diff
@@ -126,11 +144,10 @@ def execute_query(self):  # noqa: C901
             else:
                 query_data = self.order_by(query_data, query_order_by)
 
-            cost_units_value = self._mapper.report_type_map.get("cost_units_fallback", "USD")
             usage_units_value = self._mapper.report_type_map.get("usage_units_fallback")
             count_units_value = self._mapper.report_type_map.get("count_units_fallback")
             if query_data:
-                cost_units_value = query_data[0].get("cost_units")
+                # cost_units_value = query_data[0].get("cost_units")
```
adberglund, line 150:
```suggestion
```

koku/api/report/gcp/query_handler.py
```diff
@@ -3,6 +3,8 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 """GCP Query Handling for Reports."""
+from __future__ import annotations
```
adberglund, line 6: Was this a vs code auto-import? Guessing we don't want this.

abaiken: Most definitely was an auto-import 👍  I removed it - thank you!

koku/api/report/all/openshift/provider_map.py
myersCody: Hey Andrew! 

This change stemmed from some smoke test failures where the count was unusually high. I discovered that the count of the resource_ids was essentially loosing the "distinct" during the pandas implementation. Therefore in order to deduplicate the results again, I changed it to a list and [implemented some deduplication](https://github.com/project-koku/koku/pull/3708/commits/df51675ea0fda099a3bd3d1899e02dff88336d50#diff-7856c4ba653f802f9901aa185f9dfcd3754d01fe20a6fa824b6443d31b88d48aR715) for when we are calculating the totals. 

I am open to some groovy pandas alternatives if they exist, I just couldn't figure them out myself 😂

adberglund: @myersCody I think it's likely because the count term is being summed, when we want the maximum value for count. 

`aggs = {col: ["max"] if "units" in col else ["sum"] for col in annotations if col not in skip_columns}`

adberglund (approved):
It might still be worth looking later at the count stuff, but since we have a working solution here, I wouldn't change in now!

PR #3716, opened 2022-06-20
Cost 1329 initial enabled keys migration

Red-HAP:
## Jira Ticket

[COST-1329](https://issues.redhat.com/browse/COST-1329)

## Description

This change will update the PR template and will perform the initial (and completely compatible) migrations to add the `enabled boolean` column to the following tables:

* reporting_azureenabledtagkeys
* reporting_gcpenabledtagkeys
* reporting_ocpenabledtagkeys

This is being done to support tag-preprocessing as is being done for AWS.

## Testing

1. Checkout `main`
2. `make docker-down && make remove-db && make docker-up-no-build && sleep 20 && make docker-reinitdb-with-sources`
3. `psql postgresql://postgres@localhost:15432/postgres`
    1. `\d template0.reporting_azureenabledtagkeys`
    2. `\d template0.reporting_gcpenabledtagkeys`
    3. `\d template0.reporting_ocpenabledtagkeys`
    4. There will be no `enabled` column of `boolean` type on **any** of these tables.
5. `make run-migrations`
6. `psql postgresql://postgres@localhost:15432/postgres`
    1. `\d template0.reporting_azureenabledtagkeys`
    2. `\d template0.reporting_gcpenabledtagkeys`
    3. `\d template0.reporting_ocpenabledtagkeys`
    4. There will be an `enabled` column of `boolean` type on **all** of these tables.

## Notes

This change only contains the migration and an update to the PR template.

PR #3717, opened 2022-06-20
Cost 1329 azure parquet tag processing

Red-HAP:
## Jira Ticket

[COST-1329](https://issues.redhat.com/browse/COST-1329)

## Dependency

#3716 must be released to staging and prod first.

## Description

This change will utilize the `enabled` column. The `id` column will be unaffected as the `key` column is the true primary key.

The database table utilized in these changes will be `reporting_azureenabledtagkeys`.

## Testing

1. Checkout branch
2. Restart Koku
3. Koku Settings
    1. Check the DB for the enabled keys.
    2. Verify that the settings page only shows these keys as enabled
    3. Use the UI to disable one or more keys
    4. Verify that the keys are disabled on a page reload
    5. Verify in the DB that the keys in the table still exist but the `enabled` column is set to `false`
    6. Enable a key
    7. Verify that the key is enabled on page reload
    8. Verify in the DB that the key exists in the table with `enabled` set to `true`
4. Go to the cost_explorer
    1. Verify that only enabled keys are in the interface controls.
    2. If a disabled key is used in the explorer interface, data or no data should return as normally expected
    3. Disable and enable keys in the settings interface to ensure that the results are as expected in the controls and in the data.
5. Ensure that local processing works for PG or Trino modes.


InsightsDroid:
838856d9-4f8b-4b0f-a465-4645ebe0b43c

adberglund (approved):

PR #3725, opened 2022-06-27
[Cost-2026]order_by date cannot sort on a day without data

Fozzywozzybear:
Fixes [COST-2026](https://issues.redhat.com/browse/COST-2026)

Changes proposed in this PR:
Now filtering on a date without data does not remove all valid dates with data from the query. 

Testing Instructions:
1. Spin up docker, make test customer data and load it. 
2. Since Only two months of data are generated we will access a date that is out of range and has no data. 
3. Hit this endpoint http://localhost:8000/api/cost-management/v1/reports/gcp/costs/?filter[limit]=10&filter[offset]=0&group_by[gcp_project]=*&order_by[cost]=desc&order_by[date]=2022-04-30&end_date=2022-06-02&start_date=2022-04-30
4. Expected to see that 04-30 has no data but all other valid dates have data in the given range from 04-30 to 06-02

---
Additional Context:
Please reach out with any questions!


InsightsDroid:
https://url.corp.redhat.com/ibutsu-runs-bee72940-62d1-41f4-ac8d-bb1e34f0bca6

koku/api/report/aws/openshift/query_handler.py
```diff
@@ -75,6 +75,10 @@ def execute_query(self):  # noqa: C901
             order_date = None
             for i, param in enumerate(query_order_by):
                 if check_if_valid_date_str(param):
+                    # Checks to see if the date is in the query_data
+                    if not any(d["date"] == param for d in query_data):
+                        break
+                    # else we set order_date to a valid date
                     order_date = param
                     break
```
Red-HAP, line 83: **Question**

I'm wondering if this would be more efficient:

```suggestion
            order_date = None
            for i, param in enumerate(query_order_by):
                if check_if_valid_date_str(param):
                    for d in query_data:
                        found = d["date"] == param
                        if found:
                            order_date = param
                            break
                    if order_date is not None:
                        break
```

The intent here would be to stop on the first hit instead of running though the entire query_data list every time we hit a date.

Also, it seems to me that this is replicated a lot. Looks like a good candidate to be put into a common module.

Red-HAP (changes requested):

koku/api/report/aws/openshift/query_handler.py
Fozzywozzybear: I changed the function to 
                if check_if_valid_date_str(param):
                    # Checks to see if the date is in the query_data
                    if any(d["date"] == param for d in query_data):
                        # Set order_date to a valid date
                        order_date = param
                        break
I think this is the most efficient method

Red-HAP: So long as `any` will short-circuit on the first true-y value, then yes.

adberglund: https://docs.python.org/3/library/functions.html#any

Fozzywozzybear: Given with Andrews comment the any() will achieve the same goal

Red-HAP (approved):

PR #3726, opened 2022-06-27
Adding start/end dates

lcouzens:
Fixes [COST-2759](https://issues.redhat.com/browse/COST-2759)

Changes proposed in this PR:
* Adds start end dates to cache_args to enable multiple month simultaneous processing
*

Testing Instructions:
1. start up with multiple koku workers - Add scale=2 (number of workers) to your up command
2. Ingest data spanning multiple months
3. see that each worker picks up a different month for summary regardless if a different month is currently in progress

---
Additional Context:


koku/masu/processor/tasks.py
```diff
@@ -358,7 +358,7 @@ def update_summary_tables(  # noqa: C901
     """
     worker_stats.REPORT_SUMMARY_ATTEMPTS_COUNTER.labels(provider_type=provider).inc()
     task_name = "masu.processor.tasks.update_summary_tables"
-    cache_args = [schema_name, provider, provider_uuid]
+    cache_args = [schema_name, provider, provider_uuid, start_date, end_date]
```
adberglund, line 361: I think this leaves us open to running multiple tasks for the same month. How about taking the start date and replacing the day with the first of the month and just passing that in as our one date arg? That way we restrict per month.

InsightsDroid:
https://url.corp.redhat.com/ibutsu-runs-a9690560-2a06-46d4-8faa-de26545dc70d

adberglund (approved):

PR #3727, opened 2022-06-27
Cost 1329 final azure enabled tag key migration

Red-HAP:
## Jira Ticket

[COST-1329](https://issues.redhat.com/browse/COST-1329)

## Dependency

#3717  must be released to staging and prod first.

## Description

This is a migration that will remove the `id` column and unique constraint on `key` and create a primary key constraint on `key`

## Testing

1. Checkout branch
2. Run migrations `make run-migrations`
3. `psql postgresql://postgres@localhost:15432/postgres`
    1. `\d template0.reporting_azureenabledtagkeys` 
    2. Verify that there is no `id` column and the primary key constraint is on the `key` column
    3. \d template0.reporting_gcpenabledtagkeys` 
    4. Verify that the `id` column exists
    5. Exit `psql`
4. Run reverse migration `make run-migrations applabel=reporting migration=0256
6. `psql postgresql://postgres@localhost:15432/postgres`
    1. `\d template0.reporting_azureenabledtagkeys` 
    2. Verify that the `id` column exists and is the primary key and the `key` column only has a unique constraint
    3. \d template0.reporting_gcpenabledtagkeys` 
    4. Verify that the `id` column exists
    5. Exit `psql`


InsightsDroid:
https://url.corp.redhat.com/ibutsu-runs-93ea24c3-c27c-4a38-9296-ff9c6c6de5a4

koku/api/settings/settings.py
```diff
@@ -247,37 +247,37 @@ def _tag_key_handler(self, settings):
...
+                        enabled_tags_no_abbr.append(enabled_tag.split(tag_delimiter, 1)[1])
+
+                invalid_keys = [tag_key for tag_key in enabled_tags_no_abbr if tag_key not in available]
+
+                if invalid_keys:
+                    key = "settings"
+                    message = f"Invalid tag keys provided: {', '.join(invalid_keys)}."
+                    raise ValidationError(error_obj(key, message))
+
+                if provider_name in ("aws", "azure"):
```
lcouzens, line 271: Should we do something like this instead here? `if provider_name in (Provider.PROVIDER_AZURE, Provider.PROVIDER_AWS):`

Red-HAP: Nope. It's referring to a dict defined at the top of the file.

lcouzens (approved):
lgtm

lcouzens:
I'm just going to close this. If we need it lets re-open it.

PR #3733, opened 2022-07-05
Utilize managed kafka user, paswd, sasl_mech, sec_proto

Red-HAP:
## Jira Ticket

[COST-1380](https://issues.redhat.com/browse/COST-1380)

## Description

This change utilizes the **managed** Kafka username, password, sasl_mechanism, and security_protocol in the `Configurator` and when creating Kafka `Producer` and `Consumer`

## Testing

1. Checkout Branch
2. Restart Koku
3. Run tox
4. Deploy to ephemeral (should not be any errors in kafka listener, producer code

## Notes

Once approved, this should get deployed to staging for another system test.

### DIFF
--- a/koku/koku/configurator.py
+++ b/koku/koku/configurator.py
@@ -216,6 +216,26 @@ class EnvConfigurator(Configurator):
         return requestedName
 
     @staticmethod
+    def get_kafka_username():
+        """Obtain kafka username"""
+        return None
+
+    @staticmethod
+    def get_kafka_password():
+        """Obtain kafka password"""
+        return None
+
+    @staticmethod
+    def get_kafka_sasl_mechanism():
+        """Obtain kafka sasl mechanism"""
+        return None
+
+    @staticmethod
+    def get_kafka_security_protocol():
+        """Obtain kafka security protocol"""
+        return None
+
+    @staticmethod
     def get_cloudwatch_access_id():
         """Obtain cloudwatch access id."""
         return ENVIRONMENT.get_value("CW_AWS_ACCESS_KEY_ID", default=None)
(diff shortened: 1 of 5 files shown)


koku/koku/configurator.py
```diff
@@ -380,6 +400,26 @@ def get_kafka_topic(requestedName: str):
...
+
+    @staticmethod
+    def get_kafka_sasl_mechanism():
+        """Obtain kafka sasl mechanism"""
+        return LoadedConfig.kafka.brokers[0].sasl.saslMechanism
+
+    @staticmethod
+    def get_kafka_security_protocol():
+        """Obtain kafka security protocol"""
+        return LoadedConfig.kafka.brokers[0].sasl.securityProdocol
```
maskarb, line 421:
```suggestion
        return LoadedConfig.kafka.brokers[0].sasl.securityProtocol
```
It looks like this field was added in app-common-python yesterday, so we likely need to update to the latest version.

Also, `LoadedConfig.kafka.brokers[0]` contains `cacert` and `authtype` fields, I assume we need to do something with those too.

maskarb (commented):
These were added very recently and also need to be updated:
https://github.com/project-koku/koku/blob/main/koku/kafka_utils/utils.py#L21-L40

koku/koku/configurator.py
Red-HAP: Thanks!

koku/koku/configurator.py
```diff
@@ -380,6 +410,36 @@ def get_kafka_topic(requestedName: str):
...
+
+    @staticmethod
+    def get_kafka_sasl_mechanism():
+        """Obtain kafka sasl mechanism"""
+        return LoadedConfig.kafka.brokers[0].sasl.saslMechanism
+
+    @staticmethod
+    def get_kafka_security_protocol():
+        """Obtain kafka security protocol"""
+        return LoadedConfig.kafka.brokers[0].sasl.securityProtocol
```
maskarb, line 431: The issue that we are hitting right now in ephemeral is that `sasl` is not configured there. I think instead of these 4 methods, we could just use 1 that returns the sasl config that can be used directly in the Kafka configs.

Maybe something like:
```
    @staticmethod
    def get_kafka_sasl():
        """Obtain kafka sasl config"""
        conf = {}
        sasl = LoadedConfig.kafka.brokers[0].sasl
        if sasl:
            conf["security_protocol"] = sasl.securityProtocol
            conf["sasl_mechanism"] = sasl.saslMechanism
            conf["sasl_plain_username"] = sasl.username
            conf["sasl_plain_password"] = sasl.password
        return conf
```
or something similar.

maskarb (approved):

PR #3738, opened 2022-07-06
Fix semantic error on organizations endpoint in API docs 

abaiken:
## Jira Ticket

I noticed we had a semantic error on our api docs for the organizations endpoint. 

## Description
Create a OrganizationFilter parameter reference so that the parameters section doesn't point to a schema 

This change will ...

## Testing

1. Copy/paste the openapi.json into https://editor.swagger.io/ and notice that there is no longer a semantic error

## Notes

...

### DIFF
--- a/docs/specs/openapi.json
+++ b/docs/specs/openapi.json
@@ -924,9 +924,6 @@
                     "basic_auth": []
                 }],
                 "parameters": [{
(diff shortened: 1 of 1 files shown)


ddonahue007 (approved):
LGTM

PR #3747, opened 2022-07-18
fix unleash schema strategy so it handles exact match

ddonahue007:
## Jira Ticket

[COST-2811](https://issues.redhat.com/browse/COST-2811)

## Description
Fix for unleash schema-strategy


myersCody (approved):
Sweet Demo

PR #3748, opened 2022-07-19
Added ability to check for invalid request params

ddonahue007:
## Jira Ticket

[COST-2812](https://issues.redhat.com/browse/COST-2812)

## Description
Added the ability to check for bad request parameters

## Testing
Start up koku normally. no need to run create customer or load customer targets

run the following masu endpoint and you will see that any none excepted params will cause and error.

`/api/cost-management/v1/hcs_report_finalization/?bogus=nothing'`

Should see something like the following in the logs
```
[2022-07-19 12:30:08,276] INFO None {'method': 'GET', 'path': '/api/cost-management/v1/hcs_report_finalization/?bogus=nothing', 'status': 400, 'request_id': None, 'account': None, 'username': None, 'is_admin': False, 'response_time': '125.58 ms'}
```

Client response would be
```
{
    "Error": "bogus: is not a valid Request parameter.Valid params: ('month', 'year', 'provider_type', 'provider_uuid', 'schema_name')"
}
```
...

### DIFF
--- a/koku/masu/api/hcs_report_finalization.py
+++ b/koku/masu/api/hcs_report_finalization.py
@@ -31,11 +31,11 @@ def hcs_report_finalization(request):
 
     if request.method == "GET":
         params = request.query_params
-        expected_params = ("month", "year", "provider_type", "provider_uuid", "schema_name")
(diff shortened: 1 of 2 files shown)


lcouzens (approved):
lgtm

PR #3779, opened 2022-08-01
[COST-2621] remove org-id populator job from clowdapp

cgoodfred:
## Jira Ticket

[COST-2621](https://issues.redhat.com/browse/COST-2621)

## Description

Now that the job has run in prod and the code released, we can remove the job from our clowdapp.

...


maskarb (approved):
lgtm!

PR #3804, opened 2022-08-10
[COST-2867] return if no report period for delete tag usage costs

cgoodfred:
## Jira Ticket

[COST-2867](https://issues.redhat.com/browse/COST-2867)

## Description

This change will log a warning and return if there is not a report period for a given provider and usage start when deleting tag usage costs.

...


maskarb (approved):
lgtm!

PR #3809, opened 2022-08-11
COST-2682:  Summarize by a list of manifests

myersCody:
## Jira Ticket

[COST-2682](https://issues.redhat.com/browse/COST-2682)

## Description

This change will change a portion of our summary processing logic. We are used to working on monthly manifests, but a change went into GCP to where we get a manifest per day per download to track the export time of the data provided to us. 

Resummarizing by day slows down the resummary for GCP and had issues with the manifests being marked as complete. 

The changes in this PR do the following:
1. Remove an unnecessary secondary function call that was recently added. The function is only ever gonna be called once and it makes it harder to trace the code. 
2. Move the summary task call outside of the for loop for the manifest list. 
3. Update the mark manifest as complete to accept a list instead of just an id. 

Changes went in earlier to main to resummarize by a date range for our continuous sources, so this is more of a catch for some of the fallout of the gcp changes. 

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Notes

...


koku/masu/processor/tasks.py
```diff
@@ -349,6 +350,7 @@ def update_summary_tables(  # noqa: C901
     synchronous=False,
     tracing_id=None,
     ocp_on_cloud=True,
+    manifest_list=None,
 ):
```
cgoodfred, line 354: So this function now takes both a `manifest_id` and a `manifest_list`? I'm not sure I understand why both should be inputs to the function, shouldn't it just accept one or the other?

myersCody:
Smoke Tests: 
- https://ci.ext.devshift.net/job/project-koku-koku-pr-check/6302/

koku/masu/processor/tasks.py
myersCody: Sounds good I will get started on the changes. 🙇

cgoodfred: I'm not sure I know what changes you are talking about, are you going to change the function to only accept one of those two parameters?

myersCody: So, I was exploring the idea of passing a `manifest_list` into the `ReportSummaryUpdater` class. However, that started to feel like a can of worms. 

### Why we pass in manifest_id as an optional parameter

I did discover why the manifest_id is passed in to that class though, we essentially use the manifest in some of our cloud specific (AWSReportParquetSummaryUpdater) classes in the `_get_sql_inputs` method to decide the start & end date for our summary sql. We don't actually use it for GCP & OCI, but since it is used in some of the child classes we have to pass it to all of the child classes. 

### Why I needed to pass in manifest list

So, GCP is the only downloader that returns a list of manifests instead of one. Part of the work I am doing in this pull request is no longer call a summary task per manifest by moving [this section](https://github.com/project-koku/koku/blob/65cb6dfb51dcb179fcfda3a2698c1fad7c25b9a3/koku/masu/processor/orchestrator.py#L243-L254) outside of the for loop for `manifest_list` which creates our processing task.

However, now that we have one summary task for GCP, I ran into an issue where only one manifest is being marked as complete since we mark manifests as complete at the end of the summary task. My solution to this was to pass in a manifest_list variable and update method that marks them to accept a list. 

Thanks for the question @cgoodfred

I think since they are both optional parameters I am gonna leave as is for now.

lcouzens (approved):
Fantastic work here dude!

myersCody:
Smoke test results: https://ci.ext.devshift.net/job/project-koku-koku-pr-check/6315/

PR #3815, opened 2022-08-17
COST-2892: adding invoice month to GCP summary tasks

lcouzens:
## Jira Ticket

[COST-2892](https://issues.redhat.com/browse/COST-2892)

## Description 

Add GCP invoice month to summary tasks. This is because we miss month bound crossover data when running summary via standard dates. With this change we can pass in start/ends past a month boundary with the inclusion of invoice month so the full billed month is summarised rather than just a date range. 

## Testing

1. Checkout Branch
2. Restart Koku
3. Add real GCP data
4. Verify all the numbers add up correctly per month

## Notes

...

### DIFF
--- a/koku/api/resource_types/test/gcp_accounts/test_views.py
+++ b/koku/api/resource_types/test/gcp_accounts/test_views.py
@@ -3,7 +3,6 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 """Test the Resource Types views."""
-import datetime
 from unittest.mock import patch
 from uuid import uuid4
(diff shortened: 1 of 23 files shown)


lcouzens:
Failing smoke test is OCP/GCP probably failing because of https://github.com/project-koku/koku/pull/3815/files#diff-1477f4923c6c64b0cc0a78af7738c365d58ac4200fba76d2c6a57293f75f8f6aR602

myersCody (approved):

PR #3823, opened 2022-08-20
Adding the ability to disable a specific schema from processing

lcouzens:
## Jira Ticket

[COST-2903](https://issues.redhat.com/browse/COST-2903)

## Description

This change will give the ability to disable a specific schema's cloud sources from scheduling download tasks during polling. Done via unleash.

## Testing

1. Checkout Branch
2. Restart Koku
3. Add schema to disable_cloud_source_processing flag in unleash
4. Add a source under specific schema and see no data is ingested + logs say schema processing is disabled.

## Notes

...

### DIFF
--- a/.unleash/flags.json
+++ b/.unleash/flags.json
@@ -56,6 +56,25 @@
             "createdAt": "2022-08-12T21:22:00.756Z"
         },
         {
+            "name": "disable_cloud_source_processing",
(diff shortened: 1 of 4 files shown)


myersCody:
I think this may disable the download endpoint too, let me see if I can find a way to not do that real quick.

myersCody (approved):

PR #3829, opened 2022-08-24
Modify purge endpoint to not delete manifests.

myersCody:
## Jira Ticket

None

## Description

We have run into the problem where the parquet compactor has run for a customer's account. However, we do not want to download the entire month for GCP. Right now the GCP downloader is heavily tied into if manifests exist for that partition date or not.

1. So we are planning to run it for the entire month without deleting the manifest first.
^^^ This will allow us to remove all parquet & csv files for the month

2. Then we will run it to delete a specific range of time for the manifests & download just those days.  

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Notes

...

### DIFF
--- a/koku/masu/api/purge_trino_files.py
+++ b/koku/masu/api/purge_trino_files.py
@@ -68,6 +68,9 @@ def purge_trino_files(request):  # noqa: C901
 
     start_date = params.get("start_date")
     end_date = params.get("end_date") if params.get("end_date") else start_date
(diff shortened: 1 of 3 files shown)


myersCody:
I actually don't think there is a need to run smokes on this one. All the things I touched will not be tested by smoke tests anyway.

lcouzens (approved):

PR #3831, opened 2022-08-24
Disable ingest when marking provider setup_complete as false

myersCody:
## Jira Ticket

[COST-2909](https://issues.redhat.com/browse/COST-2909)

## Description

So, as part of the purge endpoint, we mark the manifest as `setup_complete=False` so that we can trigger an "initial ingest" for GCP. However, we discovered that when we were hitting save it would kick off a download for the entire month in the priority worker.

Funny enough it was also kicking off downloads for other accounts. 

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Notes

...


myersCody:
Smoke Tests: https://ci.ext.devshift.net/job/project-koku-koku-pr-check/6399/

lcouzens (approved):

PR #3835, opened 2022-08-26
COST-2914: stop tag matching if no matching enabled keys

myersCody:
## Jira Ticket

[COST-2914](https://issues.redhat.com/browse/COST-2914)

## Description

This change will ...

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Notes

...


myersCody:
One azure smoke test failure: https://ci.ext.devshift.net/job/project-koku-koku-pr-check/6416/

Gonna run it again.

koku/masu/database/gcp_report_db_accessor.py
```diff
@@ -588,6 +588,18 @@ def delete_ocp_on_gcp_hive_partition_by_day(self, days, gcp_source, ocp_source,
...
+            SELECT COUNT(*) FROM {self.schema}.reporting_gcpenabledtagkeys as gcp
+                INNER JOIN {self.schema}.reporting_ocpenabledtagkeys as ocp ON gcp.key = ocp.key;
+        """
+        with connection.cursor() as cursor:
+            cursor.db.set_schema(self.schema)
+            cursor.execute(match_sql)
+            results = cursor.fetchall()
+            if results[0][0] < 1:
+                LOG.info(f"No matching enabled keys for OCP on GCP {self.schema}")
+                return None
```
maskarb, line 602: The return statement below returns a list. Can we replace this `None` with an empty list to keep the return types the same?

koku/masu/processor/parquet/ocp_cloud_parquet_report_processor.py
```diff
@@ -152,7 +152,7 @@ def process(self, parquet_base_filename, daily_data_frames):
             if self.has_enabled_ocp_labels:
                 LOG.info("Getting matching tags from Postgres.")
                 matched_tags = self.db_accessor.get_openshift_on_cloud_matched_tags(self.bill_id, report_period_id)
-                if not matched_tags:
+                if not matched_tags or matched_tags is not None:
```
maskarb, line 155: I'm confused by this logic. Are we wanting to run this when there _are_ matched tags?

maskarb: I can't figure out when this if statement is not True.
```
>>> matched_tags = None
>>> not matched_tags
True
>>> matched_tags is not None
False
>>> matched_tags = True
>>> not matched_tags
False
>>> matched_tags is not None
True
>>> tags = []
>>> not tags
True
>>> tags is not None
True
>>> tags = ['something']
>>> not tags
False
>>> tags is not None
True
>>> 
```

koku/masu/database/gcp_report_db_accessor.py
maskarb: Also, is this the correct method to update? Shouldn't we be updating get_openshift_on_cloud_matched_tags_trino?

koku/masu/processor/parquet/ocp_cloud_parquet_report_processor.py
myersCody: ahhhhh yes! Thank you @maskarb 

It is confusing because it is suppose to be an `and` condition and not an  `or`.  I had to go back and look at my test file 😂 

```
(koku) cmyers@cmyers1-mac koku % cat test.py
matched = []
if not matched and matched is not None:
    print(matched)

matched = None
if not matched and matched is not None:
    print(matched)

matched = "ABC"
if not matched and matched is not None:
    print(matched)%
(koku) cmyers@cmyers1-mac koku % python3 test.py
[]
 
```

Essentially, we only want to run the Trino tag matching if it is an empty `[]`. Been staying up to late man 😂 

But now that you mention it, returning a None is a bit confusing, I am gonna replace it with the string `skip`. Thanks for the extra set of eyes.

myersCody: I was debating returning a `None` or the string `skip` in my head during this and I decided to return None, but forgot to change the if condition to match. :smh:

koku/masu/database/gcp_report_db_accessor.py
myersCody: So, for context on the initial download this function will return []. We check to see if we get an empty result, we then call the trino sql which takes a long time. However, if they have no enabled tag keys that match then it will always return an empty list. I see your point about about `None` being confusing though and switch it over to the string `skip` to bypass the trino sql. Thanks for the great feedback it has been super helpful lately during these long days and much appreciated. 🙇

maskarb: So I'm not really a fan of returning `"skip"` either. Can we move this check functionality to a new method and then use that result to gate calling the `get_openshift_on_cloud_matched_tags` method?

myersCody: Yeah, I think that is an approach we could implement for this let me take a swing at it and ill ping ya when I got something figured out.

myersCody: I think we will have to implement a method like that for all the other providers as well, but we should be able to add it easily. Let me check to make sure the logic of skipping the trino tag matching is consistent with the other providers real quick by checking their SQL.

myersCody:
- [ ] Reword jira issue to remove ocp on gcp specific language.

myersCody:
Smoke Test Results: https://ci.ext.devshift.net/job/project-koku-koku-pr-check/6433/

koku/masu/processor/parquet/ocp_cloud_parquet_report_processor.py
```diff
@@ -151,8 +151,9 @@ def process(self, parquet_base_filename, daily_data_frames):
             matched_tags = []
             if self.has_enabled_ocp_labels:
                 LOG.info("Getting matching tags from Postgres.")
+                matching_enabled_tags = self.db_accessor.check_for_matching_enabled_keys()
                 matched_tags = self.db_accessor.get_openshift_on_cloud_matched_tags(self.bill_id, report_period_id)
-                if not matched_tags:
+                if not matched_tags and matching_enabled_tags:
```
maskarb, line 156: Maybe I am mis-understanding the purpose of this PR, but aren't we trying to prevent the code in `get_openshift_on_cloud_matched_tags` from running if there are no enabled tags?

myersCody:
Smoke tests running here: https://ci.ext.devshift.net/job/project-koku-koku-pr-check/6439/

koku/masu/processor/parquet/ocp_cloud_parquet_report_processor.py
myersCody: We are trying to prevent the `get_openshift_on_cloud_matched_tags_trino` from running, but after thinking about it we can also prevent the `get_openshift_on_cloud_matched_tags` from running as well.

myersCody:
Smoke tests running here: https://ci.ext.devshift.net/job/project-koku-koku-pr-check/6445/

lcouzens (approved):

PR #3841, opened 2022-08-30
[COST-1840] - set health check interval to 5 seconds

maskarb:
## Jira Ticket

[COST-1840](https://issues.redhat.com/browse/COST-1840)

## Description

Set the Redis health-check-interval to 5 seconds.


chambridge (approved):
:five:

PR #3843, opened 2022-08-31
[COST-2901] AWS/Azure only generate HCS reports for last 3 days on initial ingest

cgoodfred:
## Jira Ticket

[COST-2901](https://issues.redhat.com/browse/COST-2901)

## Description

This change will make it so AWS/Azure HCS reports are properly generated on initial ingest for the entire ingestion window. It will also determine if the ingested month should be finalized and will finalize months that meet the criteria.

## Testing

1. Checkout Branch
2. Restart Koku
3. Load test customer data like normal:
       - `make create-test-customer`
       - `make load-test-customer-data`
4. Check in minio and you should see aws/azure HCS reports for this month and previous months whereas on main you only see the last 3 days. The previous months files metadata should also include the finalized flag as true.

## Notes

...


maskarb (approved):
lgtm!

PR #3849, opened 2022-09-01
Convert download interval to a set schedule

maskarb:
## Description

* some celery file cleanup: remove old comments, remove variables from pods that don't use them, convert LOG messages to prints, print the celery configs in the appropriate pods
* use a cron schedule for report downloads instead of an interval. Default is everyday at 400 and 1600. This schedule can be set thru the `REPORT_DOWNLOAD_SCHEDULE` env variable.


deploy/kustomize/patches/worker-hcs.yaml
```diff
@@ -83,16 +83,6 @@
           value: ${WORKER_HCS_WORKER_QUEUE}
         - name: WORKER_PROC_ALIVE_TIMEOUT
           value: ${WORKER_PROC_ALIVE_TIMEOUT}
-        - name: SCHEDULE_CHECK_INTERVAL
-          value: ${SCHEDULE_CHECK_INTERVAL}
-        - name: REMOVE_EXPIRED_REPORT_DATA_ON_DAY
```
adberglund, line 88: Possible it's my eyes. How are we handling the schedule for the expired data task now? Or is that only the scheduler need actually care about this? And it's not needed in the workers? Am I answering my own question?

maskarb: Yep, you're answering your own question. Only the scheduler requires this env variable.

adberglund (approved):

PR #3850, opened 2022-09-01
Update to faster query + exclude empty string

adberglund:
## Jira Ticket

[COST-2747](https://issues.redhat.com/browse/COST-2747)

## Description

This change will run a faster query to identify OCP cluster on an AWS or Azure source. The key addition is also the WHERE condition filtering out NULL and '' values for the OCP resource, which was causing us to incorrectly associate cluster with an AWS account.

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Notes

...


adberglund:
[🚬 ](https://ci.ext.devshift.net/job/project-koku-koku-pr-check/6493/)

lcouzens (approved):

PR #3853, opened 2022-09-02
Update agg_fields for instance-type reports

djnakabaale:
- update aggregable fields for instance-type reports

## Jira Ticket

- related to [COST-2643](https://issues.redhat.com/browse/COST-2643)

This change will ...

- prevent error in pandas aggregate function when there's a `nan` value in  the `count_units` column

## Testing

1. Checkout Branch
2. Restart Koku
3. Hit endpoint
    - `http://localhost:8000/api/cost-management/v1/reports/openshift/infrastructures/azure/instance-types/?filter[limit]=3&filter[service_name]=Virtual Machine&group_by[project]=*&order_by[cost]=asc&end_date=2022-08-13&start_date=2022-07-01`
    2. You should see return a `200 status code`


maskarb (approved):
lgtm!

PR #3874, opened 2022-09-13
[COST-3011] ocp label bugfix

adberglund:
## Jira Ticket

[COST-3011](https://issues.redhat.com/browse/COST-3011)

## Description

This moves us to using the "tags" column for all ocp on cloud endpoints instead of using pod labels on the "by project" views. Because we also take this summarized data and insert it into the OCP table, we were injecting cloud tags into the OCP table. This should prevent that

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Notes

...


esebesto (approved):

myersCody (approved):
LGTM!

maskarb (approved):

PR #3879, opened 2022-09-15
Add unleash flag to disable GCP resource matching

lcouzens:
## Jira Ticket

[COST-3024](https://issues.redhat.com/browse/COST-3024)

## Description

This change will add an unleash flag to disable GCP resource matching

## Testing

1. Checkout Branch
2. Restart Koku
3. Add schema to unleash
4. Add OCP on GCP sources with matching resource data and see we fall back to tag matching

## Notes

...

### DIFF
--- a/.unleash/flags.json
+++ b/.unleash/flags.json
@@ -56,7 +56,7 @@
             "createdAt": "2022-08-12T21:22:00.756Z"
         },
         {
-            "name": "disable_gcp_resource_matching",
+            "name": "cost-management.backend.disable_gcp_resource_matching",
             "description": "Toggle to disable gcp resource matching defaulting back to tag matching.",
             "type": "permission",
             "project": "default",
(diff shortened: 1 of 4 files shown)


.unleash/flags.json
```diff
@@ -55,6 +55,25 @@
             "variants": [],
             "createdAt": "2022-08-12T21:22:00.756Z"
         },
+        {
+            "name": "disable_gcp_resource_matching",
```
myersCody, line 59: So, Corey is currently working on renaming our unleash flags to the following naming convention:
```
cost-management.backend.{unleash_flag}
```
We should go ahead and rename this flag now to save some rework for him.

lcouzens: Great idea

myersCody: I can knock that out for you if you would like

myersCody (approved):

PR #3891, opened 2022-09-20
[COST-3040] provider delete ordering

adberglund:
## Jira Ticket

[COST-3040](https://issues.redhat.com/browse/COST-3040)

## Description

This change will remove the wrapped transaction.atomic, and rely exclusively on the call on the provider model delete. We are also switching to deleting the actual api_provider entry as an on_commit callback to force this to happen only after committing the change to delete all related data. 

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Notes

...


maskarb (dismissed):
lgtm!

maskarb (approved):
lgtm!

PR #3898, opened 2022-09-26
[COST-3040] Allow other months to be considered

adberglund:
## Jira Ticket

[COST-3040](https://issues.redhat.com/browse/COST-3040)

## Description

This removes a filter on the the is processing check to allow us to look for other months that are processing. The manifest creation timestamp will still be the current date, but particularly for smoke tests we generate manifests for multiple months while processing. 

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Notes

...


maskarb (approved):
lgtm!

PR #3904, opened 2022-09-28
[COST-3039] Move ProviderInfrastructureMap off of get_or_create in provider_db_accessor

cgoodfred:
## Jira Ticket

[COST-3039](https://issues.redhat.com/browse/COST-3039)

## Description

This change will switch from using `get_or_create` to filtering for the first entry that matches and using that if it exists or creating an entry if it does not for `ProviderInfrastructureMap` objects.

## Testing

1. Checkout Branch
2. Restart Koku
3. Load in OCP on cloud data such as with: `make create-test-customer` followed by `make load-test-customer-data`
4. See that data still processes normally and all is well.

## Notes

...


lcouzens:
Does it make sense to do something similar here too? https://github.com/project-koku/koku/blob/1e17328dd5691cfd8c567f0e2f1d48b95d59b66c/koku/masu/database/gcp_report_db_accessor.py#L405

cgoodfred:
> Does it make sense to do something similar here too?
> 
> https://github.com/project-koku/koku/blob/1e17328dd5691cfd8c567f0e2f1d48b95d59b66c/koku/masu/database/gcp_report_db_accessor.py#L405

Yeah, I went ahead and updated that area to move off of get_or_create as well. Good catch!

lcouzens (approved):

adberglund (approved):

PR #3912, opened 2022-09-30
[COST-3007] add node role queries to get nodes and fill in node role on topology

cgoodfred:
## Jira Ticket

[COST-3007](https://issues.redhat.com/browse/COST-3007)

## Description

This change will update the query for get_nodes_presto to determine the node role as well. Populate_node_table now uses the additional node element to fill in the nodes role on the table. After talking with Andrew and reviewing data from the demo cluster and prometheus queries, we settled on searching for the `openshift-kube-apiserver` project to determine a master and searching for the specific infra node label for infra nodes, otherwise it is a worker.

## Testing

1. Checkout Branch
2. Restart Koku
3. Load in some test data, `make create-test-customer` followed by `make load-test-customer-data`
4. The yaml changes in this PR make each source type (on prem/ocp-aws/ocp-azure/ocp-gcp) create data associating a node with each of the available node-roles
5. You should be able to connect to the DB and see the different node roles identified on the `org1234567.reporting_ocp_nodes` table.

## Notes
This branch is dependent on the migration in the PR for COST-3006 and that must go in first.
...


adberglund (approved):

djnakabaale (approved):
I see some node roles in the db

```
trino:org1234567> select node_role from reporting_ocp_nodes;
 node_role 
-----------
 infra     
 master
 worker    
 worker    
 worker    
 worker    
 worker    
 master     
 master    
```

PR #3914, opened 2022-09-30
Update development setup

samdoran:
## Description

- Update the dev setup docs related to setup since I'm going through it now.
- Change the `Makefile` to use Docker Compose v2.
- Enable buildkit so the multi-arch multi-stage image builds properly
- Remove M1 specific compose file and related documentation since it is now covered in the main README

## Testing

1. Checkout Branch
2. `make docker-down`
3. Unset `DOCKER_BUILDKIT` and `compose_file` if they are set and remove them from your `.env` file
4. Run `make docker-up`

A new `koku-base` image should be built that matches the container host architecture

Verify the architecture matches the container host architecture
```
docker ps -a --format 'table {{ .ID }}\t{{ .Names }}\t{{ .Image }}\t{{ .State }}\t{{ .Label "architecture" }}'
```

`docker compose logs -f koku-server koku-worker` should show the server and worker are operating as expected.

## Notes

- [x] Further changes can be made once we simplify the container image build for ARM (see #3909).


Makefile
```diff
@@ -26,13 +34,6 @@ OCP_PROVIDER_TEMP_DIR = $(PROVIDER_TEMP_DIR)/insights_local
 # How to execute Django's manage.py
 DJANGO_MANAGE = DJANGO_READ_DOT_ENV_FILE=True $(PYTHON) $(PYDIR)/manage.py
 
-# Docker compose specific file
-ifdef compose_file
-    DOCKER_COMPOSE = $(DOCKER)-compose -f docker-compose.yml -f $(compose_file)
```
maskarb, line 31: The way we use the `docker-compose.m1.yaml` is by calling
```
docker compose -f docker-compose.yaml -f docker-compose.m1.yaml
```
This line here is what accomplishes that, so we can't just remove this without some kind of replacement.

maskarb (changes requested):

Makefile
samdoran: I put that back in.

What do you think about just adding that file automatically for ARM systems rather than needing to set `compose_file` in the environment?

samdoran: I did it for science: https://github.com/project-koku/koku/pull/3914/commits/6b042d24b761a6db9cf4613410cb8a91b63213ed

samdoran: If that's too magic, I'll just switch back to explicitly requiring the env var and put that back in the docs as well.

maskarb: I'm a-okay with the magic route.

I looked into building multi-arch images before, but the nodes used to build trino and hive use rhel8/podman and I could never figure out if multi-arch was supported on those nodes.

docs/koku_setup_on_m1.rst
```diff
@@ -7,30 +7,14 @@ About
 
 This documentation is to guide M1 Mac users on how to successfully install and run Koku on their machines.
 
-Generally, in order to run Intel-based packages and programs on Mac with Apple silicon processors (M1), one can make `use of Rosetta`_ which simulates the Intel environment.
+Generally, in order to run Intel-based packages and programs on Mac with Apple silicon processors (M1), one can make `use of Rosetta`_ which emulates the Intel environment.
```
maskarb, line 10: Should we just remove this file entirely?

samdoran: Oh, I thought I did that. Will do.

samdoran: We should probably also remove `rosetta_on_m1.rst`. That suggests some things that can get you in trouble, namely installing Python with `brew`. I don't think having x86 binaries is really necessary.

maskarb (approved):
lgtm!

PR #3917, opened 2022-10-03
Multi stage container build build

samdoran:
## Jira Ticket

## Description

Changes the containerfile to a multi-stage build doing different things based on the image architecture. This transparently builds ARM or x86_64 images.

## Testing

1. Checkout Branch
2. Remove `compose_file` from the `.env` file
3. Run `docker compose up -d --build`
4. Verify image builds successfully and matches the host architecture: `docker image inspect -f '{{ .RepoTags }} {{ .Architecture }} {{ .Size }}' koku_base`

## Notes

Using native images allows `watchdog` to work properly but only when using Docker Desktop since it uses `grpcfuse` for bind mounts. I tested used 9p with Rancher Desktop, but inotify events from the container host don't make it processes running in the container.


Dockerfile
```diff
@@ -1,9 +1,8 @@
-FROM registry.access.redhat.com/ubi8/ubi-minimal:latest
+FROM registry.access.redhat.com/ubi8/ubi-minimal:latest AS base
 
 # PIPENV_DEV is set to true in the docker-compose allowing
 # local builds to install the dev dependencies
 ARG PIPENV_DEV=False
```
maskarb, line 5: Do we need this arg here? Or is it sufficient to only have the ARG on line 69?

maskarb (approved):
Looks good to me, just one comment/question.

docker-compose.yml
```diff
@@ -8,6 +8,7 @@ services:
       args:
         GIT_COMMIT: ${GIT_COMMIT-local-dev}
         PIPENV_DEV: "True"
+        USER_ID: ${USER_ID-1000}
```
samdoran, line 11: I added this to allow making the UID of the `koku` user match the user that owns the directory on the contraire host. This was an issue when using 9p and Rancher Desktop but wasn't necessary with Docker Desktop using grpcfuse.

Dockerfile
samdoran: I don't _think_ we need it there. That's why I was looking into the subtle different between `ARG` and `ENV`. It doesn't look like anything in that stage is running `pipenv` commands, so I think it should only be needed in the final stage.

samdoran: I went ahead and remove this and just moved the comment down.

maskarb (approved):
lgtm! Thanks for taking care of this!!

maskarb (approved):

PR #3923, opened 2022-10-05
COST-2063: Consistent numeric fields migration.

myersCody:
## Jira Ticket

[COST-2063](https://issues.redhat.com/browse/COST-2063)

## Description

This change will create a standard numeric value for our calculated fields in the database. 

## Testing

1. Checkout main
2. Load dummy data
3. Switch to this branch
4. Run migrations.

## Notes

...


koku/reporting/migrations/0260_standard_numeric_on_calculated_feilds.py
```diff
@@ -0,0 +1,1403 @@
+# Generated by Django 3.2.15 on 2022-10-05 15:40
```
maskarb, line 1: This migration will have to be re-written. We'll need the column updates consolidated into a single SQL query per table. Else this will consume all of the head space in the DB.

We did a migration similar to this a while back and we had to change the migration because it wrecked production.

maskarb (changes requested):

koku/reporting/migrations/0260_standard_numeric_on_calculated_feilds.py
myersCody: Oh good to know. I can write it up as single query per table. 👍

myersCody:
Smoke test results: https://ibutsu.apps.ocp4.prod.psi.redhat.com/runs/139faf90-2fa7-4e41-aa82-0e6ae0f5da53#classify-failures

myersCody:
The same tests failed, gonna try running them locally.

myersCody:
Current Thoughts: 
Changing the precision caused some unittest failures on our side due to no longer having an exact value match. I want to make sure we don't have the same problem with some smoke tests, so I am waiting for a successful smoke test result. However, we seem to be having some trouble with the ephemeral environment at the moment.

myersCody:
smoke results: https://ibutsu.apps.ocp4.prod.psi.redhat.com/runs/ada16e27-c7aa-4876-8c6d-2c51a3440a5c#classify-failures

maskarb (approved):

PR #3928, opened 2022-10-10
add filter on product_service to oci provider_map storage report

djnakabaale:
## Jira Ticket

[COST-3098](https://issues.redhat.com/browse/COST-3098)

## Description

This change will ...

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Notes

...

### DIFF
--- a/dev/scripts/load_test_customer_data.sh
+++ b/dev/scripts/load_test_customer_data.sh
@@ -355,8 +355,8 @@ build_all(){
   build_aws_data
   build_azure_data
   build_gcp_data
-  build_onprem_data
   build_oci_data
+  build_onprem_data
 }
 
 # ---execute---
--- a/koku/api/report/oci/provider_map.py
+++ b/koku/api/report/oci/provider_map.py
@@ -49,7 +49,7 @@ class OCIProviderMap(ProviderMap):
                     "region": {"field": "region", "operation": "icontains"},
                     "instance_type": {"field": "instance_type", "operation": "icontains"},
                 },
-                "group_by_options": ["product_service", "payer_tenant_id", "region"],
+                "group_by_options": ["product_service", "payer_tenant_id", "region", "instance_type"],
                 "tag_column": "tags",
                 "report_type": {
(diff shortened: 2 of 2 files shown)


koku/api/report/test/oci/test_oci_query_handler.py
```diff
@@ -1538,6 +1538,9 @@ def test_exclude_functionality(self, _):
                         continue
                     opt_dict = opt_dict.get(f"{exclude_opt}s")[0]
                     opt_value = opt_dict.get(exclude_opt)
+                    if "no-" in opt_value:
```
samdoran, line 1541: This will raise a `TypeError` if the `exclude_opt` key doesn't exists since `opt_value` would be `None`. Defaulting to an empty string will avoid that.
```python
opt_value = opt_dict.get(exclude_opt, "")
```

samdoran: Should this match only be on the beginning of the string? I'm not sure what real values look like but this test would match `some-rhino-sighting`.
```suggestion
                    if opt_value.startswith("no-"):
```

djnakabaale: 👍🏿 glad you pointed this one out

koku/api/report/test/oci/test_oci_query_handler.py
```diff
@@ -1597,11 +1600,12 @@ def test_multi_exclude_functionality(self, _):
                 exclude_one = None
                 exclude_two = None
                 for exclude_option in opt_list:
-                    if "no-" not in exclude_option.get(ex_opt):
+                    _exclude_option = exclude_option.get(ex_opt)
```
samdoran, line 1603: Default this to string to avoid an `AttributeError` if `ext_opt` isn't in `exclude_option`.
```suggestion
                    _exclude_option = exclude_option.get(ex_opt, "")
```

samdoran (approved):

PR #3929, opened 2022-10-11
[COST-3104] masu api for ocp on cloud processing

adberglund:
## Jira Ticket

[COST-3104](https://issues.redhat.com/browse/COST-3104)

## Description

This adds a Masu API to process OCP on Cloud parquet files. This is done during normal processing, but for a current customer bug we have been incorrectly processing this data. This API and corresponding celery task will delete parquet files for a specific month for OCP on Cloud and reprocess by grabbing the daily data from Trino and using the OCPCloudParquetReportProcessor to create new parquet files. 

## Testing

1. Checkout Branch
2. Load test customer data
3. Using trino check `select sum(lineitem_usageamount) from aws_openshift_daily;` and `select count(*) from aws_openshift_daily;`
4. Hit endpoint http://127.0.0.1:5042/api/cost-management/v1/report/process/openshift_on_cloud/?provider_uuid=0b0d5a9c-4672-4866-a796-f5ebb2b616cd&schema=org1234567&start_date=2022-08-01&end_date=2022-09-30 with dates and provider uuid for your cloud provider that has associated OCP clusters.
5. 3. Using trino check `select sum(lineitem_usageamount) from aws_openshift_daily;` and `select count(*) from aws_openshift_daily;` and verify that data is still correct

## Notes

...


koku/masu/api/process_openshift_on_cloud.py
```diff
@@ -0,0 +1,85 @@
...
+        params = {
+            "schema_name": schema_name,
+            "provider_uuid": cloud_provider_uuid,
+            "bill_date": bill_date,
+            "tracing_id": tracing_id,
+        }
+        LOG.info("Triggering process_openshift_on_cloud task with params:")
+        LOG.info(params)
+        msg = f"on queue: {queue_name}"
+        LOG.info(msg)
```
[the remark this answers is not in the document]
adberglund: Just because weird  python log standards. I think black yells at you for passing in an f string directly. I switched to % style.

lcouzens (dismissed):

koku/masu/processor/tasks.py
```diff
@@ -915,3 +920,54 @@ def remove_stale_tenants():
...
+        schema_name, "", provider_uuid, provider_type, 0, context={"tracing_id": tracing_id, "start_date": bill_date}
+    )
+    remove_files_not_in_set_from_s3_bucket(
+        tracing_id, processor.parquet_ocp_on_cloud_path_s3, 0, processor.error_context
+    )
+    for i, offset in enumerate(range(0, count, settings.PARQUET_PROCESSING_BATCH_SIZE)):
+        query_sql = (
+            f"SELECT * FROM {table_name}"
+            + where_clause
+            + f"OFFSET {offset} LIMIT {settings.PARQUET_PROCESSING_BATCH_SIZE}"
```
samdoran, line 964: This can use implicit string concatenation and avoid the (quite minor in this case but significant in other cases) performance hit caused by using the string concatenation operator. I think this makes it slightly more readable, but only slightly. 

```suggestion
            f"SELECT * FROM {table_name}"
            f"{where_clause}"
            f"OFFSET {offset} LIMIT {settings.PARQUET_PROCESSING_BATCH_SIZE}"
```


```
%timeit f"SELECT * FROM {table_name}" + where_clause + f"OFFSET {offset} LIMIT some limit"
200 ns ± 0.706 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

%timeit (f"SELECT * FROM {table_name}" f"{where_clause}" f"OFFSET {offset} LIMIT some limit")
134 ns ± 0.836 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```

koku/masu/processor/tasks.py
```diff
@@ -915,3 +920,54 @@ def remove_stale_tenants():
...
+        )
+        results, columns = execute_trino_query(schema_name, query_sql)
+        data_frame = pd.DataFrame(data=results, columns=columns)
+        for column in table_info.get(provider_type).get("date_columns"):
```
samdoran, line 968: If there is no `provider_type` key in `table_info` the second `get()` call will raise an an `AttributeError` on `None`.
```suggestion
        for column in table_info.get(provider_type, {}).get("date_columns"):
```

adberglund: True, true. I'm actually totally okay with an exception here. Never say never, but the API code is the only thing that will be calling this task and it will return a 400 if you provide an unacceptable provider / type. If we  add another OCP on Cloud type, I'd rather this get noisy and fail as a way of alerting us to to the need to add info on the specific fields for that type, rather than this silently pass and do nothing.

koku/masu/processor/tasks.py
adberglund: Gladly.

lcouzens (approved):

PR #3938, opened 2022-10-17
Fixing OCP/GCP project credit distribution

lcouzens:
## Jira Ticket

[COST-2453](https://issues.redhat.com/browse/COST-2453)

## Description

This change will fix OCP on GCP project level credit distribution

## Testing

1. Checkout Branch
2. Restart Koku
3. Load OCP on GCP data with multiple projects that have matching credits
4. See we have credits distributed correctly based from projects usage via api/cost-management/v1/reports/openshift/infrastructures/gcp/costs/?group_by[project]=*

## Notes

...


lcouzens:
Failing smoke tests:
- test_api_ocp_on_gcp_source_raw_calc
- test_api_oci_instance_endpoint_report_filter
- test_api_oci_instance_endpoint_previous_month_yesterday

All are known issues, that said the ocp/gcp one is fallout from this change specifically and requires a QE test update (I'm working on this)

koku/reporting/migrations/0261_ocpgcpcostlineitemprojectdailysummaryp_pod_credit.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 3.2.15 on 2022-10-17 15:41
```
myersCody, line 1: We should move these postgres migrations to their own pull request.

lcouzens: I _Can_ do this but is it strictly necessary especially with this additive case, and how we run migrations. Sure in stage the migration will run while the app is redeploying but that happens regardless now. As for Prod we can push out the migration separately from this code being released in app-interface anyway.

lcouzens: https://github.com/project-koku/koku/pull/3949

myersCody (approved):

myersCody (approved):

PR #3939, opened 2022-10-18
Letter case fix

radekkaluzik:
This fix is related to a bug described here: https://issues.redhat.com/browse/RHCLOUD-21435

I'm only fixing the letter case but I'm not familiar with this repo. Please double check whether or not some additional configs need to be regenerated in order to reflect this modification.


chambridge (approved):
LGTM

PR #3947, opened 2022-10-18
remove prod-sec manifest

maskarb:
* Prod-sec no longer uses this manifest, so we should no longer maintain it by removing it from the repo.


myersCody (approved):
LGTM

PR #3948, opened 2022-10-18
[COST-582] Update Azure libraries to use new authentication

samdoran:
## Jira Ticket

[COST-582](https://issues.redhat.com/browse/COST-582)

## Description

Update Azure Python libraries to use newer authentication class.

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Notes

Tests still need updating. I haven't quite figured out how to properly update the `Pipenv.lock` file yet. Still fighting with that.


samdoran:
The init signature for `CostManagementClient` was changed in https://github.com/Azure/azure-sdk-for-python/commit/48a72ddd3d15df38885abf0cd3216d793a517006. From reading the release notes, it's not clear to me that was intentional. Reading through the code, however, I don't see where `subscription_id` is used by `CostManagementClient`, so it didn't break anything per-se (only [people like me](https://github.com/Azure/azure-sdk-for-python/issues/19521) that were passing in `subscription_id` for `base_url`).

The subscription ID is only used in `ExportsOperations.list()`, which is unchanged.

maskarb:
> The init signature for `CostManagementClient` was changed in [Azure/azure-sdk-for-python@48a72dd](https://github.com/Azure/azure-sdk-for-python/commit/48a72ddd3d15df38885abf0cd3216d793a517006). From reading the release notes, it's not clear to me that was intentional. Reading through the code, however, I don't see where `subscription_id` is used by `CostManagementClient`, so it didn't break anything per-se (only [people like me](https://github.com/Azure/azure-sdk-for-python/issues/19521) that were passing in `subscription_id` for `base_url`).
> 
> The subscription ID is only used in `ExportsOperations.list()`, which is unchanged.

So, I guess we don't need the subscription id anymore?

samdoran:
> So, I guess we don't need the subscription id anymore?

That's correct. I looked through the Azure library code and it's not used directly by `CostManagementClient`. It gets the correct subscription information from the credential.

maskarb (dismissed):
lgtm!

maskarb (approved):

PR #3951, opened 2022-10-20
currency settings unleash

maskarb:
## Description

This change will hide currency settings behind unleash flag. This flag matches the ui flag used to enable/disable currency in the ui.

## Testing
1. with `cost-management.ui.currency` not enabled for the user, the first field in the settings json is:
```
[
    {
        "title": "",
        "name": "api.settings.settings.subform",
        "fields": [
            {
                "component": "plain-text",
                "label": "Enable tags and labels",
                "name": "api.settings.tag_management.form-text",
                "variant": "h2"
            },
```
2. Enabling the flag for the user, the first field in the settings json is:
```
[
    {
        "title": "",
        "name": "api.settings.settings.subform",
        "fields": [
            {
                "component": "plain-text",
                "label": "Currency",
                "name": "api.settings.currency",
                "variant": "h2"
            },
```

### DIFF
--- a/koku/api/settings/settings.py
+++ b/koku/api/settings/settings.py
@@ -4,6 +4,7 @@
 #
 """Data Driven Component Generation for Tag Management Settings."""
 import logging
+import re
 
 from django.conf import settings
 from django.test import RequestFactory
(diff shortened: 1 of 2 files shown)


myersCody (approved):
LGTM!

koku/api/settings/settings.py
```diff
@@ -105,6 +107,8 @@ def __init__(self, request):
         self.factory = RequestFactory()
         self.schema = request.user.customer.schema_name
 
+        self.unleash_context = {"userId": re.sub(r"\D+", "", self.schema)}
```
samdoran, line 110: I think the goal here is to remove everything but digits from `self.schema`. Here's a simpler and more performant approach that avoids the use of regular expressions.
```suggestion
        self.unleash_context = {"userId": "".join(c for c in self.schema if c in string.digits)}
```

PR #3968, opened 2022-10-28
add unleash kill-switch for currency

maskarb:
## Description

* Enabling the flag `cost-management.backend.currency` in unleash will disable currency in the backend.
* To disable for specific users, we should use the `userId` strategy with the `org-id` number.


maskarb:
Currency support seems to be working well, so closing this PR.

PR #3971, opened 2022-10-31
COST-3130: Update ClowApp with kube-linter ignore for min-three-replicas

chambridge:
## Jira Ticket

[COST-](https://issues.redhat.com/browse/COST-3130)

## Description

* The scheduler and sources-listener are single replica only deployments
* Add annotation to bypass kube-linter checks from AppSRE

## Notes
This only effects the ClowdApp file.


maskarb (approved):
lgtm!

PR #3978, opened 2022-11-01
[CI] Update action versions, remove deprecated commands, improve organization

samdoran:
### Description

Several actions we use rely on a deprecated version of Node.js. Update the actions to use a supported version of Node.js.

The `set-option` and `set-state` commands are deprecated. Use Environment File instead.

Consolidate workflows and add names to all steps for better presentation.

Pin all jobs to `ubuntu-20.04` for consistency.

### Testing

1. Run workflows in GitHub Actions


maskarb (dismissed):

.github/workflows/ci.yml
```diff
@@ -0,0 +1,235 @@
...
+        uses: actions/setup-python@v4.3.0
+        with:
+          python-version: '3.8'
+
+      - name: Run pre-commit checks
+        uses: pre-commit/action@v3.0.0
+        env:
+          SETUPTOOLS_USE_DISTUTILS: stdlib
+
+      - name: Check clowadd manifest
```
cgoodfred, line 28: nitpick: I think this is supposed to be clowdapp

samdoran: Yup. Thanks!

lcouzens (approved):

PR #3981, opened 2022-11-01
[COST-2659] - fix sources-cleanup pagination

maskarb:
## Jira Ticket

[COST-2659](https://issues.redhat.com/browse/COST-2659)

## Description

The previous implementation of pagination for the sources-cleanup was "double" paginating the data. It was applying [offset:offset + limit] to the DB query and then the same [offset:offset + limit] to the Response dataset. This was resulting in accurate responses.


myersCody (approved):
LGTM

PR #3985, opened 2022-11-03
COST-3031: Remove dead panels from grafana dashboard

chambridge:
Feedback from AppSRE was to remove defunct panels to avoid confusion.

## Jira Ticket

[COST-3031](https://issues.redhat.com/browse/COST-3031)

## Description
* Remove celery and volume panels that were no longer collecting data
* Fix RDS file usage chart which had an incorrect data source

image
image


maskarb (approved):
lgtm!

PR #3992, opened 2022-11-04
[COST-2659] - paginate providers-without-sources

maskarb:
## Jira Ticket

[COST-2659](https://issues.redhat.com/browse/COST-2659)

## Description

* adds pagination for `/source_cleanup/?providers_without_sources`.
* for `?missing_sources` the count reflects ALL sources in the db. The limits and offsets will now check for sources that are missing in the limit/offset range. The links now page thru all sources and show any sources missing in that range.

#### providers-without-sources (out-of-order-deletes paginates using the same mechanism)
2. Go here to see no data: [http://localhost:5042/api/cost-management/v1/source_cleanup/?limit=20&providers_without_sources](http://localhost:5042/api/cost-management/v1/source_cleanup/?limit=20&providers_without_sources)
3. Log into the DB and delete all sources:
```
postgres=# delete from api_sources ;
DELETE 11
```
4. Go here to see 11 Providers: [http://localhost:5042/api/cost-management/v1/source_cleanup/?limit=20&providers_without_sources](http://localhost:5042/api/cost-management/v1/source_cleanup/?limit=20&providers_without_sources)
5. Change the limits and offsets and see that the data is limited and that the links are correct.


adberglund (approved):
📖 📄 📄 📄

PR #3993, opened 2022-11-08
commit after deletes

maskarb:
## Description

This change will immediately commit to the database after data from a particular table is deleted.


koku/api/provider/models.py
```diff
@@ -275,7 +264,7 @@ def _set_infrastructure_id_null(self):
         # can be shared among other providers, we must set the api_provider.infrastructure_id to null
         # for the infrastructure_id targeted for delete that is shared across other providers.
         # That is a confusing sentence to describe a circular relation between tables.
-        _sql = f"""
+        _sql = """
```
samdoran, line 267: It's always bugged me how Python makes it hard to store multi-line blocks of text at the same indentation level of the containing method or function. One approach I've used is using `textwrap.dedent()`, which was added in Python 3.3.

```suggestion
        _sql = """\
            update public.api_provider p
                set infrastructure_id = %s
               from public.api_providerinfrastructuremap m
             where m.infrastructure_provider_id = %s::uuid
               and p.infrastructure_id = m.id
            ;\
        """ 
        
        cur.execute(textwrap.dedent(_sql), params)
```

The `# nova: F541` on line 274 can be removed.

samdoran (approved):

koku/api/provider/models.py
maskarb: I don't particularly like how this is formatted either, but this code is written so you can copy it into a sql file or into the terminal without dealing with formatting.

koku/api/provider/models.py
```diff
@@ -196,22 +195,13 @@ def save(self, *args, **kwargs):
...
-                with transaction.atomic():
-                    self._cascade_delete()
-                    # Make sure we commit the cascade delete before deleting the Provider record
-                    transaction.on_commit(
-                        lambda: self._delete_from_target(
-                            {"table_schema": "public", "table_name": self._meta.db_table, "column_name": "uuid"},
-                            target_values=[self.pk],
-                        )
-                    )
+                self._normalized_type = _type[: _type.index("-")] if _type.endswith("-local") else _type
```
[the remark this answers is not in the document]
maskarb: I found an even better solution 😁 :
```
self._normalized_type = _type.removesuffix("-local")
```

koku/api/provider/models.py
samdoran: Oh that's handy. I hadn't thought about that.

koku/api/provider/models.py
```diff
@@ -198,7 +198,7 @@ def delete(self, *args, **kwargs):
             with schema_context(self.customer.schema_name):
                 LOG.info(f"PROVIDER {self.name} ({self.pk}) CASCADE DELETE -- SCHEMA {self.customer.schema_name}")
                 _type = self.type.lower()
-                self._normalized_type = _type[: _type.index("-")] if _type.endswith("-local") else _type
+                self._normalized_type = _type.removesuffix("-local")
```
samdoran, line 201: Even better! I didn't know about that method.

samdoran (approved):

koku/api/provider/models.py
maskarb: `removesuffix` and `removeprefix` are new in Python 3.9!

PR #3996, opened 2022-11-09
[CI] Always run Units job, just skip some steps

samdoran:
## Description

Rather than skipping the entire units job, which is required for merges, only skip steps within the job that should not run when tests should be skipped.

## Testing

1. Run CI

## Notes

...


.github/workflows/ci.yml
```diff
@@ -151,6 +150,7 @@ jobs:
 
     steps:
       - name: Checkout
+        if: needs.changed-files.outputs.run_tests == 'true'
```
maskarb, line 153: Just a thought, but instead of adding this `if` to all the steps in `units`, what if `units` sets a variable, and we create a new job that checks the files-changes and the output of `units`?

samdoran: I wanted to avoid having another job in the test matrix and then having to change the repo settings to check the status of that job. The duplicate conditionals are a bit undesirable in the job config, but it makes for a nicer output in the Actions console.

maskarb (approved):

PR #3997, opened 2022-11-09
Cost 3199: custom lookup to allow nulls to be returned during tag exclusion

myersCody:
## Jira Ticket

[COST-3199](https://issues.redhat.com/browse/COST-3199)

## Description

Whenever excluding by tag, we were loosing the `no-{option}` in our data returned. Which was an expected behavior we would like to keep. After digging into it a bit I realized that a `IS NOT NULL` is being added at the django optimizer level. Part of what django is trying to do is handle the tri-value logic, where nulls are considered unknown
```
NOT true == false
NOT false == True
NOT unknown is still unknown
```

This [stack overflow](https://stackoverflow.com/questions/71041019/django-q-object-is-adding-extra-is-not-null) actually does a pretty good job of explaining why the is not null is added. However, the main take away is: 
```
To thus ensure that ~Q(…) is the full opposite of Q(…), it thus should add foo IS NOT NULL to the condition.
```

So, in order to bypass this, I wrote my own [django lookup](https://docs.djangoproject.com/en/3.2/howto/custom-lookups/) that will take a list of tag values and do a `NOT LIKE` on them, and then added in some django Q filters to allow emptyset values and added them to self.query_filter for when we exclude by tag. 

## Testing

1. Checkout Branch
2. Load dummy data:
```
make create-test-customer
make load-test-customer-data
```

*On main:*
```
http://localhost:8000/api/cost-management/v1/reports/aws/costs/?group_by[tag:app]=*&exclude[tag:app]=cost
```
Notice how `no-app` is not in the return

*On this branch*
```
http://localhost:8000/api/cost-management/v1/reports/aws/costs/?group_by[tag:app]=*&exclude[tag:app]=cost
http://localhost:8000/api/cost-management/v1/reports/aws/costs/?group_by[tag:app]=*&exclude[tag:app]=cost&exclude[tag:app]=analytics
```

notice how  `no-app` is in the return value. 

Try for all of our different tag endoings.

## Notes

...


myersCody:
I had two smoke test failures. `OCP on GCP` & `OCP on AWS` raw calc test. I don't think my changes should of affected those so I am going to rerun them while investigating locally.

myersCody:
These raw calc test are also failing on other PRs: https://ci.ext.devshift.net/job/project-koku-koku-pr-check/7386/

koku/api/report/queries.py
```diff
@@ -315,67 +323,76 @@ def _get_search_filter(self, filters):  # noqa C901
...
+        for tag in tag_filters:
+            tag_db_name = self._mapper.tag_column + "__" + strip_tag_prefix(tag)
+            filt = {"field": tag_db_name, "operation": "notlikelist"}
+            emptyset_filters.append({"field": tag_db_name, "operation": "isnull", "parameter": True})
+            list_ = self.parameters.get_exclude(tag, list())
+            if list_ and not ReportQueryHandler.has_wildcard(list_):
+                q_filter = QueryFilter(parameter=list_, **filt)
+                tag_exclusion_collection.add(q_filter)
+        emptyset_filters.append({"field": self._mapper.tag_column, "operation": "exact", "parameter": "{}"})
+        emptyset_composed = self._set_or_filters(emptyset_filters)
```
maskarb, line 356: How does an `or` filter behave with exclusions?

koku/reporting/provider/__init__.py
```diff
@@ -1 +1,40 @@
 # noqa
```
maskarb, line 1:
```suggestion
```

koku/reporting/provider/__init__.py
```diff
@@ -1 +1,40 @@
...
+# Custom Django Lookups
+@Field.register_lookup
+class AllowNullIcontains(Lookup):
+    """
+    This custom lookup accepts a list as the param wne will create
+    a NOT LIKE with an 'AND' between each element in the list.
+    Currently exclusively beinng used for tag exclusions.
+    """
+
+    lookup_name = "notlikelist"
```
maskarb, line 15: Is this lookup supposed to be the opposite of `icontains`? Should we just call this `noticontains`?

maskarb (commented):
A couple of questions:

koku/reporting/provider/__init__.py
myersCody: I had it originally call noticontains, then renamed it after much debate since this one takes a list. I am okay with going back to icontains.

koku/api/report/queries.py
myersCody: So, these `or_filters` are to add additional filters I needed to include the emptysets to get `no-option` to show up. So, it will add filters like `tag is null or tag = '{}'` to the sql query. However, in general, I think this could be a worthwhile discussion to see if or is needed with excludes.

myersCody: So, I continued digging into this question, and with the `noticontainslist` filter `exclude[or:tag:app]` would do the same thing as `exclude[and:tag:app]` because of the logic I added [here](https://github.com/project-koku/koku/pull/3997/files#diff-85e41f2ec3229710ac79109eb26a6463faf136a8b9acdd677d57a5b34798a6fcR29). If we change that logical operator to an `OR` than when we do multiple excludes:
```
((UPPER((reporting_awscostentrylineitem_daily_summary.tags ->> 'app')::text) NOT LIKE UPPER('%analytics%')) OR (UPPER((reporting_awscostentrylineitem_daily_summary.tags ->> 'app')::text) NOT LIKE UPPER('%catalog%'))
```
both the `analytics` & `catalog` tag will be returned in the response. So, the short answer here for tags is that it is essentially doesn't do anything. I can update the serializer to not allow `exclude[or:tag:{key}]` or we can just keep it & do the right thing on the backend. 

-------
**Outside of Tag Exclusions**

`exclude[or:{option}]={value}` is needed in order to tell django how to combine two different exclusions. For example, if you were to hit the following endpoint:
```
http://localhost:8000/api/cost-management/v1/reports/aws/costs/?group_by[account]=*&group_by[service]=*&exclude[account]=9999999999995&exclude[service]=AmazonRDS
```
Django creates sql like this:
```
AND NOT (((UPPER(reporting_awsaccountalias.account_alias::text) LIKE UPPER('%9999999999995%') 
            AND reporting_awsaccountalias.account_alias IS NOT NULL) 
            OR UPPER(reporting_aws_cost_summary_by_service_p.usage_account_id::text) LIKE UPPER('%9999999999995%')) 
            AND UPPER(reporting_aws_cost_summary_by_service_p.product_code::text) LIKE UPPER('%AmazonRDS%'))
```
Due to the `AND` this will only exclude the `AmazonRDS` when the accounts = `9999999999995` which we are about to exclude with the `NOT` making that condition useless.

However, if we change it to: `
&exclude[account]=9999999999995&exclude[or:service]=AmazonRDS`
```
AND NOT (
        ((UPPER(reporting_awsaccountalias.account_alias::text) LIKE UPPER('%9999999999995%') 
        AND reporting_awsaccountalias.account_alias IS NOT NULL)  
        OR UPPER(reporting_awscostentrylineitem_daily_summary.usage_account_id::text) LIKE UPPER('%9999999999995%') 
        OR UPPER(reporting_awscostentrylineitem_daily_summary.product_code::text) LIKE UPPER('%AmazonRDS%')))) 
```
That `AND` is flipped to an `OR` allowing us to exclude `AmazonRDS` from the return for all the accounts not just the one we excluded.

koku/reporting/provider/__init__.py
maskarb: I wasn't sure why it was named this, which is why I asked. I was only advocating for changing it if it was supposed to be the opposite of `icontains`. If it is a different lookup entirely, then let's name it whatever we want.

koku/api/report/queries.py
myersCody: **Additional Context:**
So, Michael Andrew and I had a private chat in slack regarding:
```
exclude[account]=9999999999995&exclude[service]=AmazonRDS
Does this query make sense to only exclude AmazonRDS when the account equals 9999999999995? I feel like if I was making that API request, i would want to exclude everything that is account 9999999999995 and every service that is AmazonRDS
```
After discussion we agreed that the logic is confusing, and we were making:
```
WHERE NOT (account = '9999999999995' AND service = 'AmazonRDS')
```
instead of
```
WHERE NOT account = '9999999999995' AND NOT service = 'AmazonRDS'
```
I attempted to configure django to do `WHERE NOT account = '9999999999995' AND NOT service = 'AmazonRDS'`. However, I had some issues with how django to do this. However, I was able to get it to do:
```
WHERE NOT (account = '9999999999995' OR service = 'AmazonRDS')
```

---- 

Additional query checks to make sure they were equivalent:
NOT FOO and NOT bar:
```
SELECT
    sum(reporting_aws_cost_summary_by_service_p.unblended_cost),
    reporting_aws_cost_summary_by_service_p.product_code AS service
FROM reporting_aws_cost_summary_by_service_p
LEFT OUTER JOIN reporting_awsaccountalias ON (reporting_aws_cost_summary_by_service_p.account_alias_id = reporting_awsaccountalias.id)
WHERE (
    reporting_aws_cost_summary_by_service_p.usage_start = '2022-11-07'::date
    AND NOT (((UPPER(reporting_awsaccountalias.account_alias::text) LIKE UPPER('%9999999999995%') AND reporting_awsaccountalias.account_alias IS NOT NULL)))
    AND NOT (UPPER(reporting_aws_cost_summary_by_service_p.usage_account_id::text) LIKE UPPER('%9999999999995%')))
    GROUP BY DATE_TRUNC('day', reporting_aws_cost_summary_by_service_p.usage_start), reporting_aws_cost_summary_by_service_p.product_code, COALESCE('USD', 'USD');
```
result:
```
      sum      |             service
---------------+----------------------------------
   1.000028000 | AmazonRoute53
  22.633787904 | AmazonRDS
  12.000000000 | AmazonS3
 454.259754079 | Red Hat OpenShift Service on AWS
   0.469870676 | AmazonDynamoDB
  39.625239100 | AmazonEC2
   1.200000000 | AmazonVPC
```
NOT (FOO OR BAR)
```
SELECT
    sum(reporting_aws_cost_summary_by_service_p.unblended_cost),
    reporting_aws_cost_summary_by_service_p.product_code AS service
FROM reporting_aws_cost_summary_by_service_p
LEFT OUTER JOIN reporting_awsaccountalias ON (reporting_aws_cost_summary_by_service_p.account_alias_id = reporting_awsaccountalias.id)
WHERE (
    reporting_aws_cost_summary_by_service_p.usage_start = '2022-11-07'::date
    AND
    NOT (((UPPER(reporting_awsaccountalias.account_alias::text) LIKE UPPER('%9999999999995%') AND reporting_awsaccountalias.account_alias IS NOT NULL)
            OR UPPER(reporting_aws_cost_summary_by_service_p.usage_account_id::text) LIKE UPPER('%9999999999995%')
        ))) GROUP BY DATE_TRUNC('day', reporting_aws_cost_summary_by_service_p.usage_start), reporting_aws_cost_summary_by_service_p.product_code, COALESCE('USD', 'USD');
```
result:
```
      sum      |             service
---------------+----------------------------------
   1.000028000 | AmazonRoute53
  22.633787904 | AmazonRDS
  12.000000000 | AmazonS3
 454.259754079 | Red Hat OpenShift Service on AWS
   0.469870676 | AmazonDynamoDB
  39.625239100 | AmazonEC2
   1.200000000 | AmazonVPC
```

koku/api/report/queries.py
```diff
@@ -233,15 +233,22 @@ def _build_custom_filter_list(self, filter_type, method, filter_list):
 
     def _check_for_operator_specific_filters(self, filter_collection, check_for_exclude=False):
         """Checks for operator specific fitlers, and adds them to the filter collection"""
```
maskarb, line 235: The logic in this function is getting really difficult to follow. Since this func is used to create 
```
composed_filters = self._check_for_operator_specific_filters(filters)
```
and
```
self.query_exclusions = self._check_for_operator_specific_filters(exclusions, True)
```
can we remove the exclusion stuff from this func and create a 
```
self.query_exclusions = self._check_for_operator_specific_exlusions
```
?

myersCody: Yeah! Originally we wanted to use the same functions, but I think this one as hit a point where we should separate them. Let me do that real quick and i'll ping you afterwards.

koku/api/report/queries.py
```diff
@@ -264,14 +275,16 @@ def _get_search_filter(self, filters):  # noqa C901
         # define filter parameters using API query params.
         fields = self._mapper._provider_map.get("filters")
         access_filters = QueryFilterCollection()
-        exclusions = QueryFilterCollection()
+        # exclusions = QueryFilterCollection()
```
maskarb, line 278:
```suggestion
```

maskarb (dismissed):
A couple more suggestions, but none of these are really blockers.

maskarb (approved):
lgtm! Thanks for making all these changes!

PR #4007, opened 2022-11-15
COST-3251: Add an unleash flag for charging for unallocated capacity

myersCody:
## Jira Ticket

[COST-3251](https://issues.redhat.com/browse/COST-3251)

## Description

This change will gate the charge of unallocated capacity behind an unleash flag. 

## Testing

Load our dummy customer data locally and look for these logs:

```
koku-koku-worker-1  | [2022-11-14 21:34:01,510] INFO e8c41502-8a97-4ce8-9077-31f001aac167 Summary UNLEASH check: {'schema': 'org1234567'}
koku-koku-worker-1  | [2022-11-14 21:34:01,510] INFO e8c41502-8a97-4ce8-9077-31f001aac167     Calculate unallocated capacity disabled org1234567
```
**Confirmation SQL:**

```
postgres=# select count(*) from reporting_ocpusagelineitem_daily_summary where namespace='Workers Unallocated Capacity';
 count
-------
     0
(1 row)

postgres=# select count(*) from reporting_ocpusagelineitem_daily_summary where namespace='Platform Unallocated Capacity';
 count
-------
     0
(1 row)
```
Then enable the unleash flag: `org1234567`

```
http://localhost:4242/#/features/strategies/cost-management.backend.unallocated_capacity_enabled
```

 Then resummarize:
```
http://127.0.0.1:5042/api/cost-management/v1/report_data/?provider_uuid=98389555-d145-4bbb-8091-79c996b30f8c&start_date=2022-10-01&end_date=2022-11-15&schema=org1234567
```

```
koku-koku-worker-1  | [2022-11-15 14:13:56,879] INFO 101e87a5-bc96-40bc-9bea-19eec420ef47 Summary UNLEASH check: {'schema': 'org1234567'}
koku-koku-worker-1  | [2022-11-15 14:13:56,879] INFO 101e87a5-bc96-40bc-9bea-19eec420ef47     Calculate unallocated capacity enabled org1234567
koku-koku-worker-1  | [2022-11-15 14:13:56,955] INFO 101e87a5-bc96-40bc-9bea-19eec420ef47 table_exists_trino for org1234567
```

**Confirmation SQL:**

```
postgres=# select count(*) from reporting_ocpusagelineitem_daily_summary where namespace='Platform Unallocated Capacity';
 count
-------
    90
(1 row)

postgres=# select count(*) from reporting_ocpusagelineitem_daily_summary where namespace='Workers Unallocated Capacity';
 count
-------
    90
(1 row)
```

## Notes

...

### DIFF
--- a/koku/masu/database/presto_sql/reporting_ocpusagelineitem_daily_summary.sql
+++ b/koku/masu/database/presto_sql/reporting_ocpusagelineitem_daily_summary.sql
@@ -409,6 +409,9 @@ What was selected from unallocated capacity.
 AND lids.namespace != 'Platform Unallocated Capacity'
 AND lids.namespace != 'Workers Unallocated Capacity'
  */
+
+{% if unallocated_capacity_enabled %}
+
 INSERT INTO hive.{{schema | sqlsafe}}.reporting_ocpusagelineitem_daily_summary (
     uuid,
     report_period_id,
@@ -480,7 +483,17 @@ WHERE lids.source = {{source}}
     AND lids.node IS NOT NULL
     AND lids.data_source = 'Pod'
 GROUP BY lids.node, lids.usage_start, lids.source_uuid
+
+{% else %}
+
+SELECT 1
+-- Developer Note: Jinja has a weird relationship
+-- with semicolons inside of if statements
+
+{% endif %}
+
 ;
+
 
 INSERT INTO postgres.{{schema | sqlsafe}}.reporting_ocpusagelineitem_daily_summary (
     uuid,
--- a/.unleash/flags.json
+++ b/.unleash/flags.json
@@ -205,6 +205,24 @@
             ],
             "variants": [],
             "createdAt": "2022-09-08T09:48:54.492Z"
+        },
+        {
+            "name": "cost-management.backend.unallocated_capacity_enabled",
+            "description": "A list of schemas for customers that are currently being charged for their unallocated capacity.",
+            "type": "operational",
+            "project": "default",
+            "enabled": true,
+            "stale": false,
+            "strategies": [
+                {
+                    "name": "schema-strategy",
+                    "parameters": {
+                        "schema-name": ""
+                    }
+                }
+            ],
+            "variants": [],
+            "createdAt": "2022-11-14T19:57:52.119Z"
         }
     ],
     "strategies": [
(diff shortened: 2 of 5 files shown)


koku/masu/database/presto_sql/reporting_ocpusagelineitem_daily_summary.sql
```diff
@@ -480,8 +483,18 @@ WHERE lids.source = {{source}}
     AND lids.node IS NOT NULL
     AND lids.data_source = 'Pod'
 GROUP BY lids.node, lids.usage_start, lids.source_uuid
+
+{% else %}
+
+SELECT 1
```
myersCody, line 489: I am completely open to a better solution for this, I was running into errors with trying to include a `;` inside of jinja's if block. If you do this:
```
{% if my_unleash_flag %}
<some sql here>
;
{% endif %}
```
You will the following error:
```
Jinja was looking for the following tags: 'elif' or 'else' or 'endif'.
```

However, since the unleash gate will only be in place until we notify our customers of the upcoming cost changes, I didn't want to sink too much time into the perfect solution for code that will be removed in the future. The `Select 1` allows me to move the `;` outside of the if statement making both sql & jinja happy. 

----

I also tried:
```
(Option One)
{% set my_string = ; %}
{{ my_string }}

(Option Two)
{{ ; | sqlsafe}}

(Option Three)
{{ autoescape false }}
;
{{ endautoescape }}
```

myersCody: I forgot to put context for why I included the else statement 🤦 

So, that was the first thing I tried, but it resulted in the following Trino error: 

```
TrinoUserError(type=USER_ERROR, name=SYNTAX_ERROR, message="line 12:1: mismatched input '<EOF>'.
```

I added the else statement so prevent the mismatched EOF caused moving the the `;` outside the if condition. So that both trino & jinja would be happy with the sql created by the template.

lcouzens (changes requested):
We should 100% just make separate SQL files here instead. Easier for cleanup later on + less prone to bugs when it is changed :crossed_fingers:

myersCody:
After further discussion with @lcouzens & @kholdaway we decided that notifying the customers of the upcoming change is sufficient and gating the change with unleash won't be necessary.

PR #4016, opened 2022-11-17
[COST-3277] Handle `ValueError` raised by Azure libraries

samdoran:
## Jira Ticket

[COST-3277](https://issues.redhat.com/browse/COST-3277)

## Description

Handle `ValueError` raised by Azure libraries in the same manner we handle `TypeError` and several others exceptions. The specific exception message is logged and a scrubbed message is presented in the UI.

## Testing

1. Checkout Branch
2. Restart Koku
3. Try to add an Azure source with an invalid tenant_id. Any character not in `frozenset("ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz" + "0123456789" + "-.")` is invalid.
4. Verify that "Azure client configuration error." is returned to the UI and actual exception message is logged

Or run
```
python koku/manage.py test providers.test.azure.test_provider
```

## Notes

The new Azure libraries raise `ValueError` and not just exceptions from their custom exception classes.


maskarb:
Some testing:
```
invalid tenant id (this one seems wrong):
"availability_status_error": "Azure client configuration error.",
This is the internal error that we see, vs what we are sending back to sources:
sources_client  | [2022-11-18 15:10:39,599] WARNING None Source ID: 9 Internal message: Invalid tenant id provided. You can locate your tenant id by following the instructions here: https://docs.microsoft.com/partner-center/find-ids-and-domain-names.
sources_client  | [2022-11-18 15:10:39,599] WARNING None Source ID: 9 error message: Azure client configuration error.

invalid subscription-id: (this one is a bit tricky because it is hard to tell the difference between an invalid sub-id and a missing report)
"availability_status_error": "A cost management export cannot be found. In Azure, create a daily export task for your storage account.",

invalid resource-group (same deal with the sub-id):
"availability_status_error": "A cost management export cannot be found. In Azure, create a daily export task for your storage account.",

invalid storage-account (same deal with the sub-id):
"availability_status_error": "A cost management export cannot be found. In Azure, create a daily export task for your storage account.",

invalid password (this one looks good):
"availability_status_error": "A problem has been detected with the Azure client secret for this source. Refer to the Microsoft Azure troubleshooting guide in the cost management documentation for details.",

invalid username (this one looks good):
"availability_status_error": "The client ID was entered incorrectly for this source. Edit your Azure source and verify the client ID.",

```

samdoran:
Very good point. Let me play with this and see if there's any other room for improvement.

maskarb:
I tested the storage account and resource groups again, just to be sure:
```
storage account (this looks good):

sa: totallywrong

"availability_status_error": "The storage account was entered incorrectly for this source. Edit your Azure source and verify the storage account.",

resource-group (this looks good):

rg: totallywrong

"availability_status_error": "The resource group was entered incorrectly for this source. Edit your Azure source and verify the resource group.",
```

But when I test with these, the error is different:
```
rg: RG1///

"availability_status_error": "A cost management export cannot be found. In Azure, create a daily export task for your storage account.",

sa: mysa1///

"availability_status_error": "A cost management export cannot be found. In Azure, create a daily export task for your storage account.",
```

So, it's definitely good to update the error about the missing export, given that incorrect rg and sa can lead to that error too.
```

maskarb:
Regarding the invalid sub-id, we could probably get away with adding the `InvalidSubscriptionId` into here
https://github.com/project-koku/koku/blob/main/koku/sources/sources_error_message.py#L35-L36

maskarb (approved):

PR #4020, opened 2022-11-21
Refine and cleanup local dev setup

samdoran:
## Description

- Rename all containers to have a `koku-` prefix
- Remove prefix from service name for less typing
- Use dashes instead of underscores for container names to align with default default in Compose v2
- Updates default `.env.example` file to use correct values and add missing ones
- Don't use hotfix version of `minio` since there is no ARM container for hotfix tags
- Update some deprecated `mc` commands

## Testing

1. Checkout Branch
1. Stop and remove all existing koku containers, base image, and build cache

    ```
    docker stop $(docker ps --filter name=koku -q)
    docker rm $(docker ps  -a --filter name=koku -q)
    docker rmi koku_base
    docker buildx prune -af
    ```

1. Remove existing Minio data. 

    ```
    rm -rf .trino/parquet_data
    ```

1. Source the new values in `.env.example` or update them in existing `.env` file.

1. Start new containers

    ```
    make docker-up
    ```

1. Verify services work as expected

    ```
    make docker-logs
    ```


Dockerfile.minio
```diff
@@ -1,4 +1,4 @@
-FROM minio/minio:RELEASE.2022-05-26T05-48-41Z.hotfix.15f13935a
```
maskarb, line 1: What is the reason for this change?

docker-compose.yml
```diff
@@ -1,7 +1,9 @@
 version: '3'
+name: koku
```
maskarb, line 2: Can you link to docs that describe what this field is for?

Why are we changing all of the service names in this file?

Dockerfile.minio
samdoran: The hotfix container images are not multi-arch.

docker-compose.yml
samdoran: > Can you link to docs that describe what this field is for?

The [name top-level argument](https://docs.docker.com/compose/compose-file/#name-top-level-element) sets the prefix used for each container name The docs don't explain that very well. The default value is the name of the directory container the compose file.

I mainly added this for explicit consistency. It can be omitted, though, since we explicitly set the `container_name` for almost every service.

samdoran: > Why are we changing all of the service names in this file?

For consistency and to reduce some typing.

Some services had the compose project name prefixed, such as `koku-server`, others did not, such as `db` and `services-client`. Removing the prefix from the service names makes compose commands more succinct.

`docker compose up server masu-server worker db trino`

Naming all the containers with a consistent prefix makes it clear which project a collection of containers is associated with when running more than one collection of containers.

```
> docker ps
CONTAINER ID   NAMES                 IMAGE                                           STATE     architecture   PORTS
0ef4c6415f53   sources-api-server    sources-api                                     running   aarch64        0.0.0.0:9394->9394/tcp, 0.0.0.0:3000->8000/tcp
cb71b8245e91   sources-redis         redis:5.0.4                                     running                  0.0.0.0:6378->6379/tcp
24ff6f08e1b9   sources-kafka         confluentinc/cp-kafka                           running   arm64          0.0.0.0:9092->9092/tcp, 0.0.0.0:29092->29092/tcp
0dbc7abd0f7a   sources-db            postgres:14                                     running                  0.0.0.0:15434->5432/tcp
c74090343132   sources-zookeeper     confluentinc/cp-zookeeper                       running   arm64          2181/tcp, 2888/tcp, 3888/tcp
7cfcb87eedca   koku-worker-1         koku_base                                       running   aarch64        8000/tcp, 0.0.0.0:6001->9000/tcp
f62f3c96a9f7   koku-server           koku_base                                       running   aarch64        0.0.0.0:8000->8000/tcp, 0.0.0.0:8001->9000/tcp
50e199bb9a33   masu-server           koku_base                                       running   aarch64        0.0.0.0:5042->8000/tcp, 0.0.0.0:5001->9000/tcp
217bea790e0d   koku-hive-metastore   quay.io/mskarbek/ubi-hive:3.1.2-metastore-008   running   arm64          1000/tcp, 0.0.0.0:9083->8000/tcp
c9e68d2d0b1f   koku-minio            local-minio:latest                              running   x86_64         0.0.0.0:9000->9000/tcp, 0.0.0.0:9090->9090/tcp
46e9236c4530   koku-redis            redis:5.0.4                                     running                  0.0.0.0:6379->6379/tcp
fb618c665975   koku-unleash          unleashorg/unleash-server:3.9                   running                  0.0.0.0:4242->4242/tcp
dcdd08127c18   koku-db               postgres:13                                     running                  0.0.0.0:15432->5432/tcp
```

Part of me thinks that `masu-server` should maybe be `koku-masu-server`.

Dockerfile.minio
samdoran: I would like to update the version of Trino we use locally to more closely match what we use in production. I started making changes to get it working with a newer version but it was more involved.

maskarb: If you want a native image of Trino, just pull the repo and build the image. Let's not waste our time trying to make that a multi-arch image. It's not worth the effort until we decide to change the architecture of production.

There was a reason we were using the hotfix image. Instead of reverting to an old image, we should see if a new image works ok.

samdoran: It looks like we can use up to `RELEASE.2022-10-24T18-35-07Z` without much change. Starting with `RELEASE.2022-10-29T06-21-33Z`, the gateway and filesystem mode were removed, so existing local data needs to be migrated on the the `.trino/parquest_data` dir must be removed. I'm not sure how disruptive it is having to recreate/migrate. data stored in local buckets.

PR #4026, opened 2022-11-28
COST-3199 (pt2): Update tag exclusions to handle different tag keys.

myersCody:
## Jira Ticket

This pull request is to fix a scenario I missed when working on this [pull request](https://github.com/project-koku/koku/pull/3997)

The previous implementation of multiple tag exclusions only works for if you are using the same tag key. Therefore if you go to main and hit the following url:
```
http://localhost:8000/api/cost-management/v1/reports/aws/costs/?group_by[tag:app]=*&group_by[tag:storageclass]=*&exclude[tag:app]=analytics&exclude[tag:storageclass]=bravo
```
You will noticed that `anaytics` & `bravo` are still in the endpoint returns. 

In order to get `no-{option}` to show up in the return, I was adding emptyset filters that allow nulls. However, since we had two tag keys two null statements were being added:
```
OR (reporting_awscostentrylineitem_daily_summary.tags ->> 'app') IS NULL 
OR (reporting_awscostentrylineitem_daily_summary.tags ->> 'storageclass') IS NULL 
```
Turns out that both `analytics` & `bravo` now pass those where filters. In order to fix it I had to restructure how we combined the null filters so that they would be wrapped together by django. For example:
```
OR (reporting_awscostentrylineitem_daily_summary.tags -> 'storageclass' IS NULL AND reporting_awscostentrylineitem_daily_summary.tags -> 'app' IS NULL)
```

## Description

This change will ...

## Testing

Compare main and this branch with the following url and notice that this branch actually excludes bravo & analytics now.

## Notes

...

### DIFF
--- a/koku/api/report/queries.py
+++ b/koku/api/report/queries.py
@@ -249,6 +249,7 @@ class ReportQueryHandler(QueryHandler):
         composed_filters = composed_filters & and_composed_filters & or_composed_filters
         if tag_exclusion_composed:
             composed_filters = composed_filters & tag_exclusion_composed
+            LOG.info(f"composed_filters: {composed_filters}")
         return composed_filters
 
     def _check_for_operator_specific_exlusions(self, composed_filters):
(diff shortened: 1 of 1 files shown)


koku/api/report/queries.py
```diff
@@ -249,6 +249,7 @@ def _check_for_operator_specific_filters(self, filter_collection):
         composed_filters = composed_filters & and_composed_filters & or_composed_filters
         if tag_exclusion_composed:
             composed_filters = composed_filters & tag_exclusion_composed
+            LOG.info(f"composed_filters: {composed_filters}")
```
lcouzens, line 252: Do we want/need this?

lcouzens (dismissed):

lcouzens (approved):

PR #4040, opened 2022-12-05
COST-3199: Add ephemeral fallback for unleash negative filtering 

myersCody:
## Jira Ticket

[COST-3199](https://issues.redhat.com/browse/COST-3199)

## Description

This change will add a fallback function to the negative filtering unleash flag. The `environment` value in context comes from the unleash client itself, and is not something we pass in. This change should allow the negative filtering function to return true while running in the ephemeral environment. 

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Notes

...


maskarb (approved):
lgtm!

PR #4041, opened 2022-12-05
COST-3075: Fix excluded org units being visable when filtering by root

myersCody:
## Jira Ticket

[COST-3075](https://issues.redhat.com/browse/COST-3075)

## Description

This change will fix exclude org unit when filtering by root org. 

Example url:
```
http://localhost:8000/api/cost-management/v1/reports/aws/costs/?group_by[org_unit_id]=R_001&filter[org_unit_id]=R_001&exclude[org_unit_id]=OU_002
```

## Testing

1. Load test customer data
2. On main look at the example url and see that `OU_002` is in the results.
3. Switch to this branch, run `make clear-cache` and refresh the page to notice that `OU_002` is no longer in the results. 

## Notes

...

### DIFF
--- a/koku/api/report/aws/query_handler.py
+++ b/koku/api/report/aws/query_handler.py
@@ -668,6 +668,11 @@ select coalesce(raa.account_alias, t.usage_account_id)::text as "account",
                 .distinct("org_unit_id")
             )
 
+            # This is to remove the excluded values from the sub_org_units
+            if "org_unit_id" in list(self.query_table_exclude_keys):
+                for org_id in self.parameters.get_exclude("org_unit_id"):
+                    org_unit_list.append(org_id)
+
             sub_org_unit_list = None
             if org_unit_objects:
                 sub_org_units = []
--- a/koku/api/report/test/aws/test_queries.py
+++ b/koku/api/report/test/aws/test_queries.py
@@ -2658,6 +2658,19 @@ class AWSReportQueryTest(IamTestCase):
                         with self.subTest(breakdown):
                             self.assertAlmostEqual(group_dict[breakdown]["value"], filter_dict[breakdown]["value"], 6)
 
+    @patch("api.query_params.enable_negative_filtering", return_value=True)
+    def test_group_by_and_filter_with_excluded_org_unit(self, _):
+        """Test that the cost group_by and filter match."""
+        excluded_org_unit = "OU_002"
+        url = "?group_by[org_unit_id]=R_001&filter[org_unit_id]=R_001&exclude[org_unit_id]=OU_002"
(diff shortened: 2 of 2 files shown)


myersCody:
Current Results: https://ibutsu.apps.ocp4.prod.psi.redhat.com/runs/b693370a-4397-40ad-9160-138870a52a6f#classify-failures

I think these are related to: https://gitlab.cee.redhat.com/insights-qe/iqe-cost-management-plugin/-/merge_requests/1492

koku/api/report/aws/query_handler.py
```diff
@@ -668,6 +668,11 @@ def _get_sub_org_units(self, org_unit_list):
                 .distinct("org_unit_id")
             )
 
+            # This is to remove the excluded values from the sub_org_units
+            if "org_unit_id" in list(self.query_table_exclude_keys):
+                for org_id in self.parameters.get_exclude("org_unit_id"):
+                    org_unit_list.append(org_id)
```
samdoran, line 674: `QueryParams.get_exclude()` could `None` unless a default value is passed in.
```suggestion
            if "org_unit_id" in list(self.query_table_exclude_keys):
                for org_id in self.parameters.get_exclude("org_unit_id",[]):
                    org_unit_list.append(org_id)
```

This could also be written as a comprehension which would use less memory and perform better or large data sets.

```suggestion
            if "org_unit_id" in list(self.query_table_exclude_keys):
                org_unit_list.extend(org_id for org_id in self.parameters.get_exclude("org_unit_id", []))
```

koku/api/report/test/aws/test_queries.py
```diff
@@ -2658,6 +2658,19 @@ def test_filter_and_group_by_org_unit_match(self):
                         with self.subTest(breakdown):
                             self.assertAlmostEqual(group_dict[breakdown]["value"], filter_dict[breakdown]["value"], 6)
 
+    @patch("api.query_params.enable_negative_filtering", return_value=True)
+    def test_group_by_and_filter_with_excluded_org_unit(self, _):
```
samdoran, line 2662: If we don't need to modify the `Mock` it might be better to use it as a context manager within the test method to avoid using a throwaway variable name, or give this parameter a descriptive name. Descriptive names make it easier to keep track of multiple patch objects if more are added in the future. It's also clearer to the reader what this parameter is supposed to be without have to infer.

koku/api/report/aws/query_handler.py
myersCody: Thanks for the pointer @samdoran gonna add this to my python arsenal 🙇

samdoran (approved):

PR #4044, opened 2022-12-06
COST-3050 Repeated update of timestamp for OCI source

djnakabaale:
## Jira Ticket

[COST-3050](https://issues.redhat.com/browse/COST-3050)

## Description

This change will add skipping creating manifest records for months with no reports and  a function remove duplicate manifests to prevent repeatedly updating `last_manifest_complete_date` field.

## Testing

1. Checkout Branch
2. Restart Koku
3. Load some oci data with `make create-test-customer` and `make load-test-customer-data test_source=oci`
4. Hit the sources endpoint to get the oci source id, http://127.0.0.1:8000/api/cost-management/v1/sources/
5. Hit endpoint  `http://127.0.0.1:8000/api/cost-management/v1/sources/<id>/stats/`
    1. You should see unique timestamps in `last_manifest_complete_date` field for each manifest. 
    
6. Do more things...

## Notes

Actual endpoint should be similar to this http://127.0.0.1:8000/api/cost-management/v1/sources/11/stats/


koku/masu/external/downloader/oci/oci_report_downloader.py
```diff
@@ -244,9 +244,9 @@ def _extract_names(self, assembly_id, ingest_month):
         """
 
         last_reports = self.get_last_reports(assembly_id)
-        initial_ingest = True if last_reports == {"cost": "", "usage": ""} else False
-        usage_report = last_reports["usage"] if "usage" in last_reports else ""
-        cost_report = last_reports["cost"] if "cost" in last_reports else ""
+        initial_ingest = last_reports == {"cost": "", "usage": ""}
+        usage_report = last_reports.get("usage", "")
+        cost_report = last_reports.get("cost", "")
```
samdoran, line 249: Ternary operators are very nice, but I like the way this reads much better.

koku/masu/processor/tasks.py
```diff
@@ -293,7 +294,7 @@ def remove_expired_data(schema_name, provider, simulate, provider_uuid=None, que
 
 
 @celery_app.task(name="masu.processor.tasks.summarize_reports", queue=SUMMARIZE_REPORTS_QUEUE)
-def summarize_reports(reports_to_summarize, queue_name=None, manifest_list=None):
+def summarize_reports(reports_to_summarize, queue_name=None, manifest_list=None):  # noqa C901
```
samdoran, line 297: Rather than accepting the increased complexity, try and make this simpler. It's right on the edge at 10 right now.

koku/masu/processor/tasks.py
```diff
@@ -315,6 +316,9 @@ def summarize_reports(reports_to_summarize, queue_name=None, manifest_list=None)
         ends = []
         if report and report.get("provider_type") in [Provider.PROVIDER_GCP, Provider.PROVIDER_GCP_LOCAL]:
             reports_deduplicated += deduplicate_reports_for_gcp(report_list)
+        elif report and report.get("provider_type") in [Provider.PROVIDER_OCI, Provider.PROVIDER_OCI_LOCAL]:
+            manifest_list = []
+            reports_deduplicated += deduplicate_reports_for_oci(report_list)
```
samdoran, line 321: This may not be true in newer Python versions, but using the `+=` operator uses more memory than the `.extend()` and `.append()` methods. This is because Python first has to build both lists is memory before combining them instead of modifying the existing list.
```suggestion
            reports_deduplicated.extend(deduplicate_reports_for_oci(report_list))
```

samdoran: Here is a rough example of how to keep the complexity of this method down by using a map to get the correct deduplicating function.
```python
        dedupe_func_map = {
            Provider.PROVIDER_GCP: deduplicate_reports_for_gcp,
            Provider.PROVIDER_GCP_LOCAL: deduplicate_reports_for_gcp,
            Provider.Provider_OCI: deduplicate_reports_for_oci,
            Provider.Provider_OCI_LOCAL: deduplicate_reports_for_oci,
        }
        provider_type = report.get("provider_type")
        if report and provider_type in dedupe_func_map:
            dedupe_func = dedupe_func_map.get(provider_type)
            reports_deduplicated.extend(dedupe_func())
```

koku/masu/processor/tasks.py
djnakabaale: @samdoran I truly appreciate this comment and suggestion below.

koku/masu/util/oci/common.py
```diff
@@ -161,3 +161,26 @@ def detect_type(report_path):
     else:
         report_type = "usage"
     return report_type
+
+
+def deduplicate_reports_for_oci(report_list):
+    """Remove deduplicate oci manifests"""
```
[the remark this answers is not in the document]
samdoran: I'm wondering if this function could be replaced with `set(report_list)`. That would only work if all the reports contain identical keys and values (which is how the fixture data in the tests are).

koku/masu/test/processor/test_tasks.py
```diff
@@ -351,35 +351,39 @@ def test_summarize_reports_processing_list(self, mock_update_summary):
...
-                report2_meta["provider_uuid"] = self.ocp_test_provider_uuid
+                report2_meta["provider_type"] = provider_type
+                report2_meta["provider_uuid"] = provider_uuid
                 report2_meta["manifest_id"] = 2
                 report2_meta["start"] = str(DateHelper().yesterday)
                 report2_meta["end"] = str(DateHelper().today)
-                if provider_dict.get("type") == Provider.PROVIDER_GCP:
-                    report_meta["invoice_month"] = invoice_month
+                if provider_type == Provider.PROVIDER_GCP:
+                    report2_meta["invoice_month"] = invoice_month
```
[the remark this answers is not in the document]
djnakabaale: Yes, to add `invoice_month` for the second GCP test report.

koku/masu/external/downloader/oci/oci_report_downloader.py
```diff
@@ -302,13 +302,14 @@ def get_manifest_context_for_date(self, date):
...
             assembly_id = ":".join([str(self._provider_uuid), str(invoice_month)])
             month_file_names = self._extract_names(assembly_id, month)
-            manifest_id = self._process_manifest_db_record(assembly_id, str(month), len(month_file_names), dh._now)
-            monthly_report["manifest_id"] = manifest_id
-            monthly_report["assembly_id"] = assembly_id
-            monthly_report["compression"] = manifest_dict.get("compression")
             files_list = [{"key": key, "local_file": self.get_local_file_for_report(key)} for key in month_file_names]
-            monthly_report["files"] = files_list
-            report_manifests_list.append(monthly_report)
+            if len(files_list) > 0:
```
samdoran, line 306: A truthy check is equivalent and more idiomatic.
```suggestion
            if files_list:
```

koku/masu/external/downloader/oci_local/oci_local_report_downloader.py
```diff
@@ -145,13 +145,15 @@ def get_manifest_context_for_date(self, date):
...
-            monthly_report["files"] = files_list
-            report_manifests_list.append(monthly_report)
+            if len(files_list) > 0:
+                manifest_id = self._process_manifest_db_record(assembly_id, str(month), len(month_file_names), dh._now)
+                monthly_report["manifest_id"] = manifest_id
+                monthly_report["assembly_id"] = assembly_id
+                monthly_report["compression"] = manifest_dict.get("compression", "")
+                monthly_report["files"] = files_list
+                report_manifests_list.append(monthly_report)
+        LOG.info(f"Report Manifests List: {str(report_manifests_list)}")
```
samdoran, line 156: Inside f-strings, `str()` is called by default on values inside expressions.
```suggestion
        LOG.info(f"Report Manifests List: {report_manifests_list}")
```

koku/masu/util/oci/common.py
```diff
@@ -161,3 +161,15 @@ def detect_type(report_path):
...
     return report_type
+
+
+def deduplicate_reports_for_oci(report_list):
+    """Remove duplicate oci manifests"""
+    reports_deduplicated = []
+    manifest_id_set = set()
+    for report in report_list:
+        _manifest_id = report.get("manifest_id")
+        if _manifest_id and _manifest_id not in manifest_id_set:
```
samdoran, line 172: I was curious if the walrus operator would work here, so I tried it.
```suggestion
        if (_manifest_id := report.get("manifest_id")) and _manifest_id not in manifest_id_set:
```

I don't think this is better (more readable) than what you have currently, but it's interesting.

samdoran (dismissed):

koku/masu/util/oci/common.py
djnakabaale: I'm going to lean towards readability here 😀

samdoran (approved):

PR #4049, opened 2022-12-08
[COST-3311] - Match volume to node on namespace too

adberglund:
## Jira Ticket

[COST-3311](https://issues.redhat.com/browse/COST-3311)

## Description

This change will match OCP PVCs to a node using pod AND namespace, instead of just pod. Unfortuantely, the same pod name can exist in multiple namespaces, which can be on any node, and we ended up over matching a PVC to many nodes when the pod name existed on multiple nodes. 

## Testing

1. Currently does not seem possible to get nise data to add the same pod name to multiple namespaces on different nodes.

## Notes

...


adberglund:
https://ci.ext.devshift.net/job/project-koku-koku-pr-check/7638/ 🚬

cgoodfred (approved):

PR #4060, opened 2022-12-13
[COST-3202] tag based monthly amortization

adberglund:
## Jira Ticket

[COST-3202](https://issues.redhat.com/browse/COST-3202)

## Description

This change will add handling for tag based rates with the new cost columns.
## Testing

1. Checkout Branch
2. `make create-test-customer` and `make load-test-customer-data`
3. WAIT
4. Run SQL query in org1234566 schema 
```
select node, pod_labels, sum(cost_model_cpu_cost) from reporting_ocpusagelineitem_daily_summary where monthly_cost_type = 'Node' and cost_model_rate_type = 'Supplementary' and usage_start between '2022-12-01' and '2022-12-31' group by node, pod_labels;
```
  a . See that nodes for a full month have the entire cost per node, including default rates from the updated on prem cost model json in this PR.
  b. Repeat for January and see that each node has n / 31 where n is the current day this is tested on.
 5. Run SQL query in org1234567 schema 
 ```
 select persistentvolumeclaim, volume_labels, sum(cost_model_volume_cost) from reporting_ocpusagelineitem_daily_summary where monthly_cost_type = 'PVC' and usage_start between '2022-12-01' and '2022-12-31' group by persistentvolumeclaim, volume_labels;
 ```
  a. Do the same checks as described for node costs above.
## Notes

...

### DIFF
--- a/koku/masu/test/processor/ocp/test_ocp_cost_model_cost_updater.py
+++ b/koku/masu/test/processor/ocp/test_ocp_cost_model_cost_updater.py
@@ -780,7 +780,7 @@ class OCPCostModelCostUpdaterTest(MasuTestCase):
         updater._delete_tag_usage_costs(start_date, end_date, "")
         self.assertLogs("masu.processor.ocp", level="WARNING")
 
-    def test_build_node_tag_rate_case_statement_str(self):
+    def test_build_node_tag_cost_case_statements(self):
         """Make sure a correct SQL CASE statement is generated."""
         tag_key = "app"
         dh = DateHelper()
@@ -831,7 +831,7 @@ class OCPCostModelCostUpdaterTest(MasuTestCase):
             "NULL::decimal as cost_model_volume_cost",
         )
 
-        case_dict = self.updater._build_node_tag_rate_case_statement_str(
+        case_dict = self.updater._build_node_tag_cost_case_statements(
             tag_rate_dict, start_date, default_rate_dict=default_rate_dict
         )
 
@@ -845,14 +845,14 @@ class OCPCostModelCostUpdaterTest(MasuTestCase):
         )
 
         self.updater._distribution = metric_constants.MEMORY_DISTRIBUTION
-        case_dict = self.updater._build_node_tag_rate_case_statement_str(
+        case_dict = self.updater._build_node_tag_cost_case_statements(
             tag_rate_dict, start_date, default_rate_dict=default_rate_dict
         )
 
         for actual, expected in zip(case_dict.get(tag_key), expected_case_strs):
             self.assertEqual(actual.replace("\n", "").replace(" ", ""), expected.replace("\n", "").replace(" ", ""))
 
-    def test_build_volume_tag_rate_case_statement_str(self):
+    def test_build_volume_tag_cost_case_statements(self):
         """Make sure a correct SQL CASE statement is generated."""
         tag_key = "app"
         dh = DateHelper()
@@ -881,7 +881,7 @@ class OCPCostModelCostUpdaterTest(MasuTestCase):
             """,
         )
 
-        case_dict = self.updater._build_volume_tag_rate_case_statement_str(
+        case_dict = self.updater._build_volume_tag_cost_case_statements(
             tag_rate_dict, start_date, default_rate_dict=default_rate_dict
         )
 


koku/masu/database/ocp_report_db_accessor.py
```diff
@@ -709,6 +709,64 @@ def populate_monthly_cost(  # noqa: C901
...
+        table_name = self._table_map["line_item_daily_summary"]
+        report_period = self.report_periods_for_provider_uuid(provider_uuid, start_date)
+        with schema_context(self.schema):
+            report_period_id = report_period.id
+
+        cpu_case, memory_case, volume_case = case_dict.get("allocated")
+        labels = case_dict.get("labels")
+
+        if cost_type == "Node":
+            summary_sql = pkgutil.get_data("masu.database", "sql/openshift/cost_model/monthly_cost_node_by_tag.sql")
```
samdoran, line 730: It's a good convention to prefix vars containing bytes with `b_` so it's clear which vars are bytes and which are text. Not something to do in this PR, but something to think about.

koku/masu/processor/ocp/ocp_cost_model_cost_updater.py
```diff
@@ -175,6 +170,130 @@ def _update_markup_cost(self, start_date, end_date):
...
+            """
+
+        for tag_key, tag_value_rates in rate_dict.items():
+            cpu_statement_list = ["CASE"]
+            memory_statement_list = ["CASE"]
+            for tag_value, rate_value in tag_value_rates.items():
+                label_condition = f"pod_labels->>'{tag_key}'='{tag_value}'"
+                if unallocated is True:
+                    label_condition_list = ["pod_labels = '{", f'"{tag_key}": "{tag_value}"', "}'"]
+                    label_condition = "".join(label_condition_list)
```
samdoran, line 202: This can be a single format string which returns a sting like this: `pod_labels = '{"tk": "tv"}'`
```suggestion
                    label_condition_list = ["pod_labels = '{", f'"{tag_key}": "{tag_value}"', "}'"]
                    label_condition = f'pod_labels = \'{{"{tag_key}": "{tag_value}"}}\''
```

koku/masu/processor/ocp/ocp_cost_model_cost_updater.py
```diff
@@ -175,6 +170,130 @@ def _update_markup_cost(self, start_date, end_date):
...
+                rate = get_amortized_monthly_cost_model_rate(rate_value, start_date)
+                cpu_statement_list.append(
+                    f"""
+                        WHEN {label_condition}
+                            THEN {cpu_distribution_term}
+                                * {rate}::decimal
+                    """
```
samdoran, line 209: Does it matter that all the leading white space here is preserved? Should this be run through `textwrap.dedent()`?

koku/masu/processor/ocp/ocp_cost_model_cost_updater.py
```diff
@@ -50,11 +50,6 @@ def __init__(self, schema, provider):
             self._tag_default_supplementary_rates = cost_model_accessor.tag_default_supplementary_rates
             self._distribution = cost_model_accessor.distribution
 
-    @property
-    def is_amortized(self):
```
samdoran, line 54: Any particular reason this public property was removed without a deprecation period and now the private attribute is being used directly?

This public property isn't doing much today, but allows how the value is determined to change in the future without breaking any calling code. It also provides a way to cache the value on the object if it is expensive to calculate.

koku/masu/processor/ocp/ocp_cost_model_cost_updater.py
adberglund: Nah. The SQL runs great.

koku/masu/processor/ocp/ocp_cost_model_cost_updater.py
adberglund: It was introduced in the previous PR as a hacky way to get unit tests to run. We fixed that, but forgot to remove the property until this PR. I love properties, but we don't need a depreciation period here. Nothing else uses this and we don't have 3rd parties using things where we need a period like that. The goal is to remove this check entirely but for now this let's us do both paths based on the unleash flag while we're getting this feature change out. You'd also need a cached_property to cache, otherwise it's calculated every time.

koku/masu/processor/ocp/ocp_cost_model_cost_updater.py
```diff
@@ -282,6 +400,82 @@ def _update_monthly_tag_based_cost(self, start_date, end_date):
...
+                        elif cost_type == "PVC":
+                            key_rate_dict = self._build_volume_tag_rate_case_statement_str(
+                                rate_dict, start_date, default_rate_dict
+                            )
+                            labels = self._build_labels_case_statement_str(
+                                rate_dict, "volume_labels", default_rate=default_rate_dict
+                            )
+
+                        for tag_key, case_statements in key_rate_dict.items():
+                            unallocated_case_statements = unallocated_key_rate_dict.get(tag_key, {})
```
maskarb, line 458: Will `cost_type` always be `PVC` or `Node`? It looks like there is also `Cluster`, so how does this `for` loop not throw a `variable referenced before assignment` error when iterating over the `key_rate_dict.items()` when the cost_type is `Cluster`?

Also, is `unallocated_key_rate_dict` defined anywhere other than if the cost_type is `Node`?

maskarb: Oh I see whats happening here. The first cost_type in `OCPUsageLineItemDailySummary.MONTHLY_COST_RATE_MAP` is `Node`, so `key_rate_dict` and `unallocated_key_rate_dict` are defined the first time thru the `for` loop and then re-used the next time through. Is this intentional?

adberglund: Prolly not, I shall fix.

maskarb (approved):
lgtm!

maskarb (approved):
lgtm!

PR #4068, opened 2023-01-03
drop specific sentry transactions

maskarb:
## Description

This change will add a block-list to prevent sending metrics on readiness/liveness checks and sources-status POSTs.

## Notes

This change is necessary to limit the impact cost-mgmt has against the enterprise sentry transaction quota.
...

### DIFF
--- a/koku/koku/sentry.py
+++ b/koku/koku/sentry.py
@@ -16,8 +16,9 @@ BLOCK_LIST = [
 
 
 def traces_sampler(sampling_context):
-    if sampling_context["transaction_context"]["op"] == "http.server" and any(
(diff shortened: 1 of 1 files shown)


adberglund (approved):

koku/koku/sentry.py
```diff
@@ -9,13 +9,29 @@
 
 LOG = logging.getLogger(__name__)
 
+BLOCK_LIST = [
+    "/api/cost-management/v1/status/",
+    "/api/cost-management/v1/source-status/",
+]
```
samdoran, line 15: Since these values are unique, this can be a set.
```suggestion
BLOCK_LIST = {
    "/api/cost-management/v1/status/",
    "/api/cost-management/v1/source-status/",
}
```

PR #4073, opened 2023-01-03
COST-3366/3367: Update category filter for tag endpoint.

myersCody:
## Jira Ticket

[COST-3366](https://issues.redhat.com/browse/COST-3366)
[COST-3367](https://issues.redhat.com/browse/COST-3367)

## Description

This change will fix filter[category]=* on tag endpoints and also disable the usage of category={something} on these endpoints.

## Testing

1. Checkout Branch
2. Restart Koku
3. Make load data
4. See the following throws an error http://localhost:8000/api/cost-management/v1/tags/openshift/?category=Platform
5. See tags filtered by category via http://localhost:8000/api/cost-management/v1/tags/openshift/?filter[category]=Platform

## Notes

...

### DIFF
--- a/koku/api/report/serializers.py
+++ b/koku/api/report/serializers.py
@@ -324,11 +324,13 @@ class ParamSerializer(BaseSerializer):
         """
         super().validate(data)
 
-        if data.get("category") and ("key_only" not in data.keys()):
-            # allow category to be passed in without group by for the tags endpoint
-            if data.get("category") and (not data.get("group_by") or not data.get("group_by").get("project")):
-                error = {"error": ("Category may not be used without a group_by project parameter")}
-                raise serializers.ValidationError(error)
+        if data.get("category") and ("key_only" in data.keys()):
+            error = {"error": ("Category may only be used as a filter for the tag endpoints.")}
+            raise serializers.ValidationError(error)
+
+        if data.get("category") and (not data.get("group_by") or not data.get("group_by").get("project")):
+            error = {"error": ("Category may not be used without a group_by project parameter")}
+            raise serializers.ValidationError(error)
 
         if not data.get("currency"):
             data["currency"] = get_currency(self.context.get("request"))
@@ -382,8 +384,6 @@ class ParamSerializer(BaseSerializer):
         Raises:
             (ValidationError): if field inputs are invalid
         """
-        # Categories should always be capitalized
-        value = [item.title() for item in value]
         categories = None
         db_categories = OpenshiftCostCategory.objects.values_list("name", flat=True).distinct()
         if ReportQueryHandler.has_wildcard(value):
--- a/koku/api/report/serializers.py
+++ b/koku/api/report/serializers.py
@@ -324,7 +324,7 @@ class ParamSerializer(BaseSerializer):
         """
         super().validate(data)
 
-        if data.get("category")
+        if data.get("category"):
             if "key_only" in data:
                 error = {"error": "Category may only be used as a filter for the tag endpoints."}
                 raise serializers.ValidationError(error)
--- a/koku/api/report/serializers.py
+++ b/koku/api/report/serializers.py
@@ -329,7 +329,7 @@ class ParamSerializer(BaseSerializer):
                 error = {"error": "Category may only be used as a filter for the tag endpoints."}
                 raise serializers.ValidationError(error)
 
-             if not data.get("group_by") or not data.get("group_by").get("project"):
+            if not data.get("group_by") or not data.get("group_by").get("project"):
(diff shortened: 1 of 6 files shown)


koku/api/report/serializers.py
```diff
@@ -324,11 +324,13 @@ def validate(self, data):
         """
         super().validate(data)
 
-        if data.get("category") and ("key_only" not in data.keys()):
-            # allow category to be passed in without group by for the tags endpoint
-            if data.get("category") and (not data.get("group_by") or not data.get("group_by").get("project")):
-                error = {"error": ("Category may not be used without a group_by project parameter")}
-                raise serializers.ValidationError(error)
+        if data.get("category") and ("key_only" in data.keys()):
+            error = {"error": ("Category may only be used as a filter for the tag endpoints.")}
```
samdoran, line 328: Is `error` supposed to contain a single item tuple? If it's supposed to be a string, then the parenthesis aren't necessary.

myersCody: Thanks @samdoran that was bad copy & paste. 🙇

koku/api/report/serializers.py
```diff
@@ -324,11 +324,13 @@ def validate(self, data):
         """
         super().validate(data)
 
-        if data.get("category") and ("key_only" not in data.keys()):
-            # allow category to be passed in without group by for the tags endpoint
-            if data.get("category") and (not data.get("group_by") or not data.get("group_by").get("project")):
-                error = {"error": ("Category may not be used without a group_by project parameter")}
-                raise serializers.ValidationError(error)
+        if data.get("category") and ("key_only" in data.keys()):
```
samdoran, line 327: The `dict_keys` view isn't needed here.
```suggestion
        if data.get("category") and ("key_only" in data):
```

myersCody: Makes me wonder where else we might be doing this. 🤨 I think I may push up a separate pr to fix a few other places where we do `x in y.keys()`. Thanks for pointing this out Sam, it was some new learning for me. 🙇

koku/api/report/serializers.py
samdoran: I've been bitten many times by single item tuple bugs. 😄

koku/api/report/serializers.py
samdoran: This could be fixed ad-hoc. It's probably not worth making an explicit PR just to change this.

Calling `dict.keys()` returns a `dict_keys` [view object](https://docs.python.org/3.11/library/stdtypes.html#dict-views), and then we do the test on that object. It's just an extra step that's not needed here.

koku/api/report/serializers.py
```diff
@@ -324,11 +324,13 @@ def validate(self, data):
...
-            if data.get("category") and (not data.get("group_by") or not data.get("group_by").get("project")):
-                error = {"error": ("Category may not be used without a group_by project parameter")}
-                raise serializers.ValidationError(error)
+        if data.get("category") and ("key_only" in data):
+            error = {"error": "Category may only be used as a filter for the tag endpoints."}
+            raise serializers.ValidationError(error)
+
+        if data.get("category") and (not data.get("group_by") or not data.get("group_by").get("project")):
+            error = {"error": "Category may not be used without a group_by project parameter"}
+            raise serializers.ValidationError(error)
```
samdoran, line 333: Moving the overlapping condition to the top makes this a bit easier to read.
```suggestion
        if data.get("category")
            if "key_only" in data:
                error = {"error": "Category may only be used as a filter for the tag endpoints."}
                raise serializers.ValidationError(error)

             if not data.get("group_by") or not data.get("group_by").get("project"):
                error = {"error": "Category may not be used without a group_by project parameter"}
                raise serializers.ValidationError(error)
```

I was also thinking the second `data.get("group_by")` should return `{}` rather than `None` to avoid a potential `AttributeError`, but an "or" comparison will short circuit on the first false condition, so I think it should be ok to leave that as is.

samdoran (approved):

samdoran (approved):

koku/api/report/serializers.py
```diff
@@ -324,10 +324,13 @@ def validate(self, data):
         """
         super().validate(data)
 
-        if data.get("category") and ("key_only" not in data.keys()):
-            # allow category to be passed in without group by for the tags endpoint
-            if data.get("category") and (not data.get("group_by") or not data.get("group_by").get("project")):
-                error = {"error": ("Category may not be used without a group_by project parameter")}
+        if data.get("category")
```
samdoran, line 327:
```suggestion
        if data.get("category"):
```

djnakabaale (approved):
lgtm

PR #4074, opened 2023-01-04
[COST-3370] - drop ranking field from query data

maskarb:
## Jira Ticket

[COST-3370](https://issues.redhat.com/browse/COST-3370)

## Description

This change will ensure the ranking field is dropped from the original query data.
This query:
```
/?filter[limit]=50&filter[offset]=0&filter[resolution]=monthly&filter[time_scope_units]=month&filter[time_scope_value]=-2&group_by[project]=*&order_by[request]=desc
```
would result in the  `ranks_by_day` and `data_frame` containing a `request` column. When these dataframes were merged, it resulted in a new DF with `request_x` and `request_y` fields. Further along in the query_handler, when ordering the data, we would try to sort by `request` but this field no longer existed in the data.

## Testing

1. Ingest OCP data.
2. Go to http://localhost:8000/api/cost-management/v1/reports/openshift/compute/?filter[limit]=50&filter[offset]=0&filter[resolution]=monthly&filter[time_scope_units]=month&filter[time_scope_value]=-2&group_by[project]=*&order_by[request]=desc
3. see data.
4. Go to http://localhost:8000/api/cost-management/v1/reports/openshift/compute/?filter[limit]=50&filter[offset]=0&filter[resolution]=monthly&filter[time_scope_units]=month&filter[time_scope_value]=-2&group_by[project]=*&order_by[delta]=desc&delta=request__limit
5. see data.

## Notes

...

### DIFF
--- a/koku/api/report/queries.py
+++ b/koku/api/report/queries.py
@@ -1021,9 +1021,9 @@ class ReportQueryHandler(QueryHandler):
             if rank_value not in rankings:
                 rankings.append(rank_value)
                 distinct_ranks.append(rank)
-        return self._ranked_list(data, distinct_ranks)
+        return self._ranked_list(data, distinct_ranks, list(rank_annotations.keys()))
 
-    def _ranked_list(self, data_list, ranks):
+    def _ranked_list(self, data_list, ranks, rank_field):
         """Get list of ranked items less than top.
 
         Args:
@@ -1037,7 +1037,7 @@ class ReportQueryHandler(QueryHandler):
         group_by = self._get_group_by()
         self.max_rank = len(ranks)
         # Columns we drop in favor of the same named column merged in from rank data frame
-        drop_columns = ["cost_units", "source_uuid"]
+        drop_columns = ["cost_units", "source_uuid"] + rank_field
         if self.is_openshift:
             drop_columns.append("clusters")
 
--- a/koku/api/report/queries.py
+++ b/koku/api/report/queries.py
@@ -1021,20 +1021,21 @@ class ReportQueryHandler(QueryHandler):
             if rank_value not in rankings:
                 rankings.append(rank_value)
                 distinct_ranks.append(rank)
-        return self._ranked_list(data, distinct_ranks, list(rank_annotations))
+        return self._ranked_list(data, distinct_ranks, set(rank_annotations))
 
-    def _ranked_list(self, data_list, ranks, rank_field=None):
+    def _ranked_list(self, data_list, ranks, rank_fields=None):
         """Get list of ranked items less than top.
 
         Args:
             data_list (List(Dict)): List of ranked data points from the same bucket
             ranks (List): list of ranks to use; overrides ranking that may present in data_list.
+            rank_fields (Set): the fields on which ranking is performed.
         Returns:
             List(Dict): List of data points meeting the rank criteria
 
         """
-        if not rank_field:
-            rank_field = []
+        if not rank_fields:
+            rank_fields = set()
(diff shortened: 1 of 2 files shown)


adberglund (dismissed):

koku/api/report/queries.py
```diff
@@ -1021,9 +1021,9 @@ def _group_by_ranks(self, query, data):  # noqa: C901
             if rank_value not in rankings:
                 rankings.append(rank_value)
                 distinct_ranks.append(rank)
-        return self._ranked_list(data, distinct_ranks)
+        return self._ranked_list(data, distinct_ranks, list(rank_annotations.keys()))
```
samdoran, line 1024: Passing a dictionary into `list()` will return a list of keys.
```suggestion
        return self._ranked_list(data, distinct_ranks, list(rank_annotations))
```

koku/api/report/queries.py
```diff
@@ -1021,9 +1021,9 @@ def _group_by_ranks(self, query, data):  # noqa: C901
...
                 rankings.append(rank_value)
                 distinct_ranks.append(rank)
-        return self._ranked_list(data, distinct_ranks)
+        return self._ranked_list(data, distinct_ranks, list(rank_annotations))
 
-    def _ranked_list(self, data_list, ranks):
+    def _ranked_list(self, data_list, ranks, rank_field=None):
         """Get list of ranked items less than top.
 
         Args:
```
samdoran, line 1029: New arg needs to be added to the doc string.

koku/api/report/queries.py
```diff
@@ -1075,6 +1077,10 @@ def _ranked_list(self, data_list, ranks):
         # Cross join ranks and days to get each field/rank for every day in th query
         ranks_by_day = rank_data_frame.merge(day_data_frame, how="cross")
 
+        # add the ranking column if it still exists in both dataframes
+        if rank_field and rank_field[0] in ranks_by_day.columns.intersection(data_frame.columns):
```
samdoran, line 1081: Is `rank_field` expected to be a single item list passed into this method?

maskarb: yesf

maskarb: On second look, it does appear that rank_field may contain more than 1 entry. I'll update this logic here to check for all possible entries.

maskarb: Here is a query which results in a `rank_field` containing multiple fields:
```
api/cost-management/v1/reports/openshift/compute/?filter[limit]=50&filter[offset]=0&filter[resolution]=monthly&filter[time_scope_units]=month&filter[time_scope_value]=-2&group_by[project]=*&order_by[delta]=desc&delta=request__limit
```

koku/api/report/queries.py
```diff
@@ -1075,6 +1078,11 @@ def _ranked_list(self, data_list, ranks):
         # Cross join ranks and days to get each field/rank for every day in th query
         ranks_by_day = rank_data_frame.merge(day_data_frame, how="cross")
 
+        # add the ranking columns if they still exists in both dataframes
+        rank_fields.intersection_update(set(ranks_by_day.columns.intersection(data_frame.columns)))
```
samdoran, line 1082: `set` methods accept an iterable and don't require a `set`. The `pandas.core.indexes.numeric.Int64Index` returned by the `intersection` method is iterable.
```suggestion
        rank_fields.intersection_update(ranks_by_day.columns.intersection(data_frame.columns))
```

adberglund (approved):

PR #4077, opened 2023-01-05
action: add retry for manifest sanity check

maskarb:
The Sanity CI check has failed due to rate limiting on a few occasions. This PR adds a retry to the Sanity check so we can hopefully avoid having to manually restart the check.


maskarb:
This retry doesn't work. Need to do something else for the rate-limiting.

PR #4089, opened 2023-01-10
COST-3045: Unexpected items in GCP storage during multiple groupby.

myersCody:
## Jira Ticket

[COST-3045](https://issues.redhat.com/browse/COST-3045)

## Description

This change will fix the `*` functionality for the category filter:

1) http://localhost:8000/api/cost-management/v1/tags/openshift/?filter[category]=*

Expected behavior is to return everything. I forgot that the `*` would also return None. 😓 

2) Fix unexpected items in the GCP storage during multiple groupby/filter.

As well as fix unexpected items showing up in the with multiple groupby/filter. The issue is that when we use a combination out
side of the one listed in the provider map we default to the daily summary table. When that default happens we were not using the same filters in the [where conditions](https://github.com/project-koku/koku/blob/main/koku/masu/database/sql/gcp/reporting_gcp_storage_summary_p.sql#L36) to build the storage table. So, I added these conditions to when the daily summary table is being used on the gcp storage endpoint. 

## Testing

Load test customer data:

1. Hit: http://localhost:8000/api/cost-management/v1/tags/openshift/?filter[category]=* See that it returns everything.
2. Hit: http://localhost:8000/api/cost-management/v1/tags/openshift/?filter[category]=whatever see that it returns nothing

I followed the reproduction steps left by Eva in the issue. I included examples of the endpoints:

1. [storage.txt](https://github.com/project-koku/koku/files/10393119/storage.txt)
2. [group_by_service.txt](https://github.com/project-koku/koku/files/10393120/group_by_service.txt)
3. [group_by_service_project.txt](https://github.com/project-koku/koku/files/10393121/group_by_service_project.txt)
4. [group_by_service_filter_project.txt](https://github.com/project-koku/koku/files/10393122/group_by_service_filter_project.txt)

## Notes

...


myersCody:
successful smoke tests: https://url.corp.redhat.com/ibutsu-runs-827ff32c-fb04-4925-a90b-681f30f45817

cgoodfred (approved):

PR #4098, opened 2023-01-13
COST-3217 - OCI previous month's data not ingested locally

djnakabaale:
## Jira Ticket

[COST-3217](https://issues.redhat.com/browse/COST-3217)

## Description

This change will update `oci_static_data.yml` to resolve first of month issue caused due to bad data from nise.

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Notes

...
Changes to nise were added in this pr https://github.com/project-koku/nise/pull/430


dev/scripts/nise_ymls/oci/oci_static_data.yml
```diff
@@ -3,24 +3,49 @@ generators:
...
+      consumed_quantity: 14634
       currency: USD
       compartment_name: richardprice
       tenant_id: ocid1.tenancy.oc1..EfjkUPxyZSYLvd
+      subscription_id: 25906734
+      tags:
+        tags/free-form-tag: our
   - OCIDatabaseGenerator:
       start_date: {{start_date}}
       end_date: {{end_date}}
```
samdoran, line 41: Do these keys need to be removed since they are duplicates now?

djnakabaale: I think we should keep these to be semi-consistent with other providers.

maskarb (approved):

PR #4109, opened 2023-01-23
Cost 3298

lcouzens:
## Jira Ticket

[COST-3298](https://issues.redhat.com/browse/COST-3298)

## Description

This change will add a new HCS report POST endpoint and process the new posted data.

## Testing

1. Checkout Branch
2. Restart Koku
3. Create a storage-only AWS source - data-source includes 
{
        "bucket": "my-bucket",
        "storage-only": "True"
    },
4. Post to the new endpoint a marketplace report stored in S3
    a. Endpoint: /api/cost-management/v1/ingress/reports/
    b. Example: {"provider": "9e7159f1-c5fe-4455-9881-ee730654cf84", "reports_list": ["MY_BUCKET/2023/01/18/34472771-acab-4ca4-8c6a-8721d9b91188/cc3ecf82-9960-4f02-9a31-b9163f75e6b7.csv"]}
5. See that the data is ingested correctly

## Notes
Reach out to Luke if you want real AWS buckets/run down to test this.
...


koku/api/common/permissions/hcs_access.py
```diff
@@ -0,0 +1,20 @@
...
+"""Defines the Settings Access Permissions class."""
+from rest_framework import permissions
+
+
+class HCSAccessPermission(permissions.BasePermission):
+    """Determines if a user can post HCS data."""
+
+    def has_permission(self, request, view):
+        """Check permission to view and post HCS report data."""
+        if request.user.admin:
```
chambridge, line 14: Does this mean only and Org Admin user can Post?

koku/api/report/hcs/view.py
```diff
@@ -0,0 +1,80 @@
...
+from api.common.permissions.hcs_access import HCSAccessPermission
+from api.report.hcs.serializers import HCSReportSerializer
+from reporting.hcs.models import HCSReport
+
+LOG = logging.getLogger(__name__)
+
+
+class HCSDetailView(APIView):
+    """
+    View to fetch HCS report details for specific provider
```
chambridge, line 23: I would use `source` instead of `provider` for external language consistency for the user.

koku/api/report/hcs/view.py
```diff
@@ -0,0 +1,80 @@
+#
+# Copyright 2023 Red Hat Inc.
+# SPDX-License-Identifier: Apache-2.0
+#
+"""View for HCS report posting."""
```
chambridge, line 5: I probably wouldn't use HCS everywhere its just one (and current primary target for this.

Might call it something like simplified, or minimal cloud report ingestion.

koku/api/urls.py
```diff
@@ -335,6 +337,8 @@
         ),
         name="reports-openshift-azure-instance-type",
     ),
+    path("reports/hcs/", HCSView.as_view(), name="reports-hcs"),
```
chambridge, line 340: Again wouldn't used `hcs` everywhere, or provider.

koku/masu/external/downloader/aws/aws_report_downloader.py
```diff
@@ -177,6 +176,10 @@ def _get_manifest(self, date_time):
             (Dict): A dict-like object serialized from JSON data.
 
         """
+        # Checking if initial ingest is for a HCS manifest DO NOT PROCESS
+        if "Parquet" == self.report.get("Compression"):
+            LOG.info("Skipping ingest since data is Parquet compressed and likely for HCS")
```
chambridge, line 181: So does this bypass Parquet support for now unless they are going the customer filtered path? I think that is ok, but it would be nice to support more at some point.

chambridge (commented):
left some comments around naming permissioning and first pas.

koku/api/common/permissions/hcs_access.py
lcouzens: Great question, I can't say I have actually tested it strictly but I believe that would be the case based on how others work. Since this is being posted potentially by a Lambda function we probably need to change this. :sweat_smile:

koku/api/report/hcs/view.py
lcouzens: I hate the Source/Provider stuff! We say source for a customer but we actually want the Provider UUID for our flow, so confusing! But I agree we can change this :grin:

koku/masu/external/downloader/aws/aws_report_downloader.py
lcouzens: Right this essentially means a customer can add a bucket that has Parquet reports in (used with athena) And we just skip ingest. On the assumption that they really want to be posting the minimal report instead. Are you saying we should just straight support parquet files too? That feels like a lot more work :grin:

koku/api/common/permissions/minimal_report_access.py
```diff
@@ -0,0 +1,21 @@
+#
+# Copyright 2023 Red Hat Inc.
+# SPDX-License-Identifier: Apache-2.0
+#
+"""Defines the Settings Access Permissions class."""
```
chambridge, line 5: Fix this comment please

koku/masu/external/downloader/aws/aws_report_downloader.py
chambridge: Not with this feature, but it would be nice to support parquet at some point.

koku/api/urls.py
```diff
@@ -335,6 +337,8 @@
         ),
         name="reports-openshift-azure-instance-type",
     ),
+    path("reports/minimal-reports/", MinimalReportView.as_view(), name="minimal-reports"),
```
chambridge, line 340: @adberglund thoughts on the API here?
Naming is hard I know. Do we feel that `reports/minimal-reports/` is redundant?
Do we want this to live under `reports`?
Could do something like `ingress/minimal/`. If we are happy with the above ok. I just want to make sure we've thought it through.

chambridge (commented):
couple more comments

koku/api/urls.py
lcouzens: I also hate reports reports, I went ahead and made it ingress/minimal-reports/ for now.

koku/api/common/permissions/hcs_access.py
adberglund: I do think absent a true service account, we probably want to keep this open to anyone with a login for now.

koku/api/urls.py
adberglund: I know right now we're tackling the idea of a filtered subset of data, but we could keep it broader and remove `minimal` from external reference. Something like `ingress/reports/` ?

lcouzens:
Smoke results failures are test_api_oci_ingest_single_source and test_api_aws_settings_tag_enablement, both of which @esebesto said are expected.

koku/api/ingress/reports/serializers.py
```diff
@@ -0,0 +1,59 @@
...
+            year_message = year_range[0]
+            if dh.bill_year_from_date(dh.this_month_start) not in year_range:
+                year_range.append(dh.bill_year_from_date(dh.last_month_start))
+                year_message = f"{year_range[0]} or {year_range[1]}"
+            if bill_month in month_range and bill_year in year_range:
+                interface = ProviderAccessor(source_type)
+                interface.check_file_access(data.get("source"), data.get("reports_list"))
+                return data
+            key = "bill_period"
+            message = f"Invalid bill, year and month must be {year_message} and {month_range[0]} or {month_range[1]}"
```
adberglund, line 55:
```suggestion
            message = f"Invalid bill, year must be {year_message} and month must be {month_range[0]} or {month_range[1]}"
```

koku/api/ingress/reports/view.py
```diff
@@ -0,0 +1,83 @@
...
+        try:
+            UUID(source)
+        except ValueError:
+            return Response({"Error": "Invalid source uuid."}, status=status.HTTP_400_BAD_REQUEST)
+        report_instance = self.get_object(source)
+        if not report_instance:
+            return Response({"Error": "Provider uuid not found."}, status=status.HTTP_400_BAD_REQUEST)
+
+        serializer = IngressReportsSerializer(report_instance, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
```
adberglund, line 50: I think we want to use our paginator class for this so it still has the default meta and links section.

koku/api/ingress/reports/view.py
```diff
@@ -0,0 +1,83 @@
...
+
+    permission_classes = [AllowAny]
+
+    def get(self, request, *args, **kwargs):
+        """
+        Return list of sources.
+        """
+        reports = IngressReports.objects.filter()
+        serializer = IngressReportsSerializer(reports, many=True)
+        return Response(serializer.data, status=status.HTTP_200_OK)
```
adberglund, line 66: Similarly, I think we want to use one of our paginator classes

koku/api/provider/models.py
```diff
@@ -340,6 +341,7 @@ def _get_linked_table_names(self, target_table, public_schema):
                 "type_fregex": f"_{self._normalized_type}",
                 "ocptype_fregex": f"_ocp({self._normalized_type}|all)",
                 "rpt_common_fregex": "^reporting_common_",
+                "rpt_ingress_fregex": "^reporting_ingressreports",
```
adberglund, line 344: I'm so glad we had to do this. 😑

koku/reporting/ingress/models.py
```diff
@@ -0,0 +1,52 @@
...
+            LOG.info(f"Starting data ingest task for Provider {data.get('source')}")
+            # Start check_report_updates task after Provider has been committed.
+            transaction.on_commit(
+                lambda: check_report_updates.s(
+                    provider_uuid=data.get("source"),
+                    bill_date=f"{data.get('bill_year')}{data.get('bill_month')}",
+                    ingress_reports=data.get("reports_list"),
+                    ingress_report_uuid=data.get("ingress_report_uuid"),
+                )
+                .set(queue="priority")
```
adberglund, line 50: I'm not convinced on this, but might consider keeping this as a regular download queue job. That worker has more CPU and memory than priority.

lcouzens: This was a copy/paste mistake! :smile_cat:

adberglund (dismissed):

adberglund (approved):

PR #4115, opened 2023-01-28
COST-3397: Update OCP on AWS sql

djnakabaale:
## Jira Ticket

[COST-3397](https://issues.redhat.com/browse/COST-3397)

## Description

This change will update ocp-on-aws sql to handle potential division by zero error when we have zero **node_capacity_cpu_core_hours**

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Notes

...


adberglund (commented):
Let's add the nullif to unblended_cost and markup_cost as well since they will be prone to the same divide by zero issue.

adberglund (approved):

PR #4116, opened 2023-01-30
[COST-3082] Specify AWS S3 region

samdoran:
## Jira Ticket

[COST-3082](https://issues.redhat.com/browse/COST-3082)

## Description

Allow the S3 region to be specified when adding a source.

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Notes

...


samdoran:
A couple things I would like feedback on:

1. We originally planned to add this to the `SourcesViewSet`. I implemented this in the main Koku API for simplicity. If we would rather have it in the `SourcesViewSet`, let me know and I'll do that.

1. I added an API that accepts an AWS service and returns available regions. Maybe we don't want that, or maybe that should be the main API? I also feel like this needs some sort of input validation/normalization since it's a bit fragile right now. For example, a typo in the service name or wrapping it in quotes will return an empty list of regions.

chambridge:
So I don't think we need two APIs here we only want a single API to provide s3 region support to the wizard. No need to maintain more than that. If you feel like its important to call out that this is s3 then I am fine with the `api/cost-management/v1/sources/aws-s3-regions/` path. I think it would be nice if the response followed our typical paginated response format: https://github.com/project-koku/koku/blob/main/koku/api/common/pagination.py.

I'm think I would also have not stuffed this API into the source_status.py file.

I think @maskarb found a simple way to have put this in the sources viewset.
```
    @action(methods=["get"], detail=False, permission_classes=[AllowAny], url_path="aws-regions")
    def aws_regions(self, request):
        regions = {"regions": ["us-east-1"]}
        return Response(regions)
```

Lastly the openapi spec file needs updating:
https://github.com/project-koku/koku/blob/main/docs/specs/openapi.json

samdoran:
This is great feedback. Thanks!

samdoran:
I made the changes suggested above.

Regarding updating the OpenAPI spec, is there an easy way to validate or preview that? I just modeled it off of some other endpoints but it may be incomplete.

chambridge:
> I made the changes suggested above.
> 
> Regarding updating the OpenAPI spec, is there an easy way to validate or preview that? I just modeled it off of some other endpoints but it may be incomplete.

You can use: https://editor.swagger.io/

koku/sources/api/view.py
```diff
@@ -121,6 +122,11 @@ class SourcesViewSet(*MIXIN_LIST):
     filterset_class = SourceFilter
     http_method_names = HTTP_METHOD_LIST
 
+    @action(methods=["get"], detail=False, permission_classes=[AllowAny], url_path="aws-s3-regions")
+    def aws_s3_regions(self, request):
+        regions = get_available_regions("s3")
+        return Response({"regions": regions})
```
maskarb, line 128:
```suggestion
        return ListPaginator([{"regions": regions}], request).paginated_response
```
For completeness sake, we'll want to use our Paginators here.

Which comes from:
```
from api.common.pagination import ListPaginator
```

The other methods on this endpoint do not use this Paginator, but they _should_. This was a miss back in the day which we never corrected.

maskarb (approved):
lgtm!

lcouzens:
Only failure is test_api_ocp_on_gcp_source_raw_calc which is expected.

PR #4119, opened 2023-02-01
Add tests cases for summarize_reports

samdoran:
## Jira Ticket

## Description

- Test beginning and middle of the month
- Use fixed date values to prevent spontaneous breakage

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Notes

...

### DIFF
--- a/koku/masu/test/processor/test_tasks.py
+++ b/koku/masu/test/processor/test_tasks.py
@@ -10,6 +10,7 @@ import shutil
 import tempfile
 import time
 from datetime import date
+from datetime import datetime
 from datetime import timedelta
 from decimal import Decimal
 from unittest import skip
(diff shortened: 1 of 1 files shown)


koku/masu/test/processor/test_tasks.py
```diff
@@ -406,6 +407,40 @@ def test_summarize_reports_processing_list_with_none(self, mock_update_summary):
...
+        """Test that the summarize_reports task is called once when a processing list
+        of reports with no start or end date provided."""
+        mock_update_summary.s = Mock()
+
+        reports_to_summarize = [{"manifest_id": id} for id in range(1, 5)]
+
+        current_date = datetime.now()
+        with patch(
+            "masu.external.date_accessor.DateAccessor.today",
+            return_value=datetime(current_date.year, current_date.month, 28),
```
samdoran, line 421: I use the current year and month of the test run here. I think these values should also be static, but wasn't sure if wanted to keep them dynamic.

PR #4120, opened 2023-02-01
[COST-3465] - sources: check if header item is None before decoding

maskarb:
## Jira Ticket

[COST-3465](https://issues.redhat.com/browse/COST-3465)

## Description

This change will prevent `'NoneType' object has no attribute 'decode'` when decoding Kafka headers when the value in the key:value pairs is `None`

### DIFF
--- a/koku/kafka_utils/test/test_utils.py
+++ b/koku/kafka_utils/test/test_utils.py
@@ -35,3 +35,37 @@ class KafkaUtilsTest(TestCase):
         utils.is_kafka_connected(TEST_HOST, TEST_PORT)
         connection_errors_after = WORKER_REGISTRY.get_sample_value("kafka_connection_errors_total")
         self.assertEqual(connection_errors_after - connection_errors_before, 1)
+
+    def test_extract_from_header(self):
+        """Test test_extract_from_header."""
+        encoded_value = bytes("value", "utf-8")
+        test_table = [
+            {
+                "test": [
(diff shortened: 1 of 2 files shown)


samdoran (approved):

koku/kafka_utils/test/test_utils.py
```diff
@@ -35,3 +35,37 @@ def test_kafka_connection_metrics_listen_for_messages(self, mock_start, mock_sle
         utils.is_kafka_connected(TEST_HOST, TEST_PORT)
         connection_errors_after = WORKER_REGISTRY.get_sample_value("kafka_connection_errors_total")
         self.assertEqual(connection_errors_after - connection_errors_before, 1)
+
+    def test_extract_from_header(self):
+        """Test test_extract_from_header."""
+        encoded_value = bytes("value", "utf-8")
```
samdoran, line 41: Another convention for indicating vars that hold bytes is a `b_` prefix.
```suggestion
        b_value = bytes("value", "utf-8")
```

samdoran: Bytes literals are UTF-8 if no encoding declaration exists in the file, so this could also be expressed as `b"value"`.

maskarb: I woulda committed this, but I didn't wanna wait on CI again.

PR #4122, opened 2023-02-03
COST-3050: Update deduplicate reports function for OCI

djnakabaale:
## Jira Ticket

[COST-3050](https://issues.redhat.com/browse/COST-3050)

## Description

This change will update the deduplicate function to merge reports with duplicate manifest ids but different (start, end) date pairs

## Testing

1. Checkout Branch
2. Restart Koku
3. Load some oci data with `make create-test-customer` and `make load-test-customer-data test_source=oci`
4. Hit the sources endpoint to get the oci source id, http://127.0.0.1:8000/api/cost-management/v1/sources/
5. Hit endpoint  `http://127.0.0.1:8000/api/cost-management/v1/sources/<id>/stats/`
    1. You should see unique timestamps in `last_manifest_complete_date` field for each manifest. 

## Notes

...
To test with a real/remote source, you need OCI credentials in your .env file and run 

```
 make oci-source oci_name=<source_name> bucket=<bucket_name> namespace=<namespace> region=<region>

```

and follow steps **4** and **5**


koku/masu/database/report_manifest_db_accessor.py
```diff
@@ -65,9 +65,8 @@ def mark_manifests_as_completed(self, manifest_list):
         if manifest_list:
             bulk_manifest_query = self._get_db_obj_query().filter(id__in=manifest_list)
             for manifest in bulk_manifest_query:
-                if not manifest.manifest_completed_datetime:
-                    manifest.manifest_completed_datetime = completed_datetime
-                    manifest.save()
+                manifest.manifest_completed_datetime = completed_datetime
```
adberglund, line 68: Why remove the condition here?

djnakabaale: I had added that if statement in this [previous pr](https://github.com/project-koku/koku/pull/4044)
But with it, the `manifest_completed_datetime` is not updated when there’s still more processing done for the same manifest id. I noticed this on manifests with (start, end) date pairs with ranges wider than the `TRINO_DATE_STEP` set in the env. 

I also now think it would prevent update on a re-summary.

lcouzens:
Smoke failure test_api_aws_settings_tag_enablement, shouldnt be related to these changes.

koku/masu/database/report_manifest_db_accessor.py
lcouzens: I had the same question as Andrew but I think this makes sense.

pedrolp85:
Could we add oci-smokes and retest before merging?

pedrolp85:
Seems data is ingested correctly, https://ci.ext.devshift.net/job/project-koku-koku-pr-check/8007/, awesome!

PR #4126, opened 2023-02-05
[COST-3315] partition gcp openshift daily

cgoodfred:
## Jira Ticket

[COST-3315](https://issues.redhat.com/browse/COST-3315)

## Description

This change will adjust our OCP on GCP data to be partitioned daily as well. SQL updates to match those changes. It also adjusts get_openshift_topology_for_multiple_providers to return a list of topologies that each pertain to a specific clusters information and updates to the match OpenShift resources functions to match those changes.

## Testing

1. Checkout Branch
2. Restart Koku
3. Load test customer data: `make create_test_customer` followed by `make load_test_customer_data`
4. Verify that the OCP on GCP data has loaded: http://localhost:8000/api/cost-management/v1/reports/openshift/infrastructures/gcp/costs/
5. Log into trino using the trino CLI: `trino --server localhost:8080 --catalog hive --schema org1234567 --user admin --debug`
6. Verify that the gcp_openshift_daily is now partitioned by day as well: `show create table gcp_openshift_daily;' You should see `partitioned_by = ARRAY['source','year','month','day'] ` at the end
7. smoke tests should cover calculation checks.

## Testing the transition from main to this branch:

1. checkout main
2. restart koku
3. in a local IQE setup, put a pdb at the top of the `test_api_ocp_on_gcp_source_raw_calc` test
4. run the raw calc test until it stops at the pdb
5. checkout this branch
6. run the migrate trino script: `python scripts/cji_scripts/migrate_trino.py`
7. hit the process ocp on cloud endpoint with the relevant provider, example call: 
http://localhost:5042/api/cost-management/v1/report/process/openshift_on_cloud/?provider_uuid=a3fbdeb3-d78c-4e4e-b6c4-c534ff0e52fb&schema=org1234567&start_date=2023-01-01

8. resummarize that provider for the last two months, example call:
http://localhost:5042/api/cost-management/v1/report_data/?schema=org1234567&provider_type=GCP-local&provider_uuid=ea684142-f2a2-48eb-b635-425060553fb0&start_date=2023-01-01

http://localhost:5042/api/cost-management/v1/report_data/?schema=org1234567&provider_type=GCP-local&provider_uuid=ea684142-f2a2-48eb-b635-425060553fb0&start_date=2023-02-01

9. continue with the smoke test and it should pass
## Notes

...


koku/masu/processor/parquet/ocp_cloud_parquet_report_processor.py
```diff
@@ -118,15 +127,16 @@ def _determin_s3_path(self, file_type):
     def create_ocp_on_cloud_parquet(self, data_frame, parquet_base_filename, file_number):
         """Create a parquet file for daily aggregated data."""
         # Add the OCP UUID in case multiple clusters are running on this cloud source.
```
maskarb, line 129: What does this comment mean?

koku/masu/processor/parquet/ocp_cloud_parquet_report_processor.py
```diff
@@ -118,15 +127,16 @@ def _determin_s3_path(self, file_type):
...
-                )
+                parquet_base_filename = f"{data_frame['invoice_month'].values[0]}_{uuid4()}"
         file_name = f"{parquet_base_filename}_{file_number}_{PARQUET_EXT}"
         file_path = f"{self.local_path}/{file_name}"
         self._write_parquet_to_file(file_path, file_name, data_frame, file_type=self.report_type)
-        self.create_parquet_table(file_path, daily=True)
+        if self.provider_type in {Provider.PROVIDER_GCP, Provider.PROVIDER_GCP_LOCAL}:
+            self.create_parquet_table(file_path, daily=True, partition_map=GCP_PARTITION_MAP)
+        else:
+            self.create_parquet_table(file_path, daily=True)
```
maskarb, line 139: we could make a `partition_map` property:
```
    @property
    def partition_map(self):
        """Post processor based on provider type."""
        if self.provider_type in {Provider.PROVIDER_GCP, Provider.PROVIDER_GCP_LOCAL}:
            return GCP_PARTITION_MAP
        return None
```

Then this could become:
```suggestion
        self.create_parquet_table(file_path, daily=True, partition_map=self.partition_map)
```

It would just be nice to get away from the `if GCP` specific logic when we are not in a GCP specific class.

koku/masu/processor/parquet/ocp_cloud_parquet_report_processor.py
```diff
@@ -29,6 +31,13 @@
 
 LOG = logging.getLogger(__name__)
 
+GCP_PARTITION_MAP = {
+    "source": "varchar",
+    "year": "varchar",
+    "month": "varchar",
+    "day": "varchar",
+}
```
samdoran, line 39: Global static values like this are generally better as an Enum. The interface is a bit different than a mapping, but Enums are immutable, so you don't have to worry about something changing it.
```suggestion
class GCPPartions(Enum):
    source = "varchar"
    year = "varchar"
    month = "varchar"
    day = "varchar"
```

koku/masu/processor/report_parquet_processor_base.py
```diff
@@ -124,18 +124,29 @@ def _generate_create_table_sql(self):
             sql += f"{norm_col} {col_type}"
             if idx < (len(parquet_columns) - 1):
                 sql += ","
-        sql += ",source varchar, year varchar, month varchar"
+        if partition_map:
+            # Add the specified partition columns
+            sql += ", "
+            sql += ",".join([f"{item[0]} {item[1]} " for item in list(partition_map.items())])
```
samdoran, line 130: If `partion_map` was an Enum, it would look something like this:
```suggestion
            sql += ",".join([f"{item.name} {item.value} " for item in list(partition_map])
```

koku/masu/processor/report_parquet_processor_base.py
```diff
@@ -124,18 +124,29 @@ def _generate_create_table_sql(self):
             sql += f"{norm_col} {col_type}"
             if idx < (len(parquet_columns) - 1):
                 sql += ","
-        sql += ",source varchar, year varchar, month varchar"
+        if partition_map:
+            # Add the specified partition columns
+            sql += ", "
```
samdoran, line 129: I know this is continuing an existing pattern so I wouldn't change it, but string concatenation using `+` and `+=` can be quite expensive since each intermediate string needs to be built before assignment.

Collecting these values in a mapping or iterable and doing a single `f` string or `.join()` at the end is more efficient and doesn't sacrifice legibility.

koku/masu/processor/report_parquet_processor_base.py
```diff
@@ -124,18 +124,29 @@ def _generate_create_table_sql(self):
...
+            )
         LOG.info(f"Create Parquet Table SQL: {sql}")
         return sql
 
-    def create_table(self):
+    def create_table(self, partition_map=None):
         """Create Trino SQL table."""
-        sql = self._generate_create_table_sql()
+        sql = self._generate_create_table_sql(partition_map=partition_map)
         self._execute_sql(sql, self._schema_name)
```
samdoran, line 150: Just curious if we are worried about SQL injection attacks here. It looks like most of the values are not coming from direct user input, but I'm not 100% sure.

koku/masu/util/aws/common.py
```diff
@@ -499,9 +499,12 @@ def aws_generate_daily_data(data_frame):
     return daily_data_frame
 
 
-def match_openshift_resources_and_labels(data_frame, cluster_topology, matched_tags):
+def match_openshift_resources_and_labels(data_frame, cluster_topologies, matched_tags):
     """Filter a dataframe to the subset that matches an OpenShift source."""
-    resource_ids = tuple(cluster_topology.get("resource_ids", []))
+    resource_ids = []
+    for cluster_topology in cluster_topologies:
+        resource_ids.extend(cluster_topology.get("resource_ids", []))
```
samdoran, line 506: This is another more efficient technique for flattening a list of lists using [itertools.chain](https://docs.python.org/3/library/itertools.html#itertools.chain).
```suggestion
    resource_ids = itertools.chain.from_iterable(cluster_topology.get("resource_ids", []) for cluster_topology in cluster_topologies)
```

samdoran: I tried doing this with a comprehension using unpacking, but you can't do that. TIL. 🤓 
```
tuple(*cluster_topology.get("resource_ids", []) for cluster_topology in cluster_topologies)
  Cell In [74], line 1
    tuple(*cluster_topology.get("resource_ids", []) for cluster_topology in cluster_topologies)
          ^
SyntaxError: iterable unpacking cannot be used in comprehension
```

koku/masu/util/azure/common.py
```diff
@@ -186,10 +186,13 @@ def azure_generate_daily_data(data_frame):
     return data_frame
 
 
-def match_openshift_resources_and_labels(data_frame, cluster_topology, matched_tags):
+def match_openshift_resources_and_labels(data_frame, cluster_topologies, matched_tags):
     """Filter a dataframe to the subset that matches an OpenShift source."""
-    nodes = cluster_topology.get("nodes", [])
-    volumes = cluster_topology.get("persistent_volumes", [])
+    nodes = []
+    volumes = []
```
samdoran, line 192: The same technique could be done here.

koku/masu/processor/parquet/ocp_cloud_parquet_report_processor.py
cgoodfred: This was a good move, implemented the suggestion.

koku/masu/processor/parquet/ocp_cloud_parquet_report_processor.py
cgoodfred: I don't think an Enum would be appropriate here, an enum having the same value is treated as an alias to the original definition so the items here don't behave as they should. Example from shell:
```
>>> class GCPPartitions(Enum):
...     source = "varchar"
...     year = "varchar"
...     month = "varchar"
...     day = "varchar"
... 
>>> ",".join([f"{item.name} {item.value} " for item in list(GCPPartitions)])
'source varchar '
>>> GCPPartitions.month
<GCPPartitions.source: 'varchar'>
```

koku/masu/processor/report_parquet_processor_base.py
cgoodfred: The comment above is related to this one about Enum's likely not being the best choice for this scenario.

koku/masu/util/aws/common.py
cgoodfred: implemented the change to using chain from itertools, good suggestion.

koku/masu/util/azure/common.py
cgoodfred: implemented the change to using chain from itertools, good suggestion.

koku/masu/processor/report_parquet_processor_base.py
cgoodfred: I don't believe we have to be concerned with SQL injection here, the partition map is statically defined by us and I believe the rest of the _generate_create_table_sql is also values we have defined elsewhere that we base on an expected list such as column names. I don't believe we use any user defined input here.

koku/masu/util/aws/common.py
adberglund: It's neat, but I find that to be more taxing to reason through. More mental hoops. Just my two pennies.

koku/masu/processor/parquet/ocp_cloud_parquet_report_processor.py
samdoran: Ah, that's true. I was thinking "varchar" was some sort of placeholder for the real value that would be changed in a later commit. I don't know what the actual data looks like.

cgoodfred: we are using this to build the SQL statement that is creating the table in case it doesn't exist for a new customer or we drop it and have to recreate it. The actual data processing and insertion is separate from what this is being used for.

koku/masu/util/azure/common.py
```diff
@@ -186,10 +187,17 @@ def azure_generate_daily_data(data_frame):
...
+    nodes = chain.from_iterable(cluster_topology.get("nodes", []) for cluster_topology in cluster_topologies)
+    volumes = chain.from_iterable(
+        cluster_topology.get("persistent_volumes", []) for cluster_topology in cluster_topologies
+    )
+    nodes = []
+    volumes = []
+    for cluster_topology in cluster_topologies:
+        nodes.extend(cluster_topology.get("nodes", []))
+        volumes.extend(cluster_topology.get("persistent_volumes", []))
     matchable_resources = nodes + volumes
```
samdoran, line 201: No need to do this twice.
```suggestion
    matchable_resources = list(nodes) + list(volumes)
```

koku/masu/util/aws/common.py
samdoran: I agree it's a bit harder to read, but it's idiomatic Python (standard lib and a comprehension). Splitting over two lines may increase legibility.

```python
resource_ids = (cluster_topology.get("resource_ids", []) for cluster_topology in cluster_topologies)
resource_ids = itertools.chain.from_iterable(resource_ids)
```

I don't think this crosses over into the "clever" category, which is definitely something to be avoided.

koku/masu/util/azure/common.py
cgoodfred: definitely an oversight on my part, thanks for catching that!

adberglund (approved):
🪂

esebesto:
Smoke tests results collected directly from iqe pod:
`FAILED ../../iqe_venv/lib/python3.8/site-packages/iqe_cost_management/tests/rest_api/v1/test__i_source.py::test_api_ocp_on_azure_source_raw_calc
FAILED ../../iqe_venv/lib/python3.8/site-packages/iqe_cost_management/tests/rest_api/v1/test_user.py::test_api_aws_settings_tag_enablement
= 2 failed, 648 passed, 1 skipped, 9789 deselected, 3135 warnings in 3184.49s (0:53:04) =`

Both tests are known to fail sometimes and thus are unrelated to this pr

esebesto (approved):

koku/masu/processor/report_parquet_processor_base.py
cgoodfred: This is a good note for a potential refactor of some of this, I think for now I will leave it as is per the comment.

PR #4127, opened 2023-02-06
Update alert thresholds

samdoran:
## Description

Adjust alert thresholds in panels to match the values in alerts.

Update rows to be open by default.


maskarb (commented):
While here, can you adjust `"collapsed": true,` to `false`? It would be nice if all the panels were just open when going to Grafana.

samdoran:
Oooo, I'll definitely do that.

maskarb (approved):

PR #4139, opened 2023-02-13
Use hive specific env var

adberglund:
## Jira Ticket

## Description

This change will default to sending Hive metastore data to the hive DB when you have env vars set for your koku DB. Currently hive metastore data is going to the koku DB's public schema when env vars are set for the koku DB.

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Notes

...


adberglund:
More complicated than initially thought, and was interested in an easy win

PR #4142, opened 2023-02-14
COST-2489: Allow for easy retry logic with additional context fields.

myersCody:
## Jira Ticket

[COST-2489](https://issues.redhat.com/browse/COST-2489)

## Description

This change will ...

## Testing

Testing:
Go to `aws_org_unit_crawler.py` and replace the `_init_session` function with:
```def _init_session(self):
        """
        Set or get a session client for aws organizations

        Args:
            session = aws session
        """
        context_key = "crawl_hierarchy"
        with ProviderDBAccessor(self.account.get("provider_uuid")) as provider_accessor:
            context = provider_accessor.get_additional_context()
            self.crawlable = context.get(context_key, True)
        try:
            awsarn = utils.AwsArn(self._auth_cred)
            session = utils.get_assume_role_session(awsarn)
            session_client = session.client("organizations")
            self.account_id = awsarn.account_id
            LOG.info(
                "Starting aws organizations session for crawler for account with"
                " provider_uuid: {} and account_id: {}.".format(self.account.get("provider_uuid"), self.account_id)
            )
            self._client = session_client
            self._check_if_crawlable()
        except (ClientError, BotoConnectionError):
            self.crawlable = False
            LOG.info(
                "Incomplete aws org_unit permissions for provider_uuid: {} and account_id: {}".format(
                    self.account.get("provider_uuid"), self.account_id
                )
            )
            with ProviderDBAccessor(self.account.get("provider_uuid")) as provider_accessor:
                context[context_key] = False
                provider_accessor.set_additional_context(context)
 ```
Then go to `tasks.py` and look for the `crawl_account_hierarchy` function replace this line:
```
if account.get("provider_type") == Provider.PROVIDER_AWS:
```
with:
```
if account.get("provider_type") in [Provider.PROVIDER_AWS, Provider.PROVIDER_AWS_LOCAL]:
```
These changes will allow you to trigger the AWS org unit crawler on local aws source and have it populate the additional_context to disable the crawler.
2. Now go to: http://localhost:8000/api/cost-management/v1/sources/
3. Grab the uuid for `aws_org_tree`
4. http://127.0.0.1:5042/api/cost-management/v1/crawl_account_hierarchy/?provider_uuid=df8ed78f-cb2b-4b0a-b4f4-b2dd9a8257fe
```koku-koku-worker-1  | [2023-02-14 15:08:22,584] INFO 82d0f60f-1baf-45ee-81af-2d83632ea69c Account hierarchy crawler found 1 accounts to scan
koku-koku-worker-1  | [2023-02-14 15:08:22,586] INFO 82d0f60f-1baf-45ee-81af-2d83632ea69c Starting account hierarchy crawler for type AWS-local with provider_uuid: df8ed78f-cb2b-4b0a-b4f4-b2dd9a8257fe
koku-koku-worker-1  | [2023-02-14 15:08:22,597: INFO/ForkPoolWorker-1] Found credentials in environment variables.
koku-koku-worker-1  | [2023-02-14 15:08:22,793] INFO 82d0f60f-1baf-45ee-81af-2d83632ea69c Incomplete aws org_unit permissions for provider_uuid: df8ed78f-cb2b-4b0a-b4f4-b2dd9a8257fe and account_id: None
```
1. `make clear-cache`
2. http://localhost:8000/api/cost-management/v1/sources/
```
"additional_context": {
                "crawl_hierarchy": false
            }
 ```
You will see the `additional_context` field has been populated now.

Now you can look at & manipulate the additional context field by going to:
1. http://127.0.0.1:5042/api/cost-management/v1/additional_context/?schema=org1234567&provider_uuid=df8ed78f-cb2b-4b0a-b4f4-b2dd9a8257fe

The `GET` will return the current additional context.

Try posting the url with the following examples to see the expected behavior:
The `POST` checks to see if it is a valid key, and will only update the key in the post. If you want to remove a key use the `remove_key` as your key then value is the key you want to remove.
```
{"crawl_hierarchy": "bad"} // "Error": "Invalid value supplied: key: crawl_hierarchy, value: bad."
{"crwal_heirrchy": false} // "Error": "Invalid key supplied: crwal_heirrchy"
{"crawl_hierarchy": true} // {"crawl_hierarchy": true}
{"aws_list_account_aliases": true} // {"crawl_hierarchy": true, "aws_list_account_aliases": true}
{"remove_key": "aws_list_account_aliases"} // {"crawl_hierarchy": true}
```

## Notes

...

### CODE REVIEW

File: koku/masu/api/additional_context.py [hunk 1]
```diff
@@ -0,0 +1,53 @@
...
+                        context[key] = value
+                    else:
+                        errmsg = f"Invalid value supplied: key: {key}, value: {value}."
+                elif key == "remove_key":
+                    if value in context.keys():
+                        del context[value]
+                else:
+                    errmsg = f"Invalid key supplied: {key}"
+                if errmsg:
+                    return Response({"Error": errmsg}, status=status.HTTP_400_BAD_REQUEST)
```
maskarb, line 51: This logic is a bit confusing to reason about. I think it is easier validate the keys/values first and immediately return the error:
```suggestion
                if key not in ["aws_list_account_aliases", "crawl_hierarchy", "remove_key"]:
                    return Response({"Error": f"Invalid key supplied: {key}"}, status=status.HTTP_400_BAD_REQUEST)
                if key == "remove_key" and value in context.keys():
                        del context[value]
                elif key in ["aws_list_account_aliases", "crawl_hierarchy"]:
                    if not isinstance(value, bool):
                        return Response({"Error": f"Invalid value supplied: key: {key}, value: {value}."}, status=status.HTTP_400_BAD_REQUEST)
                    context[key] = value
```

I'm also curious, why was `ADDITIONAL_CONTEXT_KEYS` made an attribute of the Provider class? Is that attribute used anywhere else? It feels like a weird place to put that when it could just be a constant in this file instead.

myersCody: I threw it in there since it related to a field in the Provider's model. I am okay with moving to to a constant in this file instead though.

maskarb (approved):
So, I am not advocating that we update how this endpoint will work, but it does feel like the way this works is very randomly specific. The goal here is to update a json field in the db, but the way this is written, you have to know exactly how this implementation works. Like to remove a key, you send a `"remove-key": "key-to-remove"` POST, but you can also send a `"key-to-update": true` which carries a completely different meaning than key removal.

There actually is a standard to updating JSON: https://www.rfc-editor.org/rfc/rfc6902.
And here is a website that describes things in a more straightforward way with examples: https://jsonpatch.com/.
Essentially JSON patch involves sending a JSON body with a specific format that tells you what operation you want to do. For example, if we wanted to update `crawl_hierarchy` to True, the POST body would be:
```
[
    { "op": "replace", "path": "/crawl_hierarchy", "value": true }
]
```
If you wanted to remove a field, you would POST:
```
[
    { "op": "remove", "path": "/crawl_hierarchy" }
]
```
And you can combine things:
```
[
    { "op": "replace", "path": "/crawl_hierarchy", "value": true },
    { "op": "remove", "path": "/aws_list_account_aliases" }
]
```

This makes things a bit more intuitive: we're providing the `op` you want to do, you're giving the field you wanna update, and the value you wanna change it to.

Again, I am not advocating that we change this PR to use this standard, but this standard DOES exist and it seems pretty applicable to this endpoint. I don't have an issue with this PR as it is now, especially considering that it is an internal endpoint. I'm just throwing the idea out there.

myersCody:
I actually really like the standard you suggested here @maskarb I think it is cleaner and easier to understand.

koku/masu/api/additional_context.py
```diff
@@ -0,0 +1,73 @@
...
+    """Checks that the opt dict is structure correctly."""
+    if not isinstance(op_dict, dict):
+        return "Post body must be a list of dictionaries."
+    for rk in ["op", "key"]:
+        if not op_dict.get(rk):
+            return f"Missing key in body ({rk})."
+    if op_dict.get("key") not in ADDITIONAL_CONTEXT_KEYS:
+        return f"Invalid key supplied: {op_dict.get('key')}"
+    if op_dict.get("op") not in ["remove", "replace"]:
+        return f"Unrecognized op: {op_dict.get('op')}"
```
maskarb, line 30: Might be better to say "Only the `remove` and `replace` operations are supported"

maskarb (approved):
lgtm!

PR #3712, opened 2022-06-15
Cost2270 RBAC changes don't invalidate the cache

Fozzywozzybear:
Fixes [COST-2270](https://issues.redhat.com/browse/COST-2270)

Changes proposed in this PR:
Enables us to change the rate at which the RBAC cache is invalidated.
The Default rate is at 300s or 5 min. 

Testing Instructions:
1. Set your RBAC_CACHE_TIMEOUT to a short time, in this example 5 
2. Enter the Django Shell : REDIS_HOST=localhost make shell
3.Run These set of commands 
```
- from django.core.cache import caches
- cache = caches['rbac']
- cache.set('my_key', 'hello, world!')
- cache.get('my_key') : expected result 'hello, world!'
```
after your set time has passed, rerun the command 
`- cache.get('my_key') : expected result should be no output`
4. Make sure your settings.CACHE has been updated 
```
- from django.conf import settings
- settings.CACHES
```
Example output 

> `{'default': {'BACKEND': 'django.core.cache.backends.dummy.DummyCache'}, 'rbac': {'BACKEND': 'django_redis.cache.RedisCache', 'LOCATION': 'redis://localhost:6379/1', 'OPTIONS': {'CLIENT_CLASS': 'django_redis.client.DefaultClient', 'IGNORE_EXCEPTIONS': True, 'MAX_ENTRIES': 1000}}, 'worker': {'BACKEND': 'django.core.cache.backends.db.DatabaseCache', 'LOCATION': 'worker_cache_table', 'TIMEOUT': 86400}}`

---
Additional Context: Please reach out to me with questions or assistance testing the PR!

PR #4159, opened 2023-02-22
[COST-2986] Add `CACHE_TIMEOUT` to deploy config

samdoran:
## Jira Ticket

[COST-2986](https://issues.redhat.com/browse/COST-2986)

## Description

This allows us to control the sources API timeout in our deployments, helping out with testing.

Related to #3712.

### DIFF
--- a/deploy/clowdapp.yaml
+++ b/deploy/clowdapp.yaml
@@ -135,6 +135,8 @@ objects:
           value: ${ENHANCED_ORG_ADMIN}
         - name: RBAC_CACHE_TIMEOUT
           value: ${RBAC_CACHE_TIMEOUT}
+        - name: CACHE_TIMEOUT
+          value: ${CACHE_TIMEOUT}
         image: ${IMAGE}:${IMAGE_TAG}
         initContainers:
         - command:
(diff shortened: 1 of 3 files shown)


deploy/clowdapp.yaml
```diff
@@ -2904,6 +2906,9 @@ parameters:
 - displayName: RBAC_CACHE_TIMEOUT
   name: RBAC_CACHE_TIMEOUT
   value: "300"
+- displayName: Middleware Timeout
+  name: CACHE_TIMEOUT
+  value: "300"
```
adberglund, line 2911: Let's do 3600 to match the settings file to make this a no-change change in cases where we don't explicitly set this var in app interface.

deploy/kustomize/base/base.yaml
```diff
@@ -424,3 +424,6 @@ parameters:
 - name: RBAC_CACHE_TIMEOUT
   displayName: RBAC_CACHE_TIMEOUT
   value: "300"
+- name: CACHE_TIMEOUT
+  displayName: Middleware Timeout
+  value: "300"
```
adberglund, line 429: Samesies

adberglund (changes requested):

deploy/clowdapp.yaml
samdoran: Doh, I meant to do that. 🤦

adberglund (approved):

PR #4167, opened 2023-02-28
COST-3540: Support new distribution options in cost model

djnakabaale:
## Jira Ticket

[COST-3540](https://issues.redhat.com/browse/COST-3540)

## Description

This change follows this [pr](https://github.com/project-koku/koku/pull/4178) to validate the `distribution_info` json data or create a default if does not exist.

## Testing

- make sure to have changes in [this pr](https://github.com/project-koku/koku/pull/4178) and run migrations

1. Checkout this branch
2. Restart Koku
3. load data with `make create-test-customer` and `make load-test-customer-data`
4. Once the data has loaded, hit this endpoint, http://127.0.0.1:8000/api/cost-management/v1/cost-models/

    <details>
      <summary>
        You should see the `distribution_info` object for all the cost models returned, similar to this:
      </summary>
      
      ```
        {
            "uuid": "b24f78ce-82f6-444a-8acd-9fb24921dabb",
            "name": "Cost Management AWS Cost Model",
            "description": "A cost model for markup on AWS costs.",
            "source_type": "Amazon Web Services",
            "created_timestamp": "2023-03-06T21:56:08.048869Z",
            "updated_timestamp": "2023-03-06T21:56:08.048889Z",
            "rates": [],
            "markup": {
                "value": "10.0000000000",
                "unit": "percent"
            },
            "distribution": "cpu",
            "distribution_info": {
                "distribution_type": "cpu",
                "platform_cost": true,
                "worker_cost": true
            },
            "currency": "USD",
            "source_uuids": [],
            "sources": [
                {
                    "uuid": "9b2cf13e-fe21-4034-9add-5625e45d6cbc",
                    "name": "Test AWS Source"
                }
            ]
        },
        {
            "uuid": "e0632b16-e0cc-45ba-9e51-f6e7cd0af67f",
            "name": "Cost Management Azure Cost Model",
            "description": "A cost model for markup on Azure costs.",
            "source_type": "Microsoft Azure",
            "created_timestamp": "2023-03-06T21:56:18.931336Z",
            "updated_timestamp": "2023-03-06T21:56:18.931410Z",
            "rates": [],
            "markup": {
                "value": "20.0000000000",
                "unit": "percent"
            },
            "distribution": "cpu",
            "distribution_info": {
                "distribution_type": "cpu",
                "platform_cost": true,
                "worker_cost": true
            },
            "currency": "USD",
            "source_uuids": [],
            "sources": [
                {
                    "uuid": "46287741-3a11-46a5-815b-ddecac816e29",
                    "name": "Test Azure Source"
                }
            ]
        },
        {
            "uuid": "9633b93e-3b2b-441d-834c-c00c7540914c",
            "name": "Cost Management GCP Cost Model",
            "description": "A cost model for markup on GCP costs.",
            "source_type": "Google Cloud Platform",
            "created_timestamp": "2023-03-06T21:56:30.410878Z",
            "updated_timestamp": "2023-03-06T21:56:30.410904Z",
            "rates": [],
            "markup": {
                "value": "30.0000000000",
                "unit": "percent"
            },
            "distribution": "cpu",
            "distribution_info": {
                "distribution_type": "cpu",
                "platform_cost": true,
                "worker_cost": true
            },
            "currency": "USD",
            "source_uuids": [],
            "sources": [
                {
                    "uuid": "5063aab6-50b8-4c08-aa3c-8f80b55f8e8e",
                    "name": "Test GCP Source"
                }
            ]
        }
    ``` 
  </details>

## Notes

...


maskarb (changes requested):
This PR needs to be split in 2: 
1. the model additions and the migrations
2. the remaining code changes

koku/cost_models/serializers.py
```diff
@@ -54,6 +55,44 @@ class MarkupSerializer(serializers.Serializer):
...
+        else:
+            _distribute_type = data.get("distribute_type", metric_constants.CPU_DISTRIBUTION)
+            valid_distribute_type = self.validate_distribute_type(_distribute_type)
+            if data == {}:
+                data["distribute_type"] = valid_distribute_type
+                data["platform_cost"] = True
+                data["worker_cost"] = True
+            else:
+                data["distribute_type"] = valid_distribute_type
+        return data
```
maskarb, line 93: Almost none of this is required. These serializers have validation built in. So for the `distribute_type`, when the serializer receives the data, it automatically validates the fields and they can only be one of the `metric_constants.DISTRIBUTION_CHOICES`. So that method can be removed entirely. As for the validation method, the only thing we need to do is ensure we have all 3 fields if the user passes in the json. This validate method can be replaced with:

```
DISTRIBUTION_OPTIONS = {"distribute_type", "worker_cost", "platform_cost"} <<<< this is a set so we can do set math

def validate(self, data):
    diff = self.DISTRIBUTION_OPTIONS.difference(data) # this gets the difference between the options and the data
    if diff == self.DISTRIBUTION_OPTIONS:  # here, if the diff is the same as the options, it means we've received an empty dict, so we just need to populate the default data and return it. It might be better to create the default as a constant somewhere and return the constant instead.
        return {
            "distribute_type": metric_constants.CPU_DISTRIBUTION,
            "platform_cost": True
            "worker_cost": True
        }
    if diff:  # if there is a diff that does _not_ match the options, that means things are missing, which is a validation error.
        distribution_info_str = ", ".join(diff)  # `join` takes an iterable and all the entries of self.DISTRIBUTION_OPTIONS are defined as strings, so there is no need to use `str()` here, or iterate over the entries
            error_msg = f"Missing distribution information: {distribution_info_str}"  # instead of giving a general, "hey one of these is missing", this error will tell the user which field is actually missing
            raise serializers.ValidationError(error_msg)
    return data  # if we made it this far, we just need to return the data we have because the individual validators have already validated the individual entries of the json
```

koku/cost_models/serializers.py
```diff
@@ -54,6 +55,44 @@ class MarkupSerializer(serializers.Serializer):
...
+        distrib_choice_list = [choice[0] for choice in metric_constants.DISTRIBUTION_CHOICES]
+        if distribute_type not in distrib_choice_list:
+            error_msg = f"{distribute_type} is an invaild distribution type"
+            raise serializers.ValidationError(error_msg)
+        return distribute_type
+
+    def validate(self, data):
+        """Run validation for distribution options."""
+
+        if not any(ele in data.keys() for ele in self.DISTRIBUTION_OPTIONS):
```
samdoran, line 78: The membership test on a dict operates on the keys.
```suggestion
        if not any(ele in data for ele in self.DISTRIBUTION_OPTIONS):
```

koku/cost_models/serializers.py
```diff
@@ -395,6 +434,8 @@ class Meta:
         choices=metric_constants.DISTRIBUTION_CHOICES, required=False, allow_blank=True
     )
 
+    distribution_info = DistributionSerializer(required=False)
```
maskarb, line 437: why do we have:
```
distribution = serializers.ChoiceField(        
    choices=metric_constants.DISTRIBUTION_CHOICES, required=False, allow_blank=True	        )
```
and 
```
distribute_type = serializers.ChoiceField(choices=metric_constants.DISTRIBUTION_CHOICES, required=False)
```
Aren't these the same thing?

koku/cost_models/serializers.py
```diff
@@ -54,6 +55,44 @@ class MarkupSerializer(serializers.Serializer):
...
+            error_msg = f"{distribute_type} is an invaild distribution type"
+            raise serializers.ValidationError(error_msg)
+        return distribute_type
+
+    def validate(self, data):
+        """Run validation for distribution options."""
+
+        if not any(ele in data.keys() for ele in self.DISTRIBUTION_OPTIONS):
+            data = data.get("distribution_info", {})
+        if data.keys() and not (sorted(self.DISTRIBUTION_OPTIONS) == sorted(data.keys())):
```
samdoran, line 80: Rather than building two intermediate lists and calling `sorted` (which can be expensive on large data sets), use `all` and the dict view iterator.
```suggestion
        if data and not all(opt in data for opt in self.DISTRIBUTION_OPTIONS):
```

koku/cost_models/serializers.py
```diff
@@ -451,6 +492,9 @@ def validate(self, data):
         if not data.get("currency"):
             data["currency"] = get_currency(self.context.get("request"))
 
+        if not data.get("distribution_info"):
+            data["distribution_info"] = self.validate_distribution_json(data)
+
```
maskarb, line 497: If we aren't requiring a `distribution_info` to be sent with the payload, why are we creating one?

maskarb (changes requested):
Some other comments:

koku/cost_models/serializers.py
```diff
@@ -54,6 +55,44 @@ class MarkupSerializer(serializers.Serializer):
...
+        if not any(ele in data.keys() for ele in self.DISTRIBUTION_OPTIONS):
+            data = data.get("distribution_info", {})
+        if data.keys() and not (sorted(self.DISTRIBUTION_OPTIONS) == sorted(data.keys())):
+            distribution_info_str = ", ".join(str(option) for option in self.DISTRIBUTION_OPTIONS)
+            error_msg = f"Missing distribution information: one of {distribution_info_str}"
+            raise serializers.ValidationError(error_msg)
+        else:
+            _distribute_type = data.get("distribute_type", metric_constants.CPU_DISTRIBUTION)
+            valid_distribute_type = self.validate_distribute_type(_distribute_type)
+            if data == {}:
```
samdoran, line 87: It's more idiomatic to use a truth value test rather than comparison.
```suggestion
            if not data:
```

koku/cost_models/serializers.py
samdoran: I like this approach much better. Using `set` simplifies things a lot here, and it makes is much simpler to provide a useful error message.

koku/cost_models/serializers.py
maskarb: After reading the JIRA, I guess we are moving the distribution into the `distribution_info` json as the `distribution_type` field? Is that correct?

koku/cost_models/serializers.py
djnakabaale: I agree with this as well, thanks @maskarb 👍🏿 
I've updated that validate method

koku/cost_models/serializers.py
```diff
@@ -54,6 +55,44 @@ class MarkupSerializer(serializers.Serializer):
     unit = serializers.ChoiceField(choices=MARKUP_CHOICES, required=False)
 
 
+class DistributionSerializer(BaseSerializer):
+    """Serializer for distribution options"""
+
+    DISTRIBUTION_OPTIONS = ("distribute_type", "worker_cost", "platform_cost")
+
+    distribute_type = serializers.ChoiceField(choices=metric_constants.DISTRIBUTION_CHOICES, required=False)
```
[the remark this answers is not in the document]
djnakabaale: its distribution_type here

koku/cost_models/serializers.py
djnakabaale: we are keeping the distribution field for now, to be removed in a followup clean up step

koku/cost_models/serializers.py
```diff
@@ -54,6 +55,28 @@ class MarkupSerializer(serializers.Serializer):
...
+    distribution_type = serializers.ChoiceField(choices=metric_constants.DISTRIBUTION_CHOICES, required=False)
+    platform_cost = serializers.BooleanField(required=False)
+    worker_cost = serializers.BooleanField(required=False)
+
+    def validate(self, data):
+        """Run validation for distribution options."""
+
+        diff = self.DISTRIBUTION_OPTIONS.difference(data)
+        if diff == self.DISTRIBUTION_OPTIONS:
+            return {"distribution_type": metric_constants.CPU_DISTRIBUTION, "platform_cost": True, "worker_cost": True}
```
maskarb, line 72: Let's include a test that hits this line. It seems we need a test that sends a `distribution_info: {}` json.

maskarb (approved):
lgtm! Thanks for making those changes!

PR #4168, opened 2023-02-28
[COST-3315] filter more on tag matching ocpgcp

cgoodfred:
## Jira Ticket

[COST-3315](https://issues.redhat.com/browse/COST-3315)

## Description

This change will filter the tag matching ocpgcp section and not waste resources joining a table if nothing matches.

## Testing

1. smokes should be sufficient for testing.

## Notes

...


adberglund (approved):
🗄️

PR #4170, opened 2023-03-01
Tweaking override end dates

lcouzens:
## Jira Ticket

## Description

This change will ...

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Notes

...


lcouzens:
OCI raw calc was the only smoke failure. Probably just a 1st of the month issue.

koku/masu/processor/ocp/ocp_report_parquet_summary_updater.py
```diff
@@ -59,6 +60,9 @@ def _get_sql_inputs(self, start_date, end_date):
                     last_day_of_month = calendar.monthrange(bill_date.year, bill_date.month)[1]
                     start_date = bill_date
                     end_date = bill_date.replace(day=last_day_of_month)
+                    dh = DateHelper()
+                    if bill_date.year == dh.today.year and bill_date.month == dh.today.month:
```
maskarb, line 64: We've already imported and used `DateAccessor` so let's just use that here. 
```
self._date_accessor.today().year
and
self._date_accessor.today().month
```

maskarb (approved):
lgtm!

PR #4182, opened 2023-03-08
COST-3570: switch clowdApp to replicas instead of minReplicas

djnakabaale:
## Jira Ticket

[COST-3570](https://issues.redhat.com/browse/COST-3570)

## Description

This change will replace minReplicas with replicas in clowdapp and in the respective patch yml files

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Notes

...


maskarb (approved):
lgtm! Let's make sure things deploy in the smoke tests before merging.

PR #4187, opened 2023-03-09
Put mocks in correct order

samdoran:
## Description

Mocks are [passed in in the order they are applied](https://docs.python.org/3/library/unittest.mock.html#quick-guide), which is bottom up. This doesn't affect current tests since they aren't asserting anything on those mocks. Still best to have these correct when/if these tests are used to make new tests that do make assertions on those mocks.

## Testing

1. Checkout Branch
2. Run `tox -e py39 -- providers.test.aws.test_aws_provider`


koku/providers/test/aws/test_aws_provider.py
```diff
@@ -255,7 +255,7 @@ def test_check_cost_report_access_fail(self, mock_boto3_client):
     @patch("providers.aws.provider._check_s3_access", return_value=True)
     @patch("providers.aws.provider._check_cost_report_access", return_value=True)
     def test_cost_usage_source_is_reachable(
-        self, mock_get_sts_access, mock_check_s3_access, mock_check_cost_report_access
+        self, mock_check_cost_report_access, mock_check_s3_access, mock_get_sts_access
```
maskarb, line 258: These were all passing because we're not using these args in the func. We could replace these with `*args` instead.

maskarb (approved):
A comment:

PR #4195, opened 2023-03-11
tech-debt: rename "Presto" to "Trino"

maskarb:
* where possible, rename "Presto" to "Trino"
* _ADD_ `TRINO_PORT` and `TRINO_HOST` -> once this PR merges and these are added to app-interface, we can remove the `PRESTO_PORT` and `PRESTO_PORT` env variables


adberglund (approved):
This has been one of those little thing nagging at the back of my mind. Thank you!

maskarb:
> This has been one of those little thing nagging at the back of my mind. Thank you!

Same here!

PR #4198, opened 2023-03-14
[COST-3563] Improve tag search to trigger tag serializer

myersCody:
## Jira Ticket

[COST-3563](https://issues.redhat.com/browse/COST-3563)

## Description

This change will modify our url substring search  we use to trigger the Tag Serializer:
-   `[tag` -> urlencoding -> `%Btag`,
-  `:tag` -> urlencoding -> `%3Atag` 

This will allow us to do things like: `filter[aws_category:env]=stage`. While I was here I also remove the tag & aws_category category from an if/else statement allowing for both grouping by tag & aws_category. This will be needed whenever we start the UI piece of this epic. 

## Testing

1. Checkout Branch
2. load test customer data

## Notes

...


myersCody:
Successful Smoke Test Run: 
- https://ci.ext.devshift.net/job/project-koku-koku-pr-check/8446/

maskarb (commented):
some comments:

koku/api/query_params.py
```diff
@@ -479,16 +479,23 @@ def _validate(self, query_params):
...
-        else:
-            qps = self.serializer(data=query_params, context={"request": self.request})
+        serializer_kwargs = {"data": query_params, "context": {"request": self.request}}
+        # the url_data property is percent encoded
+        # https://en.wikipedia.org/wiki/URL_encoding
+        # [tag: == %5Btag
+        # or:tag == or%3Atag
+        for url_encoded_prefix in ["%5B", "%3A"]:
+            if self.report_type != "tags":
+                tag_search = url_encoded_prefix + TAG_PREFIX.replace(":", "")
```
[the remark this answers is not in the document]
maskarb: Ya know, just thinking a bit more here, I bet we could get away with searching for just the `PREFIX` in the decoded string. I can't think of a situation where either `tag:` or `aws_category:` would be a valid query value. Using your example, one would need to send:
```
filter[aws_category:env]=stag:&filter[something else]
```
but I can't imagine that this is not a bad request.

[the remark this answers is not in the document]
maskarb: SIKE, that is somehow valid:
```
GET /api/cost-management/v1/reports/aws/costs/?filter[account]=stag:adfg

HTTP 200 OK
Allow: GET, HEAD, OPTIONS
Vary: X_RH_IDENTITY, Accept

{
    "meta": {
        "count": 10,
        "currency": "USD",
        "cost_type": "unblended_cost",
        "check_tags": false,
        "filter": {
            "account": [
                "stag:adfg"
            ],
```
So scratch that. We do need the `[, :` prefixes on the PREFIX.

koku/api/query_params.py
```diff
@@ -479,16 +479,23 @@ def _validate(self, query_params):
...
-                data=query_params, aws_category_keys=self.aws_category_keys, context={"request": self.request}
-            )
-        else:
-            qps = self.serializer(data=query_params, context={"request": self.request})
+        serializer_kwargs = {"data": query_params, "context": {"request": self.request}}
+        # the url_data property is percent encoded
+        # https://en.wikipedia.org/wiki/URL_encoding
+        # [tag: == %5Btag
+        # or:tag == or%3Atag
+        for url_encoded_prefix in ["%5B", "%3A"]:
```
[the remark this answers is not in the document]
samdoran: Unquoting at the boundary makes sense here. It comes in as a URL encoded string, we then decode it for internal processing, then encode it on the way out if necessary.

koku/api/report/serializers.py
```diff
@@ -180,44 +177,22 @@ def _init_tag_keys(self, field, fargs=None, fkwargs=None):
...
 
-        tag_fields = {}
-        for key in self.tag_keys:
-            if len(self.tag_keys) > 1 and "child" in fkwargs.keys():
+        fields = {}
+        prefix_keys = set()
+        if self.tag_keys is not None:
+            prefix_keys.update(self.tag_keys)
+        if self._aws_category and self.aws_category_keys:
+            prefix_keys.update(self.aws_category_keys)
```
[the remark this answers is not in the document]
myersCody: I like this suggestion! I can replace the `None` with `set()` and switch to this 👍

koku/api/query_params.py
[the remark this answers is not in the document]
myersCody: I didn't think about handling it this way! I can unquote it for the search! Great Suggestion Michael.

koku/api/query_params.py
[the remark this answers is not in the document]
myersCody: I still like the idea of removing the unnecessary replace here though, but I agree its better to keep the prefixes. 🙇

koku/api/test_query_params.py
```diff
@@ -684,8 +684,8 @@ def test_process_tag_query_params(self):
             "filter[resolution]=monthly&"
             "filter[time_scope_value]=-1&"
             "filter[time_scope_units]=month&"
-            "filter[tag:environment]=prod&"
-            "group_by[tag:app]=*"
+            "filter%5Btag:environment]=prod&"  # urlencoded search now use for tag
+            "group_by%5Btag:app]=stage"
```
samdoran, line 688: Should we have test scenarios for both URL encoded and text values?

koku/api/report/serializers.py
```diff
@@ -180,44 +177,20 @@ def _init_tag_keys(self, field, fargs=None, fkwargs=None):
...
-            if len(self.tag_keys) > 1 and "child" in fkwargs.keys():
+        fields = {}
+        prefix_keys = set()
+        prefix_keys.update(self.tag_keys)
+        prefix_keys.update(self.aws_category_keys)
+        if not prefix_keys:
+            return
+
+        for key in prefix_keys:
+            if len(prefix_keys) > 1 and "child" in fkwargs.keys():
```
samdoran, line 188:
```suggestion
            if len(prefix_keys) > 1 and "child" in fkwargs:
```

koku/api/test_query_params.py
samdoran: I still think it's worth testing URL encoded values being sent to `QueryParameters._validate()`. Maybe that's not what this was testing.

myersCody:
Smoke tests look good: https://ci.ext.devshift.net/job/project-koku-koku-pr-check/8475/

myersCody:
I think I might of thought a better way to do this. Playing with the idea now.

koku/api/query_params.py
```diff
@@ -425,25 +387,61 @@ def _set_access_ocp_all(self, provider, filter_key, access_key, raise_exception=
...
+            if not isinstance(value, (dict, list)):
+                value = [value]
+            for inner_key in value:
+                stripped_key = self._strip_prefix(inner_key, "tag", prefix_list)
+                if stripped_key in self.tag_keys:
+                    param_tag_keys.add(inner_key)
+            stripped_key = self._strip_prefix(key, "tag", prefix_list)
+            if stripped_key in self.tag_keys:
+                param_tag_keys.add(key)
+        self.tag_keys = param_tag_keys
```
maskarb, line 425: I don't quite understand why this was moved to here. Previously this chunk of code would not run if self.tag_keys was empty, now it runs regardless. I'd either switch this back, or add a check for 
```
if not self.tag_keys:
    return
```

maskarb (commented):
It looks pretty good! Just one question/comment:

koku/api/query_params.py
```diff
@@ -157,60 +160,19 @@ def __str__(self):
...
+        Args:
+            key: key passed in the query param
+            common_substring: found in all elements of prefix list of strings
+            prefix_list: possible prefixes to be found in query params
+        """
+        if common_substring not in key:
+            return key
+        for prefix in prefix_list:
+            if key.startswith(prefix):
+                return key[len(prefix) :]  # noqa: E203
```
samdoran, line 175:
```suggestion
                return key[len(prefix):] 
```

koku/api/query_params.py
```diff
@@ -425,25 +387,61 @@ def _set_access_ocp_all(self, provider, filter_key, access_key, raise_exception=
...
+        """Searches url and returns true if any prefix is found in a url."""
+        # the url_data property is percent encoded
+        # https://en.wikipedia.org/wiki/URL_encoding
+        # %5Btag == [tag:
+        # %5Bor%3Atag == [or:tag
+        decoded_url = unquote(self.url_data)
+        for prefix in prefix_list:
+            if f"[{prefix}" in decoded_url:
+                return True
+        return False
```
samdoran, line 400: An alternative solution. Not sure if it's any more readable, though.
```suggestion
        return any(f"[{prefix}" in decoded_url for prefix in prefix_list)
```

koku/api/query_params.py
```diff
@@ -425,25 +387,61 @@ def _set_access_ocp_all(self, provider, filter_key, access_key, raise_exception=
...
                 self.parameters["filter"][filter_key] = access_list
 
-    def _set_tag_keys(self):
+    def search_url_for_prefixes(self, prefix_list):
+        """Searches url and returns true if any prefix is found in a url."""
+        # the url_data property is percent encoded
+        # https://en.wikipedia.org/wiki/URL_encoding
+        # %5Btag == [tag:
+        # %5Bor%3Atag == [or:tag
+        decoded_url = unquote(self.url_data)
```
samdoran, line 396: Rather than unquoting this every time this method is called,  I think it makes more sense to store the unquoted value on the object. Or maybe inspect the `QueryDict` object directly. That may be more precise than sub string matching.

samdoran: I think I'm getting the bigger picture of how we're doing the parsing here. I'd have to see more examples of the actual values of the query string to see if `querystring_parser` is buying us much over `QueryDict().dict()` or other similar methods.

samdoran: Or if we don't need the values to be url encoded, just pass in the characters we care about as safe.

```python
@property
def url_data(self):
    """Get the url_data."""
    return self.request.GET.urlencode(safe="[]:")
```

koku/api/query_params.py
myersCody: Yeah, the db model could contain no tag keys, so I will add a `if not self.tag_keys`

----
*TLDR:  Let me provide a little context to why I moved the logic here.*

So, I noticed we were doing the same if condition twice pretty much back to back inside of the init. Once inside of the [_set_tag_keys](https://github.com/project-koku/koku/blob/main/koku/api/query_params.py#L482) and once inside of [_validate](https://github.com/project-koku/koku/blob/main/koku/api/query_params.py#L482). 

I then noticed that the only time `_set_tag_keys()` is called is only called in the init, then we call the `_validate` function which also contains our only call of `_process_tag_query_params` as long as the same if statement is true. 

```
self._set_tag_keys()  # sets self.tag_keys
self._set_aws_category_keys()
self._validate(query_params)  # sets self.parameters
```

So, I then spent a little time trying to understand the purpose of the `_process_tag_query_params` function. When we called `_set_tag_keys` we use the if condition to see if we should collect all the acceptable tag keys from the database. I then realized  the purpose of the `_process_tag_query_params` was to reduce the tag key set to be only be keys passed into the `query_params`, but at the same time validate that those keys exist in the DB. Which is why we initially set the tag keys to all keys in the db in the `_set_tag_keys`. 

I thought this manipulation of the tag key set made more sense inside of the `_set_tag_keys` method. So, I removed `_process_tag_query_params` method and passed the query_params to the `_set_tag_keys`. I took the if condition short circuit returns for the `_set_tag_keys` function. 

I then changed the if condition to be a short circuit for the `_set_tag_keys`. If it tags or if we don't find a prefix in the url, exit and leave the tag keys as an empty set. Otherwise grab the tag keys from the db and check the query_param keys.

koku/api/query_params.py
myersCody: So, this one of those times where black & flake8 are fighting each other. if I try to commit this it will change it then flake8 will complain:
```
Flake8..........................................................................Failed
- hook id: flake8
- exit code: 1

koku/api/query_params.py:175:39: E203 whitespace before ':'

Check pipenv's Pipfile.lock to be consistent with Pipfile...(no files to check)Skipped
Validate Grafana Dashboards.................................(no files to check)Skipped
(koku) cmyers@cmyers1-mac koku % git diff
diff --git a/koku/api/query_params.py b/koku/api/query_params.py
index 53589f2f..710d3298 100644
--- a/koku/api/query_params.py
+++ b/koku/api/query_params.py
@@ -172,7 +172,7 @@ class QueryParameters:
             return key
         for prefix in prefix_list:
             if key.startswith(prefix):
-                return key[len(prefix):]
+                return key[len(prefix) :]
```

myersCody:
Successful Smokes: https://ci.ext.devshift.net/job/project-koku-koku-pr-check/8569/

koku/api/query_params.py
samdoran: Well that's pretty frustrating. We should just disable E203 in our flake8 config.

samdoran (approved):

PR #4210, opened 2023-03-16
Cost 3548 distribute worker

myersCody:
## Jira Ticket

[COST-3548](https://issues.redhat.com/browse/COST-3548)

## Description

This change will ...

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Notes

...

### DIFF
--- a/koku/masu/database/ocp_report_db_accessor.py
+++ b/koku/masu/database/ocp_report_db_accessor.py
@@ -21,6 +21,7 @@ from jinjasql import JinjaSql
 from tenant_schemas.utils import schema_context
 from trino.exceptions import TrinoExternalError
 
+from api.metrics.constants import DEFAULT_DISTRIBUTION_TYPE
 from api.provider.models import Provider
 from api.utils import DateHelper
 from koku.database import SQLScriptAtomicExecutorMixin
@@ -506,7 +507,9 @@ class OCPReportDBAccessor(SQLScriptAtomicExecutorMixin, ReportDBAccessorBase):
             )
             return [(pvc[0], pvc[1], pvc[2]) for pvc in unique_pvcs]
 
-    def populate_platform_distributed_cost_sql(self, start_date, end_date, distribution, provider_uuid):
+    def populate_platform_and_worker_distributed_cost_sql(
+        self, start_date, end_date, provider_uuid, distribution_info
+    ):
         """
         Populate the platform cost distribution of a customer.
 
@@ -516,20 +519,25 @@ class OCPReportDBAccessor(SQLScriptAtomicExecutorMixin, ReportDBAccessorBase):
             distribution: Choice of monthly distribution ex. memory
             provider_uuid (str): The str of the provider UUID
         """
+        distribute_mapping = {}
+        distribution = distribution_info.get("distribution_type", DEFAULT_DISTRIBUTION_TYPE)
+        if distribution_info.get("platform_cost"):
+            distribute_mapping["distribute_platform_cost.sql"] = "platform"
+        if distribution_info.get("worker_cost"):
+            distribute_mapping["distribute_worker_cost.sql"] = "worker"
+        if not distribute_mapping:
+            return
         table_name = self._table_map["line_item_daily_summary"]
         report_period = self.report_periods_for_provider_uuid(provider_uuid, start_date)
         if not report_period:
             LOG.info(
                 f"No report period for OCP provider {provider_uuid} with start date {start_date},"
-                " skipping populate_platform_distributed_cost_sql update."
+                " skipping platform_and_worker_distributed_cost_sql update."
             )
             return
         with schema_context(self.schema):
             report_period_id = report_period.id
-
-        platform_sql = pkgutil.get_data("masu.database", "sql/openshift/cost_model/distribute_platform_cost.sql")
-        platform_sql = platform_sql.decode("utf-8")
-        platform_sql_params = {
+        default_sql_params = {
             "start_date": start_date,
             "end_date": end_date,
             "schema": self.schema,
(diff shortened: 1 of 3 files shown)


myersCody:
OCP on GCP raw calc test again: https://ci.ext.devshift.net/job/project-koku-koku-pr-check/8526/
Gonna improve codecove patch now.

koku/masu/database/ocp_report_db_accessor.py
```diff
@@ -516,41 +519,50 @@ def populate_platform_distributed_cost_sql(self, start_date, end_date, distribut
...
             provider_uuid (str): The str of the provider UUID
         """
+        distribute_mapping = {}
+        distribution = distribution_info.get("distribution_type", DEFAULT_DISTRIBUTION_TYPE)
+        if distribution_info.get("platform_cost"):
+            distribute_mapping["distribute_platform_cost.sql"] = "platform"
+        if distribution_info.get("worker_cost"):
+            distribute_mapping["distribute_worker_cost.sql"] = "worker"
+        if not distribute_mapping:
+            return
```
samdoran, line 529: Should this be if, elif, else?

koku/masu/database/ocp_report_db_accessor.py
```diff
@@ -516,41 +519,50 @@ def populate_platform_distributed_cost_sql(self, start_date, end_date, distribut
...
-            bind_params=list(platform_sql_params),
-            operation="INSERT",
-        )
+
+        LOG.debug(default_sql_params)
+        for sql_file, log_msg in distribute_mapping.items():
+            templated_sql = pkgutil.get_data("masu.database", f"sql/openshift/cost_model/{sql_file}")
+            templated_sql = templated_sql.decode("utf-8")
+            templated_sql, templated_sql_params = self.jinja_sql.prepare_query(templated_sql, default_sql_params)
+            LOG.info(
```
samdoran, line 554: Does this need `log_json()` with a tracing ID and context?

myersCody: I can start looking into add log_json to this file.

koku/masu/database/ocp_report_db_accessor.py
```diff
@@ -516,41 +520,55 @@ def populate_platform_distributed_cost_sql(self, start_date, end_date, distribut
...
+            return
         table_name = self._table_map["line_item_daily_summary"]
         report_period = self.report_periods_for_provider_uuid(provider_uuid, start_date)
         if not report_period:
-            LOG.info(
-                f"No report period for OCP provider {provider_uuid} with start date {start_date},"
-                " skipping populate_platform_distributed_cost_sql update."
-            )
+            msg = "No report period for OCP provider, skipping platform_and_worker_distributed_cost_sql update."
+            context = {"provider_uuid": provider_uuid, "start_date": {start_date}}
```
maskarb, line 535:
```suggestion
            context = {"provider_uuid": provider_uuid, "start_date": start_date}
```

koku/masu/database/ocp_report_db_accessor.py
```diff
@@ -516,41 +520,55 @@ def populate_platform_distributed_cost_sql(self, start_date, end_date, distribut
...
+
+        LOG.debug(default_sql_params)
+        for sql_file, log_msg in distribute_mapping.items():
+            templated_sql = pkgutil.get_data("masu.database", f"sql/openshift/cost_model/{sql_file}")
+            templated_sql = templated_sql.decode("utf-8")
+            templated_sql, templated_sql_params = self.jinja_sql.prepare_query(templated_sql, default_sql_params)
+            log_msg = f"Distributing {log_msg} cost."
+            context = {
+                "provider_uuid": provider_uuid,
+                "dates": f"{start_date} - {end_date}",
```
maskarb, line 560:
```suggestion
                "start_date": start_date,
                "end_date": end_date,
```
How would we be able to use `dates` to search Kibana?

koku/masu/test/database/test_ocp_report_db_accessor.py
```diff
@@ -1032,3 +1034,52 @@ def test_populate_usage_costs_new_columns_no_report_period(self):
...
+        end_date = "2000-02-01"
+        msg = "No report period for OCP provider, skipping platform_and_worker_distributed_cost_sql update."
+        context = {"provider_uuid": self.provider_uuid, "start_date": {start_date}}
+        expected = f"INFO:masu.database.ocp_report_db_accessor:{log_json(self.provider_uuid, msg, context)}"
+
+        with self.assertLogs("masu.database.ocp_report_db_accessor", level="INFO") as logger:
+            self.accessor.populate_platform_and_worker_distributed_cost_sql(
+                start_date, end_date, self.provider_uuid, {"platform_cost": True}
+            )
+            self.assertIn(expected, logger.output)
```
maskarb, line 1050: I feel like this test is entirely _too_ specific for a unit test. It should be sufficient to just make sure `No report period for OCP provider, skipping platform_and_worker_distributed_cost_sql update` was logged. All this other stuff is not necessary.

It is fine to test that something was logged when that is the only outcome given the function input, but testing the actual logged content isn't great especially if we go back and change how things are logged.

Yes, I understand that there are other tests in this file that are similar to this one, but I would argue that those are not good tests either.

koku/masu/test/database/test_ocp_report_db_accessor.py
```diff
@@ -1032,3 +1034,52 @@ def test_populate_usage_costs_new_columns_no_report_period(self):
...
+
+    @patch("masu.database.ocp_report_db_accessor.pkgutil.get_data")
+    @patch("masu.database.ocp_report_db_accessor.OCPReportDBAccessor._execute_raw_sql_query")
+    def test_populate_platform_and_worker_distributed_cost_sql_called(self, mock_sql_execute, mock_data_get):
+        """Test that the platform distribution is called."""
+        dh = DateHelper()
+        start_date = dh.this_month_start.date()
+        end_date = dh.this_month_end.date()
```
maskarb, line 1058: This code is fine, but just as an FYI, the base test class now has a `dh` attribute, so you can totally use this if you wanted:
```suggestion
        start_date = self.dh.this_month_start.date()
        end_date = self.dh.this_month_end.date()
```

maskarb (commented):
A few comments:

koku/masu/test/database/test_ocp_report_db_accessor.py
myersCody: Good point, I was just pushing a quick fix after updating to log_json. I just switched it to check if none.

maskarb: Yea thats good with me!

maskarb (approved):

samdoran (approved):

koku/masu/test/database/test_ocp_report_db_accessor.py
maskarb: Actually, thinking on this again, WE know that these params will result in the func returning `None` without doing anything, but this test does not assert that. This function returns `None` regardless. This is why I had mentioned to assert on just `No report period for OCP provider, skipping platform_and_worker_distributed_cost_sql update` because that is all we have to assert on given this path through the function.

The other reason I didn't want such a specific test assertion before was, we have a team goal to change how we log things and we still have not come up with some standard or guidance yet. So I don't see value in adding a test that checks every character of a logged message when it's possible we're gonna have to go back and re-write a test because it was overly specific.

PR #4218, opened 2023-03-20
COST - Add NGINX proxy to allow for Cost and ROS-OCP API traffic

chambridge:
* Update base with ngnix conf configmap
* Update Kustomization to pull in new nginx patch
* Update koku to change deployment names 3scale service mapping
* Add nginx patch file with deployment setup


maskarb (approved):
lgtm!

PR #4223, opened 2023-03-22
pr_check.sh - Use string comparison operators

samdoran:
## Description

SC2071 (error): > is for string comparisons. Use -gt instead.
SC2071 (error): < is for string comparisons. Use -lt instead.

https://www.shellcheck.net/wiki/SC2071

## Testing

1. Run `pr_check.sh`

## Notes

This is currently working as desired but by accident. The `<>` string comparison operators return true or false based on the lexicographical sorting of strings.

### DIFF
--- a/pr_check.sh
+++ b/pr_check.sh
@@ -185,10 +185,10 @@ if [[ $exit_code == 0 ]]; then
     fi
 fi
 
-if [[ $exit_code > 0 ]]; then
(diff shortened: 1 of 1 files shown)


maskarb:
Given that we are changing the file that runs PR checks, we should actually test it to see how it behaves in Jenkins.

maskarb:
lgtm!

PR #4225, opened 2023-03-22
[COST-3638] Trino enabled tag filter

adberglund:
## Jira Ticket

[COST-3638](https://issues.redhat.com/browse/COST-3638)

## Description

This change will filter tag keys by whether they are enabled in Trino and stop doing the operation in Postgres.

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Notes

...


lcouzens:
Since this work is in the parking lot and we have a bunch of conflicts I'm just going to close it for now. We can work from this branch if we chose too, if and when the work is picked up.

PR #4228, opened 2023-03-23
COST: Add metrics endpoint to nginx pod

chambridge:
* Add metrics endpoint to nginx pod to try and resolve prometheus up alert


maskarb (approved):
lgtm!

PR #4231, opened 2023-03-24
[COST-3590] Add source count condition to truncate

adberglund:
## Jira Ticket

[COST-3590](https://issues.redhat.com/browse/COST-3590)

## Description

This change will only allow TRUNCATES on OCP on Cloud data when a single cloud source contributes all of the data. Otherwise we must DELETE to avoid wiping data for other sources.

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Notes

...


djnakabaale (dismissed):
lgtm!

djnakabaale (approved):

PR #4233, opened 2023-03-27
order deletes to mitigate deadlocks

maskarb:
## Jira Ticket

## Description

Update our SQL deletes to use an ordered subquery and `FOR UPDATE`.
* https://www.cockroachlabs.com/blog/select-for-update/
* https://stackoverflow.com/a/27015834
* https://www.postgresql.org/docs/current/sql-select.html#SQL-FOR-UPDATE-SHARE

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Notes

...


maskarb:
We have not had many deadlock errors lately, so closing this.

PR #4247, opened 2023-03-29
[COST-3554] Add line item type to AWS line items daily table

samdoran:
## Jira Ticket

[COST-3555](https://issues.redhat.com/browse/COST-3555)

## Description

The first part of COST-3554. We need to store the line item type so that later we can get data based on the line item type.

## Testing

I'm not sure exactly.


lcouzens (approved):
Looks good, regarding testing you can try it out in ephemeral if you want to, there is a section in our release docs.

PR #4251, opened 2023-03-30
adding additional AWS column logic

lcouzens:
## Jira Ticket

## Description

This change will add additional columns + logic for the customer filtered data flow for AWS

## Testing

1. Checkout Branch
2. Restart Koku
3. Create AWS-local source
4.  Add some data with customer filtered columns (They Like Snake case over there) example `bill_billing_entity`
5. See that the data is ingested correctly

## Notes

...


koku/masu/processor/parquet/parquet_report_processor.py
```diff
@@ -329,7 +329,7 @@ def enabled_tags_model(self):
     def csv_columns(self):
         """Return the required CSV columns if we need to filter them"""
         if self.provider_type == Provider.PROVIDER_AWS:
-            return CSV_REQUIRED_COLUMNS.get(Provider.PROVIDER_AWS)
+            return CSV_REQUIRED_COLUMNS.get(Provider.PROVIDER_AWS) + CSV_REQUIRED_COLUMNS.get("AWS-custom")
```
samdoran, line 332: Before, this would return `None` or a tuple. Now it could potentially raise a `TypeError` if either results in `None`.

This should get the key directly, which will raise a more obvious `KeyError` if something is amiss.

```suggestion
            return CSV_REQUIRED_COLUMNS[Provider.PROVIDER_AWS] + CSV_REQUIRED_COLUMNS["AWS-custom"]
```

koku/masu/processor/parquet/parquet_report_processor.py
```diff
@@ -475,7 +475,16 @@ def convert_csv_to_parquet(self, csv_filename):  # noqa: C901
             col_names = pd.read_csv(csv_filename, nrows=0, **kwargs).columns
             if self.ingress_reports:
                 REQUIRED_COLS = set(CSV_REQUIRED_COLUMNS.get(self._provider_type))
+                missing_cols = False
                 if not set(col_names).issuperset(REQUIRED_COLS):
+                    missing_cols = True
+                    if missing_cols and self._provider_type in {Provider.PROVIDER_AWS, Provider.PROVIDER_AWS_LOCAL}:
+                        REQUIRED_CUSTOM_COLS = set(CSV_REQUIRED_COLUMNS.get("AWS-custom"))
```
samdoran, line 482: A `KeyError` more clearly shows a potential problem here as well.
```suggestion
                        REQUIRED_CUSTOM_COLS = set(CSV_REQUIRED_COLUMNS["AWS-custom"])
```

koku/masu/util/aws/common.py
```diff
@@ -32,6 +32,55 @@
 
 LOG = logging.getLogger(__name__)
 
+col_translation = {
```
samdoran, line 35:
```suggestion
COL_TRANSLATION = {
```

koku/masu/processor/parquet/parquet_report_processor.py
```diff
@@ -475,7 +475,16 @@ def convert_csv_to_parquet(self, csv_filename):  # noqa: C901
...
+                        missing_cols = False
+                        if not set(col_names).issuperset(REQUIRED_CUSTOM_COLS):
+                            missing_cols = True
+                            REQUIRED_COLS = REQUIRED_CUSTOM_COLS
+                if missing_cols:
```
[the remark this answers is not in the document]
lcouzens: That is true but the previous logic was also AWS specific here anyway so..... :shrug:

koku/masu/processor/parquet/parquet_report_processor.py
```diff
@@ -475,7 +484,12 @@ def convert_csv_to_parquet(self, csv_filename):  # noqa: C901
             col_names = pd.read_csv(csv_filename, nrows=0, **kwargs).columns
             if self.ingress_reports:
                 REQUIRED_COLS = set(CSV_REQUIRED_COLUMNS.get(self._provider_type))
+                missing_cols = False
                 if not set(col_names).issuperset(REQUIRED_COLS):
+                    missing_cols = True
+                    if missing_cols and self._provider_type in {Provider.PROVIDER_AWS, Provider.PROVIDER_AWS_LOCAL}:
```
cgoodfred, line 490: Since the return value for `self.col_checker` is None if its not one we have defined a provider for, we could probably just use that as the second condition for this.
```suggestion
                    if missing_cols and self.col_checker:
```

koku/masu/util/aws/common.py
```diff
@@ -518,11 +567,27 @@ def scrub_resource_col_name(res_col_name):
...
     """
+    org_columns = set(list(data_frame))
+    columns = []
+    for col in org_columns:
+        if "/" not in col:
+            if COL_TRANSLATION.get(col):
+                data_frame = data_frame.rename(columns={col: COL_TRANSLATION[col]})
+                columns.append(COL_TRANSLATION[col])
     columns = set(list(data_frame))
+    columns = set(columns)
```
samdoran, line 590: I believe this line is equivalent to the previous line, which are both equivalent to `set(data_frame)`.

koku/masu/util/aws/common.py
```diff
@@ -518,11 +567,27 @@ def scrub_resource_col_name(res_col_name):
...
     columns = set(TRINO_REQUIRED_COLUMNS).union(columns)
```
samdoran, line 591: Set methods accept an iterable, so the previous two lines could be removed in favor of 
```suggestion
    columns = set(TRINO_REQUIRED_COLUMNS).union(data_frame)
```

koku/masu/processor/parquet/parquet_report_processor.py
```diff
@@ -475,7 +485,12 @@ def convert_csv_to_parquet(self, csv_filename):  # noqa: C901
             col_names = pd.read_csv(csv_filename, nrows=0, **kwargs).columns
             if self.ingress_reports:
                 REQUIRED_COLS = set(CSV_REQUIRED_COLUMNS.get(self._provider_type))
+                missing_cols = False
                 if not set(col_names).issuperset(REQUIRED_COLS):
+                    missing_cols = True
+                    if missing_cols and self.col_checker:
```
maskarb, line 491:
```suggestion
                    if self.col_checker:
```
Total nitpick, but `missing_cols` will never be False here.

lcouzens: Whoops! :smile:

koku/masu/util/aws/common.py
```diff
@@ -518,12 +567,26 @@ def scrub_resource_col_name(res_col_name):
...
     """
-    columns = set(list(data_frame))
-    columns = set(TRINO_REQUIRED_COLUMNS).union(columns)
+    org_columns = set(list(data_frame))
+    columns = []
+    for col in org_columns:
+        if "/" not in col:
+            if COL_TRANSLATION.get(col):
+                data_frame = data_frame.rename(columns={col: COL_TRANSLATION[col]})
+                columns.append(COL_TRANSLATION[col])
```
maskarb, line 588:
```suggestion
        if "/" not in col and COL_TRANSLATION.get(col):
            data_frame = data_frame.rename(columns={col: COL_TRANSLATION[col]})
            columns.append(COL_TRANSLATION[col])
```
Last nitpick, I swear!

koku/masu/util/aws/common.py
```diff
@@ -518,12 +567,26 @@ def scrub_resource_col_name(res_col_name):
...
+
 def aws_post_processor(data_frame):
     """
     Consume the AWS data and add a column creating a dictionary for the aws tags
     """
-    columns = set(list(data_frame))
-    columns = set(TRINO_REQUIRED_COLUMNS).union(columns)
+    org_columns = set(list(data_frame))
+    columns = []
+    for col in org_columns:
```
samdoran, line 584: If the goal is to iterate over unique column names, it's possible to do that without the intermediate `list()` conversion.
```suggestion
    org_columns = data_frame.columns.unique()
    columns = []
    for col in org_columns:
```

koku/masu/processor/parquet/parquet_report_processor.py
```diff
@@ -239,6 +241,14 @@ def post_processor(self):
             post_processor = ocp_post_processor
         return post_processor
 
+    @property
+    def col_checker(self):
+        """Customer column checker based on provider type."""
+        col_checker = None
+        if self.provider_type in [Provider.PROVIDER_AWS, Provider.PROVIDER_AWS_LOCAL]:
+            col_checker = check_aws_custom_columns
+        return col_checker
```
samdoran, line 250: This interface seems odd, and it hard codes specific behavior in the base class. Passing the checker into the object constructor and raising `NotImplementedError("No custom column checker defined")` creates better boundaries between objects and would more clearly show an incorrect use of the API.

Currently, if the caller forgets to test before calling `self.col_checker()` the caller will see `TypeError: 'NoneType' object is not callable` which isn't a clear indication of the problem.

samdoran: That's ok. Let's not worry about it.

samdoran (approved):

koku/masu/test/util/aws/test_common.py
```diff
@@ -412,6 +412,25 @@ def test_aws_post_processor(self):
...
+
+        processed_data_frame = utils.aws_post_processor(data_frame)
+        if isinstance(processed_data_frame, tuple):
+            processed_data_frame, df_tag_keys = processed_data_frame
+            self.assertIsInstance(df_tag_keys, set)
+
+        self.assertIn(expected_col_one, processed_data_frame)
+        self.assertIn(expected_col_two, processed_data_frame)
+        for column in TRINO_REQUIRED_COLUMNS:
+            self.assertIn(column.replace("-", "_").replace("/", "_").replace(":", "_").lower(), processed_data_frame)
```
samdoran, line 432: Another way to replace multiple strings without using `re` or chaining `replace` calls is to use a translation table.
```suggestion
        self.assertIn(expected_col_two, processed_data_frame)
        strings_to_replace = ("-", "/", ":")
        translation = {ord(s): ord("_") for s in strings_to_replace}
        for column in TRINO_REQUIRED_COLUMNS:
            self.assertIn(column.translate(translation).lower(), processed_data_frame)
```

samdoran: Not a blocker, just an FYI.

PR #4260, opened 2023-03-31
Update nginx config to make sure recommendation traffic only goes to the ros-ocp-api

chambridge:
* Update nginx conf


maskarb (approved):
🤞🏻

PR #4271, opened 2023-04-05
[COST-3615] Improve AWS category serializer sanity to use enabled fields

myersCody:
## Jira Ticket

[COST-3615](https://issues.redhat.com/browse/COST-3615)

## Description

This change will ...

## Testing

1. Got to the function and set it to return True: 
- (Just mocking unleash here for development)
```
def enable_aws_category_settings(account):
    """Enable aws category settings."""
    account = convert_account(account)

    context = {"schema": account}
    res = bool(UNLEASH_CLIENT.is_enabled("cost-management.backend.enable_aws_category_settings", context))

    return True
```
2. Hit the settings page: http://localhost:8000/api/cost-management/v1/settings/

Notice the currently enabled values:
```
initialValue": [
                    "aws-OUs",
                    "aws-env",
                    "aws-Organization"
                ],
```
3. Post to the settings page to disable all but one:
```
{
    "api": {
        "settings": {
            "currency": "USD",
            "tag-management": {
                "enabled": [
                ]
            },
            "cost_type": "savingsplan_effective_cost",
            "aws-category-management": {
                "enabled": [
                    "aws-OUs",
                ]
            }
        }
    }
}
```
4. Make sure you can still hit the enabled key:
- `http://localhost:8000/api/cost-management/v1/reports/aws/costs/?group_by[aws_category:OUs]=value`

5. Try one of the disabled keys:
- `http://localhost:8000/api/cost-management/v1/reports/aws/costs/?group_by[aws_category:env]=value`

See:
```
{
    "group_by": {
        "aws_category:env": [
        ]
    }
}
```
6. Reenable the key you just tried through a post to the settings endpoint:
```
postgres=# select * from reporting_awsenabledcategorykeys;
     key      | enabled
--------------+---------
 OUs          | t
 env          | t
 Organization | f
 CostCenter   | f
 Charge type  | f
 ```
 7. Rehit the same endpoint and watch the Unsupported parameter or invalid value go away: 
 ```
 {
    "meta": {
        "count": 10,
        "currency": "USD",
        "cost_type": "savingsplan_effective_cost",
        "check_tags": false,
        "group_by": {
            "aws_category:env": [
                "value"
            ]
        },
 ```
 

## Notes

...

### CODE REVIEW

File: koku/api/query_params.py [hunk 1]
```diff
@@ -416,23 +417,36 @@ def _set_tag_keys(self, query_params):
...
     def _set_aws_category_keys(self, query_params):
-        """Set the valid aws_category keys"""
+        """Set the valid self.aws_category keys.
+
+        The aws_category_keys variable is used in the report serializer
+        to update the valid field names list. Any key added to this set
+        will not a trigger the unsupport parameter or invalid value error.
+        """
         prefix_list = [AWS_CATEGORY_PREFIX, AND_AWS_CATEGORY_PREFIX, OR_AWS_CATEGORY_PREFIX]
         self.aws_category_keys = set()
```
samdoran, line 427: I should have caught this earlier. Avoid creating new attributes in methods. This should be in the constructor so the attribute is always available.

koku/api/query_params.py
```diff
@@ -416,23 +417,36 @@ def _set_tag_keys(self, query_params):
...
-                if AWS_CATEGORY_PREFIX in inner_key:
-                    param_aws_category_keys.add(inner_key)
-                    # TODO: Right now we accept any key passed in.
-                    # However, we should rewrite this to only
-                    # allow acceptable categorites keys based off
-                    # keys found in the DB. (like above)
-        self.aws_category_keys = param_aws_category_keys
+                stripped_key = self._strip_prefix(inner_key, AWS_CATEGORY_PREFIX, prefix_list)
+                if stripped_key in enabled_category_keys:
+                    self.aws_category_keys.add(inner_key)
```
samdoran, line 449: I feel like this could be reduced to a recursive function, but I've been messing around with it for an hour and haven't come up with a good solution. :shipit:

samdoran (changes requested):
Move `self.aws_category_keys` to the constructor.

maskarb (approved):

samdoran (approved):

PR #4274, opened 2023-04-07
[COST-3614] AWS category type ahead

myersCody:
## Jira Ticket

## Description

This change will ...

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Notes

...


koku/api/resource_types/aws_category/serializers.py
```diff
@@ -0,0 +1,18 @@
+#
+# Copyright 2023 Red Hat Inc.
+# SPDX-License-Identifier: Apache-2.0
+#
+"""Serializer for aws category resource types."""
+import logging
+
+from rest_framework import serializers
+
+LOG = logging.getLogger(__name__)
```
maskarb, line 10: We should not import logging and create this logger if we are not going to use it. If you use this locally for debugging, that is fine if that is how you work, but don't leave it here. It is unused code.

koku/api/resource_types/test/aws_category/test_views.py
```diff
@@ -0,0 +1,67 @@
...
+    @RbacPermissions({"aws.account": {"read": ["*"]}})
+    def test_aws_categories_view_with_wildcard(self):
+        """Test aws categories return."""
+        with schema_context(self.schema):
+            expected = AWSEnabledCategoryKeys.objects.filter(enabled=True).values_list("key", flat=True)
+            url = reverse("aws-categories")
+            response = self.client.get(url, **self.headers)
+            self.assertEqual(response.status_code, status.HTTP_200_OK)
+            json_result = response.json()
+            LOG.info(json_result)
```
maskarb, line 44: We don't need to log anything in unit tests, so let's remove these and the logger.

maskarb (commented):
This PR looks pretty good overall. I have a couple of nitpicks:

koku/api/resource_types/aws_category/serializers.py
myersCody: This was honestly just a sloppy copy & paste error, so I went back to where I copied it from and removed it from there. Which made me wonder where else we import logging and don't use it. Which lead me down a rabbithole, but I think I got all instances where we import it and don't use it.

koku/api/resource_types/test/aws_category/test_views.py
myersCody: Same as above, your comment lead me down an audit and I removed the logger from the unittest files.

maskarb (approved):
lgtm! Thanks for removing all those instances of the unused logger!

myersCody:
successful smokes: https://ci.ext.devshift.net/job/project-koku-koku-pr-check/9030/

maskarb (approved):

PR #4294, opened 2023-04-12
[COST-3703] sonarqube: require GET requests

maskarb:
## Jira Ticket

[COST-3703](https://issues.redhat.com/browse/COST-3703)

## Description

This change will ...

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Notes

...


maskarb:
Closing because these api endpoints only accept `GET` requests as-is, so there is no point in these changes. We acknowledged the Sonarqube hotspots as safe.

PR #4308, opened 2023-04-13
[COST-3715] Handle unexpected error message data type

samdoran:
## Jira Ticket

[COST-3715](https://issues.redhat.com/browse/COST-3715)

The data in `err_dict[err_key]` could be something other than a list or it could be empty. Catch and handle the exception so a good error message is returned to the customer.

## Description

Return a good error message when the data is not a list.

## Testing

```
tox -e py39 -- sources.test.test_sources_error_message
```

### DIFF
--- a/koku/sources/sources_error_message.py
+++ b/koku/sources/sources_error_message.py
@@ -80,9 +80,9 @@ class SourcesErrorMessage:
             try:
                 err_body = err_dict.get(err_key, [None]).pop()
             except TypeError:
-                err_msg = str(self._error.detail).encode().decode("UTF-8")
+                err_msg = str(self._error.detail)
             else:
-                err_msg = err_body.encode().decode("UTF-8")
+                err_msg = err_body
 
         return err_key, err_msg
 
--- a/koku/sources/sources_error_message.py
+++ b/koku/sources/sources_error_message.py
@@ -68,12 +68,14 @@ class SourcesErrorMessage:
         if isinstance(self._error, ValidationError):
             err_dict = self._error.detail
             LOG.info(f"validation error: {self._error}. Validation detail {err_dict}")
-            err_key = list(err_dict.keys()).pop()
-            err_body = err_dict.get(err_key, []).pop()
-            if err_body:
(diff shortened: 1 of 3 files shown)


samdoran:
Very good point. I just assumed it was AWS. I'll just make a generic "bucket missing" message.

koku/sources/sources_error_message.py
```diff
@@ -68,12 +76,14 @@ def _extract_from_validation_error(self):
...
             LOG.info(f"validation error: {self._error}. Validation detail {err_dict}")
-            err_key = list(err_dict.keys()).pop()
-            err_body = err_dict.get(err_key, []).pop()
-            if err_body:
-                err_msg = err_body.encode().decode("UTF-8")
-            else:
+            err_key = list(err_dict).pop()
+            try:
+                err_body = err_dict.get(err_key, []).pop()
+            except (TypeError, IndexError):
```
maskarb, line 82: The whole purpose of this `_extract_from_validation_error` is to pull the `ErrorDetail` out and ship that back to Sources. The `err_msg` we return here is displayed in the UI, so we cannot send back the full `err_body`.

maskarb, line 82: What raises a TypeError here?

Also, we can avoid the IndexError by giving the default list a value, such as `[None]`.
```suggestion
            err_key = list(err_dict).pop()
            err_body = err_dict.get(err_key, [None]).pop()
            if err_body:
                err_msg = err_body.encode().decode("UTF-8")
```

samdoran: `{}.pop()` raises a `TypeError`. This was the source of the original error. We assumed the data would always be a list and (because of another bug) we're getting a dict instead.

> Also, we can avoid the IndexError by giving the default list a value, such as [None]

Yup, that would work.

samdoran: We still need to catch and handle the `TypeError`, though.

samdoran: Any idea what the `.encode.decode('uff-8')` here accomplishes?

PR #4310, opened 2023-04-13
[COST-3614] Add filter options to aws category resource types endpoint.

myersCody:
## Jira Ticket

[COST-3614](https://issues.redhat.com/browse/COST-3614)

## Description

This change will add filters to the resource_types endpoint for aws categories.

## Testing

load aws data:
1. http://localhost:8000/api/cost-management/v1/resource-types/aws-categories/?account=9999999999990
2. http://localhost:8000/api/cost-management/v1/resource-types/aws-categories/?limit=1
3. http://localhost:8000/api/cost-management/v1/resource-types/aws-categories/?key=env
4. http://localhost:8000/api/cost-management/v1/resource-types/aws-categories/?value=prod

## Notes

...

### DIFF
--- a/docs/specs/openapi.json
+++ b/docs/specs/openapi.json
@@ -3167,16 +3167,16 @@
                 "summary": "Obtain mapping of key and values for AWS Categories.",
                 "operationId": "listResourcesAwsCategories",
                 "parameters": [{
-                        "$ref": "#/components/parameters/QueryOffset"
+                        "$ref": "#/components/parameters/QueryLimit"
                     },
                     {
-                        "$ref": "#/components/parameters/QueryLimit"
+                        "$ref": "#/components/parameters/QueryKey"
                     },
                     {
                         "$ref": "#/components/parameters/QueryValue"
                     },
                     {
-                        "$ref": "#/components/parameters/QueryOrder"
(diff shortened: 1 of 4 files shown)


myersCody:
> One super general question, this endpoint is very different from all of the other resource types endpoints, both in the format of the response and the way search and limit work. Is that intentional?
> 
> I know some of the resource types endpoints are used for things like user access, does the return here being different impact that?

This is a great question Corey, so all of the other resource_type endpoints only return values. However, this is the only resource_types endpoint that returns a list of key, value pairs. So I tried to model it as best as I could after the way we do the tags endpoint instead, which is why the return is so different.

I open to suggestions here, but this was the best I could figure out to get the desired functionality. As for the user_access part, I don't believe it will affect it.

cgoodfred (approved):

chambridge:
> One super general question, this endpoint is very different from all of the other resource types endpoints, both in the format of the response and the way search and limit work. Is that intentional?
> 
> I know some of the resource types endpoints are used for things like user access, does the return here being different impact that?

No, this shouldn't impact that as wasn't added to the `/api/cost-management/v1/resource-types/` endpoint. So its purely available only for type-ahead.

PR #4320, opened 2023-04-19
Fix container not found sentry error

lcouzens:
## Jira Ticket

[COST-3735](https://issues.redhat.com/browse/COST-3735)

## Description

This change will handle exceptions for Container not found errors. This case can show its face when a customer creates a cost export then creates a source immediately afterwards; before the cost export has had chance to run and create the container.

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Notes

...


lcouzens:
smokes passed

PR #4324, opened 2023-04-21
[COST-3746] limit distributed_cost param to only group by project

myersCody:
## Jira Ticket

[COST-3746](https://issues.redhat.com/browse/COST-3746)

## Description

This change will limit the distributed_cost param options to group by project. 

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Notes

...


koku/api/report/ocp/serializers.py
```diff
@@ -129,6 +130,16 @@ class OCPQueryParamSerializer(ReportQueryParamSerializer):
     FILTER_SERIALIZER = OCPFilterSerializer
     EXCLUDE_SERIALIZER = OCPExcludeSerializer
 
+    def to_internal_value(self, data):
+        """Send to internal value."""
+        if delta_value := data.get("delta"):
+            if isinstance(delta_value, str):
```
cgoodfred, line 136: Is it possible for `delta_value` to be anything other than a string here?

myersCody: Yes, if you pass in more than two delta params:
```
?delta=distributed_cost&delta=cost
```
It will turn it into a list causing the param value to be `["distributed_cost", "cost"]`, Which causes an error when passed into the `.get()`. More than one delta is not supported anyway, so this will throw a validation error instead.

myersCody:
100% smoke passed: https://ci.ext.devshift.net/job/project-koku-koku-pr-check/9252/

cgoodfred (approved):

PR #4331, opened 2023-04-24
Add a flag to show if overhead cost is distributed.

myersCody:
## Jira Ticket

## Description

This change will ...

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Notes

...


koku/api/report/ocp/query_handler.py
```diff
@@ -152,6 +165,7 @@ def _format_query_response(self):
 
         """
         output = self._initialize_response_output(self.parameters)
+        output = self._check_if_distributed_overhead(output)
```
samdoran, line 168: Rather than mutate the passed in data, a clearer interface would be to return a boolean and have the caller set the value.
```suggestion
        output = self._check_if_distributed_overhead(output)
        output["distributed_overhead"] = self._check_if_distributed_overhead(output)
```

maskarb: In this case though, we only set `output["distributed_overhead"]` when `self._report_type == "costs_by_project"`... Though, I suppose that bit can be moved here instead of inside of _check_if_distributed_overhead.

myersCody: I went ahead and moved the logic here instead.

lcouzens (approved):

PR #4335, opened 2023-04-26
[COST-3715] Handle unexpected error message data type

samdoran:
## Jira Ticket

[COST-3715](https://issues.redhat.com/browse/COST-3715)

The data in `err_dict[err_key]` could be something other than a list or it could be empty. Catch and handle the exception so a good error message is returned to the customer.

## Description

Return a good error message when the data is not a list.

## Testing

```
tox -e py39 -- sources.test.test_sources_error_message
```


maskarb (approved):
lgtm!

PR #4336, opened 2023-04-27
sonarcloud: add duplicate lines exclusions

maskarb:
* add more exclusions for code coverage and duplicate line counting


sonar-project.properties
```diff
@@ -21,4 +21,5 @@ sonar.exclusions=**/test_*.py,**/test/**/*,**/*.html,**/*.yml,**/*.yaml,**/*.jso
 sonar.tests=koku/
 sonar.test.inclusions=**/test_*.py,**/test/**/*
 
-sonar.coverage.exclusions=**/migrations/*.py,**/manage.py,**/settings.py
+sonar.coverage.exclusions=**/migrations/*.py,**/manage.py,**/settings.py,**/celery.py,**/configurator.py,**/feature_flags.py,**/probe_server.py,**/urls.py
```
samdoran, line 24: Most of these files look like operational code that should be covered by tests. Granted, the `.coveragerc` already omits a bit more than this.

Should we try and make this exclusions list match `.coveragerc` removing an cruft? Maybe that's what you already did with this list.

samdoran (approved):

PR #4337, opened 2023-04-27
[COST-3720] Update confluent kafka and don't create a new producer every time

cgoodfred:
## Jira Ticket

[COST-3720](https://issues.redhat.com/browse/COST-3720)

## Description

This change will update packages and not create a new producer every time we need to send a message. A new producer will be created and then re-used to send the messages. 

Removing `"message.timeout.ms"` as there is a default set and switching to using poll to do this asynchronously. the first message the producer sends can take over a second to send and then the messages after are lightning fast based on my ephemeral testing. More details on what that setting is here: https://github.com/confluentinc/librdkafka/blob/master/CONFIGURATION.md 

## Testing

1. Reserve an ephemeral namespace
2. Deploy this branch to an ephemeral namespace
4. Port forward and create an OCP source in ephemeral
5. Port forward and send an OCP report in ephemeral
6. Validate that the source gets created and data is in the db using the ephemeral terminal or by port forwarding and checking endpoints.

## Notes

...


koku/kafka_utils/utils.py
```diff
@@ -57,7 +58,7 @@ def get_consumer(conf_settings, address=Config.INSIGHTS_KAFKA_ADDRESS):  # pragm
 
 def _get_producer_config(address, conf_settings):  # pragma: no cover
     """Return Kafka Producer config"""
-    producer_conf = {"bootstrap.servers": address, "message.timeout.ms": 1000}
+    producer_conf = {"bootstrap.servers": address, "message.timeout.ms": 3000}
```
maskarb, line 61:
```suggestion
    producer_conf = {"bootstrap.servers": address}
```
Let's just remove this setting since there is a default value.

koku/koku/notifications.py
```diff
@@ -96,10 +96,10 @@ def send_notification(self, msg):
         """
         producer = get_producer()
         producer.produce(Config.NOTIFICATION_TOPIC, value=msg, callback=delivery_callback)
-        # Wait up to 1 second for events. Callbacks will be invoked during
+        # Wait up to 3 seconds for events. Callbacks will be invoked during
         # this method call if the message is acknowledged.
         # `flush` makes this process synchronous compared to async with `poll`
-        producer.flush(1)
+        producer.flush(3)
```
maskarb, line 102: We can just make this `.poll(0)`. We don't need to block to confirm the message was delivered. We don't even retry to send the message if it fails, so there is no point in waiting for confirmation.

maskarb (changes requested):

koku/kafka_utils/utils.py
cgoodfred: Removed

koku/koku/notifications.py
cgoodfred: Switched again to using `.poll(0)` in the places we send a message.

koku/kafka_utils/utils.py
```diff
@@ -66,10 +67,14 @@ def _get_producer_config(address, conf_settings):  # pragma: no cover
...
     """Create a Kafka producer."""
+    global PRODUCER
+    if PRODUCER:
+        return PRODUCER
     if conf_settings is None:
         conf_settings = {}
     conf = _get_producer_config(address, conf_settings)
-    return Producer(conf)
+    PRODUCER = Producer(conf)
+    return PRODUCER
```
samdoran, line 77: This looks like we want `Producer` to be a singleton. Rather than use a global, make a class that returns a singleton.

samdoran: Here's an example Singleton metaclass:
```python
class SingletonMeta(type):
    """Metaclass for classes that wish to implement Singleton functionality.
    If an instance of a class already exists, it is returned. Otherwise a new
    instance is created and returned in a thread safe manner.
    """

    _instance = None
    _rlock = RLock()

    def __call__(cls, *args, **kwargs):
        if cls._instance is not None:
            return cls._instance

        with cls._rlock:
            if cls._instance is None:
                print("Creating with lock")
                cls._instance = super().__call__(*args, **kwargs)

        return cls._instance


class ProducerSingleton(Producer, metaclass=SingletonMeta):
    ...
```

Then `get_producer` becomes much simpler.

```python
def get_producer(conf_settings=None, address=Config.INSIGHTS_KAFKA_ADDRESS):  # pragma: no cover
    """Create a Kafka producer."""
    if conf_settings is None:
        conf_settings = {}
    conf = _get_producer_config(address, conf_settings)
    return ProducerSingleton(conf)
```

samdoran: Example of this in action:

```shell
# Using Singleton meta class

In [3]: id(get_producer())
Out[3]: 4451232640

In [4]: id(get_producer())
Out[4]: 4451232640

In [5]: id(get_producer())
Out[5]: 4451232640

In [6]: id(get_producer())
Out[6]: 4451232640
```

```shell
# Behavior without changes in this PR

In [2]: id(get_producer())
Out[2]: 4426746336

In [3]: id(get_producer())
Out[3]: 4426821120

In [4]: id(get_producer())
Out[4]: 4426775888
```

koku/masu/util/common.py
```diff
@@ -570,3 +571,28 @@ def convert_account(account):
...
+        with cls._lock:
+            # The first thread to acquire the lock, reaches this conditional,
+            # goes inside and creates the Singleton instance. Once it leaves the
+            # lock block, a thread that might have been waiting for the lock
+            # release may then enter this section. But since the Singleton field
+            # is already initialized, the thread won't create a new object.
+            if cls not in cls._instances:
+                instance = super().__call__(*args, **kwargs)
+                cls._instances[cls] = instance
+        return cls._instances[cls]
```
samdoran, line 598: I've been scrutinizing the idea of using a class var to store the singleton instance and I think there may be an issue with the example I shared earlier: I believe `cls` is the `__call__()` is the class using the metaclass, not the metaclass itself. It seems to work, but I need to look at why a bit more.

samdoran: I think it may be clearer to use an `__init__` in the metaclass and explicitly store the singleton instance of a class on the class itself, not the singleton metaclass. I'll come up with an example.

samdoran (changes requested):

koku/masu/util/common.py
maskarb: We really need to wrap this PR up. Do we need this singleton class? I'd argue since this is the first time we've used a singleton, let's just use the global variable since it was working. If we need to add a new singleton in the future, then maybe we can implement this metaclass and update things here, but I don't see the value in adding it now since all we are doing is delaying this PR.

samdoran: Using a Singleton is the right move here. Using a global is not thread safe and is susceptible to mutation whereas a Singleton is not.

I worked out what this implementation is doing and it's storing `_instances` on the subclass object itself rather than in the Singleton class attribute. That's fine but not obvious.

I was thinking it would read better and make more sense to store the single on the subclass singleton instance itself, but I don't want to hold up the PR for that. The current implementation works just fine (but `RLock` should be used instead of `Lock`).

koku/masu/util/common.py
```diff
@@ -13,6 +13,8 @@
 from itertools import groupby
 from os import remove
 from tempfile import gettempdir
+from threading import Lock
```
samdoran, line 16: This can be removed and the type hint updated.
```suggestion
```

koku/masu/util/common.py
```diff
@@ -570,3 +572,28 @@ def convert_account(account):
...
     return {key: value for key, value in dictionary.items() if key in keys_to_keep}
+
+
+class SingletonMeta(type):
+    """
+    This is a thread-safe implementation of Singleton.
+    """
+
+    _instances = {}
+    _lock: Lock = RLock()
```
samdoran, line 583:
```suggestion
    _lock: RLock = RLock()
```

samdoran (approved):

maskarb (approved):

koku/masu/util/common.py
samdoran: Here's an example (for future reference or just science) of storing the singleton instance on the instance instead of in a dict in the metaclass object.

```python
# This is @sivel's implementation. Took me a while to get my head around it.
class Singleton(type):
    def __init__(cls, name, bases, dct):
        super().__init__(name, bases, dct)
        cls.__instance = None
        cls.__rlock = RLock()

    def __call__(cls, *args, **kwargs):
        if cls.__instance is not None:
            return cls.__instance

        if cls.__instance is None:
            with cls.__rlock:
                cls.__instance = super().__call__(*args, **kwargs)

        return cls.__instance
```

Name mangling keeps those on the object instance but tucked away.

```
In [7]: msc = MySingletonClass()

In [8]: msc._Singleton__instance
Out[8]: <__main__.MySingletonClass at 0x1082b2670>

In [9]: msc._Singleton__rlock
Out[9]: <unlocked _thread.RLock object owner=0 count=0 at 0x1053d49c0>

In [11]: msc is MySingletonClass()
Out[11]: True
```

This is more about organization than being better or worse than the existing implementation. The only minor difference is that each singleton object instance get its own `RLock` instance.

PR #4340, opened 2023-04-28
Compose SQL query instead of using format string

samdoran:
## Description

Resolves https://sonarcloud.io/project/security_hotspots?id=project-koku_koku&hotspots=AYd3BpTcNHnPNrOk3v0u

## Testing

```
tox -- api.report.test.ocp.test_ocp_query_handler
```

### DIFF
--- a/koku/api/report/queries.py
+++ b/koku/api/report/queries.py
@@ -36,6 +36,7 @@ from django.db.models.functions import Coalesce
 from django.db.models.functions import Concat
 from django.db.models.functions import RowNumber
 from pandas.api.types import CategoricalDtype
+from psycopg2 import sql
 
 from api.currency.models import ExchangeRateDictionary
 from api.models import Provider
@@ -1023,7 +1024,8 @@ class ReportQueryHandler(QueryHandler):
         """
         descending = True if self.order_direction == "desc" else False
         tag_column, tag_value = tag.split("__")
-        return OrderBy(RawSQL("%s -> %s", (tag_column, tag_value)), descending=descending)
+        query = sql.SQL("{tag_column} -> %s").format(tag_column=sql.Identifier(tag_column))
+        return OrderBy(RawSQL(query, (tag_value,)), descending=descending)
 
     def _percent_delta(self, a, b):
         """Calculate a percent delta.
--- a/koku/api/report/queries.py
+++ b/koku/api/report/queries.py
@@ -1023,7 +1023,7 @@ class ReportQueryHandler(QueryHandler):
         """
         descending = True if self.order_direction == "desc" else False
         tag_column, tag_value = tag.split("__")
-        return OrderBy(RawSQL(f"{tag_column} -> %s", (tag_value,)), descending=descending)
+        return OrderBy(RawSQL("%s -> %s", (tag_column, tag_value)), descending=descending)
 
     def _percent_delta(self, a, b):
         """Calculate a percent delta.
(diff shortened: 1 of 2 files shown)


koku/api/report/queries.py
```diff
@@ -1023,7 +1023,7 @@ def get_tag_order_by(self, tag):
         """
         descending = True if self.order_direction == "desc" else False
         tag_column, tag_value = tag.split("__")
-        return OrderBy(RawSQL(f"{tag_column} -> %s", (tag_value,)), descending=descending)
+        return OrderBy(RawSQL("%s -> %s", (tag_column, tag_value)), descending=descending)
```
maskarb, line 1026: I'm honestly not sure that this will ever work. In general, you cannot parameterize column names or table names. I could be wrong, but I really don't think this will work.

samdoran: I was afraid of that. 😞

samdoran: Ok, I understand why this wouldn't work now. I wonder if we could use composition with a `psycopg2.sql.Identifier` object. I'm not sure how that would work with Django's `RawSQL` though. Might have to run the whole thing through `psycopg2.sql.SQL` then render that to a string and pass that to `RawSQL`.

samdoran: Made at attempt using composition. If passing the `Composed` object doesn't work, we should be able to call `.as_string()` and pass that to `RawSQL`. I believe that would be the safe way to do use table and column names (or any SQL identifier, really) in a raw SQL query.

maskarb: The thing that I am unclear about here is what the risk is. _DO_ we even need to make a change here?

samdoran: My understanding is a user could provide a value for `tag_column` of something like `'; DROP table foo;'` since the value is coming from the user-provided parameters. I'm not sure, though. This level SQLI injection protection is quite new to me.

https://github.com/project-koku/koku/blob/32d521bfa1db5e10308f93343f3c798ad73bf06a/koku/api/report/queries.py#L625

maskarb: I've tried to add a ridiculous SQL injection here, and our serializers protect us. Granted, I didn't try anything beyond `?group_by[tag:;%20DROP%20table%20foo;]=*`, but this returns a 400:
```
HTTP 400 Bad Request
Allow: GET, HEAD, OPTIONS
Vary: X_RH_IDENTITY, Accept

{
    "group_by": {
        "tag:; DROP table foo;": [
        ]
    }
}
```
I think we can say this is a Safe operation. But if there is a better way to construct this query, maybe we try.

maskarb: Actually, looking at the code further, this is indeed completely safe. This is where we generate the `{tag_column}__{tag_value}`: 

https://github.com/project-koku/koku/blob/main/koku/api/report/queries.py#L662

So the `tag_column` is NOT user input at all. It is actually a property of the provider_map:

https://github.com/project-koku/koku/blob/main/koku/api/report/provider_map.py#L91-L96

The `tag_value` _IS_ user input, but that is parameterized by the query already, so that is fine.

samdoran: Sweet. Thanks for digging into this.

It would be nice to figure out if this strategy of using composition would work for identifiers. The tests are failing for reasons I expected. I think we would need to call the `as_string()` method before passing it to Django's RawSQL, but I couldn't figure out how to get the correct context object to pass to that method.

PR #4343, opened 2023-05-02
[COST-3773] Properly update group by data when there is overlap between groups

samdoran:
## Jira Ticket

[COST-3773](https://issues.redhat.com/browse/COST-3773)

## Description

- Simplify code that looks for overlap in keys by passing dict views to set constructor and method
- Append dictionary of data to existing list of dictionaries when there is overlap in the keys
- Update tests to cover this code path
- Use duck typing to simplify the code

## Testing

```
tox -- api.report.test.test_queries.ReportQueryUtilsTest.test_group_data_by_list
```

### DIFF
--- a/koku/api/report/queries.py
+++ b/koku/api/report/queries.py
@@ -733,6 +733,7 @@ class ReportQueryHandler(QueryHandler):
 
         Args:
             group_by_list (List): list of strings to group data by
+            group_index (Int): Position in group_by_list that contains group to group by
             data    (List): list of query results
         Returns:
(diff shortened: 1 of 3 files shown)


samdoran:
Sorting before grouping seems to break quite a few tests.

samdoran:
I'm still hitting that sorting bug. I was hoping to avoid having to fix that now.

samdoran:
I think I got it sorted. 😉

adberglund (approved):

PR #4344, opened 2023-05-03
[COST-3554] Calculate and store AWS amortized cost

samdoran:
## Jira Ticket

## Description

Calculates AWS amortized cost based on LineItemType and unblended cost, falling back to savings plan effective cost.

## Testing

TBD

## Notes

- [ ] Use `calculated_amortized_cost` in UI code instead of `savingsplan_savingsplaneffectivecost`

### DIFF
--- a/koku/masu/processor/aws/aws_report_parquet_processor.py
+++ b/koku/masu/processor/aws/aws_report_parquet_processor.py
@@ -26,7 +26,7 @@ class AWSReportParquetProcessor(ReportParquetProcessorBase):
             "lineitem_unblendedrate",
             "lineitem_blendedcost",
             "lineitem_blendedrate",
-            "savingsplan_savingsplaneffectivecost",
+            "calculated_amortized_cost",
             "pricing_publicondemandrate",
             "pricing_publicondemandcost",
         ]
--- a/koku/masu/processor/aws/aws_report_parquet_processor.py
+++ b/koku/masu/processor/aws/aws_report_parquet_processor.py
@@ -26,7 +26,7 @@ class AWSReportParquetProcessor(ReportParquetProcessorBase):
             "lineitem_unblendedrate",
             "lineitem_blendedcost",
             "lineitem_blendedrate",
-            "calculated_amortized_cost",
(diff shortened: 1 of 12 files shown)


koku/masu/processor/aws/aws_report_parquet_processor.py
```diff
@@ -26,7 +26,7 @@ def __init__(self, manifest_id, account, s3_path, provider_uuid, parquet_local_p
             "lineitem_unblendedrate",
             "lineitem_blendedcost",
             "lineitem_blendedrate",
-            "savingsplan_savingsplaneffectivecost",
+            "calculated_amortized_cost",
```
adberglund, line 29: We want to keep the `savingsplan_savingsplaneffectivecost` as the calculated value is just being sent to Postgres, and not in Trino.

koku/masu/util/aws/common.py
```diff
@@ -646,7 +646,7 @@ def aws_generate_daily_data(data_frame):
             "lineitem_blendedcost": ["sum"],
             "pricing_publicondemandcost": ["sum"],
             "pricing_publicondemandrate": ["max"],
-            "savingsplan_savingsplaneffectivecost": ["sum"],
+            "calculated_amortized_cost": ["sum"],
```
adberglund, line 649: Here we don't want to make a change, as the calculated cost doesn't exist in the raw data. This function just rolls up raw hourly data to daily data.

adberglund (changes requested):

koku/masu/util/aws/common.py
samdoran: Ok, so where do we need to return `calculated_amortized_cost` then?

samdoran:
It looks like this needs to be changed in the UI:

https://github.com/project-koku/koku-ui/blob/88ab53d4ed02987116fbbb940f947ba1f9adf7b4/src/routes/views/components/costType/costType.tsx#L44

I think the `AWSProviderMap` will then return the correct value from the database.

adberglund (approved):

PR #4347, opened 2023-05-03
PVC_DIR cleanup

maskarb:
* fix some fallout from removing the testing/pvc_dir directory -> move the .gitkeep to the correct directory


djnakabaale (approved):
Thanks for cleaning this up.

PR #4355, opened 2023-05-05
Remove scheduled task that removes stale tenants

samdoran:
## Description

No longer remove stale tenants regularly. This is currently removing tenants incorrectly since `remote_stale_tenants` is joining on `tenant_id` and we switched to `org_id`.

### DIFF
--- a/koku/koku/celery.py
+++ b/koku/koku/celery.py
@@ -193,12 +193,6 @@ app.conf.beat_schedule["get_daily_currency_rates"] = {
     "schedule": crontab(hour=1, minute=0),
 }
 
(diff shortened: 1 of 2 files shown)


adberglund (dismissed):
Can optionally delete the task definition as well, but this will protect us at least.

samdoran:
I was thinking it'd be good to keep that code in case we want to run it manually. But I should at least update the SQL.

cgoodfred (approved):

djnakabaale (approved):

myersCody (approved):

PR #4366, opened 2023-05-11
[COST-3800] Handle invalid data in OCP order by serializer

maskarb:
## Jira Ticket

[COST-3800](https://issues.redhat.com/browse/COST-3800)

## Description

We are getting strings where we are expecting a dictionary. Account for this and handle it appropriately.

## Testing

```
curl -vfsSL localhost:8000/api/cost-management/v1/reports/openshift/volumes/'?order_by=&start_date="zj\{\{print+"3042"+"4354"\}\}zj"' | tail -n 1 | jello
```

This should return a status code of 400 with useful errors.

```
*   Trying 127.0.0.1:8000...
* Connected to localhost (127.0.0.1) port 8000 (#0)
> GET /api/cost-management/v1/reports/openshift/volumes/?order_by=&start_date="zj{{print+"3042"+"4354"}}zj" HTTP/1.1
> Host: localhost:8000
> User-Agent: curl/7.87.0
> Accept: */*
>
* Mark bundle as not supporting multiuse
< HTTP/1.1 400 Bad Request
< Date: Thu, 18 May 2023 13:46:47 GMT
< Server: WSGIServer/0.2 CPython/3.9.15
< Vary: X_RH_IDENTITY, Accept, Origin
< Allow: GET, HEAD, OPTIONS
< X-Frame-Options: DENY
< Content-Length: 173
< X-Content-Type-Options: nosniff
< Referrer-Policy: same-origin
<
{ [173 bytes data]
* Connection #0 to host localhost left intact
{
  "start_date": [
    "Date has wrong format. Use one of these formats instead: YYYY-MM-DD."
  ],
  "order_by": {
    "non_field_errors": [
      "Invalid data. Expected a dictionary, but got str."
    ]
  }
}
```


myersCody (approved):
Thanks @maskarb !

samdoran:
I'm still working on this fix and adding tests.

koku/api/report/ocp/serializers.py
```diff
@@ -40,11 +44,15 @@ class OCPOrderBySerializer(OrderSerializer):
...
         """Send to internal value."""
-        if not isinstance(data, dict):
-            return super().to_internal_value(data)
-        for serializer_key, internal_key in DISTRIBUTED_COST_INTERNAL.items():
-            if serializer_key in data.keys():
-                data[internal_key] = data.pop(serializer_key)
+        try:
+            keys = data.keys()
+        except AttributeError as exc:
+            LOG.error(f"Unexpected data encountered during serialization: {exc}")
```
maskarb, line 50: What do we gain by logging this error? The user supplied invalid input. All this error will do is spam our alert channel.

This was also why I simply checked if it was a dict. If it is not a dict, then we don't need to modify the internal value at all.

maskarb: Another idea is to do this:
```
        data = super().to_internal_value(data)
        for serializer_key, internal_key in DISTRIBUTED_COST_INTERNAL.items():
            if serializer_key in data.keys():
                data[internal_key] = data.pop(serializer_key)
        return data
```
use super first to validate the data structure, then modify what we have after we know it is valid.

samdoran: The logging there was to avoid swallowing the exception details. But you're right this would be way too noisy.

samdoran: I was trying to avoid the "super sandwich". That looked funny even though it worked.

Doesn't the data need to be run through `super().to_internal_value(data)` _after_ the keys in `data` are potentially changed?

maskarb: Hmm, honestly, this might be the wrong function entirely. We should probably be using `to_representation` here instead of `to_internal_value`:

https://www.django-rest-framework.org/api-guide/serializers/#overriding-serialization-and-deserialization-behavior

Need to think about this one more.

maskarb: I clearly got very distracted here. Anyway, I don't think a try/except is necessarily correct here. These methods are meant to validate data, and I don't think a "ask for forgiveness" pattern is really doing that. I could be down for adding nesting if you're not a fan of the super sandwich:
```
        if isinstance(data, dict):
            for serializer_key, internal_key in DISTRIBUTED_COST_INTERNAL.items():
                if serializer_key in data.keys():
                    data[internal_key] = data.pop(serializer_key)
        return super().to_internal_value(data)
```

samdoran: Looking at what's happening inside `to_internal_value()` it looks like we get primitive validation of the contents of a given field to make sure it contains values in `OrderSerializer.ORDER_CHOICES`.

I think you're correct: an instance check makes the most sense in validation code. I'll go back to basically what you had initially.

samdoran: Thanks for scrutinizing this. I learned some things about DRF and like the solution we ended up with.

samdoran:
Any ideas for other test scenarios? Should I add an integration test that sends data through a request to make sure we get the expected error code?

adberglund (approved):

samdoran:
I added one more test to make sure we return a 400.

lcouzens (approved):

PR #4371, opened 2023-05-15
[COST-3801] Account for values that can be None

djnakabaale:
## Jira Ticket

[COST-3801](https://issues.redhat.com/browse/COST-3801)

## Description

This change will add conditional checks for potential `None` values on some of the date variables, before trying to convert them to date strings.

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Notes

...


koku/api/utils.py
```diff
@@ -513,7 +514,7 @@ def get_months_in_date_range(
             (
                 dt_start.strftime(date_format),
                 dt_end.strftime(date_format),
-                dt_invoice_month.strftime(invoice_date_format),
+                dt_invoice_month.strftime(invoice_date_format) if dt_invoice_month else None,
```
samdoran, line 517: Is returning `None` here ok or should this be a a string?

samdoran: > Passing around invoice month as None is actually ok here.

Awesome. Thanks for verifying.

PR #4375, opened 2023-05-17
Automatically append a slash to URLs

samdoran:
## Description

Automatically append a trailing slash to API calls.

Related to #266
## Testing

```
curl -L localhost:8000/api/cost-management/v1/settings | jello
```


lcouzens (approved):
LGTM as long as smokes are happy.

PR #4378, opened 2023-05-18
[COST-3554] Add column to store amortized cost with markup

samdoran:
## Jira Ticket

## Description

Customers can add a markup value. We need to store the calculated amortized cost with markup.


koku/masu/database/trino_sql/reporting_awscostentrylineitem_daily_summary.sql
```diff
@@ -76,6 +76,7 @@ SELECT uuid() as uuid,
     cast(unblended_cost * {{markup | sqlsafe}} AS decimal(24,9)) as markup_cost,
     cast(blended_cost * {{markup | sqlsafe}} AS decimal(33,15)) as markup_cost_blended,
     cast(savingsplan_effective_cost * {{markup | sqlsafe}} AS decimal(33,15)) as markup_cost_savingsplan
+    cast(calculated_amortized_cost * {{markup | sqlsafe}} AS decimal(33,9)) as markup_cost_amortized
```
cgoodfred, line 79: Any particular reason to use 33,9 for the decimal when we use 33,15 in most places? We're adding a column to the select but not the insert here, we probably want to insert this column into the db too.

koku/reporting/provider/aws/models.py
```diff
@@ -154,8 +154,9 @@ class Meta:
     blended_cost = models.DecimalField(max_digits=24, decimal_places=9, null=True)
     markup_cost_blended = models.DecimalField(max_digits=33, decimal_places=15, null=True)
     savingsplan_effective_cost = models.DecimalField(max_digits=24, decimal_places=9, null=True)
-    calculated_amortized_cost = models.DecimalField(max_digits=33, decimal_places=9, null=True)
     markup_cost_savingsplan = models.DecimalField(max_digits=33, decimal_places=15, null=True)
+    calculated_amortized_cost = models.DecimalField(max_digits=33, decimal_places=9, null=True)
+    markup_cost_amortized = models.DecimalField(max_digits=33, decimal_places=9, null=True)
```
cgoodfred, line 159: All of these tables are being updated with these fields but I don't see any changes to the SQL to actually put values into these columns, is that coming in a separate PR?

koku/masu/database/trino_sql/reporting_awscostentrylineitem_daily_summary.sql
```diff
@@ -100,6 +101,7 @@ FROM (
         max(lineitem_blendedrate) as blended_rate,
         sum(lineitem_blendedcost) as blended_cost,
         sum(savingsplan_savingsplaneffectivecost) as savingsplan_effective_cost,
+        sum(calculated_amortized_cost) as calculated_amortized_cost,
```
cgoodfred, line 104: The next step we do here is to actually calculate `calculated_amortized_cost` as it isn't present on the `aws_line_items_daily` table, so this won't work as expected and I don't think we want to use the same name twice anyway since both lines are `as calculated_amortized_cost`.

koku/masu/database/trino_sql/reporting_awscostentrylineitem_daily_summary.sql
cgoodfred: And theres a missing comma at the end of line 78 above

koku/reporting/provider/aws/models.py
samdoran: No, I guess I need to update some SQL too.

koku/masu/database/trino_sql/reporting_awscostentrylineitem_daily_summary.sql
samdoran: Hmm, I need to read this more carefully and understand what it's doing.

samdoran: Oh geez, I know what this is doing. This line is totally wrong. I'll remove it.

koku/reporting/provider/aws/models.py
cgoodfred: Yeah, so most of these are our UI summary tables, and we actually keep a SQL file per UI summary table that has the same name as the `db_table` in the class meta. These SQL files are how we are populating those tables. For example, here is the file for the `AWSComputeSummaryByAccountP` model: https://github.com/project-koku/koku/blob/main/koku/masu/database/sql/aws/reporting_aws_compute_summary_by_account_p.sql

samdoran:
I discovered we are missing `calculated_amortized_cost` for AWS filtered by OpenShift. I'll have to update those SQL files as well.

koku/reporting/provider/aws/models.py
samdoran: Think I finally wrapped my head around all the places the SQL needs updating. I pushed up the changes but I've got more to do for OCP on AWS.

koku/masu/database/trino_sql/reporting_ocpawscostlineitem_daily_summary.sql
```diff
@@ -383,6 +409,8 @@ SELECT aws.uuid as aws_uuid,
         max(aws.blended_cost) * cast({{markup}} as decimal(33,15)) as markup_cost_blended,
         max(aws.savingsplan_effective_cost) as savingsplan_effective_cost,
         max(aws.savingsplan_effective_cost) * cast({{markup}} as decimal(33,15)) as markup_cost_savingsplan,
+        max(aws.calculated_amortized_cost) as calculated_amortized_cost,
+        max(aws.calculated_amortized_cost) * cast({{markup}} as decimal(33,9)) as calculated_amortized_cost,
```
lcouzens, line 413: This looks wrong, I dont think we should be returning calculated_amortized_cost for both here this one should probably be markup_cost_amortized I'm guessing
```suggestion
        max(aws.calculated_amortized_cost) * cast({{markup}} as decimal(33,9)) as markup_cost_amortized,
```

lcouzens (commented):
one little change and I think we are good

scripts/cji_scripts/migrate_trino.py
```diff
@@ -119,20 +119,22 @@ def main():
...
+    }
 
     for schema in schemas:
         CONNECT_PARAMS["schema"] = schema
-        # logging.info(f"*** Adding column to tables for schema {schema} ***")
-        # add_columns_to_table(columns_to_add, "reporting_ocpawscostlineitem_project_daily_summary", CONNECT_PARAMS)
-        logging.info(f"*** Dropping tables {tables_to_drop} for schema {schema} ***")
-        drop_tables(tables_to_drop, CONNECT_PARAMS)
+        logging.info(f"*** Adding column to tables for schema {schema} ***")
+        add_columns_to_table(columns_to_add, "reporting_awscostentrylineitem_daily_summary", CONNECT_PARAMS)
```
cgoodfred, line 134: We don't need to add columns to this table, this is a postgres only table. We just go from `aws_line_items_daily` into postgres in that SQL without a middle step: https://github.com/project-koku/koku/blob/main/koku/masu/database/trino_sql/reporting_awscostentrylineitem_daily_summary.sql#L1

koku/reporting/provider/aws/openshift/models.py
```diff
@@ -98,8 +98,9 @@ class Meta:
     blended_cost = models.DecimalField(max_digits=33, decimal_places=15, null=True)
     markup_cost_blended = models.DecimalField(max_digits=33, decimal_places=15, null=True)
     savingsplan_effective_cost = models.DecimalField(max_digits=33, decimal_places=15, null=True)
-    calculated_amortized_cost = models.DecimalField(max_digits=33, decimal_places=9, null=True)
     markup_cost_savingsplan = models.DecimalField(max_digits=33, decimal_places=15, null=True)
+    calculated_amortized_cost = models.DecimalField(max_digits=33, decimal_places=9, null=True)
+    markup_cost_amortized = models.DecimalField(max_digits=33, decimal_places=9, null=True)
```
cgoodfred, line 103: You did awesome updating the AWS UI Summary Table SQL. Since were adding these columns to OCP on AWS, we gotta do the same thing in that SQL too. Only thing to keep in mind on this one is its trino SQL and not postgres.

cgoodfred (changes requested):
I think were almost there, just a few things related to those latest updates and we should be good. Thanks for slogging through all these changes.

koku/reporting/provider/aws/openshift/models.py
samdoran: I _think_ I updated the correct SQL files.

cgoodfred (approved):
LGTM, thanks for all the updates Sam!

PR #4387, opened 2023-05-23
Unleash: check env before setting header

maskarb:
## Description

This change will set the header to `f"Bearer {settings.UNLEASH_TOKEN}"` in Stage and Prod, otherwise it will be `f"{settings.UNLEASH_TOKEN}"`.


kholdaway:
Great work @maskarb!

PR #4401, opened 2023-05-30
[COST-3794] Add new API for managing custom cost groups

samdoran:
## Jira Ticket

[COST-3794](https://issues.redhat.com/browse/COST-3794)

## Description

Add new API with PUT/GET/DELETE methods to control what projects belong to which cost groups. Currently there is only one cost group, Platform. In the future, we will all customers to create custom cost groups and add projects to them.

## Testing
Get current cost groups

```
curl -sSL localhost:8000/api/cost-management/v1/settings/cost-groups/
```

Get just non-default projects

```
curl -gsSL 'localhost:8000/api/cost-management/v1/settings/cost-groups/?filter[default]=false'
```

Add new projects to platform projects:

```
curl -X "PUT" "http://localhost:8000/api/cost-management/v1/settings/cost-groups/" \
     -H 'Content-Type: application/json; charset=utf-8' \
     -d '[
    {
        "project_name": "analytics",
        "group": "Platform"
    }
]'
```

Delete projects from cost groups. Default projects _will not be_ deleted.

```
curl -X "DELETE" "http://localhost:8000/api/cost-management/v1/settings/cost-groups/" \
     -H 'Content-Type: application/json; charset=utf-8' \
     -d '[
    {
        "project_name": "analytics",
        "group": "Platform"
    }
]'
```


lcouzens:
So the initial code looks good for the most part. I think we need some updated testing instructions though (Might be because this is still WIP? But it is not in draft so :shrug:) The other part we are still missing is the re-summary step, to trigger a re-summary task for the openshfit provider/source. That will populate the category_id and thus update the platform cost distribution.

I did manually test this out locally and it did seem to work so great job there! In short after you add/remove a project to the platform category we just need to trigger `update_summary_tables` task.

samdoran:
>  I think we need some updated testing instructions though

I added instructions for testing GET, PUT, and DELETE. Sorry I didn't add those before.

> The other part we are still missing is the re-summary step

I added resummarization on PUT or DELETE. `update_summary_tables` requires `schema_name`, `provider_type`, and `provider_uuid`. The schema name is readily available from the request. I user `Provider.PROVIDER_OCP` for `provider_type` though I'm not sure that's correct. I was even less clear on how to get the provider UUID(s). I ran a query to get provider UUIDs based on the customer and type.

I could use some help making sure I'm scheduling this task correctly..

myersCody (approved):

PR #4409, opened 2023-06-01
standardize `schema_name`

maskarb:
## Jira Ticket

## Description

This change will ...

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Notes

...


maskarb:
This PR is not worth updating. We should still standardize on something in the future, but that can be an effort for another day.

PR #4415, opened 2023-06-02
Warn on Invalid ARN

chambridge:
* Add try/except for Invalid ARN
* Add test case for InvalidARN catch


esebesto:
I looked at the last smoke test results and both failures are unrelated to the PR - they are caused by eph env issues which often happens after ~ 1,5 - 2 hrs of run,  resulting in timeouts in ingestion/resummaries  
```
 tests.rest_api.v1.test_user.test_api_aws_settings_tag_enablement
 tests.rest_api.v1.test_user.test_api_aws_settings_categories_enablement
```

samdoran (approved):

PR #4418, opened 2023-06-05
[COST-3688] update trino sql logging

maskarb:
## Jira Ticket

[COST-3688](https://issues.redhat.com/browse/COST-3688)


lcouzens (approved):

koku/api/common/__init__.py
```diff
@@ -12,9 +12,9 @@ def error_obj(key, message):
     return {key: [_(message)]}
 
 
-def log_json(tracing_id, message, context=None, **kwargs):
+def log_json(tracing_id="", msg="", context=None, **kwargs):
```
samdoran, line 15: How about making `msg` a required kwarg only parameter instead of an option positional arg?
```python
def log_json(tracing_id="", *, msg, context=None, **kwargs):
```

PR #4443, opened 2023-06-09
re-enabled cancel or terminate lock info pids via masu

cgoodfred:
## Jira Ticket

[COST-4441](https://issues.redhat.com/browse/COST-4441)

## Description

This change will enable the functionality to cancel/terminate PIDs via the masu db performance endpoints.

## Testing

1. Checkout Branch
2. Create a source or two, doesn't need data, just to exist so `make create-test-customer` is fine.
3. Visit the lock-info db-performance page and verify there are no locks: http://localhost:5042/api/cost-management/v1/db-performance/lock-info/
4. Open a PSQL window to run a little SQL in and run the following:
```
BEGIN;
UPDATE public.api_provider SET paused = false;
```
5. Go to the sources endpoint (http://localhost:8000/api/cost-management/v1/sources/) and decide on a source to try and delete, trigger a delete on that source.
6. Go back to the lock-info page and see information on a deadlock showing.
7. Cancel the PID for the source delete using the provided button after clicking through the verification prompts.
8. The page should refresh and the block should be cleared. The source delete page will now show information about being cancelled due to user request: 
```
"Source removal resulted in UNKNOWN error: OperationalError: canceling statement due to user request\nCONTEXT:  while deleting tuple (0,6) in relation \"api_provider\"\n"
```

## Notes

...


lcouzens:
Closing this since we closed the jira issue for this.

PR #4451, opened 2023-06-09
Create daily files with deterministic names

chambridge:
* Each file should have a deterministic name ignoring the incoming filename
* If daily content spans multiple files they will be output into files suffixed with a counter

### DIFF
--- a/koku/masu/external/downloader/gcp/gcp_report_downloader.py
+++ b/koku/masu/external/downloader/gcp/gcp_report_downloader.py
@@ -59,6 +59,7 @@ def create_daily_archives(
     start_date,
     context={},
     ingress_reports=None,
+    ingress_report_counter=None,
 ):
     """
     Create daily CSVs from incoming report and archive to S3.
@@ -98,8 +99,8 @@ def create_daily_archives(
             date_range = {"start": min(days), "end": max(days), "invoice_month": str(invoice_month)}
             partition_dates = invoice_month_data.partition_date.unique()
             for partition_date in partition_dates:
-                if ingress_reports and ingress_reports.partition_date_counter.get(partition_date) is None:
-                    ingress_reports.partition_date_counter[partition_date] = 0
+                if ingress_reports and ingress_report_counter and ingress_report_counter.get(partition_date) is None:
+                    ingress_report_counter.partition_date_counter[partition_date] = 0
                 partition_date_filter = invoice_month_data["partition_date"] == partition_date
                 invoice_partition_data = invoice_month_data[partition_date_filter]
                 start_of_invoice = dh.invoice_month_start(invoice_month)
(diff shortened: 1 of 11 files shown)


koku/masu/external/downloader/gcp/gcp_report_downloader.py
```diff
@@ -97,13 +98,18 @@ def create_daily_archives(
...
                 partition_date_filter = invoice_month_data["partition_date"] == partition_date
                 invoice_partition_data = invoice_month_data[partition_date_filter]
                 start_of_invoice = dh.invoice_month_start(invoice_month)
                 s3_csv_path = get_path_prefix(
                     account, Provider.PROVIDER_GCP, provider_uuid, start_of_invoice, Config.CSV_DATA_TYPE
                 )
                 day_file = f"{invoice_month}_{partition_date}_{file_name}"
+                if ingress_reports:
+                    day_file = f"{invoice_month}_{partition_date}_{partition_date_counter[partition_date]}"
+                    partition_date_counter[partition_date] = partition_date_counter[partition_date] + 1
```
samdoran, line 112:
```suggestion
                    partition_date_counter[partition_date] +=  1
```

koku/masu/external/downloader/gcp/gcp_report_downloader.py
```diff
@@ -97,13 +98,18 @@ def create_daily_archives(
             date_range = {"start": min(days), "end": max(days), "invoice_month": str(invoice_month)}
             partition_dates = invoice_month_data.partition_date.unique()
             for partition_date in partition_dates:
+                if partition_date_counter.get(partition_date) is None:
```
samdoran, line 101: Unless we need an explicit check for `None`, a truthiness check is more idiomatic.
```suggestion
                if not partition_date_counter.get(partition_date):
```

samdoran (approved):
Just some minor nit picks. Overall strategy seems sound.

koku/masu/external/downloader/gcp/gcp_report_downloader.py
chambridge: It was a choice since its a numeric counter where 0 could be a value and not necessarily intended to be considered false technically resetting it to 0 would be fine, but the expectation is that the absence here would be an explicit None.

lcouzens:
Based on my testing/investigation, regardless how we want to handle file names our side will not prevent duplicate data being sent from the cloud function.

chambridge:
> Based on my testing/investigation, regardless how we want to handle file names our side will not prevent duplicate data being sent from the cloud function.

Can you provide more context as to what duplication happens outside of this? The only thing I had been thinking about over the weekend was that we might want to also enforce a maximum size (number of rows) so we knew we would wipe out all older files (only wouldn't work if somehow we got less data on a later day, but that hardly seems to make sense).

koku/masu/external/downloader/gcp/gcp_report_downloader.py
samdoran: That makes perfect sense.

lcouzens (dismissed):
I think I got myself confused and this may actually work.

PR #4452, opened 2023-06-10
[COST-3847] Create separate processing queues for large customers

chambridge:
## Jira Ticket

[COST-3847](https://issues.redhat.com/browse/COST-3847)

## Description

This change will ...
* Leverage unleash flag with the is_customer_large method
* Route large customers to *_XL queues
* Add new deployments to support the new XL queues
* Capture prometheus metrics for new queues
* Route Masu endpoints to XL queue if a large customer

## Testing

1. Deploy to ephemeral environment
```
bonfire namespace reserve --duration 24h
bonfire deploy hccm --set-template-ref koku=cost-3847-large-customer-queue --set-parameter koku/IMAGE_TAG=pr-4452-1021a7b   --no-remove-resources koku \
  --no-remove-resources presto \
  --no-remove-resources hive-metastore
```
2. Kick off smoke tests `oc apply -f cji-smoke.yml`
```
apiVersion: cloud.redhat.com/v1alpha1
kind: ClowdJobInvocation
metadata:
  name: cost-mgmt-smoke
spec:
  appName: koku
  testing:
    iqe:
      ui:
        enabled: false
      marker: "cost_smoke"
      dynaconfEnvName: "clowder_smoke"
      filter: "test_api"
```
3. Gather results for test when going through regular pipeline and workers ✅ 
`= 194 failed, 4098 passed, 25 skipped, 1200 deselected, 1838 xfailed, 19677 warnings, 3426 errors in 12723.95s (3:32:03) =` and verify no traffic in xl workers
4. Connect to unleash to setup large customer check
`oc port-forward service/env-${NAMESPACE}-featureflags 4242:4242`
5. Add `cost-management.backend.large-customer` to default project as "Operational" feature and set it to "On" for "development" and "production" values (essentially meaning it will always return true.
6. Create new CJI (just change the name) to rerun smokes but now the flow should go through the XL queues and workers
7. Gather results for tests and validate traffic in XL workers ✅ 
`= 87 failed, 4208 passed, 25 skipped, 1200 deselected, 4672 xfailed, 20906 warnings, 589 errors in 16127.07s (4:28:47) =`
verify that while the Orchestrator.prepare() logic does hit the priority queue the download_xl and summary_xl queues and workers handle the traffic and the regular workers had no traffic.

### DIFF
--- a/deploy/clowdapp.yaml
+++ b/deploy/clowdapp.yaml
@@ -1443,6 +1443,177 @@ objects:
           enabled: false
         public:
           enabled: false
+    - name: clowder-worker-cost-model-xl
+      podSpec:
+        command:
+        - /bin/bash
+        - -c
+        - |
+          PYTHONPATH=${APP_HOME} celery -A koku worker --without-gossip -E -l $CELERY_LOG_LEVEL -Q $WORKER_QUEUES
+        env:
+        - name: CLOWDER_ENABLED
+          value: ${CLOWDER_ENABLED}
+        - name: AWS_ACCESS_KEY_ID
+          valueFrom:
+            secretKeyRef:
+              key: aws-access-key-id
+              name: koku-aws
+              optional: false
+        - name: AWS_SECRET_ACCESS_KEY
+          valueFrom:
+            secretKeyRef:
+              key: aws-secret-access-key
+              name: koku-aws
+              optional: false
+        - name: GOOGLE_APPLICATION_CREDENTIALS
+          value: ${GOOGLE_APPLICATION_CREDENTIALS}
+        - name: APP_POD_NAME
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.name
+        - name: DEVELOPMENT
+          value: ${DEVELOPMENT}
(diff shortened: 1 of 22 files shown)


lcouzens (dismissed):
LGTM, The only comment I have is we should probably scale all these deployments to 0 in app-interface for Ephemeral so our PR-check deployments dont take an age to spin up + use extra resources. We probably want/need to have a discussion on what exactly the resources should look like for stage/prod too. :+1:

chambridge:
I can add some vars ahead of merging. Are we ok with me reusing `cost-management.backend.large-customer`? Any concerns about this being combined with the caching minimize concurrent processing for an account? We can certainly revisit or come back later and either switch this logic or remove the concurrent processing flow. Just was one decision point I wanted to have some conversation around.

lcouzens:
Its a good question! Personally I'm ok with reusing the large_customer flag for now at least. I was also thinking along similar lines as you with removing the concurrent flow, least I was kinda hoping this queue idea could replace that anyway.

lcouzens (approved):

PR #4365, opened 2023-05-10
[COST-3554] Use `calculated_amortized_cost`

samdoran:
## Jira Ticket

[COST-3554](https://issues.redhat.com/browse/COST-3554)

## Description

Use `calculated_amortized_cost` in the settings and serializer.

Depends on #4378.

## To Do

- [x] Write data migration to change existing settings value in `user_settings` table.

PR #4459, opened 2023-06-13
[COST-3554] Add data migration to update user settings

samdoran:
## Jira Ticket

## Description

Add a data migration to update user settings to use the new field for amortized cost when that was selected in user settings.

Putting this in a separate PR so that code changes can be done separately from the migration. This makes it easy to perform the data migration in prod when UI changes are ready to release..

Depends on #4365


koku/reporting/migrations/0290_update_settings.py
```diff
@@ -0,0 +1,29 @@
...
+    UserSettings = apps.get_model("reporting", "UserSettings")
+    if user_settings := UserSettings.objects.first():
+        if user_settings.settings.get("cost_type") == "savingsplan_effective_cost":
+            user_settings.settings["cost_type"] = "calculated_amortized_cost"
+            user_settings.save()
+
+
+def revert(apps, schema_editor):
+    UserSettings = apps.get_model("reporting", "UserSettings")
+    if user_settings := UserSettings.objects.first():
```
myersCody, line 15: You may need to refresh my mind about how user settings will work, but will there only ever be one row for each schema?

samdoran: Correct. It's just one row and the `settings` column is a JSON field that contains the settings values.

maskarb (approved):

myersCody (approved):

PR #4463, opened 2023-06-14
[COST-3925] simplify sql params for logging

maskarb:
## Jira Ticket

[COST-3925](https://issues.redhat.com/browse/COST-3925)

## Description

* remove redundant `DateAccessor()` and `JijaSQL()` from classes that inherit from `ReportDBAccessorBase`.
* create a `_prepare_and_execute_raw_sql_query` func that accepts the sql template and sql params and calls `JinjaSQL.prepare_query`. This allows us to skip calling prepare-query before `_execute_raw_sql_query`.
* All of the calls to `_execute_raw_sql_query` in the AWSReportDBAccessor were replaced with calls to `_prepare_and_execute_raw_sql_query`. The other report db accessors can be updated after this PR is merged.
* adding logging into the `_prepare_and_execute_raw_sql_query`
* create an `extract_context_from_sql_params` func that pulls specific fields out of the sql params and gives us a context dict for logging.


lcouzens (approved):

koku/masu/database/report_db_accessor_base.py
```diff
@@ -112,9 +113,32 @@ def _get_db_obj_query(self, table, columns=None):
                 query = table.objects.all()
             return query
 
+    def extract_context_from_sql_params(self, sql_params: dict):
+        return {
+            "schema": sql_params.get("schema"),
+            "start_date": sql_params.get("start") or sql_params.get("start_date"),
+            "end_date": sql_params.get("end") or sql_params.get("end_date"),
```
samdoran, line 120: You could also use nested `get()` methods.
```suggestion
            "start_date": sql_params.get("start", sql_params.get("start_date")),
            "end_date": sql_params.get("end", sql_params.get("end_date")),
```

koku/masu/database/report_db_accessor_base.py
```diff
@@ -112,9 +113,32 @@ def _get_db_obj_query(self, table, columns=None):
...
+
+    def _prepare_and_execute_raw_sql_query(self, table, tmp_sql, tmp_sql_params=None, operation="UPDATE"):
+        """Prepare the sql params and run via a cursor."""
+        if tmp_sql_params is None:
+            tmp_sql_params = {}
+        LOG.info(
+            log_json(
+                msg=f"triggering {operation}",
+                table=table,
+                context=self.extract_context_from_sql_params(tmp_sql_params),
```
samdoran, line 133: You can also just splat in the kwargs.
```suggestion
                **self.extract_context_from_sql_params(tmp_sql_params),
```

samdoran (approved):
Minor style suggestions but no blockers.

koku/masu/database/report_db_accessor_base.py
maskarb: Normally I would go with this, but I don't want a weird edge case where the `sql_params` contains: `{"start": None, "start_date": a-date, "end": None, "end_date": another-date}` because the `sql_params.get("start")` will return `None` here, regardless of what the default is.

PR #4464, opened 2023-06-14
[COST-3796] Migration for AWS cost category settings move

myersCody:
## Jira Ticket

[COST-3796](https://issues.redhat.com/browse/COST-3796)

## Description

This change will add a uuid to the aws cost category model. 

## Testing
1. Checkout main
2. make create-test-customer
3. make load-test-customer-data test_source=AWS
4. check database:
```
postgres=# \dt reporting_aws*
Did not find any relation named "reporting_aws*".
postgres=# set search_path=org1234567;
SET
postgres=# select * from reporting_awsenabledcategorykeys;
     key      | enabled 
--------------+---------
 CostCenter   | t
 Organization | t
 Charge type  | t
 OUs          | t
 env          | t
```
5. checkout this branch
6. make run-migrations
7. check database:
```
postgres=# set search_path=org1234567;
SET
postgres=# select * from reporting_awsenabledcategorykeys;
     key      | enabled |                 uuid                 
--------------+---------+--------------------------------------
 CostCenter   | t       | 090e4301-f58c-4427-9a17-8cdafe8c1bb1
 Organization | t       | 24c3460e-4054-466d-a0ba-9b3af388dbfe
 Charge type  | t       | 229f56ba-8d70-4748-b641-c3bb6cb25243
 OUs          | t       | 1d67bd29-7165-46ab-987a-ed803c2c239f
 env          | t       | 90ac1f84-adee-4c60-8b2d-55e0acadec75
```
8. Use current settings to update enabled keys:
URL: http://localhost:8000/api/cost-management/v1/settings/
Type: POST
Body:
```
{
    "api": {
        "settings": {
            "currency": "USD",
            "tag-management": {
                "enabled": [
                ]
            },
            "cost_type": "savingsplan_effective_cost",
            "aws-category-management": {
                "enabled": [
                    "aws-env"
                ]
            }
        }
    }
}
```
9. Check database:
```
postgres=# select * from reporting_awsenabledcategorykeys;
     key      | enabled |                 uuid                 
--------------+---------+--------------------------------------
 env          | t       | 90ac1f84-adee-4c60-8b2d-55e0acadec75
 CostCenter   | f       | 090e4301-f58c-4427-9a17-8cdafe8c1bb1
 Organization | f       | 24c3460e-4054-466d-a0ba-9b3af388dbfe
 Charge type  | f       | 229f56ba-8d70-4748-b641-c3bb6cb25243
 OUs          | f       | 1d67bd29-7165-46ab-987a-ed803c2c239f
```
10. Check that new keys can be added after migrationis run, with a fresh db on this branch:
```
make create-test-customer
make load-test-customer-data test_source=AWS
```
11. Check Logs:
```
koku-koku-worker-1  | [2023-06-14 14:47:16,082] INFO bde8d5a5-243a-4866-a4c7-1da6bb52ef54 Creating enabled key records: awsenabledtagkeys.
koku-koku-worker-1  | [2023-06-14 14:47:16,084] INFO bde8d5a5-243a-4866-a4c7-1da6bb52ef54 Create batch 1: batch_size 7
koku-koku-worker-1  | [2023-06-14 14:47:16,087] INFO bde8d5a5-243a-4866-a4c7-1da6bb52ef54 Creating enabled key records: awsenabledcategorykeys.
koku-koku-worker-1  | [2023-06-14 14:47:16,088] INFO bde8d5a5-243a-4866-a4c7-1da6bb52ef54 Create batch 1: batch_size 5
koku-koku-worker-1  | [2023-06-14 14:47:16,128] INFO bde8d5a5-243a-4866-a4c7-1da6bb52ef54 Checking for schema
koku-koku-worker-1  | [2023-06-14 14:47:16,178] INFO bde8d5a5-243a-4866-a4c7-1da6bb52ef54 Checking for table
koku-koku-worke
```
12. Check category keys:
```
postgres=# select * from reporting_awsenabledcategorykeys;
     key      | enabled |                 uuid                 
--------------+---------+--------------------------------------
 env          | t       | 8f7e988d-488e-43df-9b1f-ad8deca3983d
 Charge type  | t       | 485a9bfe-c787-43bf-8417-def83eb15f0d
 Organization | t       | 7f8c5688-dc92-4b12-8d4f-c882faa65e67
 OUs          | t       | b6e8303b-6d58-49cc-9962-ab419e6b13c9
 CostCenter   | t       | 30bf5072-d2dd-4d07-9a4d-9d02a4c2048d
```

## Notes

...


koku/reporting/provider/aws/models.py
```diff
@@ -283,6 +283,7 @@ class Meta:
         indexes = [models.Index(fields=["key", "enabled"], name="aws_enabled_category_key_index")]
         db_table = "reporting_awsenabledcategorykeys"
 
+    uuid = models.UUIDField(default=uuid4, editable=False, unique=True, null=False)
```
maskarb, line 286: What is the purpose of this field?

myersCody: The addition of the uuid is part of the new design for enabling & disabling keys through the new settings page APIs. Admittedly, we are changing the aws categories enablement here to be consistent with how we will be enabling tags in the future. 

Our old enabling/disabling tag workflow required you to know the tag key & the provider type of the key. However, we are now introducing uuids to these tables so that we can pass a list of uuids to the enable/disable endpoints for enablement. 

Examples of this can be found in this pull request: https://github.com/project-koku/koku/pull/4466

maskarb: I guess I still don't understand the point. The `key` is the primary key for this table, which automatically makes it unique. Where will we be getting the uuids to enable the keys? and why can we not just enable keys thru the primary key?

lcouzens: I have the same question here.

samdoran: The missing piece here is that provider specific tables will be [combined into one table](https://github.com/project-koku/koku/compare/main...COST-3752-tags-settings-api?diff=split) with all keys for all providers. That introduces the potential for collisions in the `key` field.

cgoodfred: Sam mentioned the tag part having a collision on provider which is why a UUID was added there. The intended purpose of adding it here was to keep the API response and settings flow for enabling/disabling Cost Categories consistent with how Tags are enabled/disabled, both GET responses would include a UUID for each record that could then be used to enable/disable the relevant record. 

Its feasible to just use the key here, it's worth asking Dan if the consistency of a UUID brings him value or not as I would imagine the UI is how people will interact with their settings.

maskarb: If we're trying to make things consistent, that is fine. I think we need to update the table here such that the UUID is the PK instead of the key. It just doesn't make sense to add a random UUID if we don't use it as a PK.

myersCody: Makes the migrations a bit more complicated to use a new PK, but I finally got it figured out: https://github.com/project-koku/koku/pull/4464/commits/40e21f8e86eae3e84964786e9e188d9eae2c9d03

koku/reporting/migrations/0290_awsenabledcategorykeys_uuid.py
```diff
@@ -0,0 +1,49 @@
+# Generated by Django 3.2.19 on 2023-06-20 19:04
```
maskarb, line 1: The steps in this migration are far more complex than I was expecting and I am not sure we should ever try to run this much complexity in a single migration. Creating a new table, moving everything to it, dropping the old table, and then renaming the new table to the old table seems like way too many steps in one migration. Also, why are all of those steps necessary? 

If you set the model to this:
```
class AWSEnabledCategoryKeys(models.Model):
    """A collection of the current enabled category keys."""

    class Meta:
        """Meta for AWSCategoryTagKeys."""

        indexes = [models.Index(fields=["key", "enabled"], name="aws_enabled_category_key_index")]
        db_table = "reporting_awsenabledcategorykeys"

    uuid = models.UUIDField(default=uuid4, editable=False, unique=True, primary_key=True)
    key = models.CharField(max_length=253)
    enabled = models.BooleanField(default=True)
```

and the migration to this:
```
import uuid

from django.db import migrations
from django.db import models


def create_uuid(apps, schema_editor):
    awsenabledcategorykeys = apps.get_model("reporting", "awsenabledcategorykeys")
    for category in awsenabledcategorykeys.objects.all():
        category.uuid = uuid.uuid4()
        category.save()


class Migration(migrations.Migration):

    dependencies = [
        ("reporting", "0288_auto_20230615_1658"),
    ]

    operations = [
        migrations.AddField(
            model_name="awsenabledcategorykeys",
            name="uuid",
            field=models.UUIDField(default=uuid.uuid4, editable=False, null=True),
        ),
        migrations.RunPython(create_uuid, reverse_code=migrations.RunPython.noop),
        migrations.AlterField(
            model_name="awsenabledcategorykeys",
            name="key",
            field=models.CharField(max_length=253),
        ),
        migrations.AlterField(
            model_name="awsenabledcategorykeys",
            name="uuid",
            field=models.UUIDField(default=uuid.uuid4, editable=False, primary_key=True, serialize=False, unique=True),
        ),
    ]

```

you end up with this:
```
postgres=# \d+ reporting_awsenabledcategorykeys;
                            Table "org1234567.reporting_awsenabledcategorykeys"
 Column  |          Type          | Collation | Nullable | Default | Storage  | Stats target | Description 
---------+------------------------+-----------+----------+---------+----------+--------------+-------------
 key     | character varying(253) |           | not null |         | extended |              | 
 enabled | boolean                |           | not null |         | plain    |              | 
 uuid    | uuid                   |           | not null |         | plain    |              | 
Indexes:
    "reporting_awsenabledcategorykeys_uuid_0cd0c5ec_pk" PRIMARY KEY, btree (uuid)
    "reporting_awsenabledcategorykeys_key_enabled_idx" btree (key, enabled)
    "reporting_awsenabledcategorykeys_key_idx" btree (key varchar_pattern_ops)
    "reporting_awsenabledcategorykeys_uuid_0cd0c5ec_uniq" UNIQUE CONSTRAINT, btree (uuid)
Access method: heap
```
which is what we are after. With this migration, we create the new UUID field, populate it, and then swap the keys in place.

myersCody: The key will still need to be unique but that was an easy enough change to add in, I was trying to create the uuid with:
```
migrations.RunSQL("UPDATE reporting_awsenabledcategorykeys SET uuid = uuid_generate_v4();"),
```
But kept running into duplicate kye errors, changing that to python seemed to be the trick. 🙇

lcouzens (dismissed):
Good discussions get good results :smiley:

myersCody:
Unit test are failing because bakery is inserting more copies of the same key than expected for the tests:
```
test_postgres=# select * from reporting_awsenabledcategorykeys;
  key  | enabled |                 uuid                 
-------+---------+--------------------------------------
 ninja | t       | a6849b49-255e-49cb-bd73-f9c99e52c5a4
 ninja | t       | 3fb7521f-436d-4bab-ab2b-57b5afaf73bf
 ninja | t       | d2a284b8-8390-408d-b8e9-eef017b969cc
 env   | t       | 11dab19e-d511-43b8-956c-cce4691f4da7
```
Since `key` used to be the [primary key](https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-PRIMARY-KEYS), it was automatically getting a `unique` constraint. Preventing multiple copies of the same key from existing. 

I was curious to why I didn't see this problem with our dummy data, and our [create_enabled_keys](https://github.com/project-koku/koku/blob/main/koku/masu/util/common.py#L339) method has some built in logic to prevent duplicate keys from being entered into the DB. However, I am not sure if we want to rely on this function from preventing duplicate keys for AWS categories, so I am going to add a constraint to the database to prevent this as well for potential future changes around how we insert enabled keys.

PR #4478, opened 2023-06-16
[COST-3082] Correctly add AWS sources with region information

samdoran:
## Jira Ticket

[COST-3082](https://issues.redhat.com/browse/COST-3082)

## Description

Switch to using the boto3 Client API instead of Resources. No new features are being added to the Resources API and certain resources do not use the region information.

Update signatures to pass region information where it is needed.

Do not specify region for the Cost and Usage Resource client since that is only available in us-east-1.

## Testing

1. Checkout Branch
2. Restart Koku
3. Post an AWS source to `http://localhost:8000/api/cost-management/v1/sources/`


lcouzens (approved):

lcouzens:
AWS smokes passed 100%

PR #4494, opened 2023-06-22
Clean up bad cluster records

lcouzens:
## Jira Ticket

[COST-3793](https://issues.redhat.com/browse/COST-3793)

## Description

This change will clean up old duplicate records when trying to fetch single clusters from the reporting_ocp_clusters table.

## Testing

1. Checkout Branch
2. Restart Koku
3. add ocp cluster
4. Add additional bad/dup entry in the DB
5. Update the cluster_alias entry (source name for your real ocp source) OR manually trigger 'populate_cluster_table' method

## Notes

...

### DIFF
--- a/koku/masu/database/ocp_report_db_accessor.py
+++ b/koku/masu/database/ocp_report_db_accessor.py
@@ -1004,7 +1004,7 @@ class OCPReportDBAccessor(SQLScriptAtomicExecutorMixin, ReportDBAccessorBase):
             LOG.info(log_json(msg="fetching entry in reporting_ocp_cluster", provider_uuid=provider.uuid))
             clusters = OCPCluster.objects.filter(provider_id=provider.uuid)
             if clusters.count() > 1:
-                clusters_to_delete = clusters.exclude(cluster_alias=provider.name)
+                clusters_to_delete = clusters.exclude(cluster_alias=cluster_alias)
                 LOG.info(
                     log_json(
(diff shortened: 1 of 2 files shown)


koku/masu/database/ocp_report_db_accessor.py
```diff
@@ -1002,7 +1002,21 @@ def populate_cluster_table(self, provider, cluster_id, cluster_alias):
...
+                            log_json(
+                                msg="attempting to delete duplicate entry in reporting_ocp_cluster",
+                                provider_uuid=provider.uuid,
+                            )
+                        )
+                        cluster_iter.delete()
+                    else:
+                        cluster = cluster_iter
+            else:
+                cluster = clusters.first()
```
maskarb, line 1019: Of course, there are always many ways to do something, but we can simplify this a bit by utilizing the ORM:
```
            if clusters.count() > 1:
                clusters_to_delete = clusters.exclude(cluster_alias=provider.name)
                LOG.info(
                    log_json(
                        msg="attempting to delete duplicate entry in reporting_ocp_cluster",
                        provider_uuid=provider.uuid,
                    )
                )
                clusters_to_delete.delete()

            cluster = clusters.first()
```

This way, there is less branching, and more of making the db do all the work here.

cgoodfred (approved):

cgoodfred (approved):

PR #4495, opened 2023-06-22
[COST-1142] Exact filter matching

lcouzens:
## Jira Ticket

[COST-1142](https://issues.redhat.com/browse/COST-1142)

## Description

This change will add the filter option to exact match on data types

## Testing

1. Checkout Branch
2. Restart Koku
3. make create/load test customer data
4. Go to any API endpoint and try adding filter[exact:something]=EXACT_NAME
5. The returned data should only include exact matches 

## Notes
Project match example: http://localhost:8000/api/cost-management/v1/reports/openshift/costs/?group_by[project]=*&filter[exact:project]=openshift-ocp_prod

Side point, this DOES NOT add support for Exact tag matching. Thats a bigger feat and we should create a new issue specifically if we want to try and tackle that.

This also misses out project support on the OCP on ALL endpoint. For some reason that endpoint does not default to the `OCPALLCostLineItemProjectDailySummaryP` table, this means the namespace filtering fails due to the table having an array. We _could_ switch to this table instead, BUT there is a bunch of fields missing that will likely break other things. We can add support later if we want but again I suggest creating a follow up issue since its going to be a large piece of work. Plus the UI today doesn't even make use of these anyway.
...


koku/api/report/test/all/openshift/test_query_handler.py
```diff
@@ -218,6 +218,15 @@ def test_exclude_functionality(self):
...
+        exclude_opts.remove("account")
+        exclude_opts.remove("az")
+        exclude_opts.remove("instance_type")
+        exclude_opts.remove("region")
+        exclude_opts.remove("service")
+        exclude_opts.remove("storage_type")
+        exclude_opts.remove("product_family")
+        exclude_opts.remove("cluster")
+        exclude_opts.remove("node")
         for exclude_opt in exclude_opts:
```
maskarb, line 230: What does this test do? Doesn't this make `exclude_opts` an empty list?

lcouzens:
100% smokes passed on previous run

koku/api/report/test/all/openshift/test_query_handler.py
```diff
@@ -289,6 +250,10 @@ def test_multi_exclude_functionality(self):
         """Test that the exclude feature works for all options."""
         exclude_opts = list(OCPAllExcludeSerializer._opfields)
         exclude_opts.remove("source_type")
+        exclude_opts.remove("account")
+        exclude_opts.remove("az")
+        exclude_opts.remove("instance_type")
+        exclude_opts.remove("storage_type")
```
samdoran, line 256: A comprehension is more efficient for filtering a list.
```suggestion
        exclude = {"source_type", "account", "az", "instance_type", "storage_type"}
        exclude_opts = [opfield for opfield in OCPAllExcludeSerializer._opfields if opfield not in exclude]
```

koku/api/report/test/aws/openshift/test_ocp_aws_query_handler.py
```diff
@@ -487,10 +487,10 @@ def test_exclude_functionality(self):
         """Test that the exclude feature works for all options."""
         exclude_opts = list(OCPAWSExcludeSerializer._opfields)
         # az needed to be tested separate cause
-        # no info was available in the response for az
+        # no info was available in the response for az, inbstance_type, stoprage_type
```
samdoran, line 490:
```suggestion
        # no info was available in the response for az, instance_type, storage_type
```

maskarb (approved):

samdoran (approved):

koku/api/query_filter.py
```diff
@@ -51,6 +51,8 @@ def __init__(
         self.logical_operator = "or"
         if logical_operator:
             self.logical_operator = logical_operator
+        if logical_operator == "exact":
+            self.operation = "exact"
```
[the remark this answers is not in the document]
maskarb: maybe, but you should see how we call this thing...

PR #4500, opened 2023-06-26
Add development docs for cost model.

myersCody:
## Jira Ticket

This is to start a section for development docs related to cost model. Feel free to push up any changes or suggestions you may have to this branch. I just wanted to get some quick ideas down. 

...


lcouzens (changes requested):
I either overlooked this or we are missing tag based rates too. Do you think its worth including that here too? This is looking great otherwise!

myersCody:
> I either overlooked this or we are missing tag based rates too. Do you think its worth including that here too? This is looking great otherwise!

I agree that there are other elements of the cost model that could for sure be captured in this doc. However, my intention behind this PR is to get this doc started. I was a bit worried it would turn into me having to document all the features of cost models, so I chose a specific place to start. However, I am okay with opening a tech debt to capture other elements that we want in this doc, but I also don't want to spend all my time documenting features we added to the project 3 years ago 😂

lcouzens (approved):

PR #4502, opened 2023-06-27
COST-3647: Add DVO ignore annotations as XL workers may not always be at 3+ replicas

chambridge:
* Add DVO ignore annotations as XL workers may not always be at 3+ replicas

## Jira Ticket

[COST-3647](https://issues.redhat.com/browse/COST-3647)


maskarb (approved):
lgtm!

PR #4507, opened 2023-06-28
[COST-3468] Adding max group by options

lcouzens:
## Jira Ticket

[COST-3468](https://issues.redhat.com/browse/COST-3468)

## Description

This change will limit the number of group bys supported

## Testing

1. Checkout Branch
2. Restart Koku
3. load test customer data
4. hit any API endpoint with more than 2 group_by options.

## Notes
Today you _can_ use more than two group_bys but this tends to build expensive queries and often gives weird unreliable results. Also the UI only supports a single group by today, so two is still useful via the API I could of made this one. :rofl: 
...


lcouzens:
Smokes passed 100%

myersCody (approved):

PR #4526, opened 2023-06-30
[COST-3978] Modify RBAC settings to handle read & write for new apis

myersCody:
## Jira Ticket

[COST-3978](https://issues.redhat.com/browse/COST-3978)

## Description

This will deprecate the current access class for the DDF settings, and create a new permission class that will accept read and write access. 

## Testing
Follow along with the [unittests ](https://github.com/project-koku/koku/compare/COST-3978-settings-rbac#diff-7439a1a743ab8e913c409acc9f9d40a071360ecdbb0499fa688f95b603446310R70-R120) to set your permissions locally using the DEVELOPMENT_IDENTITY and hit the corresponding endpoints to confirm expected behavior.

## Notes

...


myersCody:
successful smokes: https://ci.ext.devshift.net/job/project-koku-koku-pipeline/958/

koku/api/settings/test/aws_category_keys/test_view.py
```diff
@@ -64,3 +65,56 @@ def test_put_bad_request(self):
...
         self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+
+
+class SettingsAWSCategoryRBACTest(TestAwsCategoryClass):
+    """Test case for RBAC permissions to access settings."""
+
+    NO_ACCESS = {"aws.account": {"read": ["*"]}, "aws.organizational_unit": {"read": ["*"]}}
+    READ = {"settings": {"read": ["*"]}}
+    WRITE = {"settings": {"write": ["*"]}}
+    READ_WRITE = {"settings": {"read": ["*"], "write": ["*"]}}
```
samdoran, line 76: These are local variables not globals, so it may make more sense to keep them as lowercase.

koku/api/common/permissions/settings_access.py
```diff
@@ -11,6 +12,32 @@ class SettingsAccessPermission(permissions.BasePermission):
...
+        """Check permission based on the defined access."""
+        if settings.ENHANCED_ORG_ADMIN and request.user.admin:
+            return True
+
+        if not request.user.access:
+            return False
+
+        if request.method in permissions.SAFE_METHODS:
+            setting_read = request.user.access.get(self.resource_type, {}).get("read", [])
+            if setting_read:
```
samdoran, line 25: nit: an assignment expression could be used here.
```suggestion
            if setting_read := request.user.access.get(self.resource_type, {}).get("read", []):
```

koku/api/common/permissions/settings_access.py
```diff
@@ -11,6 +12,32 @@ class SettingsAccessPermission(permissions.BasePermission):
...
+                return True
+        else:
+            setting_write = request.user.access.get(self.resource_type, {}).get("write", [])
+            if "*" in setting_write:
```
samdoran, line 29:
```suggestion
            if "*" in setting_write := request.user.access.get(self.resource_type, {}).get("write", []):
```

samdoran (approved):

koku/api/common/permissions/settings_access.py
myersCody:
```
    if "*" in setting_write := request.user.access.get(self.resource_type, {}).get("write", []):
       ^
SyntaxError: cannot use assignment expressions with comparison
```

koku/api/common/permissions/settings_access.py
myersCody: Since we are just returning true, I don't think we need the assignment actually. 

```
if request.user.access.get(self.resource_type, {}).get("read", []):
```
Should work just find for us.

koku/api/settings/test/aws_category_keys/test_view.py
myersCody: Valid point, let me change this.

koku/api/common/permissions/settings_access.py
samdoran: Ah, that's right. I forgot about some of the weirdness around assignment expressions. Need the wrap the expression in parenthesis.
```python
if "*" in (setting_write := request.user.access.get(self.resource_type, {}).get("write", []))
    ...
```

koku/api/common/permissions/settings_access.py
samdoran: Absolutely. This is much better.

koku/api/user_settings/views.py
```diff
@@ -42,6 +44,7 @@ def __init__(self, message):
 @dataclass
 class SettingParamsHandler:
     setting: str
+    request: object
```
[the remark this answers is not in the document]
samdoran: The `get_param` and `update_param` are `Callable`.
```python
import typing as t
# ...
     get_param: t.Callable = field(init=False)
     update_param: t.Callable = field(init=False)
```

koku/api/user_settings/views.py
```diff
@@ -56,38 +59,43 @@ def __post_init__(self):
...
+        serializer = UserSettingUpdateCurrencySerializer(schema_name, self.request.data)
         if serializer.is_valid(raise_exception=True):
-            set_currency(schema_name, request_data.get(self.setting))
+            set_currency(schema_name, self.request.data.get(self.setting))
             invalidate_view_cache_for_tenant_and_all_source_types(schema_name)
             return Response(status=status.HTTP_204_NO_CONTENT)
 
-    def retrieve_user_settings(self, request):
-        users_setting = {"settings": {self.setting: self.get_param(request)}}
+    def retrieve_user_settings(self):
```
samdoran, line 76: Maybe use `get_` instead of `retrieve_`.
```suggestion
    def get_user_settings(self):
```

koku/api/user_settings/views.py
```diff
@@ -56,38 +59,43 @@ def __post_init__(self):
...
+        return [permission() for permission in permission_classes]
 
     def get(self, request, *args, **kwargs):
         """Gets a list of users current settings."""
         if not kwargs:
             user_settings = get_account_settings(request)
         else:
-            param_handler = SettingParamsHandler(kwargs["setting"])
-            user_settings = param_handler.retrieve_user_settings(request).data
+            param_handler = SettingParamsHandler(kwargs["setting"], request)
```
samdoran, line 97: Are we guaranteed to have a `settings` key in `kwargs`? There may be a potential for a `KeyError` here.

koku/api/user_settings/test/test_views.py
```diff
@@ -135,3 +136,75 @@ def test_no_userset_cost_type(self):
...
+
+    no_access = {"aws.account": {"read": ["*"]}, "aws.organizational_unit": {"read": ["*"]}}
+    read = {"settings": {"read": ["*"]}}
+    write = {"settings": {"write": ["*"]}}
+    read_write = {"settings": {"read": ["*"], "write": ["*"]}}
+
+    # deprecated permissions
+    @RbacPermissions(no_access)
+    def test_no_access_to_get_request(self):
+        for acct_setting in self.account_settings.keys():
```
samdoran, line 157: Keys are iterated over by default. There's no need to get the keys view (and it saves a couple of calls).
```suggestion
        for acct_setting in self.account_settings:
```

koku/api/user_settings/test/test_views.py
```diff
@@ -135,3 +136,75 @@ def test_no_userset_cost_type(self):
...
+            with self.subTest(acct_setting):
+                url = self.base_url + f"{acct_setting}/"
+                client = APIClient()
+                response = client.get(url, **self.headers)
+                self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+    @RbacPermissions(read)
+    def test_read_accesss_to_get_request(self):
+        for acct_setting in self.account_settings.keys():
```
samdoran, line 166:
```suggestion
        for acct_setting in self.account_settings:
```

koku/api/user_settings/test/test_views.py
```diff
@@ -135,3 +136,75 @@ def test_no_userset_cost_type(self):
...
+                url = self.base_url + f"{acct_setting}/"
+                client = APIClient()
+                data = {acct_setting.replace("-", "_"): value}
+                response = client.put(url, data, format="json", **self.headers)
+                self.assertEqual(response.status_code, status.HTTP_204_NO_CONTENT)
+
+    # Deprecated permissions
+    @RbacPermissions(write)
+    def test_write_on_get_request(self):
+        for acct_setting in self.account_settings.keys():
```
samdoran, line 196:
```suggestion
        for acct_setting in self.account_settings:
```

koku/api/user_settings/test/test_views.py
```diff
@@ -135,3 +136,75 @@ def test_no_userset_cost_type(self):
...
+            with self.subTest(acct_setting=acct_setting):
+                url = self.base_url + f"{acct_setting}/"
+                client = APIClient()
+                response = client.get(url, **self.headers)
+                self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+    @RbacPermissions(read_write)
+    def test_read_and_write_on_get_request(self):
+        for acct_setting in self.account_settings.keys():
```
samdoran, line 205:
```suggestion
        for acct_setting in self.account_settings:
```

koku/api/user_settings/views.py
```diff
@@ -56,38 +59,43 @@ def __post_init__(self):
...
         return Response(users_setting)
 
 
 class AccountSettings(APIView):
     """Settings views for all user settings."""
 
-    permission_classes = [permissions.AllowAny]
+    def get_permissions(self):
+        if self.request.method.lower() == "get":
+            permission_classes = [DeprecatedSettingsAccessPermission]
```
samdoran, line 87: Is there a more precise way to tell when the request is coming from the DDF settings page?  Any indicator in the request? Keying off the request method seems like it could lead to using the incorrect settings class.

samdoran (commented):
Just a handful of minor things.

koku/api/user_settings/views.py
myersCody: I couldn't find a more precise way to tell other than the method. However, these are very short lived changes too. The Deprecated class is to allow backwards compatibility, while QE & UI kick the tires on the new settings api & permissions. We will remove the deprecated settings when we release the new settings. I can add it as a comment in our transition plan.

koku/api/user_settings/views.py
myersCody: Yes! If kwargs exist, we are guaranteed a setting kwarg because the way the paths are set up. 
```
path("account-settings/", AccountSettings.as_view(), name="account-settings"),
path("account-settings/<str:setting>/", AccountSettings.as_view(), name="get-account-setting"),
```

koku/api/user_settings/views.py
samdoran: Sounds good. Just wanted to pause and consider.

samdoran (approved):

PR #4544, opened 2023-07-06
Constrain kombu

samdoran:
## Description

Kombu 5.3 fixed bugs that incorrectly cast UUID and datetime objects to strings. We expect those to be strings in many places.

https://github.com/celery/kombu/pull/1575
https://github.com/celery/kombu/pull/1515
https://github.com/celery/kombu/releases/tag/v5.3.0

Closes #4539.


maskarb:
I think we should update kombu instead. That is why this was opened:
https://issues.redhat.com/browse/COST-3997.

PR #4548, opened 2023-07-07
[COST-2665] Clowder now applying annotations from deployment spec level

chambridge:
## Jira Ticket

[COST-2665](https://issues.redhat.com/browse/COST-2665)

## Description

This change will ...

* Moved annotations up to spec level
* Left existing pod level annotations as it seems to have applied for older versions

See: https://github.com/RedHatInsights/clowder/pull/729

## Testing

1. Deploy PR to ephemeral environment with bonfire
```
bonfire namespace reserve --duration 24h
bonfire deploy hccm --set-template-ref koku=cost-2665-kube-lint-ignore --set-parameter koku/IMAGE_TAG=pr-4548-latest   --no-remove-resources koku \
  --no-remove-resources presto \
  --no-remove-resources hive-metastore
```
2. Note annotations on deployments.

### DIFF
--- a/deploy/clowdapp.yaml
+++ b/deploy/clowdapp.yaml
@@ -412,7 +412,11 @@ objects:
           enabled: false
         public:
           enabled: false
-    - name: clowder-masu
+    - metadata:
+        annotations:
+          ignore-check.kube-linter.io/minimum-three-replicas: This deployment uses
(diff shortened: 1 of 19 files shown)


chambridge:
Annotation example on deployment post bonfire deploy to ephemeral:
image

lcouzens (approved):

PR #4552, opened 2023-07-07
[COST-3984] Do not update summary tables if there are no AWS bills

samdoran:
## Jira Ticket

[COST-3984](https://issues.redhat.com/browse/COST-3984)

## Description

Skip updating the AWS summary tables if we don't have AWS bill data.

## Testing
 
TBD

## Notes

Instead of stopping _all_ processing, only skip the summary tables update. The markup cost and UI summary tables will still be updated. We had some discussion and are thinking that will be ok.

### DIFF
--- a/koku/masu/processor/ocp/ocp_cloud_parquet_summary_updater.py
+++ b/koku/masu/processor/ocp/ocp_cloud_parquet_summary_updater.py
@@ -261,7 +261,7 @@ class OCPCloudParquetReportSummaryUpdater(PartitionHandlerMixin, OCPCloudUpdater
                 end_date,
             )
 
-            aws_bill_ids = [str(bill.id) for bill in aws_bills]
+            aws_bill_ids = [bill.id for bill in aws_bills]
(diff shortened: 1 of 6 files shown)


koku/masu/processor/ocp/ocp_cloud_parquet_summary_updater.py
```diff
@@ -247,8 +261,8 @@ def update_aws_summary_tables(self, openshift_provider_uuid, aws_provider_uuid,
                 end_date,
             )
 
-            aws_bill_ids = [str(bill.id) for bill in aws_bills]
-            current_aws_bill_id = aws_bills.first().id if aws_bills else None
+            aws_bill_ids = [bill.id for bill in aws_bills]
+            current_aws_bill_id = aws_bill_ids[0] if aws_bills else None
```
maskarb, line 265:
```suggestion
            current_aws_bill_id = aws_bill_ids[0]
```
The if else None can be removed now because of the check above.

PR #4556, opened 2023-07-10
[COST-3890/3892] Data Extraction and Kafka Messaging for SUBS

cgoodfred:
## Jira Ticket

[COST-3890](https://issues.redhat.com/browse/COST-3890)
[COST-3892](https://issues.redhat.com/browse/COST-3892)

Also addresses and ask from SWATCH to send product ids appropriately based on tags
[COST-4062](https://issues.redhat.com/browse/COST-4062)

## Description

This change will add identification of RHEL metering records and processing of those records to S3, followed by another task to process S3 reports to a kafka message on a different worker.

## Testing

1. Checkout Branch
2. Restart Koku and include a subs worker `make docker-up-min-with-subs`
3. Create and Load test customer data, the AWS source yaml was updated to contain a specific tag on an EC2 generator that causes the record to be identified as RHEL metered. This tag is case insensitive. REQUIRED TAG FOR IDENTIFICATION: `resourceTags/user:com_redhat_rhel: true`
4. in the logs you should see a message that says `beginning subs rhel extraction`, this indicates it is attempting to process the AWS report and identify relevant records.
5. There will be a following log of `1 file(s) uploaded to s3 for rhel metering` if relevant records were found.
6. Once the process to kafka task has run, you will see a log of `iterating over records and sending kafka messages` followed by a count of the number of messages sent `sent 1438 kafka messages for subs`.

## Notes
LOCAL NOTES:
- Locally the bucket uploaded to is the regular minio koku-bucket. the path will be something like this with the org as a top level path in the bucket: 
```
koku-bucket/org1234567/AWS/source=fee1eb86-d751-49c4-a8ff-be1a5d2f48ea/date=2023-07-17/subs_9281881a-2963-48f4-a27b-6f5a5397bfe3_0.csv
```
- Unless you spin up kafka, you will see an error related to connecting to kafka but it will not prevent processing


cgoodfred:
Smokes run: https://ci.ext.devshift.net/job/project-koku-koku-pipeline/1362/

Ran the failures by QE and they are unrelated.

koku/masu/prometheus_stats.py
```diff
@@ -136,10 +136,9 @@
     registry=WORKER_REGISTRY,
     multiprocess_mode="livesum",
 )
-
 SUBS_TRANSMISSION_BACKLOG = Gauge(
     "subs_transmission_backlog",
-    "Number of celery tasks in the SUBS Data Transmission queue",
+    "Number of celery tasks in the SUBS Data Tranmission queue",
```
maskarb, line 141:
```suggestion
    "Number of celery tasks in the SUBS Data Transmission queue",
```

koku/masu/processor/orchestrator.py
```diff
@@ -306,7 +308,8 @@ def start_manifest_processing(  # noqa: C901
                 summary_task = summarize_reports.s(
                     manifest_list=manifest_list, ingress_report_uuid=self.ingress_report_uuid
                 ).set(queue=SUMMARY_QUEUE)
-                async_id = chord(report_tasks, group(summary_task, hcs_task))()
+                subs_task = collect_subs_report_data_from_manifest.s().set(queue=SUBS_EXTRACTION_QUEUE)
+                async_id = chord(report_tasks, group(summary_task, hcs_task, subs_task))()
```
maskarb, line 312: I'm wondering if we need to do this, or if this can be a scheduled task instead?

koku/subs/tasks.py
```diff
@@ -0,0 +1,117 @@
...
+    schema_name = convert_account(schema_name)
+    context = {"schema_name": schema_name}
+    LOG.info(log_json(msg="enable_subs_messaging context", context=context))
+    return bool(
+        UNLEASH_CLIENT.is_enabled("cost-management.backend.subs-data-messaging", context) or settings.ENABLE_SUBS_DEBUG
+    )
+
+
+@celery_app.task(name="subs.tasks.collect_subs_report_data_from_manifest", queue=SUBS_EXTRACTION_QUEUE)
+def collect_subs_report_data_from_manifest(reports_to_subs_summarize):
```
maskarb, line 68: Given that this task goes to the `EXTRACTION` queue, is this task "extracting" data?

```suggestion
def extract_subs_data_from_reports(reports_to_extract):
```

koku/subs/tasks.py
```diff
@@ -0,0 +1,117 @@
...
+                LOG.debug(log_json(tracing_id, msg="skipping report, no manifest found.", context=context))
+                continue
+            start_date, end_date = date_tuple
+        if start_date is None:
+            start_date = DateAccessor().today() - datetime.timedelta(days=2)
+        if end_date is None:
+            end_date = DateAccessor().today()
+        if isinstance(start_date, str):
+            start_date = datetime.datetime.strptime(start_date, "%Y-%m-%d")
+            end_date = datetime.datetime.strptime(end_date, "%Y-%m-%d")
```
maskarb, line 105: I still don't like any of this data logic here. Given a report, we should have a function that accepts that report and returns the start and end dates that we require. All this logic here is messy and hard to follow.

koku/subs/subs_data_messenger.py
```diff
@@ -0,0 +1,103 @@
...
+                LOG.info(
+                    log_json(
+                        self.tracing_id,
+                        msg="iterating over records and sending kafka messages",
+                        context=self.context,
+                    )
+                )
+                msg_count = 0
+                for row in reader:
+                    msg = self.build_subs_msg(
```
maskarb, line 54: I suppose this is fine for now, but are these the columns we will be getting for Azure and GCP too?

maskarb (commented):
some comments:

koku/subs/subs_data_extractor.py
```diff
@@ -0,0 +1,170 @@
...
+from reporting.provider.aws.models import TRINO_LINE_ITEM_TABLE as AWS_TABLE
+
+LOG = logging.getLogger(__name__)
+
+TABLE_MAP = {
+    Provider.PROVIDER_AWS: AWS_TABLE,
+}
+
+
+def get_subs_s3_client():  # pragma: no cover
```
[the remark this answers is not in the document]
cgoodfred: This is a good point, [I updated the masu utils function](https://github.com/project-koku/koku/pull/4556/files#diff-d4383ec8727dc9a534391503b1149fbbc2559d181d979fd89129bd41cf7e0a87L388) to take arguments and its related function calls and used that here.

koku/subs/subs_data_extractor.py
```diff
@@ -0,0 +1,170 @@
...
+        )
+
+    def update_latest_processed_time(self, year, month, end_time):
+        """Update the latest processing time for a provider"""
+        with schema_context(self.schema):
+            subs_obj = SubsLastProcessed.objects.filter(
+                source_uuid=self.provider_uuid, year=year, month=month
+            ).first() or SubsLastProcessed.objects.create(
+                source_uuid=TenantAPIProvider.objects.get(uuid=self.provider_uuid), year=year, month=month
+            )
```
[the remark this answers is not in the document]
cgoodfred: I was thinking about the times we had run into timing issues with get_or_create in the past, but since I made the `source_uuid`, `year` and `month` unique together, that should be fine. Updated to use `get_or_create` here.

As for the `TenantAPIProvider` lookup, I had forgotten that inserting with Foreign keys using django models was different and `source_uuid` for the insert required a reference to the object but `source_uuid_id` could be the matching UUID. I have updated that accordingly.

koku/subs/subs_data_extractor.py
```diff
@@ -0,0 +1,170 @@
...
+
+    def extract_data_to_s3(self, start_date, end_date):
+        """Process new subs related line items from reports to S3."""
+        LOG.info(log_json(self.tracing_id, msg="beginning subs rhel extraction", context=self.context))
+        month = start_date.strftime("%m")
+        year = start_date.strftime("%Y")
+        # if there is no latest time for this month, we need to include the first hour of the month
+        latest_timestamp = self.determine_latest_processed_time_for_provider(year, month) or (
+            datetime(start_date.year, start_date.month, start_date.day) - timedelta(hours=1)
+        )
```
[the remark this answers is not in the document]
cgoodfred: I was operating under the assumption that if there was not a `last_processed_time` it was due to initial ingestion, that was a poor assumption and I have adjusted the date logic to always pass in the month start because if there is not a latest processed time for the month, we want to process the entire month. 

The reason for the `- timedelta(hours=1)` is the query is just GT not GTE so we don't have to manipulate the latest timestamp, so we go back one hour to the previous month as the timestamp so the 0th hour of the current month is checked. I have updated the comment here as well to be more descriptive.

koku/subs/subs_data_messenger.py
```diff
@@ -0,0 +1,103 @@
...
+LOG = logging.getLogger(__name__)
+
+
+class SUBSDataMessenger:
+    def __init__(self, context, schema_name, tracing_id):
+        self.context = context
+        self.tracing_id = tracing_id
+        self.schema_name = schema_name
+        self.s3_client = get_subs_s3_client()
+        subs_cust = Customer.objects.filter(schema_name=schema_name).first()
```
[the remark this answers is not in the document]
cgoodfred: Switched to using `Customer.objects.get(schema_name=schema_name)` here.

koku/masu/prometheus_stats.py
cgoodfred: Fixed the wording.

koku/subs/tasks.py
cgoodfred: Updated name appropriately.

koku/subs/tasks.py
cgoodfred: I have reworked the date logic into a standalone function and reduced the complexity heavily, we now only get the start of the month associated with the manifest or start date and process based on month start since we either need to process all of the current month or the latest timestamp.

koku/subs/subs_data_messenger.py
cgoodfred: We will need different columns when it comes time for Azure/GCP, I think for now this function can stay as is and when its time to add another provider, we can take into account what field names to use.

koku/masu/processor/orchestrator.py
cgoodfred: I'm not sure I'm in support of a scheduled task here. I think having it be a part of the tasks triggered after a download handles initial ingestion of a new source better since we process multiple months here.

koku/subs/subs_data_extractor.py
```diff
@@ -0,0 +1,160 @@
...
+            log_json(
+                self.tracing_id,
+                msg=f"identified {total_count} matching records for metered rhel",
+                context=self.context,
+            )
+        )
+        upload_keys = []
+        filename = f"subs_{self.tracing_id}_"
+        sql_file = f"trino_sql/{self.provider_type.lower()}_subs_summary.sql"
+        query_sql = pkgutil.get_data("subs", sql_file)
```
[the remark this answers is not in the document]
cgoodfred: updated to follow this suggestion and `log_ref` updated to just be the SQL file

koku/subs/tasks.py
maskarb: Sweet, thanks.

djnakabaale (commented):
tested locally and things are looking good 👍🏾

here are some comments:

djnakabaale (approved):
Thanks addressing all the feedback! lgtm 👍🏾

PR #4561, opened 2023-07-11
[COST-4015] Dynamially set large customers based on provider count

lcouzens:
## Jira Ticket

[COST-4015](https://issues.redhat.com/browse/COST-4015)

## Description

This change will dynamically set if a customer/schema should be classed as large or not.

## Testing

1. Checkout Branch
2. Restart Koku
3. load test customer data
4. hit the download endpoint
5. select * from api_customer; and see large_customer column set to True
6. update LARGE_CUSTOMER_PROVIDER_COUNT in Config above your provider count
7. hit download again
8. select * from api_customer; and see large_customer column set to False

## Notes

...


lcouzens:
This was just a POC and can be revisited in the future.

PR #4565, opened 2023-07-12
[COST-4012] Don't check serviceinfo2 for Azure HCS filter

cgoodfred:
## Jira Ticket

[COST-4012](https://issues.redhat.com/browse/COST-4012)

## Description

This change will remove the check of serviceinfo2 from the Azure filter as it is a legacy field and the information we care about is being matched via other criteria.

Opened an [MR to update the failing](https://gitlab.cee.redhat.com/insights-qe/iqe-cost-management-plugin/-/merge_requests/1690) smoke test as well since the data covered is different.

## Testing

1. Checkout Branch
2. Restart Koku
3. Create and Load the test customer data for all source types since the SQL was updated to remove the sqlsafes: `make create-test-customer` followed by `make load-test-customer-data`
4. Data should process fine and minio should have HCS files for Azure, AWS, and GCP.

## Notes

...


djnakabaale (approved):
thanks for cleaning up the sql, lgtm 👍🏾

koku/hcs/database/report_db_accessor.py
```diff
@@ -69,7 +69,7 @@ def get_hcs_daily_summary(self, date, provider, provider_uuid, sql_summary_file,
 
             sql_params = {
                 "provider_uuid": provider_uuid,
-                "year": date.year,
+                "year": date.strftime("%Y"),
                 "month": date.strftime("%m"),
```
maskarb, line 73:
```suggestion
                "year": date.year,
                "month": date.month,
```
Why don't we do this?

maskarb: Ahhh because `.month` and `.year` don't return strings. Got it.

cgoodfred: The partitions here are varchars and not integers, using `date.year` or `date.month` get integers back so the comparison fails.
```
>>> type(datetime.today().month)
<class 'int'>
```
This is how the comparison fails: 
```
trino:org1234567> select * from aws_line_items where month = 5;
Query 20230712_192853_03620_44kzd failed: line 1:42: Cannot apply operator: varchar = integer
```

cgoodfred:
The failing smoke has been addressed in an MR to the cost plugin to update the test.

PR #4570, opened 2023-07-13
[COST-3990] Add write check to user_access endpoint 

myersCody:
## Jira Ticket

[COST-3990](https://issues.redhat.com/browse/COST-3990)

## Description

This change will add the read_only to the user_access endpoint.

## Testing

1. Load test customer data.
2. Modify your RBAC identify to give yourself different access settings.
3. Hit: `http://localhost:8000/api/cost-management/v1/user-access/`
4. See expected access.

## Notes

...


myersCody:
Successful Smokes: https://ci.ext.devshift.net/job/project-koku-koku-pipeline/1252/

koku/api/user_access/view.py
```diff
@@ -28,121 +28,88 @@ class UIFeatureAccess:
...
+    def access_key_mapping(self):
+        mapping = {}
+        mapping["aws"] = ["aws.account", "aws.organizational_unit"]
+        mapping["ocp"] = ["openshift.cluster", "openshift.node", "openshift.project"]
+        mapping["azure"] = ["azure.subscription_guid"]
+        mapping["gcp"] = ["gcp.account", "gcp.project"]
+        mapping["oci"] = ["oci.payer_tenant_id"]
+        mapping["ibm"] = ["ibm.account"]
+        mapping["cost_model"] = ["cost_model"]
+        mapping["settings"] = ["settings"]
```
maskarb, line 67:
```suggestion
        mapping = {
            "aws": ["aws.account", "aws.organizational_unit"],
            "azure": ["azure.subscription_guid"],
            "gcp": ["gcp.account", "gcp.project"],
            "ibm": ["ibm.account"],
            "oci": ["oci.payer_tenant_id"],
            "ocp": ["openshift.cluster", "openshift.node", "openshift.project"],
            "cost_model": ["cost_model"],
            "settings": ["settings"],
        }
        mapping["any"] = mapping["aws"] + mapping["azure"] + mapping["gcp"] + mapping["oci"] + mapping["ocp"]
        return mapping

```

koku/api/user_access/view.py
```diff
@@ -28,121 +28,88 @@ class UIFeatureAccess:
...
-
-    access_keys = (
-        AWSUserAccess.access_keys
-        + AzureUserAccess.access_keys
-        + GCPUserAccess.access_keys
-        + OCIUserAccess.access_keys
-        + OCPUserAccess.access_keys
-    )
+        if len(self.access_key_list) == 1:
+            # For backwards compatability.
```
maskarb, line 103: What does `# For backwards compatability.` tell us? Backwards compatibility with what?

koku/api/user_access/view.py
```diff
@@ -164,26 +131,14 @@ def get(self, request, **kwargs):
...
-                admin_user, source_type.get("pre_release_feature")
-            )
-            data.append({"type": source_type.get("type"), "access": access_granted})
+        ui_feature_access = UIFeatureAccess(user_access, admin_user)
+        is_valid_param = ui_feature_access.set_access_key_list(query_params.get("type"))
+        if not is_valid_param:
+            return Response({f"Unknown source type: {query_params.get('type')}"}, status=status.HTTP_400_BAD_REQUEST)
+        data = ui_feature_access.generate_data()
+        if isinstance(data, dict):
+            # backwards compatability
```
maskarb, line 140: Same question as above. Backwards compatible with what?

koku/api/user_access/view.py
```diff
@@ -28,121 +28,88 @@ class UIFeatureAccess:
...
     """
 
-    def __init__(self, access):
+    PRERELEASE = ["ibm"]
+
+    def __init__(self, access, admin_user, access_key_list=None):
         """Class Constructor.
 
         Args:
             access (dict) - an RBAC dict; see: koku.koku.middleware.IdentityHeaderMiddleware
```
maskarb, line 40: this list needs to be updated.

koku/api/user_access/view.py
```diff
@@ -28,121 +28,88 @@ class UIFeatureAccess:
...
 
-    def access(self, admin_user, pre_release_feature):
+    def _check_access(self, pre_release_feature, access_key):
         """Access property returns whether the user has the requested access.
 
         Return:
-            (bool)
+            (access, read_only)
+            acesss = bool or None
+            read_only = bool or none
```
maskarb, line 81: These returns are a bit confusing. Why would `read_only` ever be `None`? Also, `access` never returns `None`.

koku/api/user_access/view.py
```diff
@@ -28,121 +28,88 @@ class UIFeatureAccess:
...
+            if self._get_access_value(rbac_setting, "write"):
+                return True, False
+            elif self._get_access_value(rbac_setting, "read"):
+                return True, True
+        return False, None
+
+    def generate_data(self):
+        data = []
+        for access_key in self.access_key_list:
+            pre_release_feature = bool(access_key in self.PRERELEASE)
```
maskarb, line 98:
```suggestion
            pre_release_feature = access_key in self.PRERELEASE
```
This type-cast is redundant.

koku/api/user_access/view.py
```diff
@@ -28,121 +28,88 @@ class UIFeatureAccess:
...
-
-        access_keys (list) - a list of access keys to check. must be implemented by sub-classes.
-                             multiple keys will be ORed; i.e. if a user has access to anything in the list,
-                             this API returns True
     """
 
-    def __init__(self, access):
+    PRERELEASE = ["ibm"]
+
+    def __init__(self, access, admin_user, access_key_list=None):
```
maskarb, line 35: Where do we pass `access_key_list` into this init?

koku/api/user_access/view.py
```diff
@@ -28,121 +28,88 @@ class UIFeatureAccess:
...
 
         Args:
             access (dict) - an RBAC dict; see: koku.koku.middleware.IdentityHeaderMiddleware
 
         """
+        if not access_key_list:
+            self.access_key_list = self.access_key_mapping.keys()
         self.access_dict = access if access else {}
+        self.admin_user = admin_user
+        self.access_key_list = []
```
maskarb, line 46: I'm a little confused here. I can't seem to find where we pass in `access_key_list`, so the first `if not access_key_list` is always true, then we set `self.access_key_list = self.access_key_mapping.keys()`. But then right here we do `self.access_key_list = []`. We set, then immediately unset this value.

koku/api/user_access/view.py
```diff
@@ -164,26 +131,14 @@ def get(self, request, **kwargs):
...
-
-        data = []
-        for source_type in self._source_types:
-            access_granted = False
-            access_granted = source_type.get("access_class")(user_access).access(
-                admin_user, source_type.get("pre_release_feature")
-            )
-            data.append({"type": source_type.get("type"), "access": access_granted})
+        ui_feature_access = UIFeatureAccess(user_access, admin_user)
+        is_valid_param = ui_feature_access.set_access_key_list(query_params.get("type"))
```
maskarb, line 135: I think the naming convention here is confusing. I think we need a validating method that determines if `query_params.get("type")` is valid. I think this `set_access_key_list` method is trying to be too clever.

you could define:
```
def validate_type(self, query_param):
   if not query_param:
      return True
   return query_param in self.access_key_mapping
```
which can then be used:
```
ui_feature_access = UIFeatureAccess(user_access, admin_user)
if param_type := query_params.get("type"):
   if not ui_feature_access.validate_type(param_type)
       return Response({f"Unknown source type: {query_params.get('type')}"}, status=status.HTTP_400_BAD_REQUEST)
   ui_feature_access.update_access_key_list(param_type)
...
```

maskarb (commented):
It's nice to see a lot of these unnecessary classes cleaned up!

Lot's of feedback here:

koku/api/user_access/view.py
maskarb: It could also be worth making this dict a constant in this file. As a class property, this thing has to be recreated every single time the property is accessed which seems unnecessary.

koku/api/user_access/view.py
maskarb: In this case, `no access` is the same as `False`.

koku/api/user_access/view.py
myersCody: I missed this in my clean up 🙇

koku/api/user_access/view.py
maskarb: Actually, now I'm confused, why does `read-only: True` equate to write access?

myersCody: Sorry, I have been thinking about his since I commented it and chatting with Dan about it. I think the real requirement here is to understand if the user has write access. So, I changed the PR to `write` which will make this boolean make way more sense.

koku/api/user_access/view.py
```diff
@@ -18,6 +18,36 @@
...
+    "cost_model": ["cost_model"],
+    "settings": ["settings"],
+}
+ACCESS_KEY_MAPPING["any"] = (
+    ACCESS_KEY_MAPPING["aws"]
+    + ACCESS_KEY_MAPPING["azure"]
+    + ACCESS_KEY_MAPPING["gcp"]
+    + ACCESS_KEY_MAPPING["oci"]
+    + ACCESS_KEY_MAPPING["ocp"]
+)
```
samdoran, line 49: Modifying a global on every import seems messy. It might be cleaner to change the global to a frozen dataclass instance.

```python
@dataclasses.dataclass
class AccessMapping:
    aws: tuple[str, str] = ("aws.account", "aws.organizational_unit")
    azure: tuple[str] = ("azure.subscription_guid",)
    gcp: tuple[str, str] = ("gcp.account", "gcp.project")
    ibm: tuple[str] = ("ibm.account",)
    oci: tuple[str] = ("oci.payer_tenant_id",)
    ocp: tuple[str, str, str] = ("openshift.cluster", "openshift.node", "openshift.project")
    cost_model: tuple[str] = ("cost_model",)
    settings: tuple[str] = ("settings",)
    any: tuple[str, ...] = None

    def __post_init__(self):
        result = []
        for field in dataclasses.fields(self):
            if field.name == "any":
                continue

            result.extend(getattr(self, field.name))

        self.any = tuple(result)

    def __getitem__(self, item):
        return getattr(self, item)

    def get(self, item, default=None):
        return getattr(self, item, default)

ACCESS_KEY_MAPPING = AccessMapping()
```

koku/api/user_access/view.py
```diff
@@ -28,121 +58,74 @@ class UIFeatureAccess:
...
+
+    def __init__(self, access, admin_user):
         """Class Constructor.
 
         Args:
             access (dict) - an RBAC dict; see: koku.koku.middleware.IdentityHeaderMiddleware
+            admin_user (boolean) - Indicates if the user is an admin
 
         """
+        self.access_key_list = ACCESS_KEY_MAPPING.keys()
```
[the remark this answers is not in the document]
samdoran: If this is a dataclass, you can convert it to a dict to easy get a list of fields.
```suggestion
        self.access_key_list = list(dataclasses.asdict(ACCESS_KEY_MAPPING))
```

samdoran (changes requested):

koku/api/user_access/view.py
```diff
@@ -28,121 +58,74 @@ class UIFeatureAccess:
...
-
-        access_keys (list) - a list of access keys to check. must be implemented by sub-classes.
-                             multiple keys will be ORed; i.e. if a user has access to anything in the list,
-                             this API returns True
     """
 
-    def __init__(self, access):
+    PRERELEASE = ["ibm"]
+
+    def __init__(self, access, admin_user):
```
samdoran, line 65: If access is optional (it is set to an empty dict later on), it should default to `None` here.
```suggestion
    def __init__(self, admin_user, access=None):
```

myersCody: access here is coming directly from `request.user.access`. I think empty dict, is to convert the potential None value to something usable in this class. I am not 100% certain cause that line was added 2 years ago though 🤨

maskarb (dismissed):

koku/api/user_access/test/test_view.py
```diff
@@ -12,6 +12,42 @@
...
+        "settings": {"read": [], "write": []},
+    }
+    for key, value in rbac_dict.items():
+        rbac_defaults[key] = value
+    return rbac_defaults
+
+
+def build_expected_ouput(testing_dict={}, default_access=False, default_write=False):
+    """Helper function to allow you to build expected outputs bases on permissions."""
+    expected_output = []
```
samdoran, line 38:
```suggestion
def build_expected_ouput(testing_dict=None, default_access=False, default_write=False):
    """Helper function to allow you to build expected outputs bases on permissions."""
    if testing_dict is None:
        testing_dict = {}

    expected_output = []
```

koku/api/user_access/test/test_view.py
```diff
@@ -22,607 +59,255 @@ def setUp(self):
...
-        }
-    )
+    @RbacPermissions(build_rbac_permissions({"aws.account": {"read": ["*"]}}))
     def test_aws_view_read(self):
         """Test user-access view with aws read wildcard permission."""
         url = reverse("user-access")
         response = self.client.get(url, **self.headers)
+        testing_matrix = {"any": {"access": True}, "aws": {"access": True}}
+        expected_output = build_expected_ouput(testing_matrix)
+        for result in response.data.get("data"):
```
samdoran, line 69: This could be `None` resulting in a `TypeError`. It might be better to do a direct key lookup so that the test will fail with a `KeyError` is `data` is missing.

samdoran (approved):

PR #4576, opened 2023-07-14
[COST-3974] Switch data process to use new unified enabled tag table

myersCody:
## Jira Ticket

[COST-3974](https://issues.redhat.com/browse/COST-3974)

## Description

This change will switch our data processing to use new unified enable tag limit. 

## Testing

1. Load test customer data

## Notes

...


koku/masu/util/common.py
```diff
@@ -325,43 +326,61 @@ def batch(iterable, start=0, stop=None, _slice=1):
...
+                    enabled_tag_count = EnabledTagKeys.objects.filter(enabled=True).count()
+                if Config.ENABLED_TAG_LIMIT <= 0:
+                    # If the tag limit is zero or less, the limit is disabled.
+                    LOG.info(log_json(msg="tag limit is disabled", context=ctx))
+                elif enabled_tag_count < Config.ENABLED_TAG_LIMIT:
+                    LOG.info(log_json(msg="tag limit is enabled", context=ctx))
+                else:
+                    LOG.info(log_json(msg="tag limit has been reached"), content=ctx)
+                    enabled_value = False
+            ctx["tags_enagled"] = enabled_value
```
samdoran, line 370:
```suggestion
            ctx["tags_enabled"] = enabled_value
```

koku/api/query_params.py
```diff
@@ -362,9 +363,12 @@ def _set_tag_keys(self, query_params):
             # we do not need to fetch the tags for tags report type.
             # we also do not need to fetch the tags if a tag prefix is not in the URL
             return
-        for tag_model in self.tag_handler:
-            with tenant_context(self.tenant):
-                self.tag_keys.update(tag_model.objects.values_list("key", flat=True).distinct())
+        with tenant_context(self.tenant):
+            # TODO: See if we can just use an in here instead of looping through provider types.
```
myersCody, line 367:
```suggestion
```

koku/api/tags/gcp/openshift/queries.py
```diff
@@ -27,12 +26,16 @@ class OCPGCPTagQueryHandler(GCPTagQueryHandler, OCPTagQueryHandler):
         {
             "db_table": OCPGCPTagsSummary,
             "db_column_period": "cost_entry_bill__billing_period",
-            "annotations": {"enabled": Exists(GCPEnabledTagKeys.objects.filter(key=OuterRef("key")))},
+            "annotations": {
+                "enabled": Exists(EnabledTagKeys.objects.filter(provider_type="GCP").filter(key=OuterRef("key")))
```
maskarb, line 30: This is probably fine, but everywhere else, we use `provider_type=provider` where `provider = Provider.provider-type`. So maybe here we can use `provider_type=Provider.PROVIDER_GCP`?
```suggestion
                "enabled": Exists(EnabledTagKeys.objects.filter(provider_type=Provider.PROVIDER_GCP).filter(key=OuterRef("key")))
```
and same for Provider.PROVIDER_OCP below?

koku/api/tags/view.py
```diff
@@ -80,7 +81,6 @@ def get(self, request, **kwargs):
 
     def validate_key(self, key):
         """Validate that tag key exists."""
-        for handler in self.tag_handler:
-            if handler.objects.filter(key=key).exists():
-                return True
+        if EnabledTagKeys.objects.filter(provider_type__in=self.tag_providers, key=key).exists():
+            return True
         return False
```
maskarb, line 86: You can remove this branching:
```suggestion
        return EnabledTagKeys.objects.filter(provider_type__in=self.tag_providers, key=key).exists()
```

koku/masu/test/util/azure/test_azure_post_processor.py
```diff
@@ -99,7 +99,7 @@ def test_finalize_post_processing(self):
         self.post_processor.finalize_post_processing()
 
         with schema_context(self.schema):
-            tag_key_count = AzureEnabledTagKeys.objects.filter(key__in=expected_tag_keys).count()
+            tag_key_count = EnabledTagKeys.objects.filter(provider_type="Azure", key__in=expected_tag_keys).count()
```
maskarb, line 102: We need to be consistent in how we filter. In the aws file, we used `Provider.PROVIDER_AWS`, but here and GCP/OCI, we're using strings. Pick one or the other.

koku/masu/test/util/test_common.py
```diff
@@ -387,21 +386,25 @@ def test_batch_value_error(self):
 
     def test_create_enabled_keys_aws(self):
         with schema_context(self.schema):
-            orig_keys = [{"key": e.key, "enabled": e.enabled} for e in AWSEnabledTagKeys.objects.all()]
-            AWSEnabledTagKeys.objects.all().delete()
+            orig_keys = [
+                {"key": e.key, "enabled": e.enabled, "provider_type": "AWS"}
+                for e in EnabledTagKeys.objects.filter(provider_type="AWS")
+            ]
+            EnabledTagKeys.objects.filter(provider_type="AWS").delete()
```
maskarb, line 393: Same as the above comment, pick `"AWS"` or `Provider.PROVIDER_AWS` everywhere.

koku/masu/util/aws/aws_post_processor.py
```diff
@@ -31,6 +37,29 @@ def handle_user_defined_json_columns(data_frame, columns, column_prefix):
...
+            for batch_num, new_batch in enumerate(batch(new_keys, _slice=500)):
+                batch_size = len(new_batch)
+                LOG.info(
+                    log_json(msg="create batch", batch_number=(batch_num + 1), batch_size=batch_size, context=ctx)
+                )
+                for ix in range(batch_size):
+                    new_batch[ix] = enabled_keys_model(key=new_batch[ix])
+                enabled_keys_model.objects.bulk_create(new_batch, ignore_conflicts=True)
+        else:
+            LOG.info(log_json(msg="no enabled keys added", context=ctx))
```
maskarb, line 60: nitpick: but we should avoid all the logic buried in an `if` statement and the `else` just logs a message
```suggestion
        if not new_keys:
            LOG.info(log_json(msg="no enabled keys added", context=ctx))
            return
            
        LOG.info(log_json(msg="creating enabled key records", context=ctx))
        for batch_num, new_batch in enumerate(batch(new_keys, _slice=500)):
            batch_size = len(new_batch)
            LOG.info(
                log_json(msg="create batch", batch_number=(batch_num + 1), batch_size=batch_size, context=ctx)
            )
            for ix in range(batch_size):
                new_batch[ix] = enabled_keys_model(key=new_batch[ix])
            enabled_keys_model.objects.bulk_create(new_batch, ignore_conflicts=True)
```

koku/masu/util/common.py
```diff
@@ -325,43 +326,61 @@ def batch(iterable, start=0, stop=None, _slice=1):
...
+                log_json(msg="create tag batch", batch_number=(batch_num + 1), batch_size=batch_size, context=ctx)
+            )
+            for ix in range(batch_size):
+                new_batch[ix] = EnabledTagKeys(key=new_batch[ix], provider_type=provider_type, enabled=enabled_value)
+            enabled_tag_count += batch_size
+            EnabledTagKeys.objects.bulk_create(new_batch, ignore_conflicts=True)
 
-def update_enabled_keys(schema, enabled_keys_model, enabled_keys):
 
+# TODO: Remove with settings deprecation
```
maskarb, line 381: Do we have a JIRA to track this TODO?

koku/masu/util/ocp/ocp_post_processor.py
```diff
@@ -36,6 +36,9 @@ def process_openshift_labels(label_string):
     Returns:
         (dict): The JSON dictionary made from the label string
 
+    Dev Note:
+        You can use 'kube_node_labels` in prometheus to see the values.
+
```
maskarb, line 41: This is not a complete statement. There are 5 metrics we query to gather all the labels. Let's just remove this and we can reference the operator if we need to see what queries to run to see labels.

koku/reporting/migrations/0303_tags_data.py
```diff
@@ -0,0 +1,51 @@
...
+    )
+
+    EnabledTagKeys.objects.using(db_alias).bulk_create(new_records, ignore_conflicts=True)
+
+
+def unmigrate_tags(apps, schema_editor, provider):
+    EnabledTagKeys = apps.get_model("reporting", "EnabledTagKeys")
+    db_alias = schema_editor.connection.alias
+
+    EnabledTagKeys.objects.using(db_alias).filter(provider_type__exact=provider).delete()
```
maskarb, line 23: Do we need this? The reverse-migration should be a truncate table, or just a no-op. Why delete individual provider types from the table?

maskarb (changes requested):

koku/reporting/migrations/0303_tags_data.py
myersCody: @samdoran took lead on making sure that this migration could be rerun if needed. Although this may look silly, what we want to accomplish is to reinsert the tags IF there is a problem. Right now our default is "on conflict do nothing". So this delete is to capture any state changes between the first & second run of the migration. With the steps we have taken to turn off the download workers and ddf, this may be non trivial, but I want to let @samdoran provide his thoughts too.

koku/masu/util/ocp/ocp_post_processor.py
```diff
@@ -36,6 +36,9 @@ def process_openshift_labels(label_string):
     Returns:
         (dict): The JSON dictionary made from the label string
 
+    Dev Note:
+        You can use 'kube_node_labels` in prometheus to see the values.
```
myersCody, line 40:
```suggestion
        You can reference the operator here to see what queries to run 
        in prometheus to see the labels. 
```

koku/masu/util/common.py
myersCody, line 381:
```suggestion
# TODO: Remove with settings deprecation COST-3797
```

koku/masu/test/util/azure/test_azure_post_processor.py
myersCody: Yeah, That is a by product of having three different engineers work on the same branch 😂 

Ill get these these cleaned up.

koku/reporting/migrations/0303_tags_data.py
maskarb: So, what I am advocating for is just deleting everything in the table if we need to roll this back. That is the outcome from a reverse migration _anyway_, so instead of filtering and deleting, we could just truncate the table on a rollback. That should be faster and less error prone.

koku/api/settings/settings.py
```diff
@@ -157,7 +157,12 @@ def _build_enable_key_form(self, sub_form_fields, key_type):
         enabled_objs = []
         for providerName in provider_params:
             enabled_key_model = provider_params[providerName]["enabled_model"]
```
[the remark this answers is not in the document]
myersCody: Sure, but this entire thing is getting deleted soon...

koku/reporting/migrations/0303_tags_data.py
samdoran: I originally had it just deleting everything in the table but wasn't sure if we needed the ability to selectively delete keys by provider. I also wasn't sure how to have one single rollback operation for all of the data migrations. Since I was calling `unmigrate_tags` for every single provider doing the same thing, I made it accept a provider param. I could do some clever things in the rollback function to ensure it only runs once if called multiple times, but maybe there is an official way to do that.

maskarb: This what you are after:
https://www.postgresql.org/docs/current/sql-truncate.html

`TRUNCATE TABLE {schema}.reporting_enabledtagkeys`

We can call this every time and there won't be a performance hit. But, 🤷🏻 , we shouldn't have to roll back the migration anyway.

koku/api/tags/aws/openshift/queries.py
```diff
@@ -28,13 +27,21 @@ class OCPAWSTagQueryHandler(AWSTagQueryHandler, OCPTagQueryHandler):
             "db_table": OCPAWSTagsSummary,
             "db_column_period": "cost_entry_bill__billing_period",
             "annotations": {
-                "enabled": Exists(AWSEnabledTagKeys.objects.filter(key=OuterRef("key")).filter(enabled=True))
+                "enabled": Exists(
+                    EnabledTagKeys.objects.filter(provider_type=Provider.PROVIDER_AWS)
+                    .filter(key=OuterRef("key"))
+                    .filter(enabled=True)
+                )
```
maskarb, line 34: Just for future reference, _all_ `.filter().filter().filter()` can be combined into one.

maskarb (approved):
lgtm

koku/masu/util/common.py
```diff
@@ -325,52 +326,74 @@ def batch(iterable, start=0, stop=None, _slice=1):
...
-                enabled_keys_model.objects.bulk_create(new_batch, ignore_conflicts=True)
-    if not changed:
-        LOG.info(log_json(msg="no enabled keys added", context=ctx))
-
-    return changed
-
+        new_tags = list(
+            set(enabled_tags)
+            - {k for k in EnabledTagKeys.objects.filter(provider_type=provider_type).values_list("key", flat=True)}
+        )
```
samdoran, line 350: You can save the creation of an additional set object by using a method and a generator expression.
```suggestion
        new_tags = list(
            set(enabled_tags).difference(
                k for k in EnabledTagKeys.objects.filter(provider_type=provider_type).values_list("key", flat=True)
        )
```

samdoran: You could also leave this as a `set` since it gets cast to a list in `batch()`

koku/masu/util/common.py
```diff
@@ -325,52 +326,74 @@ def batch(iterable, start=0, stop=None, _slice=1):
...
+                    enabled_value = False
+            ctx["tags_enabled"] = enabled_value
+            batch_size = len(new_batch)
+            LOG.info(
+                log_json(msg="create tag batch", batch_number=(batch_num + 1), batch_size=batch_size, context=ctx)
+            )
+            for ix in range(batch_size):
+                new_batch[ix] = EnabledTagKeys(key=new_batch[ix], provider_type=provider_type, enabled=enabled_value)
+            enabled_tag_count += batch_size
+            EnabledTagKeys.objects.bulk_create(new_batch, ignore_conflicts=True)
```
samdoran, line 378: This is modifying the list while iterating over it. That works fine here because the size of the list isn't changing, but it would be safer to build a different object to pass to `bulk_create()`.
```suggestion
            new_records = (EnabledTagKeys(key=key, provider_type=provider_type, enabled=enabled_value) for key in new_batch)
            enabled_tag_count += batch_size
            EnabledTagKeys.objects.bulk_create(new_records, ignore_conflicts=True)
```

koku/api/tags/aws/openshift/queries.py
samdoran: I've always wondered why Django lets you chain filters like this. Is there a performance difference between a single filter called and chained filter calls? I guess Django provides this interface to make long queries more legible but I don't know for sure.

samdoran: Looking at the [docs](https://docs.djangoproject.com/en/dev/topics/db/queries/#chaining-filters) it seems like chaining filters only makes sense if you need to filter on the intermediate QuerySet. Otherwise it's just as readable to put the kwargs on different lines.

samdoran: I would like to see the difference in the SQL that's generated, for science.

koku/reporting/migrations/0303_tags_data.py
samdoran: I pushed up a change. Hopefully we won't have to use it. 🤞

koku/api/tags/aws/openshift/queries.py
samdoran: I fixed these in this file. Thanks for the feedback.

samdoran (approved):

PR #4581, opened 2023-07-15
[COST-4026] Correct refresh XL queue name

samdoran:
## Jira Ticket

[COST-4026](https://issues.redhat.com/browse/COST-4026)

## Description

Correct the refresh XL queue name used when reporting stats.


lcouzens (approved):
Nice!

PR #4582, opened 2023-07-16
[COST-3891/COST-3893] Add workers for SUBS  data extraction and transmission

djnakabaale:
## Jira Ticket

[COST-3891](https://issues.redhat.com/browse/COST-3891)
[COST-3893](https://issues.redhat.com/browse/COST-3893)

## Description

This change will add workers for SWATCH data extraction and transmission.

## Testing

1. Checkout Branch
2. Restart Koku
3. Hit the [celery_queue_lengths](http://127.0.0.1:5042/api/cost-management/v1/celery_queue_lengths/) endpoint.

    - You should see the subs queues in the response.
    ```
    {        
        ...

        "subs_extraction": 0,
        "subs_transmission": 0
    }


## Notes

...

The minimum replicas for these workers are set to `"0"` at the moment.


deploy/kustomize/patches/worker-subs-extraction.yaml
```diff
@@ -0,0 +1,194 @@
...
+          value: ${ENABLE_SUBS_DEBUG}
+        - name: CELERY_LOG_LEVEL
+          value: ${CELERY_LOG_LEVEL}
+        - name: KOKU_LOG_LEVEL
+          value: ${KOKU_LOG_LEVEL}
+        - name: UNLEASH_LOG_LEVEL
+          value: ${UNLEASH_LOG_LEVEL}
+        - name: PROMETHEUS_MULTIPROC_DIR
+          value: ${PROMETHEUS_DIR}
+        - name: REQUESTED_BUCKET
```
cgoodfred, line 52: These workers would need access to the subs bucket similar to [what we did here](https://github.com/project-koku/koku/pull/4230/files#diff-20dabfb7f9b39f8ad15f9faa13f99e8f0a81a8f146b383294aaeada5ab76c359L53-L54 ) for the ROS bucket

cgoodfred (changes requested):

deploy/kustomize/patches/worker-subs-extraction.yaml
djnakabaale: SUBS bucket is added. Thanks for pointing it out.

cgoodfred: I think we just need it [to the object store](https://github.com/project-koku/koku/pull/4230/files#diff-23b38de8fd89583f8a61f3426216dc745c12d84068fd5fff6f259075ab186269R39-R41) and [give a default like we](https://github.com/project-koku/koku/pull/4230/files#diff-23b38de8fd89583f8a61f3426216dc745c12d84068fd5fff6f259075ab186269R224-R227) did for ROS (for ephemeral basically) so its filled in for the clowdapp:
```
- name: REQUESTED_SUBS_BUCKET
   value: '${S3_SUBS_BUCKET_NAME}'
```

cgoodfred: and we should also add the [kafka topic like we did here for ROS](https://github.com/project-koku/koku/blob/main/deploy/kustomize/base/base.yaml#L37)

djnakabaale: Thanks again Corey, I've added these changes!

deploy/kustomize/patches/worker-subs-transmission.yaml
```diff
@@ -0,0 +1,196 @@
...
+    displayName: Worker Queue
+    name: WORKER_SUBS_TRANSMISSION_WORKER_QUEUE
+    required: true
+    value: "subs_transmission"
+- op: add
+  path: /parameters/-
+  value:
+    displayName: SUBS Transmission debug
+    name: ENABLE_SUBS_DEBUG
+    value: "False"
```
cgoodfred, line 196: Rather than define this twice, we should define `ENABLE_SUBS_DEBUG` in the base.yaml and then use it in the worker patches.

deploy/kustomize/patches/worker-subs-extraction.yaml
```diff
@@ -0,0 +1,196 @@
...
+    displayName: Worker Queue
+    name: WORKER_SUBS_EXTRACTION_WORKER_QUEUE
+    required: true
+    value: "subs_extraction"
+- op: add
+  path: /parameters/-
+  value:
+    displayName: SUBS Extraction debug
+    name: ENABLE_SUBS_DEBUG
+    value: "False"
```
cgoodfred, line 196: Rather than define this twice, we should define ENABLE_SUBS_DEBUG in the base.yaml and then use it in the worker patches.

cgoodfred (commented):
Didn't notice this the first time, but I think we should be good after this.

cgoodfred (approved):

PR #4586, opened 2023-07-17
Add settings to rbac python file.

myersCody:
## Jira Ticket

## Description

This change will ...

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Notes

...

### DIFF
--- a/koku/koku/test_rbac.py
+++ b/koku/koku/test_rbac.py
@@ -21,13 +21,16 @@ from koku.rbac import RbacConnectionError
 from koku.rbac import RbacService
 
 
-def create_default_access(access_dict=None, default_write=None):
+def create_expected_access(access_dict=None, default_write=None, default_read=None):
     """Helper function for defaulting access permissions."""
     if access_dict is None:
         access_dict = {}
-       
-   if default_write is None:
-       default_write = {"write": [], "read": []}, default_read={"read": []}
+
+    if default_write is None:
+        default_write = {"write": [], "read": []}
+    if default_read is None:
+        default_read = {"read": []}
+
     default = {
         "cost_model": default_write,
         "settings": default_write,
@@ -314,14 +317,14 @@ class RbacServiceTest(TestCase):
         """Test handling exception _get_operation used in apply access method."""
         processed_acls = {"*": [{"operation": "*", "resources": ["1", "3"]}]}
         res_access = _apply_access(processed_acls)
-        expected = create_default_access()
+        expected = create_expected_access()
         self.assertEqual(res_access, expected)
         mock_get_operation.assert_called()
 
     def test_apply_access_none(self):
         """Test handling none input for apply access method."""
         res_access = _apply_access(None)
-        expected = create_default_access()
+        expected = create_expected_access()
         self.assertEqual(res_access, expected)
 
     def test_apply_access_all_wildcard(self):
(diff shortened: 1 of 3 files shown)


maskarb (commented):
This looks good, but we ought to add back the `if settings.ENHANCED_ORG_ADMIN and request.user.admin:` to the `SettingsAccessPermission` class.

koku/koku/test_rbac.py
```diff
@@ -20,6 +20,28 @@
...
+        "gcp.project": default_read,
+        "openshift.cluster": default_read,
+        "openshift.node": default_read,
+        "openshift.project": default_read,
+        "ibm.account": default_read,
+        "oci.payer_tenant_id": default_read,
+    }
+    for key, value in access_dict.items():
+        default[key] = value
+    return default
```
samdoran, line 42: You can use the `update()` method to modify the values in `default` or the `|` operator to return a new dictionary with merged values. 
```suggestion
    return default | access_dict
```

koku/koku/test_rbac.py
```diff
@@ -20,6 +20,28 @@
 from koku.rbac import RbacConnectionError
 from koku.rbac import RbacService
 
+
+def create_default_access(access_dict={}, default_write={"write": [], "read": []}, default_read={"read": []}):
+    """Helper function for defaulting access permissions."""
```
samdoran, line 25:
```suggestion
def create_default_access(access_dict=None, default_write=None):
    """Helper function for defaulting access permissions."""
    if access_dict is None:
        access_dict = {}
       
   if default_write is None:
       default_write = {"write": [], "read": []}, default_read={"read": []}
```

samdoran (changes requested):

maskarb:
> > This looks good, but we ought to add back the `if settings.ENHANCED_ORG_ADMIN and request.user.admin:` to the `SettingsAccessPermission` class.
> 
> I am pretty certain this would cause the current ddf settings to have a 403 again. The only reason that got resolved yesterday was because I removed the `settings.ENHANCED_ORG_ADMIN` check.
> 
> So, what I am trying to do is build settings permissions that will work for the current DDF settings & for the new settings endpoint.

I am 99% sure the reason we saw the 403 was because the `settings` in resource types was missing. An org admin has the `Cost Administrator` role automatically which includes the `cost-management:*:*` access. When we look up this access, we convert the wildcards [here](https://github.com/project-koku/koku/blob/main/koku/koku/rbac.py#L119) to give that user access to all the resource types. Since settings was missing, we were returning the 403.

maskarb (approved):

koku/koku/test_rbac.py
```diff
@@ -20,6 +20,34 @@
...
 from koku.rbac import RbacService
 
+
+def create_expected_access(access_dict=None, default_write=None, default_read=None):
+    """Helper function for defaulting access permissions."""
+    if access_dict is None:
+        access_dict = {}
+
+    if default_write is None:
+        default_write = {"write": [], "read": []}
```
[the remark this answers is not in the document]
myersCody: So, I tried your suggestion of combining them, but the test starting failing. I need to be able to set them independently in certain places in the test:

https://github.com/project-koku/koku/pull/4586/files#diff-e99710acde7df448c3ce00e9fdf21f96a49c90549781c51d57f9a4292423df13R374

PR #4589, opened 2023-07-20
[COST-3808] add external-id to sources integration

maskarb:
## Jira Ticket

[COST-3808](https://issues.redhat.com/browse/COST-3808)

## Description

This change will grab the external-id from the sources `extras` field when it is present.

### DIFF
--- a/koku/sources/sources_http_client.py
+++ b/koku/sources/sources_http_client.py
@@ -220,7 +220,7 @@ class SourcesHTTPClient:
 
         username = auth_data.get("username")
         if username:
-            return {"role_arn": username, "external_id": auth_data.get("external_id")}
+            return {"role_arn": username, "external_id": auth_data.get("extra", {}).get("external_id")}
 
         raise SourcesHTTPClientError(f"Unable to get AWS roleARN for Source: {self._source_id}")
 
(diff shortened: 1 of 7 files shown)


koku/sources/sources_http_client.py
```diff
@@ -221,20 +218,9 @@ def _get_aws_credentials(self, _):
...
-
-        auth_id = auth_data.get("id")
-        auth_internal_url = (
-            f"{self._internal_url}/{ENDPOINT_AUTHENTICATIONS}/{auth_id}?expose_encrypted_attribute[]=password"
-        )
-        auth_internal_response = self._get_network_response(auth_internal_url, "Unable to get AWS RoleARN")
-        password = auth_internal_response.get("password")
-        if password:
-            return {"role_arn": password}
+            return {"role_arn": username, "external_id": auth_data.get("extra", {}).get("external_id")}
```
samdoran, line 222: Is it ok for `external_id` to be `None`?

maskarb: for now, yes.

lcouzens (approved):
lgtm

samdoran (approved):

myersCody:
@maskarb You can still merge if you want, I just didn't know if you saw eva comment about how hot fix smokes doesn't add any aws sources. So I went ahead and rekick them off for you in case you wanted to rerun them so that way you don't have to wait.

PR #4598, opened 2023-07-25
[COST-3996] Unleash flag to disable tag enablement on settings.

myersCody:
## Jira Ticket

[COST-3996](https://issues.redhat.com/browse/COST-3996)

## Description

This change will ...

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Notes

...

### DIFF
--- a/.unleash/flags.json
+++ b/.unleash/flags.json
@@ -2,6 +2,22 @@
     "version": 1,
     "features": [
         {
+            "name": "cost-management.backend.disable-ddf-tag-form",
+            "description": "Toggle to disabled tag data driven form for settings.",
+            "type": "killswitch",
+            "project": "default",
+            "enabled": false,
+            "stale": false,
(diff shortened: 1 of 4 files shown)


.unleash/flags.json
```diff
@@ -1,6 +1,22 @@
 {
     "version": 1,
     "features": [
+        {
+            "name": "cost-management.backend.disable-ddf-tag-form",
```
maskarb, line 5: I wouldn't even bother putting this flag here. Aren't we only using this flag once in prod?

maskarb (approved):

.unleash/flags.json
myersCody: Yeah, I have this flag and the cost categories flag as part of the deprecated clean up issues. So it will be removed relatively soon.

PR #4606, opened 2023-08-01
drop date override from OCP flow

lcouzens:
## Jira Ticket

## Description

This change will ...

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Notes

...


myersCody:
So, I have been trying to follow the start & end dates through the code base here to make sure your check won't have any fall out. 

I started by looking at this [section of code](https://github.com/project-koku/koku/blob/main/koku/masu/processor/tasks.py#L502-L506), we use the start & end date returned from the `update_summary_tables` (which calls the function you modify) to pass to the `get_openshift_on_cloud_infra_map`:
```
def get_openshift_on_cloud_infra_map(self, start_date, end_date, tracing_id):
        """Get cloud infrastructure source and OpenShift source mapping."""
        if self._provider.type not in Provider.OPENSHIFT_ON_CLOUD_PROVIDER_LIST:
            return {}

        try:
            start_date, end_date = self._format_dates(start_date, end_date)
            context = {
                "schema": self._schema,
                "provider_uuid": self._provider_uuid,
                "start_date": start_date,
                "end_date": end_date,
            }
            LOG.info(log_json(tracing_id, msg="getting OCP-on-Cloud infra map", context=context))
            return self._ocp_cloud_updater.get_infra_map(start_date, end_date)
        except Exception as ex:
            raise ReportSummaryUpdaterCloudError(str(ex)) from ex
```
So this function will `return {}`  due to the `in Provider.OPENSHIFT_ON_CLOUD_PROVIDER_LIST:` check. 

Since we won't have an `ocp_on_cloud_infra_map` we can skip ahead to the next place the start & end dates are used and that is to `update_cost_model_cost`:
```
linked_tasks = update_cost_model_costs.s(
            schema, provider_uuid, start_date, end_date, tracing_id=tracing_id
        ).set(queue=queue_name or fallback_update_cost_model_queue)
```
I think currently we would update the entire previous month cost model costs. I don't think this logic is intentional, so I think removing this section of code is fine.

myersCody (approved):

PR #4614, opened 2023-08-05
Cost 3991 option 4 drop unrequired columns

lcouzens:
## Jira Ticket

## Description

This change will clean up some columns that we have set as required which potentially shouldnt be. Fall out from daily archive changes!

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Notes
Azure actually has at least 3 different cost export types WRT column names and data structure within them. This change _should_ now handle them. However! There is a small Nise change that probably needs to accompany this to aid smoke test. We currently have the incorrect formatting in Nise which made me overlook this in the first place! Its entirely possible there is OTHER cost export types out their we might need to handle too. So glad Azure can keep things consistent for people!! 

This also kinda requires some Nise/populator tweaks to clean up some jank
https://github.com/project-koku/nise-populator/pull/79
https://github.com/project-koku/nise/pull/462

Because of the nise update this is likely going to break smoke tests until thats releases and in IQE
...


koku/masu/external/downloader/aws/aws_report_downloader.py
```diff
@@ -57,16 +57,16 @@ def get_initial_dataframe_with_date(
...
         invoice_bill = "bill_invoice_id"
         time_interval = "identity_time_interval"
         optional_cols = ["resource_tags", "cost_category"]
-        base_cols = utils.INGRESS_ALT_COLUMNS
-    use_cols = com_utils.fetch_optional_columns(local_file, base_cols, optional_cols, tracing_id, context)
-    data_frame = pd.read_csv(local_file, usecols=use_cols)
+    # drop tag/categry columns that only contain null values.
+    drop_cols = com_utils.list_null_optional_columns(local_file, optional_cols, tracing_id, context)
+    headers = [*pd.read_csv(local_file, nrows=1)]
+    data_frame = pd.read_csv(local_file, usecols=[c for c in headers if c not in drop_cols])
```
cgoodfred, line 69: I feel like we're opening this CSV so many times using pandas. We open it twice right here after opening the entire CSV and doing some processing on it in `list_null_optional_columns`. 

Could we not just open it once and drop the null columns instead of opening it and doing processing on it before we then open a subset of columns?

koku/masu/external/downloader/azure/azure_report_downloader.py
```diff
@@ -58,17 +58,22 @@ def get_initial_dataframe_with_date(
...
+            data_frame = pd.read_csv(local_file, usecols=["Date"], nrows=1)
+            time_interval = "Date"
+            date_format = "%Y-%m-%d"
+        except ValueError:
+            time_interval = "date"
+            date_format = "%m/%d/%Y"
+    # drop tag columns that only contain null values.
+    drop_cols = com_utils.list_null_optional_columns(local_file, ["tags"], tracing_id, context)
+    headers = [*pd.read_csv(local_file, nrows=1)]
+    data_frame = pd.read_csv(local_file, usecols=[c for c in headers if c not in drop_cols])
```
cgoodfred, line 75: Same thing as the AWS one, we're using pandas and reading the same CSV so many times here. I think we could cut down on how many CSV reads we are doing.

koku/masu/util/azure/common.py
```diff
@@ -61,12 +59,9 @@
     "EffectivePrice",
     "CostInBillingCurrency",
     "ConsumedService",
-    "ResourceId",
-    "OfferId",
     "AdditionalInfo",
     "ServiceInfo1",
     "ServiceInfo2",
-    "ResourceName",
```
cgoodfred, line 69: Why are we removing some of the columns from here?

koku/masu/external/downloader/azure/azure_report_downloader.py
lcouzens: Same thing

koku/masu/util/azure/common.py
lcouzens: these can go back now :+1: thanks

PR #4615, opened 2023-08-08
Ford OCP/GCP potentially matching on "" fix

lcouzens:
## Jira Ticket

[COST-4070](https://issues.redhat.com/browse/COST-4070)

## Description

This change will update our matching node/volume logic to not include "" empty string values.

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Notes
Also updated some sql params for better logging.

Improves :hankey: like this NONE NONE NONE :smile: 
{'message': 'executed trino sql', 'tracing_id': '', 'schema': None, 'start_date': None, 'end_date': None, 'invoice_month': None, 'source_type': None, 'provider_uuid': None, 'cluster_id': None, 'log_ref': 'schema_exists_trino', 'running_time': 0.02856588363647461}
...


koku/masu/database/gcp_report_db_accessor.py
```diff
@@ -307,8 +307,14 @@ def get_gcp_topology_trino(self, source_uuid, start_date, end_date, invoice_mont
...
         """
-
-        return self._execute_trino_raw_sql_query(sql, log_ref="get_gcp_topology_trino")
+        sql_params = {
+            "schema": self.schema,
+            "start": start_date,
+            "end": end_date,
+            "provider_uuid": source_uuid,
+            "invoice_id": invoice_month_date,
+        }
```
maskarb, line 316: Adding these sql_params strictly for logging makes this code harder to understand because it _seems_ like these should be substituted into the sql (but they are not). I think it would make more sense for us to make these actual jinja templates too just so that this code and the use of these params makes sense. This template could be:
```
        sql = """
            SELECT source,
                billing_account_id,
                project_id,
                project_name,
                service_id,
                service_description,
                location_region
            FROM hive.{{schema | sqlsafe}}.gcp_line_items as gcp
            WHERE gcp.source = '{{source_uuid}}'
                AND gcp.year = '{{year}}'
                AND gcp.month = '{{month}}'
                AND gcp.usage_start_time >= {{start_date}}
                AND gcp.usage_start_time < date_add('day', 1, {{end_date}})
            GROUP BY source,
                billing_account_id,
                project_id,
                project_name,
                service_id,
                service_description,
                location_region
        """
        sql_params = {
            "schema": self.schema,
            "start_date": start_date,
            "end_date": end_date,
            "source_uuid": source_uuid,
            "year": invoice_month_date.strftime("%Y"),
            "month": {invoice_month_date.strftime("%m")},
            "invoice_id": invoice_month_date,
        }
```

That, OR we add a `context` param to the `_execute_trino_raw_sql_query` just so that it is clear that those params are passed strictly for logging and not to be used as sql_params.

lcouzens: Yeh your not wrong! I was just trying to force them into the logging dirty like (without changing everything) just to see if it helped follow things a little better for Ford. :smile:

koku/masu/util/gcp/common.py
```diff
@@ -84,7 +84,11 @@ def match_openshift_resources_and_labels(data_frame, cluster_topologies, matched
 
         if resource_id_df.any():
             LOG.info("Matching OpenShift on GCP by resource ID.")
-            ocp_matched = resource_id_df.str.contains("|".join(matchable_resources))
+            matching_check = "|".join(matchable_resources)
+            if matching_check:
+                ocp_matched = resource_id_df.str.contains(matching_check)
+            else:
+                continue
```
cgoodfred, line 91: nitpick: we could reverse the check here
```suggestion
            if not matching_check:
                continue
            ocp_matched = resource_id_df.str.contains(matching_check)
```

koku/masu/database/gcp_report_db_accessor.py
lcouzens: Context it is :+1:

koku/masu/util/gcp/common.py
lcouzens: My brain would not let me think of this and I tried really hard! LACK OF SLEEP SUCKS!! I know you would pick it up though! Thank you!! :heart:

cgoodfred (approved):

cgoodfred (approved):

PR #4616, opened 2023-08-08
[COST-3991] - Option 3 - usecols for daily archives

lcouzens:
## Jira Ticket

## Description

This change will address current alerts while still including the required list of columns + making use of usecols when opening csv's into pandas dataframes. The means we only read the columns from the csv file if we require them.

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Notes
This will still need a nise tweak to bring the columns in-line with the real reports I've seen. https://github.com/project-koku/nise/pull/463
I started an google doc with these column differences. All where based on REAL Azure reports.
...


koku/masu/util/azure/common.py
```diff
@@ -35,8 +35,6 @@
     "PreTaxCost",
     "ConsumedService",
     "ResourceType",
-    "InstanceId",
-    "OfferId",
     "AdditionalInfo",
```
cgoodfred, line 38: so why are we removing these columns again?

cgoodfred: I think I'm just struggling with why some of these changes are necessary, Azure Docs don't indicate a column called `SubscriptionGuid` exists, just `SubscriptionId`: https://learn.microsoft.com/en-us/azure/cost-management-billing/automate/understand-usage-details-fields

I'm also not sure we can rely on specific casing like `MeterSubCategory` vs `MeterSubcategory` since the docs above state casing can be inconsistent:
```
Some fields might differ in casing and spacing between account types. Older versions of pay-as-you-go cost details files have separate sections for the statement and daily cost.
```

I know Azure has different report formats such as MCA vs EA vs PayG but where are we getting the information for these different columns from?

cgoodfred: Related to the casing differences, in the post processors we try to ignore casing to be as consistent as possible like here: https://github.com/project-koku/koku/blob/main/koku/masu/util/azure/azure_post_processor.py#L61-L85

Would it be possible to handle all kinds of casing like we do in the post processor rather than create more case variants to match?

cgoodfred: I think you were actually handling the casing differences like I was hoping and I was just overlooking it. I think removing the lower is less consistent here: https://github.com/project-koku/koku/pull/4616/commits/6c4757f0d65f3f34062e9971dcbdff75ebaabfb3

That one is totally my bad, I apparently need to get my eyes checked

PR #4623, opened 2023-08-09
[COST-4097] Settings aws category keys use or for same filter.

myersCody:
## Jira Ticket

[COST-4097](https://issues.redhat.com/browse/COST-4097)

## Description

This change will enforce using ors on multiple filters of the same type. 

## Testing

http://localhost:8000/api/cost-management/v1/settings/aws_category_keys/?filter[key]=e&filter[key]=ion

```
{
            "uuid": "3cdcd9b1-2d65-4927-89ab-72b2c699da01",
            "key": "env",
            "enabled": false
        },
        {
            "uuid": "0d45b6ea-d58b-4a28-be96-da783a1220d8",
            "key": "Organization",
            "enabled": true
        }
```

## Notes

...


koku/api/settings/utils.py
```diff
@@ -2,10 +2,18 @@
 # Copyright 2021 Red Hat Inc.
 # SPDX-License-Identifier: Apache-2.0
 #
+import typing as t
+
+import django.core.exceptions
+import django_filters
+import rest_framework.exceptions
```
maskarb, line 9: Why are we not importing what we need and use? We almost always import specifics, not entire modules, such as:
```
from rest_framework.exceptions import ValidationError
from django_filters.filters import MultipleChoiceFilter
from django.core.exceptions import ValidationError as DjangoValidationError
```

koku/api/settings/tags/view.py
```diff
@@ -36,78 +30,7 @@ class SettingsTagFilter(django_filters.rest_framework.FilterSet):
     class Meta:
         model = EnabledTagKeys
         fields = ("enabled", "uuid")
-
-    def _get_order_by(
```
myersCody, line 40: This pr is mostly to move this logic to the utils so that I could use it for the aws_cost_category keys, since @samdoran and I spent time getting multiple filters functional for tags.

djnakabaale (dismissed):
lgtm 👍🏾

maskarb (dismissed):
thanks for updating those imports!

PR #4626, opened 2023-08-10
[COST-4101] tech debt remove ocp on gcp summarization by node.

myersCody:
## Jira Ticket

[COST-4101](https://issues.redhat.com/browse/COST-4101)

## Description

This change will remove a code path that we no longer use. The original issue mentioned this would only be a stop gap solution: https://issues.redhat.com/browse/COST-2992

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Notes

...


lcouzens (approved):
Another great clean up PR! :bow: Thanks Cody.

PR #4627, opened 2023-08-11
usecols for AWS archives

lcouzens:
## Jira Ticket

## Description

This change will ...

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Notes

...

### DIFF
--- a/koku/masu/external/downloader/aws/aws_report_downloader.py
+++ b/koku/masu/external/downloader/aws/aws_report_downloader.py
@@ -3,6 +3,7 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 """AWS Report Downloader."""
+import copy
 import datetime
 import json
 import logging
@@ -57,10 +58,18 @@ def get_initial_dataframe_with_date(
     """
     invoice_bill = "bill/InvoiceId"
     time_interval = "identity/TimeInterval"
-    data_frame = pd.read_csv(local_file)
-    if invoice_bill not in data_frame.columns:
+    optional_cols = ["resourcetags", "costcategory"]
+    base_cols = copy.deepcopy(utils.INGRESS_REQUIRED_COLUMNS)
+    try:
+        data_frame = pd.read_csv(local_file, usecols=[invoice_bill], nrows=1)
+    except ValueError:
         invoice_bill = "bill_invoice_id"
         time_interval = "identity_time_interval"
+        optional_cols = ["resource_tags", "cost_category"]
+        base_cols = copy.deepcopy(utils.INGRESS_ALT_COLUMNS)
+    use_cols = com_utils.fetch_optional_columns(local_file, base_cols, optional_cols, tracing_id, context)
+    LOG.info(log_json(tracing_id, msg="pandas read csv with following usecols", usecols=use_cols, context=context))
+    data_frame = pd.read_csv(local_file, usecols=use_cols)
     if data_frame[invoice_bill].any() or not com_utils.check_setup_complete(provider_uuid):
         process_date = start_date
         ReportManifestDBAccessor().mark_s3_parquet_to_be_cleared(manifest_id)


koku/masu/external/downloader/aws/aws_report_downloader.py
```diff
@@ -103,30 +111,36 @@ def create_daily_archives(
...
-        utils.copy_local_report_file_to_s3_bucket(
-            tracing_id, s3_csv_path, day_filepath, day_file, manifest_id, start_date, context
-        )
-        daily_file_names.append(day_filepath)
+    LOG.info(log_json(tracing_id, msg="pandas read csv with following usecols", usecols=use_cols, context=context))
+    with pd.read_csv(local_file, usecols=use_cols, chunksize=settings.PARQUET_PROCESSING_BATCH_SIZE) as reader:
+        for i, data_frame in enumerate(reader):
+            if data_frame.empty:
+                continue
+            data_frame = pd.read_csv(local_file, usecols=use_cols)
```
myersCody, line 122: I think you meant to remove this line.

koku/masu/external/downloader/aws/aws_report_downloader.py
```diff
@@ -103,30 +111,35 @@ def create_daily_archives(
...
-        daily_data = data_frame[data_frame[time_interval].str.match(date)]
-        day_file = ReportManifestDBAccessor().update_and_get_day_file(date, manifest_id)
-        day_filepath = f"{directory}/{day_file}"
-        daily_data.to_csv(day_filepath, index=False, header=True)
-        utils.copy_local_report_file_to_s3_bucket(
-            tracing_id, s3_csv_path, day_filepath, day_file, manifest_id, start_date, context
-        )
-        daily_file_names.append(day_filepath)
+    LOG.info(log_json(tracing_id, msg="pandas read csv with following usecols", usecols=use_cols, context=context))
+    with pd.read_csv(local_file, usecols=use_cols, chunksize=settings.PARQUET_PROCESSING_BATCH_SIZE) as reader:
```
myersCody, line 118: So, I have an idea that I wondering if it would be okay with it if I ran with it after this gets merged?

Part of me is wondering if we just want to steal the an entire with statement we use [here](https://github.com/project-koku/koku/blob/main/koku/masu/processor/parquet/parquet_report_processor.py#L522-L524). That way we can do the csv converters ahead of time too. I wonder if this would fix some of the weird date stuff you have to do later on. 

I think there are more elements in the PostProcessor that we could pull out and create a PreProcessor class?

lcouzens: Yeh I was actually thinking something similar too with the converters. Certainly seems silly opening things thn opening them again to sort the converters.

koku/masu/external/downloader/aws/aws_report_downloader.py
```diff
@@ -72,7 +80,7 @@ def get_initial_dataframe_with_date(
                 s3_csv_path, start_date, end_date, manifest_id, context, tracing_id
             )
             ReportManifestDBAccessor().set_manifest_daily_start_date(manifest_id, process_date)
-    return data_frame, time_interval, process_date
+    return use_cols, time_interval, process_date
```
myersCody, line 83: This is a nitpick and not critical, but this function no longer returns a dataframe, but is still named: `get_initial_dataframe_with_date`.

myersCody (approved):

koku/masu/external/downloader/aws/aws_report_downloader.py
```diff
@@ -103,30 +111,35 @@ def create_daily_archives(
...
+            data_frame = data_frame[data_frame[time_interval].str.contains("|".join(dates))]
+            for date in dates:
+                daily_data = data_frame[data_frame[time_interval].str.match(date)]
+                day_file = ReportManifestDBAccessor().update_and_get_day_file(date, manifest_id)
+                day_filepath = f"{directory}/{day_file}_{i}.csv"
+                daily_data.to_csv(day_filepath, index=False, header=True)
+                utils.copy_local_report_file_to_s3_bucket(
+                    tracing_id, s3_csv_path, day_filepath, day_file, manifest_id, start_date, context
+                )
+                daily_file_names.append(day_filepath)
```
maskarb, line 142: Here and in the Azure method, we do a LOT of stuff with dates, and it's difficult to follow. When we read CSV files, we can pass the `parse_dates` arg which is a list of columns containing dates. Then we have an actual date-time object stored in the df instead of a string. We can also make the date the index of the df which makes getting the rows we want easier.

https://www.dataquest.io/blog/datetime-in-pandas/

I'm not saying that we need to update anything, but it might be worth seeing if parsing the dates when we read the CSV gives us a dataframe that is easier to work with.

maskarb (commented):
Some comments:

koku/masu/external/downloader/aws/aws_report_downloader.py
```diff
@@ -57,10 +58,17 @@ def get_initial_dataframe_with_date(
...
+    base_cols = copy.deepcopy(utils.INGRESS_REQUIRED_COLUMNS)
+    try:
+        data_frame = pd.read_csv(local_file, usecols=[invoice_bill], nrows=1)
+    except ValueError:
         invoice_bill = "bill_invoice_id"
         time_interval = "identity_time_interval"
+        optional_cols = ["resource_tags", "cost_category"]
+        base_cols = copy.deepcopy(utils.INGRESS_ALT_COLUMNS)
+        data_frame = pd.read_csv(local_file, usecols=[invoice_bill], nrows=1)
+    use_cols = com_utils.fetch_optional_columns(local_file, base_cols, optional_cols, tracing_id, context)
```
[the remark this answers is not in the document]
lcouzens: So the following line we are doing `data_frame[invoice_bill].any()` to check if we have an invoice id in the report :+1:

PR #4629, opened 2023-08-14
adjust required AWS columns

lcouzens:
## Jira Ticket

## Description

This change will remove "productinstancetype", "productvcpu", "productmemory" from required columns and add them as optionl. QE have at least one account that does not include these in their real report.

## Testing

1. Checkout Branch
2. Restart Koku
3. Create an AWS source
4. generate some data without these optional columns included
5. see everything ingests correctly.

Test 2
Also check the reverse works, E.I those optional columns do in fact get collected and processed correctly.
 
## Notes
This also addresses a minor bug sending files to S3 without `.csv` extensions
...


maskarb (approved):
The more we struggle with column names and whatnot, the more I think we have an issue with how we are doing things. It seems we are now trying to maintain several lists of column names with specific formats. This does not feel maintainable or understandable.

We could use `usecols` in a way that eliminates specific formatting. We could use a regex pattern:
```
pattern = re.compile('[^a-zA-Z0-9]')
```
which eliminates everything but alphanumeric characters. Then it could be used in `usecols`:
```
usecols=lambda x: re.sub(pattern, '', x).lower()

>>> df = pd.read_csv(local_file, parse_dates=list(res), usecols=lambda x: re.sub(pattern, '', x).lower())
>>> df
    InvoiceSectionName            AccountName                              AccountOwnerId                        SubscriptionId  ... paygCostInBillingCurrency paygCostInUsd exchangeRatePricingToBilling exchangeRateDate
```
Using usecols this way still maintains the column names as they are. Of course, I don't know the full implication of reading columns this way, but I'm struggling to understand the benefit of maintaining these lists like we are today.

Anyway, just some thoughts. The PR looks fine otherwise.

PR #4632, opened 2023-08-14
[COST-4108] Use pandas to parse date during azure downloader

myersCody:
## Jira Ticket

[COST-4108](https://issues.redhat.com/browse/COST-4108)

## Description

This change will fix the azure local downloader by using the padas to convert the column to a datetime for comparison.

## Testing

1. Checkout Branch
2. make load-test-customer-data test_source=Azure
3. Load our azure test source

## Notes

...

### DIFF
--- a/koku/masu/external/downloader/azure/azure_report_downloader.py
+++ b/koku/masu/external/downloader/azure/azure_report_downloader.py
@@ -57,18 +57,12 @@ def get_processing_date(
         tracing_id (str): The tracing id
     """
     dh = DateHelper()
-    date_format = "%Y-%m-%d %H:%M:%S"
-    time_interval = "UsageDateTime"
-    try:
-        pd.read_csv(local_file, usecols=[time_interval], nrows=1)
-    except ValueError:
-        time_interval = "Date"
-        date_format = "%Y-%m-%d"
-        try:
-            pd.read_csv(local_file, usecols=[time_interval], nrows=1)
-        except ValueError:
-            time_interval = "date"
-            date_format = "%m/%d/%Y"
+    time_interval = None
+    columns = pd.read_csv(local_file, nrows=0)
+    for interval in ["UsageDateTime", "Date", "date"]:
+        if interval in columns:
+            time_interval = interval
+            break
     # Azure does not have an invoice column so we have to do some guessing here
     if (
         start_date.year < dh.today.year
@@ -87,7 +81,7 @@ def get_processing_date(
                 s3_csv_path, start_date, end_date, manifest_id, context, tracing_id
             )
             ReportManifestDBAccessor().set_manifest_daily_start_date(manifest_id, process_date)
-    return time_interval, process_date, date_format
+    return time_interval, process_date
 
 
 def create_daily_archives(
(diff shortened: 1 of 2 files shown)


koku/masu/external/downloader/azure/azure_report_downloader.py
```diff
@@ -57,18 +57,12 @@ def get_processing_date(
...
-            pd.read_csv(local_file, usecols=[time_interval], nrows=1)
-        except ValueError:
-            time_interval = "date"
-            date_format = "%m/%d/%Y"
+    time_interval = None
+    columns = pd.read_csv(local_file, nrows=0)
+    for interval in ["UsageDateTime", "Date", "date"]:
+        if interval in columns:
+            time_interval = interval
+            break
```
maskarb, line 65: We should probably look at the columns, specifically:
```suggestion
    columns = pd.read_csv(local_file, nrows=0).columns
    for interval in ["UsageDateTime", "Date", "date"]:
        if interval in columns:
            time_interval = interval
            break
```
Alternatively, you could do some a set operation (intersection) to get the column you're looking for:
```suggestion
time_interval = pd.read_csv(local_file, nrows=0).columns.intersection({"UsageDateTime", "Date", "date"})[0]
```
though this method is likely more difficult to comprehend at first glance.

koku/masu/external/downloader/azure/azure_report_downloader.py
```diff
@@ -118,30 +112,34 @@ def create_daily_archives(
...
+                "start": result_df.get(time_interval).min().strftime(DATE_FORMAT),
+                "end": result_df.get(time_interval).max().strftime(DATE_FORMAT),
                 "invoice_month": None,
             }
+
+            directory = os.path.dirname(local_file)
             for date in dates:
-                day_path = datetime.datetime.strptime(date, date_format).strftime(DATE_FORMAT)
-                daily_data = data_frame[data_frame[time_interval].str.match(date)]
+                daily_data = data_frame.loc[(result_df[time_interval] == date)]
```
maskarb, line 141: Again, with the index set to the date we are interested in, this look up becomes easier:
```suggestion
                daily_data = data_frame.loc[date]
```

koku/masu/external/downloader/azure/azure_report_downloader.py
```diff
@@ -118,30 +112,34 @@ def create_daily_archives(
...
+            if result_df.empty:
                 return [], {}
-            directory = os.path.dirname(local_file)
+
+            dates = result_df[time_interval].unique()
             date_range = {
-                "start": datetime.datetime.strptime(min(dates), date_format).strftime(DATE_FORMAT),
-                "end": datetime.datetime.strptime(max(dates), date_format).strftime(DATE_FORMAT),
+                "start": result_df.get(time_interval).min().strftime(DATE_FORMAT),
+                "end": result_df.get(time_interval).max().strftime(DATE_FORMAT),
```
maskarb, line 135: If you do end up setting and sorting the index, then you can get the dates thru simple indexing, rather than using min and max:
```
>>> data_frame.index[0]
Timestamp('2023-08-01 00:00:00')
>>> data_frame.index[-1]
Timestamp('2023-08-14 00:00:00')
>>> data_frame.index[-1].strftime(DATE_FORMAT)
'2023-08-14'
```

```suggestion
                "start": data_frame.index[0].strftime(DATE_FORMAT),
                "end": data_frame.index[-1].strftime(DATE_FORMAT),
```

koku/masu/external/downloader/azure/azure_report_downloader.py
```diff
@@ -118,30 +112,34 @@ def create_daily_archives(
...
-                csv_date = datetime.datetime.strptime(interval, date_format)
-                # Adding end here so we dont bother to process future incomplete days (saving plan data)
-                if csv_date >= process_date and csv_date <= end_date:
-                    dates.add(interval)
-            if not dates:
+
+            # Adding end here so we dont bother to process future incomplete days (saving plan data)
+            result_df = data_frame.loc[
+                (data_frame[time_interval] >= process_date) & (data_frame[time_interval] <= end_date)
+            ]
```
maskarb, line 128: If we set and sort the index on the time_interval, it potentially makes the next operations easier to write
```suggestion
            data_frame = data_frame.set_index(time_interval).sort_index()

            # Adding end here so we dont bother to process future incomplete days (saving plan data)
            data_frame = data_frame.loc[process_date:end_date]
```

and to save on memory, I would not create the `result_df`. We ought to just overwrite `data_frame`.

koku/masu/external/downloader/azure/azure_report_downloader.py
```diff
@@ -118,30 +112,34 @@ def create_daily_archives(
...
+
+            directory = os.path.dirname(local_file)
             for date in dates:
-                day_path = datetime.datetime.strptime(date, date_format).strftime(DATE_FORMAT)
-                daily_data = data_frame[data_frame[time_interval].str.match(date)]
+                daily_data = data_frame.loc[(result_df[time_interval] == date)]
+                day_path = pd.to_datetime(date).strftime(DATE_FORMAT)
                 day_file = ReportManifestDBAccessor().update_and_get_day_file(day_path, manifest_id)
                 day_filepath = f"{directory}/{day_file}_{i}.csv"
                 daily_data.to_csv(day_filepath, index=False, header=True)
```
maskarb, line 145: If we've set the index to the date, this probably has to be updated to write the date to file. Maybe `index=True`? Not 100% sure though.

maskarb (commented):
Some comments:

koku/masu/external/downloader/azure/azure_report_downloader.py
maskarb: Of course, these messages are out of order. The above makes sense if you read this comment first:
https://github.com/project-koku/koku/pull/4632#discussion_r1293949450

koku/masu/external/downloader/azure/azure_report_downloader.py
maskarb: Of course, these messages are out of order. The above makes sense if you read this comment first:
https://github.com/project-koku/koku/pull/4632#discussion_r1293949450

koku/masu/external/downloader/azure/azure_report_downloader.py
lcouzens: Plenty of options here. :smile: So originally I was avoiding loading all the columns but since we have not implemented usecols _yet_ that might not actually matter here for now. That being said regarding casings, I have seen another report with lower `usagedatestart`. With that in mind we _could_ do something like this:

`cols = ["usagedatetime", "date"]`
`df = pandas.read_csv(local_file, usecols=lambda x: x.lower() in cols, nrows=0)`
 `df.columns[0]`

That way we could handle all the column header case types. :+1:

koku/masu/external/downloader/azure/azure_report_downloader.py
myersCody: I just decided to tell it to not drop the column when it creates the index cause I needed it to create the `dates` var.

maskarb: The dates could have been pulled from the index, but not dropping it also works.

maskarb (dismissed):
lgtm!

lcouzens (approved):
Really appreciate you doing this Cody!! Thank you.

PR #4637, opened 2023-08-17
[COST-4111] Use source_type instead of provider_type in settings tag api

myersCody:
## Jira Ticket

[COST-4111](https://issues.redhat.com/browse/COST-41111)

## Description

This change will use source_type instead of provider type in the tag settings api. The `to_field_name` was not getting picked up, so I made some changes to check for it. 

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Notes

...

### DIFF
--- a/koku/api/settings/tags/serializers.py
+++ b/koku/api/settings/tags/serializers.py
@@ -20,6 +20,12 @@ class SettingsTagSerializer(serializers.Serializer):
     class Meta:
         model = EnabledTagKeys
 
+    def to_representation(self, instance):
+        representation = super().to_representation(instance)
+        add_source_type = {"source_type": instance.provider_type, **representation}
+        add_source_type.pop("provider_type")
+        return add_source_type
+
 
 class ListUUIDSerializer(serializers.ListField):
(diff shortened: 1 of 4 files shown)


dlabrecq (dismissed):

koku/api/settings/tags/serializers.py
```diff
@@ -20,6 +20,12 @@ class SettingsTagSerializer(serializers.Serializer):
     class Meta:
         model = EnabledTagKeys
 
+    def to_representation(self, instance):
+        representation = super().to_representation(instance)
+        add_source_type = {"source_type": instance.provider_type, **representation}
+        add_source_type.pop("provider_type")
+        return add_source_type
```
maskarb, line 27: This is a nitpick, but `to_representation` acts on the entire instance, not just the source-type. I would use something more generic than the `add_source_type` variable name, such as `data`:
```suggestion
    def to_representation(self, instance):
        data = super().to_representation(instance)
        data["source_type"] = data.pop("provider_type")
        return data
```

maskarb (approved):
Lgtm! Just one nitpick:

maskarb (approved):
lgtm!

PR #4644, opened 2023-08-21
[COST-4120] Fix capacity instance counts.

myersCody:
## Jira Ticket

[COST-4120](https://issues.redhat.com/browse/COST-4120)

## Description

This change will change the capacity instance counts for OCP. 

Previously we calculated in the same manner as what we did for capacity, where we summed the values over time. However instead we are just finding the max value. There is a big change to the OCP cluster instance counts:

_if you have 5 nodes on a cluster, we should be summing the number of cores on those nodes. So if a cluster has 5 nodes, and each node has 10 CPU, i would expect to see 50 CPU there_

Since I needed to create the mapping I needed to do a different query that allowed me to group by node. Therefore, I move the cluster annotations to their place in the provider map. 

## Testing

1. Checkout Branch
2. load-test-customer-data
3. Check over all capacity:

With these recent changes the total capacity between cluster & node for the same time range should be the same:

*Memory*
1. Cluster
```
http://localhost:8000/api/cost-management/v1/reports/openshift/memory/
```
Example:
```
 "capacity": {
                "value": 14016.0,
                "units": "GB-Hours",
                "unused": 8130.85,
                "unused_percent": 58.011201484018265,
                "count": 584.0,
                "count_units": "GB"
            },
```
1. Node
```
http://localhost:8000/api/cost-management/v1/reports/openshift/memory/?group_by[node]=*
```
Example:
```
"capacity": {
                "value": 14016.0,
                "units": "GB-Hours",
                "unused": 8130.85,
                "unused_percent": 58.011201484018265,
                "count": 584.0,
                "count_units": "GB"
            },
```

## Notes

...


lcouzens (changes requested):
A couple of small comments, also we should probably include our licensing blurb + a mini description in the new files. For example:
`#
# Copyright 2023 Red Hat Inc.
#
#
"""Dataclass for calculating cluster capacity values."""`

The only other thing id mention as more of a side note is unit tests. I know updating those query handler ones mostly covers things end to end, but I will say I think we should try and write more targeted tests in general, in this case directly for the dataclasses themselves. Like I said, more of a side note, maybe its overkill here and we dont want to do it, but it could be useful to directly validate things going in/out of these classes is all.

All in all great work on this! :+1:

koku/api/report/ocp/provider_map.py
```diff
@@ -301,17 +301,18 @@ def __init__(self, provider, report_type):
                             "request": Sum("pod_request_cpu_core_hours"),
                             "limit": Sum("pod_limit_cpu_core_hours"),
                         },
-                        "capacity_aggregate": {
+                        "capacity_dataclass": {
```
maskarb, line 304: I don't quite understand this name change. This variable IS an aggregate, not a dataclass.

koku/api/report/ocp/query_handler.py
```diff
@@ -326,80 +263,27 @@ def execute_query(self):  # noqa: C901
...
-                if self.resolution == "monthly":
-                    usage_start = usage_start.strftime("%Y-%m")
-                cap_value = entry.get(cap_key, 0)
-                _capacity.add(cap_value, usage_start, level_value)
-                _capacity.add_instance_count(entry.get("capacity_count"), usage_start, level_value)
-        return _capacity
-
-    def _get_cluster_capacity(self, query_data, _capacity):
-        """Calculate the cluster capacity."""
+            _class = NodeCapacity if is_grouped_by_node(self.parameters) else ClusterCapacity
```
maskarb, line 276: What if we are `filter`ing by Node?

koku/api/report/ocp/utils.py
```diff
@@ -0,0 +1,23 @@
+from decimal import Decimal
```
maskarb, line 1: We really shouldn't create utils files for a single function. I would put this in one of the `{cluster|node}_capacity` files, and just import it in the other.

maskarb (commented):
Some comments.

koku/api/report/ocp/query_handler.py
myersCody: We have only ever shown node capacity when grouping by node, it was the requirement for the original epic.

koku/api/report/ocp/utils.py
myersCody: Personally, I think utils is a good place to highlight functions that we use in a multitude of places to perform a specific task. For example, this function is used in the following files:
1. cluster_capacity.py
2. node_capacity.py
3. query_handler.py (we use it to formulate the unused pieces for the query_sum)

For me part of the purpose of this file is to make it obvious that changes to this function impact other areas of the codebase. Whereas if I put it in the node_capacity file, its not clear that the function is used outside of node capacity unless you search the codebase. Granted this is really  just a preference of mine for how I think we should scaffold reusable code snip its. 

However, I'll follow your lead here if you feel strongly about removing the utils file.

koku/api/report/ocp/provider_map.py
myersCody: I was just trying to indicate this section of the provider map is used in the dataclasses. I can undo this naming change though.

koku/api/report/ocp/utils.py
samdoran: Common code in a shared module is a good way to keep the code base organized. Lots of `utils.py` on the other hand just makes it harder to find things. Since this function is used across multiple modules, it would make sense to have it in a common location. Maybe `koku/api/utils.py`? Or just `ocp/__init__.py`?

There's nothing _techincally_ wrong with keeping the function in one of the `{cluster|node}_capacity` files. It just makes for a funny looking import from a neighboring module. Purely aesthetic.

myersCody: Yeah, I didn't see a lot of instances of code being in the  `__init__.py` for this project. I agree a lot of `utils.py` files is a bit much, just wanted to call out its common code. 

I think for now a comment can just highlight that its common code and we can move this one along. Be would be nice to get it off my list of things I have to check each day.

maskarb (approved):

lcouzens (approved):

maskarb (approved):

myersCody:
Hit a timeout during smoke run: 
- https://ci.ext.devshift.net/job/project-koku-koku-pipeline-pr-check-main/649/

PR #4649, opened 2023-08-23
[COST-4125] - adding usecols to AWS downloader flow

lcouzens:
## Jira Ticket

[COST-4125](https://issues.redhat.com/browse/COST-4125)

## Description

This change will add usecols to the AWS downloader flow to filter on columns in our recommended list, shaving off unnecessary columns and also saving us some memory.

## Testing

1. Checkout Branch
2. Restart Koku
3. Make create-test-customer
4. Make load-test-customer-data
5. Everything should process normally

Optionally add a bonus column that DOES NOT match anything in your dataset to `RECOMMENDED_COLUMNS`
Rerun the above and see we continue to process the data even if there is a column missing.

## Notes
Locally this change took our large 10G data set down to 2G

Bogus column example
`{'message': 'pandas read csv with following usecols', 'tracing_id': '20230823T030139Z', 'provider_uuid': '94261544-19ae-491e-b714-ab6f08074cdf', 'provider_type': 'AWS', 'account': 'org1234567', 'usecols': {'BOGUS COL', 'bill/InvoiceId', .................}}`

This `BOGUS COL` was not in my dataset but we processed the data fine

`copy_local_report_file_to_s3_bucket: data/csv/org1234567/AWS/source=94261544-19ae-491e-b714-ab6f08074cdf/year=2023/month=08 /testing/data/processing/org1234567/aws/redhat-cost-management-bucket-0bddd459765c3f6b/2023-08-09_0_0.csv`
...


koku/masu/util/common.py
```diff
@@ -534,11 +534,22 @@ def fetch_optional_columns(local_file, current_columns, fetch_columns, tracing_i
...
         try:
-            data_frame = pd.read_csv(local_file, usecols=lambda col: col.lower().startswith(fetch_column.lower()))
-            data_frame = data_frame.dropna(axis=1, how="all")
-            fetch_cols = data_frame.columns
-            for col in fetch_cols:
-                current_columns.add(col)
+            data_frame = pd.read_csv(
+                local_file, usecols=lambda col: col.lower().startswith(fetch_column.lower()), nrows=0
+            )
+            header_chunks = chunk_columns(data_frame.columns, settings.PANDAS_COLUMN_BATCH_SIZE)
```
maskarb, line 540: Isn't `fetch_column` a single column? If that is the case, do we need to do the column batching?

lcouzens: So this is doing starts with `reousrceTags` and for AWS that can match hundreds of columns depending on tag counts for a customer. Since aws has a column per tag. I ran one of our large customers and not batching we used around 5G, but with batching we stayed under 2G. :smile:

maskarb (approved):

PR #4652, opened 2023-08-24
[COST-4118] - Adding retry logic to ocp on cloud sql query

lcouzens:
## Jira Ticket

[COST-4118](https://issues.redhat.com/browse/COST-4118)

## Description

This change will add some retry logic to ocp on cloud infra map queries.

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Notes

...


samdoran:
I think we should hold off on merging this and instead focus on better coordination and timing of the tasks between workers. Retry logic makes for resilient code in general, but this particular problem is about concurrency, not external service reliability or throttling.

lcouzens:
After investigation this may help the alert but will like mask unintentional file deletions.

PR #4657, opened 2023-08-25
[COST-4124] Improve /settings/tags/ API

samdoran:
## Jira Ticket

[COST-4124](https://issues.redhat.com/browse/COST-4124)
[COST-3585](https://issues.redhat.com/browse/COST-3585)

## Description
- Support multiple choices for the `key` field
- Support both `provider_type` and `source_type` for both types of URL param syntax 
- Return a 400 status code when invalid filter parameters are given

## Testing

1. Checkout Branch
2. Load test customer data
3. `tox -- api.settings.test.tags`

### DIFF
--- a/koku/api/settings/test/tags/test_api.py
+++ b/koku/api/settings/test/tags/test_api.py
@@ -68,12 +68,20 @@ class TagsSettings(IamTestCase):
 
     def test_get_tags_filtering(self):
         test_matrix = (
+            # Custom syntax
             {"filter": {"filter[source_type]": "AWS"}, "key": "source_type", "expected": {"AWS"}},
             {"filter": {"filter[source_type]": ["AWS", "GCP"]}, "key": "source_type", "expected": {"AWS", "GCP"}},
             {"filter": {"filter[key]": "env"}, "key": "key", "expected": {"env", "Environment"}},
             {"filter": {"filter[enabled]": True}, "key": "enabled", "expected": {True}},
             {"filter": {"filter[enabled]": "true"}, "key": "enabled", "expected": {True}},
             {"filter": {"filter[enabled]": "false"}, "key": "enabled", "expected": {False}},
+            # DRF Syntax
+            {"filter": {"source_type": "AWS"}, "key": "source_type", "expected": {"AWS"}},
+            {"filter": {"provider_type": ["AWS", "GCP"]}, "key": "source_type", "expected": {"AWS", "GCP"}},
+            {"filter": {"key": "env"}, "key": "key", "expected": {"env", "Environment"}},
+            {"filter": {"enabled": True}, "key": "enabled", "expected": {True}},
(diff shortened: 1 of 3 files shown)


koku/api/settings/test/tags/test_api.py
```diff
@@ -68,12 +68,20 @@ def test_get_tags(self):
...
+            # Custom syntax
             {"filter": {"filter[source_type]": "AWS"}, "key": "source_type", "expected": {"AWS"}},
             {"filter": {"filter[source_type]": ["AWS", "GCP"]}, "key": "source_type", "expected": {"AWS", "GCP"}},
             {"filter": {"filter[key]": "env"}, "key": "key", "expected": {"env", "Environment"}},
             {"filter": {"filter[enabled]": True}, "key": "enabled", "expected": {True}},
             {"filter": {"filter[enabled]": "true"}, "key": "enabled", "expected": {True}},
             {"filter": {"filter[enabled]": "false"}, "key": "enabled", "expected": {False}},
+            # DRF Syntax
+            {"filter": {"source_type": "AWS"}, "key": "source_type", "expected": {"AWS"}},
+            {"filter": {"provider_type": ["AWS", "GCP"]}, "key": "source_type", "expected": {"AWS", "GCP"}},
```
dlabrecq, line 80: Don't believe we need both source_type and provider_type. Why support both?

dlabrecq (commented):
I'm confused by "Support multiple choices for the key field". The UI can filter by multiple keys; for example, using `filter[key]=admin&filter[key]=access`. This appears to be working today, so curious what's changed? 

Codey also updated the API to return a 400 status for unsupported params like `filter[provider_type]=aws` -- this already works as expected.

koku/api/settings/test/tags/test_api.py
```diff
@@ -133,16 +151,21 @@ def test_tags_order_by(self):
             self.assertEqual(important_values, test_case["expected"])
 
     def test_tags_order_by_invalid(self):
+        test_cases = (
+            {"description": "Standard DRF syntax", "params": {"order_by": "NOPE"}},
+            {"description": "Custom order_by syntax", "params": {"order_by[NOPE]": "asc"}},
```
dlabrecq, line 156: Does this API support two syntaxes; `order_by=key` and `order_by[key]`?

koku/api/settings/test/tags/test_api.py
samdoran: For completeness. Long term, I think we should change the model to avoid needing this.

koku/api/settings/test/tags/test_api.py
samdoran: Yes. The default syntax, `order_by=field` we get for free from Django/DRF. Supporting the dictionary syntax, `order_by[key]`, requires extra work.

PR #4658, opened 2023-08-26
[COST-4133] Fix old ocp/cloud daily parquet file names

lcouzens:
## Jira Ticket

[COST-4133](https://issues.redhat.com/browse/COST-4133)

## Description

This change will address parquet files not getting named with the correct date for the contained daily data.

Before this change we were writing ocp/cloud daily parquet files with random days in each _batched_ file BUT always used the LAST base parquet file name. This meant all files looked like this for the entire processed range `2023-08-26_0_{batch_num}_.parquet` but also these files could contain data for ANY day in the range not just a single day.

Before example:
2023-08-13_0_0_.parquet
2023-08-13_0_1_.parquet
2023-08-13_0_2_.parquet

Now we create correctly dated files and write the relevant dates to said files!

For example:
2023-08-23_0_0.parquet
2023-08-23_0_1.parquet
2023-08-24_0_0.parquet
2023-08-24_0_1.parquet
2023-08-25_0_0.parquet
2023-08-25_0_1.parquet
2023-08-26_0_0.parquet  
2023-08-26_0_1.parquet  

These files above only contain data for the date in the file name.

## Testing

1. Checkout Branch
2. Restart Koku
3. Make create-test-customer + make load-test-customer-data
4. See the files in `koku-bucket/data/parquet/daily/org1234567/AWS/openshift/source=9ee6fc64-7e69-4f45-a041-9b984ca58828/year=2023/month=08` For example are named correctly and contain the correct data relevant to the files name.

If your running with multiple files across multiple workers its also worth checking the manifest entry for file counts in our report_tracker:

`{"2023-08-22_p": 1, "2023-08-23_p": 1, "2023-08-24_p": 1, "2023-08-25_p": 1, "2023-08-26_p": 1}`

## Notes
The `_p` in the report_tracker above is to distinguish between csv/parquet file counts
...

### DIFF
--- a/koku/masu/database/report_manifest_db_accessor.py
+++ b/koku/masu/database/report_manifest_db_accessor.py
@@ -309,6 +309,20 @@ class ReportManifestDBAccessor(KokuDBAccess):
             manifest.save(update_fields=["report_tracker"])
             return f"{day}_{counter}.csv"
 
+    def update_and_get_parquet_batch_counter(self, day, manifest_id):
+        """This is needed for OCP on Cloud filtered daily parquet files"""
+        with transaction.atomic():
+            # With split payloads, we could have a race condition trying to update the `report_tracker`.
+            # using a transaction and `select_for_update` should minimize the risk of multiple
+            # workers trying to update this field at the same time by locking the manifest during update.
+            manifest = CostUsageReportManifest.objects.select_for_update().get(id=manifest_id)
+            if not manifest.report_tracker.get(day):
+                manifest.report_tracker[day] = 0
+            counter = manifest.report_tracker[day]
+            manifest.report_tracker[day] = counter + 1
+            manifest.save(update_fields=["report_tracker"])
+            return counter
+
     def get_manifest_list_for_provider_and_date_range(self, provider_uuid, start_date, end_date):
         """Return a list of GCP manifests for a date range."""
         manifests = (
--- a/koku/masu/processor/parquet/ocp_cloud_parquet_report_processor.py
+++ b/koku/masu/processor/parquet/ocp_cloud_parquet_report_processor.py
@@ -135,9 +135,8 @@ class OCPCloudParquetReportProcessor(ParquetReportProcessor):
         """Create a parquet file for daily aggregated data."""
         if self._provider_type in {Provider.PROVIDER_GCP, Provider.PROVIDER_GCP_LOCAL}:
             if data_frame.first_valid_index() is not None:
-                parquet_base_filename = (
-                    f"{data_frame['invoice_month'].values[0]}{parquet_base_filename[parquet_base_filename.find('_'):]}"
-                )
+                # TODO This may case us to overwrite partial GCP files
+                parquet_base_filename = f"{data_frame['invoice_month'].values[0]}_{parquet_base_filename}"
         file_name = f"{parquet_base_filename}_{file_number}{PARQUET_EXT}"
         file_path = f"{self.local_path}/{file_name}"
         self._write_parquet_to_file(file_path, file_name, data_frame, file_type=self.report_type)
@@ -190,7 +189,11 @@ class OCPCloudParquetReportProcessor(ParquetReportProcessor):
             base_filename = f"{usage_date}_p"
             counter = ReportManifestDBAccessor().update_and_get_parquet_batch_counter(base_filename, manifest_id)
             # Parquet base filename dates here DO NOT match the data written to them
-            base_file_date = parquet_base_filename.split("_")[0]
+            split_base_name = parquet_base_filename.split("_")
+            base_file_date = split_base_name[0]
+            if self.provider_type in (Provider.PROVIDER_GCP, Provider.PROVIDER_GCP_LOCAL):
+                # GCP filenames start with an invoice month before the date
+                base_file_date = f"{base_file_date}_{split_base_name[1]}"
             non_date_base_name = parquet_base_filename.removeprefix(base_file_date)
             usage_day_file_name = f"{usage_date}{non_date_base_name}"
             self.start_date = usage_date
--- a/koku/masu/processor/parquet/ocp_cloud_parquet_report_processor.py
+++ b/koku/masu/processor/parquet/ocp_cloud_parquet_report_processor.py
@@ -181,7 +181,6 @@ class OCPCloudParquetReportProcessor(ParquetReportProcessor):
         }
         date_field = date_fields[self.provider_type]
         unique_usage_days = data_frame[date_field].unique()
-
         for usage_day in unique_usage_days:
             usage_date = pd.to_datetime(usage_day).date()
             # Uniqueify manifest report counter dict entry with p for parquet (dont colide with csvs)
--- a/koku/masu/processor/parquet/ocp_cloud_parquet_report_processor.py
+++ b/koku/masu/processor/parquet/ocp_cloud_parquet_report_processor.py
@@ -138,7 +138,7 @@ class OCPCloudParquetReportProcessor(ParquetReportProcessor):
                 parquet_base_filename = (
                     f"{data_frame['invoice_month'].values[0]}{parquet_base_filename[parquet_base_filename.find('_'):]}"
                 )
-        file_name = f"{parquet_base_filename}_{file_number}_{PARQUET_EXT}"
+        file_name = f"{parquet_base_filename}_{file_number}{PARQUET_EXT}"
         file_path = f"{self.local_path}/{file_name}"
         self._write_parquet_to_file(file_path, file_name, data_frame, file_type=self.report_type)
         self.create_parquet_table(file_path, daily=True, partition_map=self.partition_map)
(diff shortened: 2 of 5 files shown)


koku/masu/processor/parquet/ocp_cloud_parquet_report_processor.py
```diff
@@ -234,14 +242,21 @@ def process(self, parquet_base_filename, daily_data_frames):
...
-                    self.create_ocp_on_cloud_parquet(openshift_filtered_data_frame, parquet_base_filename, i)
+            if self.provider_type in (
+                Provider.PROVIDER_GCP,
+                Provider.PROVIDER_GCP_LOCAL,
+                Provider.PROVIDER_AWS,
+                Provider.PROVIDER_AWS_LOCAL,
+                Provider.PROVIDER_AZURE,
+                Provider.PROVIDER_AZURE_LOCAL,
+            ):
+                self.create_partitioned_ocp_on_cloud_parquet(
```
lcouzens, line 258: Note this function already had the logic to handle AWS/Azure files correctly splitting them into daily files (Sure the naming was wrong, till now) but we never passed them into this for some reason.

koku/masu/processor/parquet/ocp_cloud_parquet_report_processor.py
```diff
@@ -234,14 +242,21 @@ def process(self, parquet_base_filename, daily_data_frames):
                 cluster_topology = accessor.get_openshift_topology_for_multiple_providers(ocp_provider_uuids)
             # Get matching tags
             matched_tags = self.get_matched_tags(ocp_provider_uuids)
-            for i, daily_data_frame in enumerate(daily_data_frames):
-                openshift_filtered_data_frame = self.ocp_on_cloud_data_processor(
-                    daily_data_frame, cluster_topology, matched_tags
-                )
+            daily_data_frames = pd.concat(daily_data_frames, ignore_index=True)
```
lcouzens, line 245: Our current processing flow, during convert to parquet we store daily_data_frames in a list all in memory then previously iterated over that list. The issue there was the df's in this list of df's likely contained dates for any of the range we processed, meaning our batching {i} count would not be consistent between runs.

Since we are storing all these df's in memory anyway we might as well concat them together and process them all at the same time. That way inside `create_partitioned_ocp_on_cloud_parquet` when we partition the df on usage dates we actually get all the dates for the current open df's in a single pass. Simplifying the batch numbers here and between multiple workers.

koku/masu/processor/parquet/ocp_cloud_parquet_report_processor.py
```diff
@@ -185,9 +185,17 @@ def create_partitioned_ocp_on_cloud_parquet(self, data_frame, parquet_base_filen
...
         for usage_day in unique_usage_days:
             usage_date = pd.to_datetime(usage_day).date()
+            # Uniqueify manifest report counter dict entry with p for parquet (dont colide with csvs)
+            # We use the manifest to hold dated file counts for multiple workers processing the same dates in files
+            base_filename = f"{usage_date}_p"
+            counter = ReportManifestDBAccessor().update_and_get_parquet_batch_counter(base_filename, manifest_id)
+            # Parquet base filename dates here DO NOT match the data written to them
+            base_file_date = parquet_base_filename.split("_")[0]
+            non_date_base_name = parquet_base_filename.removeprefix(base_file_date)
+            usage_day_file_name = f"{usage_date}{non_date_base_name}"
```
lcouzens, line 195: One side affect we may need to double check is OCP on GCP! We mentioned recently that the Normal flow for GCP (non ingress) does NOT overwrite parquet files because the data is unique based on what we gather from big query. This may in fact overwrite GCP files now for this flow since the file names will not be (current day date)+batch_num.

That being said the ingress flow WAS likely expecting files to be overwritten here. And working through this I'm fairly sure thats the duplication issue I've been looking at for you know who :car: 

If both of the above is the case we may want to just add a GCP ingress check here for NOW to handle both methods. I will say I do think all of this needs a rethink in the grand scheme of things eventually though.

lcouzens: I have confirmed this is why :car: data was being duplicated in gcp_openshift_daily!

lcouzens:
Smokes passed

koku/masu/processor/parquet/ocp_cloud_parquet_report_processor.py
lcouzens: I also confirmed that the non ingress flow _shouldnt_ overwrite files because of our naming convention here: `day_file = f"{invoice_month}_{partition_date}_{file_name}"` We append the daily file name to the end of the file, and this is used in all our parquet base naming.

koku/masu/database/report_manifest_db_accessor.py
```diff
@@ -309,6 +309,20 @@ def update_and_get_day_file(self, day, manifest_id):
             manifest.save(update_fields=["report_tracker"])
             return f"{day}_{counter}.csv"
 
+    def update_and_get_parquet_batch_counter(self, day, manifest_id):
+        """This is needed for OCP on Cloud filtered daily parquet files"""
+        with transaction.atomic():
```
samdoran, line 314: What happens on a collision? Does one worker wait then retry? I'm not familiar with exactly how these locks work.

koku/masu/processor/parquet/ocp_cloud_parquet_report_processor.py
```diff
@@ -234,14 +244,21 @@ def process(self, parquet_base_filename, daily_data_frames):
...
+                Provider.PROVIDER_AWS,
+                Provider.PROVIDER_AWS_LOCAL,
+                Provider.PROVIDER_AZURE,
+                Provider.PROVIDER_AZURE_LOCAL,
+            ):
+                self.create_partitioned_ocp_on_cloud_parquet(
+                    openshift_filtered_data_frame, parquet_base_filename, manifest_id
+                )
+            else:
+                self.create_ocp_on_cloud_parquet(openshift_filtered_data_frame, parquet_base_filename, file_number=0)
```
myersCody, line 264: Do we need this else if we are calling the previous method for AWS, Azure, and GCP.

koku/masu/database/report_manifest_db_accessor.py
lcouzens: Correct this is the same logic our counters use for csv files already

koku/masu/processor/parquet/ocp_cloud_parquet_report_processor.py
lcouzens: Probably not, but I didnt want to remove all the code for all of these yet. Id rather fix and clean up later.

koku/masu/processor/parquet/ocp_cloud_parquet_report_processor.py
```diff
@@ -182,12 +181,23 @@ def create_partitioned_ocp_on_cloud_parquet(self, data_frame, parquet_base_filen
...
+            # Uniqueify manifest report counter dict entry with p for parquet (dont colide with csvs)
+            # We use the manifest to hold dated file counts for multiple workers processing the same dates in files
+            base_filename = f"{usage_date}_p"
+            counter = ReportManifestDBAccessor().update_and_get_parquet_batch_counter(base_filename, manifest_id)
+            # Parquet base filename dates here DO NOT match the data written to them
+            split_base_name = parquet_base_filename.split("_")
+            base_file_date = split_base_name[0]
+            if self.provider_type in (Provider.PROVIDER_GCP, Provider.PROVIDER_GCP_LOCAL):
+                # GCP filenames start with an invoice month before the date
+                base_file_date = f"{base_file_date}_{split_base_name[1]}"
```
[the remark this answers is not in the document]
lcouzens: We are only doing a single split here so I'm not too worried about this for now :smile:

koku/masu/processor/parquet/ocp_cloud_parquet_report_processor.py
lcouzens: So we are overwriting the original `daily_data_frames` var so memory _shouldnt_ change at all. I did a quick test locally and it didnt seem to make my pod memory go up at all.

koku/masu/processor/parquet/ocp_cloud_parquet_report_processor.py
cgoodfred: Following up on this, the create partitioned OCP on Cloud parquet function was implemented because we partition GCP by `day` as well so the files needed to contain the specific records pertaining to the usage start date. What do we gain here by sending AWS/Azure through this flow? Doesn't this just create more individual parquet files for trino to have to open later?

lcouzens: So the gain here, for AWS/Azure at least would be when deleting the parquet files for our relevant daily processing window we delete the correct corresponding files and data. The partitioned processing within this function is breaking those files out based on their usage dates. :slightly_smiling_face: So this means the files can be named and populated correctly for our deletion on consecutive daily runs.

myersCody (approved):
👍

PR #4659, opened 2023-08-27
[COST-4118] - Clear S3 files before downloading

lcouzens:
## Jira Ticket

[COST-4118](https://issues.redhat.com/browse/COST-4118)

## Description

This change will clear S3 files for processing ranges at the manifest creation level BEFORE download tasks are started!

## Testing

1. Checkout Branch
2. Restart Koku
3. Create OCP on Azure/AWS sources + data
4. See we process the full month
5. Create new Azure/AWS files
6. Hit download
7. See in the logs we clear the processing range of S3 files BEFORE triggering the download tasks
8. Additionally add some sleeps around the deletions to confirm selected files are removed

## Notes

...

### DIFF
--- a/koku/api/utils.py
+++ b/koku/api/utils.py
@@ -197,6 +197,22 @@ class DateHelper:
             return in_date.replace(microsecond=0, second=0, minute=0, hour=0, day=month_end)
         return in_date.replace(day=month_end)
 
+    def latest_date_in_month(self, in_date):
+        """Last day of giving month date or current date (today)."""
+        if isinstance(in_date, str):
+            in_date = parser.parse(in_date).date()
+        if in_date.year < self.today.year or in_date.month < self.today.month:
+            latest_month_date = in_date.replace(day=self.days_in_month(in_date))
+        else:
+            latest_month_date = self.now.date()
+        return latest_month_date
+
+    def datetime_from_date(self, in_date):
+        """convert date to datetime."""
+        if isinstance(in_date, str):
+            in_date = parser.parse(in_date).date()
+        return datetime.datetime.fromordinal(in_date.toordinal())
+
     def next_month(self, in_date):
         """Return the first of the next month from the in_date.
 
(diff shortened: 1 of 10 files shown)


chambridge (commented):
Looks fine to me.

Two points of clarification that would be helpful. Your PR mentioned sleeps but I don't see anything about that in the code is that just stale info? Also I noted that you removed the s3_csv_path, I'm not clear why you don't need to delete both parquet and CSV files now?

koku/api/utils.py
```diff
@@ -197,6 +197,22 @@ def month_end(self, in_date):
...
+            latest_month_date = in_date.replace(day=self.days_in_month(in_date)).date()
+        else:
+            latest_month_date = self.now.date()
+        return latest_month_date
+
+    def datetime_from_date(self, in_date):
+        """convert date to datetime."""
+        if isinstance(in_date, str):
+            in_date = parser.parse(in_date).date()
+        return datetime.datetime.fromordinal(in_date.toordinal())
```
samdoran, line 214: It's cleaner to use `datetime.combine()` and avoid the conversion to an integer.
```suggestion
        return datetime.datetime.combine(in_date, datetime.time())
```

koku/masu/processor/orchestrator.py
```diff
@@ -33,11 +34,14 @@
...
 from masu.processor.tasks import SUMMARIZE_REPORTS_QUEUE_XL
 from masu.processor.worker_cache import WorkerCache
+from masu.util.aws.common import get_or_clear_daily_s3_by_date
 from masu.util.common import check_setup_complete
+from masu.util.common import get_path_prefix
 from subs.tasks import extract_subs_data_from_reports
 from subs.tasks import SUBS_EXTRACTION_QUEUE
 
 LOG = logging.getLogger(__name__)
+dh = DateHelper()
```
samdoran, line 44: I don't think this needs to be a global. An instance attribute seems more appropriate.

If this does end up being global, it should be all caps to indicate that.

PR #4664, opened 2023-08-30
Fix name overwrites once and for all

lcouzens:
## Jira Ticket

## Description

This change will ...

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Notes

...


koku/masu/util/aws/common.py
```diff
@@ -796,7 +796,7 @@ def clear_s3_files(csv_s3_path, provider_uuid, start_date, metadata_key, metadat
             try:
                 existing_object = obj_summary.Object()
                 metadata_value = existing_object.metadata.get(metadata_key)
-                if metadata_value != metadata_value_check:
+                if str(metadata_value) != str(metadata_value_check):
```
maskarb, line 799: What is the type cast for?

lcouzens: We had an issue where we were comparing an int to str. Just trying to prevent that.

maskarb (dismissed):

chambridge (dismissed):

chambridge (approved):

PR #4670, opened 2023-09-06
[COST-3797] Deprecate Settings endpoint

myersCody:
## Jira Ticket

[COST-3797](https://issues.redhat.com/browse/COST-3797)

## Description

This change will deprecate our old settings endpoint. The same functionality is still available to users, just found in different endpoints.

Steps I took to Deprecate:
1. Renamed the api folder to have the “deprecated_” prefix so that we know everything in that directory would be considered dead code in the future.
    - koku/api/settings/settings.py → koku/api/deprecated_settings/settings.py
2. Checked unleash flags in deprecated code, and removed any unnecessary flags.
3. Moved any common functions that were no longer necessary into the deprecated_ folder. [Example](https://github.com/project-koku/koku/pull/4670/files#diff-856c37712ae23d09edf02c072ac4652e41b0561af47796b099c5d512c1db9684R34-R83)
4. Create `deprecated_` versions of any functions functions till being used. still in use so that we we don't have to worry about backwards compatibility as we move forward. Example: `deprecated_set_cost_type`

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Notes

...


lcouzens (commented):
Removing code is the best!!! Love a good clean up. We have no testing instructions on this PR, but I assume its a case of `Hey we shouldnt have regressions from removing dead code`. That being said I guess we need to work through all the smoke test fails and rectify those, either we did break something of the tests need updating along side. :bow:

koku/koku/middleware.py
```diff
@@ -105,35 +101,6 @@ def __init__(self, dikt):
         super().__init__(data)
 
 
-class KokuTenantSchemaExistsMiddleware(MiddlewareMixin):
-    """A middleware to check if schema exists for Tenant."""
-
-    def process_exception(self, request, exception):
-        if isinstance(exception, (Tenant.DoesNotExist, ProgrammingError)):
```
esebesto, line 112: we may need this? and remove just the inner if block?

myersCody: Yes, thank you @esebesto

I got a little too heavy handed with the removal of the middleware. 🙇

myersCody:
> Removing code is the best!!! Love a good clean up. We have no testing instructions on this PR, but I assume its a case of `Hey we shouldnt have regressions from removing dead code`. That being said I guess we need to work through all the smoke test fails and rectify those, either we did break something of the tests need updating along side. 🙇

Yup! Got a little too heavy handed with the middleware removal. Thats what I get for rushing to get this done before my PTO. Thanks everyone!

samdoran:
From looking at the request logs, we have a few users still using the `/settings/` API. I'm wondering if we should return a 410 or a warning header.

myersCody:
Developer Note: 

There is some obvious duplication of code in this one. However, the idea is that all we need to do to remove the deprecated code in the future is delete the `deprecated_settings/` directory. I have moved the logic that is still be used by the new settings to the `user_settings/` directory.

samdoran (commented):
Is there a compelling reason for moving everything to user_settings? I was actually thinking of moving the code in `user_settings` into `settings`.

koku/api/common/deprecate_view.py
```diff
@@ -0,0 +1,83 @@
...
+# Example:
+# class SettingsView(APIView):
+#     """
+#     View to interact with settings for a customer.
+#     """
+#     deprecation_datetime = datetime(2023, 9, 22)
+#     sunset_datetime = datetime(2023, 1, 10)
+#     link = "https://github.com/project-koku/koku/pull/4670"
+
+# Then import deprecate_view wrapper and it to the view in the
```
samdoran, line 39:
```suggestion
# Then import deprecate_view wrapper and add it to the view in the
```

koku/api/common/deprecate_view.py
```diff
@@ -0,0 +1,83 @@
...
+        if hasattr(self.viewclass, "sunset_datetime"):
+            sunset_datetime = getattr(self.viewclass, "sunset_datetime")
+            response["Sunset"] = sunset_datetime.strftime(HTTP_DATE_FORMAT)
+            if sunset_datetime < DateHelper(True).now:
+                self.sunset_endpoint = True
+        if hasattr(self.viewclass, "deprecation_datetime"):
+            deprecation_datetime = getattr(self.viewclass, "deprecation_datetime")
+            response["Deprecation"] = deprecation_datetime.strftime(HTTP_DATE_FORMAT)
+        if hasattr(self.viewclass, "link"):
+            response["Link"] = getattr(self.viewclass, "link")
```
samdoran, line 63: This can be simplified a bit.
```suggestion
        if sunset_datetime := getattr(self.viewclass, "sunset_datetime", None):
            response["Sunset"] = sunset_datetime.strftime(HTTP_DATE_FORMAT)
            if sunset_datetime < DateHelper(True).now:
                self.sunset_endpoint = True
        if deprecation_datetime := getattr(self.viewclass, "deprecation_datetime", None):
            response["Deprecation"] = deprecation_datetime.strftime(HTTP_DATE_FORMAT)
        if link := getattr(self.viewclass, "link", None):
            response["Link"] = link
```

koku/api/common/deprecate_view.py
samdoran: This seems like it should be the doctoring for the `DeprecationEndpoint` class.

koku/api/common/deprecate_view.py
```diff
@@ -0,0 +1,83 @@
...
+# path("settings/", deprecate_view(SettingsView.as_view()), name="settings"),
+
+
+@dataclass
+class DeprecateEndpoint:
+    viewclass: t.Callable
+    sunset_endpoint: bool = field(init=False, default=False)
+
+    def extract_data_from_view(self, response):
+        """Checks that the view class has the correct attributes and adds headers."""
```
samdoran, line 52: "Extract" makes me think data will be returned but this method is mutating a passed in object.
```suggestion
    def update_response_headers(self, response):
        """Add sunset, deprecation, and link headers to the response if set on self.viewclass"""
```

koku/api/common/deprecate_view.py
myersCody: Thats a good point! Its a better place for this.

koku/api/common/deprecate_view.py
```diff
@@ -0,0 +1,79 @@
...
+        link = "https://github.com/project-koku/koku/pull/4670"
+
+    Then import deprecate_view wrapper and add it to the view in the
+    urls.py
+
+    Example:
+        path("settings/", deprecate_view(SettingsView.as_view()), name="settings"),
+    """
+
+    viewclass: t.Callable
```
samdoran, line 46: This should be an APIVew object not a Callable. The Callable type is for functions or other callable code, not instances of objects.
```suggestion
    viewclass: APIView # from rest_framework.view import APIView
```

samdoran (dismissed):

myersCody:
Hit a timeout during smoke tests:
https://ci.ext.devshift.net/job/project-koku-koku-pipeline-pr-check-main/653/

samdoran (approved):

PR #4675, opened 2023-09-08
[COST-4182] Return status code 412 when enabled tags limit is exceeded

samdoran:
## Jira Ticket

[COST-4182](https://issues.redhat.com/browse/COST-4182)

## Description

Include limit and number of enabled tags in the response body to allow for localization of the message.

## Testing

1. Checkout Branch
2. Load customer data
3. Set `TAG_ENABLED_LIMIT` to a small number such as 5
4. Try to enabled a tag.

```python
import json
import urllib.request

BASE_URL = "http://localhost:8000/api/cost-management/v1/settings/tags"


def get_disabled_tag_ids():
    url = f"{BASE_URL}/?enabled=0&limit=1000"
    with urllib.request.urlopen(url) as response:
        response = json.loads(response.read())
    
    return [item["uuid"] for item in response["data"]]


def enable_tags(ids):
    headers = {
        "Content-Type": "application/json",
    }
    url = f"{BASE_URL}/enable/"
    b_data = json.dumps({"ids": ids}).encode('utf-8')
    req = urllib.request.Request(url, headers=headers, data=b_data, method="PUT")
    with urllib.request.urlopen(req):
        pass
        
enable_tags(get_disabled_tag_ids())
```


chambridge:
Do we need to update the openapi spec file?

samdoran:
Yup, I will update the docs.

maskarb (approved):

PR #4677, opened 2023-09-10
[COST-4188] - Allow split azure ingress files

lcouzens:
## Jira Ticket

[COST-4188](https://issues.redhat.com/browse/COST-4188)

## Description

This change will allow multiple split files to be sent via Azure's customer filtered ingress flow, we also added an unlesah flag to skip column checks

## Testing

1. Checkout Branch
2. Restart Koku
3. Add azure storage only source
4. Post multiple split files for your source
5. make sure they all process correctly

## Notes

...


chambridge (approved):
LGTM

PR #4680, opened 2023-09-12
add steps to build koku-minio using Dockerfile.minio

pedrolp85:
## Jira Ticket

Note: I run into some error while using pre-commit linting, please run linting yourself before merging

## Description

This is a suggested change to  include the steps to build koku-minio using Dockerfile.minio in this repo

## Testing

BEFORE:
`make docker-up-min-with-subs `
runs the following docker-compose commands:

```
/usr/bin/docker-compose build koku-base
/usr/bin/docker-compose up -d db
/usr/bin/docker-compose up -d unleash
/usr/bin/docker-compose up -d --scale koku-worker=1 redis koku-server masu-server koku-worker trino hive-metastore
/usr/bin/docker-compose up -d --scale subs-worker=1 subs-worker
```

leading to an error in the koku-minio container:
```

73b29e7c6433   local-minio:latest                                   "/usr/bin/docker-ent…"   25 seconds ago   Exited (1) 24 seconds ago                                                                                              kokuminio
((koku) ) [plopezpe@fedora koku]$ docker logs 73b29e7c6433
ERROR Unable to initialize backend: mkdir /data/.minio.sys: permission denied
```

AFTER:
`make docker-up-min-with-subs`
```

/usr/bin/docker-compose build koku-base
/usr/bin/docker-compose up -d db
/usr/bin/docker-compose up -d unleash
/usr/bin/docker-compose up -d --scale koku-worker=1 redis koku-server masu-server koku-worker trino hive-metastore
/usr/bin/docker-compose up --build -d trino hive-metastore
/usr/bin/docker-compose up -d --scale subs-worker=1 subs-worker
```

The instruction /usr/bin/docker-compose up --build -d trino hive-metastore
will use the Dockerfile.minio to build kokuminio because of set up dependencies in docker-compose(hive-> create-parquet-bucket-> koku-minio)

## Notes

...

PR #4682, opened 2023-09-13
Ensure consistent permissions for bind mounts

samdoran:
## Description

Closes #4680 

Explicitly create and set the permissions on directories that are bind mounted into containers.

On Linux, the Docker daemon runs as `root`. Any bind mounted directories that do not exist on the host are created and owned by `root`. On macOS, the Docker daemon is running in user space. Bind mounted directories are created on the host owned by the current user. This results in intermittent permissions errors.

Add a `USER_ID` arg to the `Makefile`, Docker Compose file, and minio Dockerfile to ensure consistent ownership of bind mounted files on Linux and macOS.

Avoid bind mounting the PostgreSQL data directory directly to allow the entry point script to chown the data directory and avoid some permissions issues (using a [volume](https://docs.docker.com/storage/volumes/) instead of a bind mount would still be far better).

Change `make` targets that clear out persistent data to delete the _contents_ of directories and not the directories themselves.

## Testing

1. Checkout Branch
2. Completely nuke the local environment
```
make docker-down
make delete-trino # docker-down should run this, but sometimes the preceding make target fails
make delete-trino-data
make delete-db
```
3. Recreate the environment
```
make docker-up-min
make create-test-customer
make load-test-customer-data
```
4. Profit


pedrolp85:
Created a new venv in a new directory and test this one

Volumes and bind mounts have now consistent uid and gid with other files in the repo

drwxr-xr-x 1 plopezpe plopezpe     34 sep 14 09:04 .trino
-rw-r--r-- 1 plopezpe plopezpe    278 sep 14 08:59 pgadmin_servers.json.example
drwxr-xr-x 1 plopezpe plopezpe     16 sep 14 09:04 pg_data
drwxr-xr-x 1 plopezpe plopezpe     42 sep 14 08:59 pg_init

Spin it up with `make docker-up-min-trino` and intest ocp and aws data and all ok
Also `make docker-up-min-with-subs` runs kokuminio container with no issues and swatch test passed inlocal

pedrolp85 (approved):

maskarb (approved):

PR #4690, opened 2023-09-19
[COST-4209] Fix Masu download endpoint query params

maskarb:
## Jira Ticket

[COST-4209](https://issues.redhat.com/browse/COST-4209)

## Description

The download endpoint should:
1. when providing a provider_uuid arg, will override the polling timestamp
2. when providing a provider_type, will only return those providers, and will respect the polling_timestamp and batch-size

This PR fixes the /download endpoint to respect the above criteria. It also adds some logging to show args passed to the orchestrator and how many accounts will be polled.

## Testing

set polling timer to a long period:
```
POLLING_TIMER=86400  # 24 hours
```

```
make create-test-customer;
```

some endpoints to test:
```
these will only poll the providers when they are older than the polling timestamp
(ocp only works when DEVELOPMENT=True)
http://localhost:5042/api/cost-management/v1/download/?provider_type=OCP
http://localhost:5042/api/cost-management/v1/download/?provider_type=AWS
http://localhost:5042/api/cost-management/v1/download/?provider_type=AWS-local

this always polls the provider:
http://localhost:5042/api/cost-management/v1/download/?provider_uuid={a uuid}

```

## Notes

...

### DIFF
--- a/koku/masu/processor/orchestrator.py
+++ b/koku/masu/processor/orchestrator.py
@@ -67,7 +67,6 @@ class Orchestrator:
         self.queue_name = queue_name
         self.ingress_reports = kwargs.get("ingress_reports")
         self.ingress_report_uuid = kwargs.get("ingress_report_uuid")
-        self._polling_accounts = Provider.polling_objects.get_accounts()
         self._summarize_reports = kwargs.get("summarize_reports", True)
 
     def get_polling_batch(self):
--- a/koku/masu/processor/orchestrator.py
+++ b/koku/masu/processor/orchestrator.py
@@ -70,15 +70,13 @@ class Orchestrator:
         self._summarize_reports = kwargs.get("summarize_reports", True)
 
     def get_polling_batch(self):
-        if self.scheduled:
-            providers = Provider.polling_objects.get_polling_batch(settings.POLLING_BATCH_SIZE)
+        if not self.scheduled and self.provider_uuid:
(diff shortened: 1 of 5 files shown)


koku/masu/processor/orchestrator.py
```diff
@@ -67,12 +67,18 @@ def __init__(
         self.queue_name = queue_name
         self.ingress_reports = kwargs.get("ingress_reports")
         self.ingress_report_uuid = kwargs.get("ingress_report_uuid")
-        self._polling_accounts = Provider.polling_objects.get_accounts()
         self._summarize_reports = kwargs.get("summarize_reports", True)
 
     def get_polling_batch(self):
+        if not self.scheduled and self.provider_uuid:
+            providers = Provider.objects.filter(uuid=self.provider_uuid)
```
maskarb, line 74: Using the `objects` Manager overrides the polling-timestamp.

koku/masu/processor/orchestrator.py
```diff
@@ -67,12 +67,18 @@ def __init__(
...
+        else:
+            filters = {}
+            if self.provider_type:
+                filters["type"] = self.provider_type
```
maskarb, line 78: setting this filter allows us to filter on the provider-type and respect the polling-timestamps by using the `polling_objects` Manager.

cgoodfred (approved):

PR #4693, opened 2023-09-20
just delete bucket contents

maskarb:
* revert to deleting the contents of the buckets instead of the bucket itself
* chmod recursively
* delete pg_data/data instead of pg_data/* because it deletes the `.gitignore`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -173,7 +173,7 @@ delete-trino:
 	@$(PREFIX) rm -rf $(TOPDIR)/.trino/trino/*
 
 delete-trino-data:
-	@$(PREFIX) rm -rf $(TOPDIR)/.trino/parquet_data/{*,.minio*}/**
+	@$(PREFIX) rm -rf $(TOPDIR)/.trino/parquet_data/koku-bucket/*
 
 delete-redis-cache:
 	$(DOCKER) exec -it koku_redis redis-cli -n 1 flushall
--- a/Makefile
+++ b/Makefile
@@ -429,6 +429,7 @@ docker-up-min-trino: docker-up-min docker-trino-up
 
 docker-up-min-trino-no-build: docker-up-min-no-build docker-trino-up-no-build
 
+
 ### Source targets ###
 ocp-source-from-yaml:
 #parameter validation
--- a/Makefile
+++ b/Makefile
@@ -401,7 +401,7 @@ docker-iqe-api-tests: docker-reinitdb _set-test-dir-permissions delete-testing
 docker-iqe-vortex-tests: docker-reinitdb _set-test-dir-permissions delete-testing
 	./testing/run_vortex_api_tests.sh
 
-CONTAINER_DIRS = $(TOPDIR)/pg_data $(TOPDIR)/.trino/{parquet_data,trino}
+CONTAINER_DIRS = $(TOPDIR)/pg_data/data $(TOPDIR)/.trino/{parquet_data,trino}
 docker-host-dir-setup:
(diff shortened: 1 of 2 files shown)


samdoran:
> delete pg_data/data instead of pg_data/* because it deletes the .gitignore

That's only true if the `dotglob` shell option is enabled But we will need to do something similar for a different reason.

The init script that runs in the Postgres container image requires the data directory to be completely empty or it will fail to load. I was trying to avoid having to create the directory after a checkout, but it seems like that's unavoidable. `git` doesn't allow you to commit empty directories and Postgres requires a completely empty data directory. Quite a pickle.

Makefile
```diff
@@ -214,7 +214,7 @@ make-migrations:
 	$(DJANGO_MANAGE) makemigrations api reporting reporting_common cost_models
 
 delete-db:
-	$(PREFIX) rm -rf $(TOPDIR)/pg_data/*
+	$(PREFIX) rm -rf $(TOPDIR)/pg_data/data
```
samdoran, line 217: Just delete the contents of the directory.
```suggestion
	$(PREFIX) rm -rf $(TOPDIR)/pg_data/data/*
```

Makefile
```diff
@@ -406,7 +406,7 @@ docker-host-dir-setup:
 	$(DOCKER_COMPOSE) build --no-cache koku-minio
 	mkdir -p -m 0755 $(CONTAINER_DIRS) 2>&1 > /dev/null
 	chown -R $(USER_ID) $(CONTAINER_DIRS)
-	chmod 0755 $(CONTAINER_DIRS)
+	chmod -R 0755 $(CONTAINER_DIRS)
```
samdoran, line 409: We will need to create an empty subdir within `pg_data`
```makefile
CONTAINER_DIRS = $(TOPDIR)/pg_data/data $(TOPDIR)/.trino/{parquet_data,trino}
```

samdoran: Also, the  data mount in the `docker-compose.yml` file needs to change.

```patch
diff --git docker-compose.yml docker-compose.yml
index 430062cf..1546ce7a 100644
--- docker-compose.yml
+++ docker-compose.yml
@@ -452,7 +452,7 @@ services:
     ports:
       - 15432:5432
     volumes:
-      - ./pg_data:/var/lib/${DATABASE_DATA_DIR-postgresql}
+      - ./pg_data/data:/var/lib/${DATABASE_DATA_DIR-postgresql}/data
       - ./pg_init:/docker-entrypoint-initdb.d
     command:
       # This command give more precise control over the parameter settings

```

samdoran: This note from the [readme](https://hub.docker.com/_/postgres) explains what's going on:

> If the data volume you're using is a filesystem mountpoint (like with GCE persistent disks), or remote folder that cannot be chowned to the postgres user (like some NFS mounts), or contains folders/files (e.g. lost+found), Postgres initdb requires a subdirectory to be created within the mountpoint to contain the data.

Makefile
```diff
@@ -173,7 +173,7 @@ delete-trino:
 	@$(PREFIX) rm -rf $(TOPDIR)/.trino/trino/*
 
 delete-trino-data:
-	@$(PREFIX) rm -rf $(TOPDIR)/.trino/parquet_data/{*,.minio*}
+	@$(PREFIX) rm -rf $(TOPDIR)/.trino/parquet_data/{*,.minio*}/**
```
samdoran, line 176: Let's just change this to deleting everything in the bucket directory.
```suggestion
	@$(PREFIX) rm -rf $(TOPDIR)/.trino/parquet_data/koku-bucket/*
```

PR #4696, opened 2023-09-22
Improve PostgreSQL and minio container builds

samdoran:
## Description

Make the UID and GID of the `postgres` user in the container match those of the current user. This makes the PostgreSQL data files easy to manage from the container host. Build a PostgreSQL image with the correct UID, GID, and our entry point script baked in. The fewer bind mounts the better.

Use environment variables to configure the UID and GID of the minio container. There is an entry point script that sets up the user account, so we don't need to maintain a separate container file.

Move container files into `dev/containers` . This keeps the repo root clean and makes builds easier.

## Testing

1. Checkout Branch
2. Delete containers and local data

```
make docker-down
make delete-db
make delete-trino
make delete-trino-data
rm -rf pg_data/*
rm -rf .trino/parquet_data/.minio.sys
```
3. Start it all back up: `make docker-up-min`
4. Check that things came up ok `make docker-logs`
5. Load some data `make load-test-customer-data`

## Notes

I would like to move all the dev container related files and data to `dev/containers`.


docker-compose.yml
```diff
@@ -569,38 +573,37 @@ services:
...
-    image: local-minio:latest
-    build:
-        context: .
-        dockerfile: Dockerfile.minio
-        args:
-          USER_ID: ${USER_ID-1000}
-    container_name: kokuminio
+  minio:
+    container_name: koku-minio
+    image: minio/minio:RELEASE.2022-05-26T05-48-41Z.hotfix.15f13935a
```
maskarb, line 578: I know that this is the image we were using previously, but maybe we should update this to a newer image? It's definitely not important, but a nice to have since we're already changing these files around.

samdoran: I'm totally ok with updating the image. Just didn't want to change too much at once.

samdoran: The hotfix tags don't have an ARM build anyway. I've been using a more recent image locally for quite a while.

lcouzens (dismissed):

samdoran:
One other thing I would like to do is unify the service and container names. All containers would have a `koku-` prefix and use dashes instead of underscores.  Service names would not have the `koku-` prefix. This is a bit disruptive because it requires updating any values in the `.env` file locally.

Here is what the containers look like currently:

```
trino
hive-metastore
koku-koku-worker-1
koku_server
masu_server
koku-koku-base-1
koku-koku-create-parquet-bucket-1
kokuminio
koku_redis
koku-unleash
koku_db
```

This is what I would like to change them to:
```
koku-trino
koku-hive-metastore
koku-worker-1
koku-masu-server
koku-server
koku-create-parquet-bucket
koku-base
koku-minio
koku-redis
koku-unleash
koku-db
```

I can do that in a separate PR if it's less disruptive.

docker-compose.yml
```diff
@@ -1,4 +1,5 @@
 version: '3'
+name: koku
```
lcouzens, line 2: This is causing issues for me :suspect: 

```
(koku) [lcouzens@fedora-desktop koku]$ make docker-up-min-no-build && make docker-logs
/usr/bin/docker-compose up -d db
ERROR: The Compose file './docker-compose.yml' is invalid because:
'name' does not match any of the regexes: '^x-'

You might be seeing this error because you're using the wrong Compose file version. Either specify a supported version (e.g "2.2" or "3.3") and place your service definitions under the `services` key, or omit the `version` key and place your service definitions at the root of the file to use version 1.
For more on the Compose file format versions, see https://docs.docker.com/compose/compose-file/
make: *** [Makefile:374: docker-up-db] Error 1

samdoran: That seems like you're using the legacy version of Docker Compose. What does `docker compose` version show? I'm guessing that failed and falls back to `docker-compose`.

lcouzens: Thats a fair point, I probably do need to update stuff :shark:

lcouzens (approved):

PR #4697, opened 2023-09-22
ingress-flow-check-customer-for-source

lcouzens:
## Jira Ticket

## Description

This change will check that the source used for posting reports is actually a source accessible by the customer.

More importantly, prevent customers posting data over other customers sources. (If they guess the provider uuid)

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Notes

...


koku/api/ingress/reports/view.py
```diff
@@ -86,6 +86,7 @@ def post(self, request):
             "reports_list": request.data.get("reports_list"),
             "bill_year": request.data.get("bill_year"),
             "bill_month": request.data.get("bill_month"),
+            "customer": request.user.customer_id,
```
chambridge, line 89: Should we log the data object here for debugability? Should we get the schema since the customer id isn't a super visible value?

chambridge (commented):
Couple comments.

koku/api/ingress/reports/view.py
lcouzens: So I was looking at that but the silly models wouldnt let me :frowning_face: I was probably just over thinking/doing something wrong. Happy to change it if I can figure it out.

chambridge: I don't see any logging, but I guess we will be able to use gabi to make sure the schema value from the request makes sense.

chambridge (approved):
👍

koku/api/ingress/reports/serializers.py
```diff
@@ -37,49 +37,49 @@ class Meta:
             "sources_id",
             "bill_year",
             "bill_month",
+            "schema_name",
```
samdoran, line 40: Does a serialized needed to be added for this new field?

lcouzens: Correct. It wouldnt pass along the schema data without adding it here.

koku/api/ingress/reports/serializers.py
```diff
@@ -37,49 +37,49 @@ class Meta:
...
-                        f"Reports for billing month {bill_month} are currently already being processed. "
-                        "New reports cannot be processed until the current reports are completed."
-                    )
-                    raise serializers.ValidationError(error_obj(key, message))
-                return data
-            key = "bill_period"
-            message = (
-                f"Invalid bill, year must be {year_message} and month must be {month_range[0]} or {month_range[1]}"
-            )
+        if source.type.lower() not in PROVIDER_LIST:
```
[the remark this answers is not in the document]
lcouzens: Valid point, that said if we don't have a source the post would fail before this  check anyway :+1:

myersCody (approved):

PR #4699, opened 2023-09-23
[COST-4224] Only read currency API data from a good response

samdoran:
## Jira Ticket

[COST-4224](https://issues.redhat.com/browse/COST-4224)

## Description

Check the response before decoding the response data. If the response is unsuccessful, there will be no data, which results in an exception from the .json() method. This obscures the real problem, which is that the API did not return data.

Add retry with exponential backoff for resiliency.

Correct broken tests.

## Testing

`tox -- masu.test.celery.test_tasks`


koku/masu/test/celery/test_tasks.py
```diff
@@ -147,23 +149,42 @@ def test_sync_data_to_customer_max_retry(self, mock_sync, mock_data_export_reque
...
-            status_code=201, json=lambda: {"result": "success", "rates": {"AUD": 1.37, "CAD": 1.25, "CHF": 0.928}}
-        )
+
+        mock_response = Mock(status_code=201)
+        # fmt: off
+        mock_response.json.return_value = {
+            "result": "success",
+            "rates": {"AUD": 1.37, "CAD": 1.25, "CHF": 0.928}
+        }
+        # fmt: on
```
maskarb, line 181: Are these formatting commands necessary? Could we do:
```suggestion
        mock_response.json.return_value = {
            "result": "success",
            "rates": {"AUD": 1.37, "CAD": 1.25, "CHF": 0.928},
        }
```

maskarb (commented):
A comment and a suggestion:

koku/masu/test/celery/test_tasks.py
samdoran: black was collapsing into one line since it's not quite as wide is line 200. I thought it was easier on the reader having the two test cases look visually similar, so I had to disable the formatting.

koku/masu/test/celery/test_tasks.py
```diff
@@ -147,23 +149,42 @@ def test_sync_data_to_customer_max_retry(self, mock_sync, mock_data_export_reque
...
 
     # Check to see if Error is raised on wrong URL
     @patch("masu.celery.tasks.requests")
     def test_error_get_currency_conversion_rates(self, mock_requests):
-        mock_requests.get.side_effect = Exception("error")
-        with self.assertRaises(Exception) as e:
-            tasks.get_daily_currency_rates()
-            self.assertIn("Couldn't pull latest conversion rates", str(e.exception))
+        mock_session = Mock()
+        mock_session.get.side_effect = HTTPError("Raised intentionally")
```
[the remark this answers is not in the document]
samdoran: Oh neat. I didn't know about `requests-mock`. That may be a lot cleaner. It took me a bit to get this test working and it was a bit complicated. I'll see what it looks like using `requests-mock`.

koku/masu/test/celery/test_tasks.py
maskarb: Black is collapsing it because of the missing trailing comma after the rates dict (I think).

samdoran: That was it.

koku/masu/test/celery/test_tasks.py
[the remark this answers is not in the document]
samdoran: This was _way nicer_ with `requests_mock`. Thanks for the suggestion!

PR #4708, opened 2023-09-29
[COST-4235] Properly handle null values in the tags settings API

samdoran:
## Jira Ticket

[COST-4235](https://issues.redhat.com/browse/COST-4235)

## Description

Add some basic validation to the now somewhat misnamed non-validating filter. Switch to using `AllChoicesFilter` since that is more in line with what we want and it has the side effect of treating null values as an invalid choice.

## Testing

1. Checkout Branch
2. Restart Koku
3. `curl localhost:8000/api/cost-management/v1/settings/tags/source_type=%00`


lcouzens (approved):
Nice

PR #4717, opened 2023-10-04
update trino and hive images for local dev

maskarb:
* update trino ARM images
* update trino configuration
* move trino/hadoop configs into dev/containers and update volume mounts


samdoran (dismissed):

myersCody:
I tried ingesting a source locally with this branch and ran into this issue:
```
koku-koku-worker-1  |     raise self._process_error(response["error"], response.get("id"))
koku-koku-worker-1  | trino.exceptions.TrinoUserError: TrinoUserError(type=USER_ERROR, name=NOT_SUPPORTED, message="Unsupported Trino column type (varchar) for Parquet column ([node_role] optional int32 node_role)", query_id=20231010_195704_00060_5giaz)
koku-koku-worker-1  | [2023-10-10 19:57:04,859] ERROR 28fb8e2d-d833-4c72-94a2-7da5c8e3c963 62 Task failed: TrinoUserError(type=USER_ERROR, name=NOT_SUPPORTED, message="Unsupported Trino column type (varchar) for Parquet column ([node_role] optional int32 node_role)", query_id=20231010_195704_00060_5giaz)
```

myersCody (dismissed):

dev/containers/hadoop/hadoop-config/core-site.xml
samdoran, line 1: Since the hadoop config is used by the trino container, it should go in `/dev/containers/trino/etc`. Then you may be able to bind mount `dev/containers/trino/etc/:/etc/trino/`.

maskarb: I can move the hadoop file, but they all need to be mounted individually because `/etc/trino` contains files that are wiped out if mounted this way. Is there a way to merge the files in the container with the files from the volume mount?

bacciotti (approved):

dev/containers/hadoop/hadoop-config/core-site.xml
samdoran: The bind mount would replace `/etc/trino` in the container. There is now way to merge the contents.

We are mounting a little than half of the config files. We could copy the remaining config files to the container host.
```
/etc/trino/
├── catalog
│   ├── blackhole.properties
│   ├── config.yaml
│   ├── hive.properties  # current bind mount
│   ├── jmx.properties
│   ├── memory.properties
│   ├── tpcds.properties
│   └── tpch.properties
├── config.properties  # current bind mount
├── hadoop-config
│   └── core-site.xml  # current bind mount
├── jvm.config # current bind mount
├── log.properties  # current bind mount
└── node.properties
```

PR #4727, opened 2023-10-09
add the ability to purge a specific queue of tasks

lcouzens:
## Jira Ticket

[COST-4287](https://issues.redhat.com/browse/COST-4287)

## Description

This change will give us the ability to clear a specific queue of tasks without needing to clear everything via our internal endpoints.

## Testing

1. Checkout Branch
2. Restart Koku
3. Make load some data or add a bunch of summary tasks or w/e to a queue 
4. Hit http://localhost:5042/api/cost-management/v1/clear_celery_queues/?queue=priority OR whatever queue you want to clear. See that queue gets set to 0

## Notes

...

### DIFF
--- a/koku/masu/api/running_celery_tasks.py
+++ b/koku/masu/api/running_celery_tasks.py
@@ -68,17 +68,25 @@ def clear_celery_queues(request):
     if request.method == "GET":
         params = request.query_params
         clear_all = params.get("clear_all")
-        LOG.info(f"Clearing all queues parameter: {clear_all}")
+        queue = params.get("queue")
 
         # clear default_celery queue
         purged_tasks = app.control.purge()
 
         # clear all queues
         if clear_all:
+            LOG.info(f"Clearing all queues parameter: {clear_all}")
             queue_lengths = list(collect_queue_metrics().values())
             purged_tasks += sum(queue_lengths)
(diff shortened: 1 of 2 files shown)


koku/masu/api/running_celery_tasks.py
```diff
@@ -71,18 +71,26 @@ def clear_celery_queues(request):
     if request.method == "GET":
         params = request.query_params
         clear_all = params.get("clear_all")
-        LOG.info(f"Clearing all queues parameter: {clear_all}")
+        queue = params.get("queue")
 
         # clear default_celery queue
         purged_tasks = app.control.purge()
```
maskarb, line 77: I presume if we want to clear a specific queue, we don't want to run this?

lcouzens: So this one actually only clears the celery queue. Honestly I'm not sure why we even have this, I feel like it does nothing. But at the same time this one gets run when you just hit `http://localhost:5042/api/cost-management/v1/clear_celery_queues/` regardless of whether you pass in `?clear_all` or another queue.

lcouzens: I think maybe we just remove this line, since its highly unlikely that we want to clear the celery queue ever anyway.

djnakabaale (approved):
lgtm 👍🏾

PR #4733, opened 2023-10-14
fix OCP post processor volume labels

maskarb:
## Jira Ticket

## Description

The volume labels in the dataframes we write to parquet are in the `persistentvolume_labels` and `persistentvolumeclaim_labels`, not the `volume_labels` column.

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Notes

...


lcouzens:
Is there a jira issue for this? And I assume this has no regression affects on the all_labels stuff that Cody has been working on? Since that collates pod_labels and volume_labels together, so does this change mean volume_labels translates to persistentvolume_labels and persistentvolumeclaim_labels instead?

lcouzens (approved):

PR #4735, opened 2023-10-16
[COST-4289] - Skip OCP clusters without a matching billing report date for OCP on Cloud processing

lcouzens:
## Jira Ticket

[COST-4289](https://issues.redhat.com/browse/COST-4289)

## Description

If an OCP cluster does not have billing data for the currently processing Cloud bill date we should skip matching and summary tasks,

This change will add an additional check during OCP on Cloud resource/tag matching to only collect matches for clusters that match the report period currently being ingested.

This change also adds a similar check just before triggering OCP on Cloud summary tasks so we only run summary on the OCP clusters that actually have a matching report period.

## ADDITIONAL INFO

This likely does NOT directly solve the OOMing errors during matching.

## Testing

1. Checkout Branch
2. Restart Koku
3. Create multiple OCP providers + load data
4. Create Cloud provider (in my case I used GCP storage only, just so I could easily resend the same data. You can just recreate the GCP source between runs or generate fresh data)
5. Load GCP data that correlates with both clusters
6.  Everything should run normally and correlate, you should see something like this in the logs PER CLUSTER: ```{'message': 'updating ocp on cloud', 'tracing_id': '2023-10-01|b49ef598-369b-4c0e-a5ee-0825e2fc948e', 'schema': 'org1234567', 'ocp_provider_uuid': '40302d5b-d1bf-49da-8e63-bdc17fefafd7', 'provider_type': 'GCP', 'provider_uuid': '7aa5714b-1728-43b9-a966-361e7966d068', 'start_date': '2023-10-01', 'end_date': '2023-10-16', 'manifest_id': 21, 'queue_name': None}```
7. Connect to PG DB and update a cluster report period (Fake that its an old cluster no longer reporting) ```update org1234567.reporting_ocpusagereportperiod set report_period_start = '2023-08-01 00:00:00+00' where report_period_end = date('2023-11-01') and provider_id = '40302d5b-d1bf-49da-8e63-bdc17fefafd7';```
8. Recreate your GCP source or gen new data (Bare in mind if you recreate the source you ideally want to only generate data for the current month, since we only changed the report period for the current month on the cluster above. Or you could update both records)
9. This time in the logs you should on see matching + summary for the cluster that we did not change the report period on. (Since it should still match)
10. You should also see the following log messages ```no matching report periods available for cluster - skipping OCP on GCP parquet processing``` during OCP on GCP resource/tag matching AND ```no matching report periods available for cluster - removing from OCP on GCP summary list``` just before summary tasks are triggered.

## Notes
For our large customers they can have 50+ clusters (some of which are stale, not still sending reports) prior to this code we would collect and iterate over all of them regardless whether we actually have data for the currently processing report period for those clusters. Additionally we also trigger OCP on Cloud summary for ALL these OCP clusters too.
...


koku/masu/processor/ocp/ocp_cloud_parquet_summary_updater.py
```diff
@@ -119,7 +119,27 @@ def get_infra_map(self, start_date, end_date):
...
+        summary_infra_map = {}
+        with OCPReportDBAccessor(self._schema) as accessor:
+            # Only include OCP providers for OCP on Cloud summary if we have a matching report period
+            for ocp_provider_uuid in infra_map:
+                ctx = {
+                    "schema": self._schema,
+                    "ocp_provider_uuid": ocp_provider_uuid,
+                    "provider_uuid": infra_map[ocp_provider_uuid][0],
+                    "provider_type": infra_map[ocp_provider_uuid][1],
+                }
```
djnakabaale, line 132: nitpick suggestion:  could consider building this in the else where it's used

koku/masu/processor/ocp/ocp_cloud_parquet_summary_updater.py
```diff
@@ -119,7 +119,27 @@ def get_infra_map(self, start_date, end_date):
...
+                if accessor.report_periods_for_provider_uuid(ocp_provider_uuid, start_date):
```
djnakabaale, line 133: Question:
> our large customers they can have 50+ clusters

could this be making multiple calls to the db as we call this func inside the for loop?

djnakabaale (commented):
I gotta couple comments:

koku/masu/processor/ocp/ocp_cloud_parquet_summary_updater.py
lcouzens: Correct, its been called once for each distinct OCP source.

koku/masu/processor/ocp/ocp_cloud_parquet_summary_updater.py
lcouzens: true

PR #4732, opened 2023-10-14
add remote attach option for celery pods and auto-reload celery

maskarb:
## Description

This change will create a new django command `reload_celery` which taps into the Django auto-reload functionality to trigger a celery reload. This works by killing the old celery processes and starting a new one. This also adds a debug option that can be used to attach a remote debugger to the celery worker.

#### auto reload:
1. Checkout Branch
2. Restart Koku
3. watch worker logs: `docker compose logs -f koku-worker`
4. change a file, save it, and watch the worker restart (it might be slow)

#### remote attach debugging with vscode:
1. do this with a clean db
2. using this launch.json config:
```
        {
            "name": "Python: Celery Attach",
            "type": "python",
            "request": "attach",
            "port": 5678,
            "host": "localhost",
            "pathMappings": [
                {
                    "localRoot": "${workspaceFolder}/koku",
                    "remoteRoot": "."
                }
            ]
        },
```
3. set the env variable `DEBUG_ATTACH=True`
4. start koku-worker (be sure to reload env variables) and watch logs
5. observe the logs for the worker and see:
```
koku-koku-worker-1  | Waiting for debugger attach on port 5678
```
6. start the `Python: Celery Attach` debug session and watch the worker continue loading.
7. set a breakpoint [here](https://github.com/project-koku/koku/blob/main/koku/masu/processor/orchestrator.py#L62)
8. create providers and ingest data:
```
make create-test-customer; make load-test-customer-data test_source=aws
```
9. observe the debugger stop on the breakpoint.

PR #4761, opened 2023-11-02
Add a quick rancher doc about restarting celery worker.

myersCody:
## Jira Ticket

## Description

This change will ...

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Notes

...

### DIFF
--- a/docs/rancher.md
+++ b/docs/rancher.md
@@ -1,0 +1,37 @@
+# Koku Development Documentation
+
+## Debugging Celery Workers and Remote Attach with VSCode
+
+### Celery Worker Auto-Reload Issue
+
+One of the issues we've encountered in our development workflow is related to Celery workers not auto-reloading with Rancher. This can be inconvenient as it requires manual intervention to reload the worker when making changes.
+
+**Issue:** The Celery worker does not auto-reload with Rancher, leading to development challenges.
+
+**Solution:** We need to address this issue to streamline our development process.
+
+### Remote Attach Debugging with Visual Studio Code (VSCode)
+
+Debugging Celery workers in Koku can be made more efficient by setting up remote debugging with VSCode. Here are the steps to enable remote debugging:
+
+1. **Start with a clean database.** Ensure that you have a fresh database to work with.
+
+2. **VSCode Configuration:**
+   - Create a `launch.json` configuration in VSCode with the following content:
+
+   ```json
+   {
+       "name": "Python: Celery Attach",
+       "type": "python",
+       "request": "attach",
+       "port": 5678,
+       "host": "localhost",
+       "pathMappings": [
+           {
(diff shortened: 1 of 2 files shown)


docs/rancher.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Debugging Celery Workers and Remote Attach with VSCode
+
+### Celery Worker Auto-Reload Issue
+
+One of the issues we've encountered in our development workflow is related to Celery workers not auto-reloading with Rancher. This can be inconvenient as it requires manual intervention to reload the worker when making changes.
+
+**Issue:** The Celery worker does not auto-reload with Rancher, leading to development challenges.
+
+**Solution:** We need to address this issue to streamline our development process.
```
maskarb, line 11: I would remove this whole section because we've fixed the auto-reload by tapping into Django auto-reload instead of relying on watchdog.

docs/rancher.md
```diff
@@ -0,0 +1,37 @@
...
+
+### Remote Attach Debugging with Visual Studio Code (VSCode)
+
+Debugging Celery workers in Koku can be made more efficient by setting up remote debugging with VSCode. Here are the steps to enable remote debugging:
+
+1. **Start with a clean database.** Ensure that you have a fresh database to work with.
```
maskarb, line 17:
```suggestion
1. **Update environment variables.** Set `DEBUG_ATTACH=True` and reload the `.env` file.
```
A clean db is not a requirement to use the debugger, but setting this value is.

docs/rancher.md
```diff
@@ -0,0 +1,37 @@
...
+       "host": "localhost",
+       "pathMappings": [
+           {
+               "localRoot": "${workspaceFolder}/koku",
+               "remoteRoot": "."
+           }
+       ]
+   }
+
+More info can be found here: https://github.com/project-koku/koku/pull/4732
```
maskarb, line 37:
````suggestion
3. Start the `koku-worker` with the updated `.env` variables.
4. Observe the worker logs and wait for the message:
```
koku-koku-worker-1  | Waiting for debugger attach on port 5678
```
5. In VSCode, in the *Run and Debug* view, select the `Python: Celery Attach` configuration and start the debug session.
6. Once the debugger attaches to the container, the worker will resume starting like normal.
7. Set breakpoints and debug away!

**Note:** If code changes are made, the container will auto-reload and the debugger will detach! It will be necessary to re-attach the debugger to continue debugging.

More info can be found here: https://github.com/project-koku/koku/pull/4732

````

maskarb (commented):
It's definitely a good idea to add docs. However, we don't need anything specific for Rancher. Let's change the name of this file to something like `vscode-debugging` and then we can add more info about other configurations here when we get to it.

PR #4762, opened 2023-11-03
Applying str data type to the reading CSV to avoid leading zeros bug.

bacciotti:
## Jira Ticket

[COST-4363](https://issues.redhat.com/browse/COST-4363)

## Description

This change applies the string (str) data type to Azure CSV files, preventing the loss of leading zeros.

## Testing

1. Checkout Branch
2. Restart Koku
3. Run **test_create_daily_archives_check_leading_zeros** test

## Notes

I'm not sure if it would be beneficial but it could avoid future bugs.


maskarb (approved):

samdoran:
This is a good change. Even though I don't think it's fixing a known bug, it maintains integrity of the text when we're just splitting into smaller CSV files.

It looks like pandas keep UUIDs as strings. This is still a very good fix, though. Explicit is better than implicit.

<details>
  <summary>Pandas type inference</summary>
  
```
In [31]: f = io.StringIO("UUID,int,leading zero int\n0fc067a1-65d2-46da-b24b-7a9cbe2c69bd,134513413,000123453\n")

In [32]: df = pandas.read_csv(f)

In [33]: df
Out[33]:
                                   UUID        int  leading zero int
0  0fc067a1-65d2-46da-b24b-7a9cbe2c69bd  134513413            123453

In [34]: df.dtypes
Out[34]:
UUID                object
int                  int64
leading zero int     int64
dtype: object

In [35]: f.seek(0)
Out[35]: 0

In [36]: df = pandas.read_csv(f, dtype='str')

In [37]: df
Out[37]:
                                   UUID        int leading zero int
0  0fc067a1-65d2-46da-b24b-7a9cbe2c69bd  134513413        000123453

In [38]: df.dtypes
Out[38]:
UUID                object
int                 object
leading zero int    object
dtype: object
```
</details>

PR #4763, opened 2023-11-03
[COST-4405] support case insensitive queries for user-access endpoints

maskarb:
## Jira Ticket

[COST-4405](https://issues.redhat.com/browse/COST-4405)

## Description

This change will enable case-insensitive queries on the user-access endpoints.

## Testing

Both of these should work:

1. http://localhost:8000/api/cost-management/v1/user-access/?type=OCP
2. http://localhost:8000/api/cost-management/v1/user-access/?type=ocp
```
HTTP 200 OK
Allow: GET, HEAD, OPTIONS
Vary: X_RH_IDENTITY, Accept

{
    "data": true,
    "access": true,
    "write": true
}
```

## Notes

...


samdoran (approved):

samdoran:
It might be good to have tests cases for upper/lower/mixed case inputs. That's not a blocker, though.

dlabrecq (approved):

PR #4776, opened 2023-11-14
[COST-4430] Handle service account in API identity header

chambridge:
* Check for type in identity header
* Gather data from service account object

## Jira Ticket

[COST-4430](https://issues.redhat.com/browse/COST-4430)

## Description

This change will our ability to handle requests from service accoutn objects

## Testing

1. Checkout Branch
2. Restart Koku
3. Set `DEVELOPMENT_IDENTITY` object with a service account object
```
DEVELOPMENT_IDENTITY='{ "identity": { "account_number": "10001", "org_id": "1234567", "type": "ServiceAccount", "service_account": { "client_id": "93930000", "username": "service_account_user_dev", } }, "entitlements": { "cost_management": { "is_entitled": "True" } } }'
```
4. Hit endpoint with identity header that contains a service account object

### DIFF
--- a/koku/koku/middleware.py
+++ b/koku/koku/middleware.py
@@ -341,11 +341,20 @@ class IdentityHeaderMiddleware(MiddlewareMixin):
 
         account = json_rh_auth.get("identity", {}).get("account_number")
         org_id = json_rh_auth.get("identity", {}).get("org_id")
-        user = json_rh_auth.get("identity", {}).get("user", {})
-        username = user.get("username")
-        email = user.get("email")
-        is_admin = user.get("is_org_admin")
+        token_type = str(json_rh_auth.get("identity", {}).get("type", "user")).lower()
+        user = None
+        email = None
+        is_admin = False
         req_id = None
+        if token_type == "user":
+            user = json_rh_auth.get("identity", {}).get("user", {})
+            username = user.get("username")
+            email = user.get("email")
+            is_admin = user.get("is_org_admin")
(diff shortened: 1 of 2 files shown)


koku/koku/middleware.py
```diff
@@ -341,11 +341,20 @@ def process_request(self, request):  # noqa: C901
...
+        user = None
+        email = None
+        is_admin = False
         req_id = None
+        if token_type == "user":
+            user = json_rh_auth.get("identity", {}).get("user", {})
+            username = user.get("username")
+            email = user.get("email")
+            is_admin = user.get("is_org_admin")
+        else:
```
maskarb, line 354: Should we always assume that a token_type that is not `user` will always be `service-account`?

maskarb (commented):
There is probably something in here we need to update too:
https://github.com/project-koku/koku/blob/main/koku/api/provider/provider_builder.py#L82-L107

koku/koku/middleware.py
chambridge: If we do and it doesn't contain a service account object then we will end up with a 401, but if you would prefer an `elseif` I can do that.

maskarb: Eh, I'm fine with a 401. I just wanted to verify that is handled correctly when we receive neither `user` nor `service-account`.

lcouzens (approved):

maskarb (approved):

PR #4780, opened 2023-11-15
[COST-4210] more ProviderDBAccessor removal

maskarb:
## Jira Ticket

[COST-4210](https://issues.redhat.com/browse/COST-4210)

## Description

This change will remove `ProviderDBAccessor` from everywhere except the `OCPCloudUpdaterBase` which are more involved changes. That will be updated in a follow-up PR.


koku/masu/external/ros_report_shipper.py
```diff
@@ -142,8 +142,7 @@ def send_kafka_message(self, msg):
 
     def build_ros_msg(self, presigned_urls, upload_keys):
         """Gathers the relevant information for the kafka message and returns the message to be delivered."""
-        with ProviderDBAccessor(self.provider_uuid) as provider_accessor:
-            cluster_alias = provider_accessor.get_provider_name()
+        cluster_alias = get_cluster_id_from_provider(self.provider_uuid)
```
cgoodfred, line 145: Is this still getting the cluster alias here? Doesn't this return the cluster uuid instead?

lcouzens (dismissed):

koku/masu/external/ros_report_shipper.py
lcouzens: good point this might need to be 
```suggestion
        cluster_id = get_cluster_id_from_provider(openshift_provider_uuid)
        cluster_alias = get_cluster_alias_from_cluster_id(cluster_id)
```

samdoran (commented):
It may be helpful to reformat the log messages to use `log_json` and put the parameters in context instead of string in the message. Not a blocker, just thinking it would be helpful to filter on some of those values by having them in distinct fields.

koku/masu/external/accounts/hierarchy/aws/aws_org_unit_crawler.py
```diff
@@ -106,30 +97,26 @@ def _crawl_org_for_accounts(self, ou, prefix, level):
...
+                f"{self.provider.uuid}, and account_id: {self.account_id}"
             )
 
     def _check_if_crawlable(self):
         """Checks to see if the account is crawlable."""
         context_key = "crawl_hierarchy"
-        with ProviderDBAccessor(self.account.get("provider_uuid")) as provider_accessor:
-            context = provider_accessor.get_additional_context()
-            self.crawlable = context.get(context_key, True)
+        context = self.provider.additional_context or {}
```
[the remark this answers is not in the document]
maskarb: You're missing that this is a nullable field:
```
additional_context = JSONField(null=True, default=dict)
```

We always add fields as nullable.

cgoodfred (approved):

lcouzens (approved):

PR #4789, opened 2023-11-17
One approach to solve null values for subscription_name

bacciotti:
## Jira Ticket

[COST-4313](https://issues.redhat.com/browse/COST-4313)

## Description

This change will replace the null values with `subscription_guid` value.

## Testing

1. Checkout Branch
2. Restart Koku
3. Hit endpoint 
`http://localhost:8000/api/cost-management/v1/reports/azure/costs/?group_by[subscription_guid]=*&filter[limit]=100`
4. Check if there are null values on some `subscription_name` keys


koku/api/report/queries.py
```diff
@@ -1202,6 +1202,10 @@ def _ranked_list(self, data_list, ranks, rank_fields=None):  # noqa C901
             account_aliases = account_aliases.drop_duplicates(subset="account")
         data_frame.drop(columns=drop_columns, inplace=True, errors="ignore")
         data_frame = data_frame.merge(ranks_by_day, how="right", on=merge_on)
+        data_frame["subscription_name"] = data_frame.apply(
+            lambda row: row["subscription_guid"] if pd.isna(row["subscription_name"]) else row["subscription_name"],
+            axis=1,
+        )
```
maskarb, line 1208: This is `subscription_name/subscription_guid` Azure specific, but we are in a base class here. So I think we need to have a `if self.is_azure`, similar to the way we have `self.is_aws` just a couple lines above here. If we don't have a `is_azure` property, then we will need to add one.

lcouzens (approved):

lcouzens (approved):

PR #4800, opened 2023-11-22
[COST-4443] Update data model for cost groups.

myersCody:
## Jira Ticket

[COST-4443](https://issues.redhat.com/browse/COST-4443)

## Description

This change will pull out the namespace field from the `reporting_ocp_cost_category` model into its own table. The purpose of this change is so that we can enforce a one to many relationship between the namespace and categories. With the namespace array the responsibility of maintaining that relationship is left up to the developers. 

This change will only add the new model and add the correct defaults to the table. 

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Notes

...


koku/reporting/provider/ocp/models.py
```diff
@@ -299,6 +299,19 @@ class Meta:
...
 
+class OpenshiftCostCategoryNamespace(models.Model):
+    """Namespaces to bucket to category."""
+
+    class Meta:
+        """Meta for cost category namespaces."""
+
+        db_table = "reporting_ocp_cost_category_namespace"
+
+    namespace = models.TextField(unique=True)
```
maskarb, line 310: Just thinking about this a little bit, this field doesn't _really_ represent "namespace". Like, `openshift-%` is not a namespace. I wonder if we should name this something like "namespace_pattern"?

maskarb (approved):
LGTM! Just one comment:

PR #4813, opened 2023-12-04
[COST-4389] - Add the ability to reprocess csv reports

lcouzens:
## Jira Ticket

[COST-4389](https://issues.redhat.com/browse/COST-4389)

## Description

This change will add a new masu endpoint + a new customer downloader flow in order to reprocess CSV files for a provider. Essentially this means we can grab CSV files stored in S3 and reprocess them into relevant parquet files.

This was put together to assist with the trino dtype changes but can be used outside of those changes.

## Testing

1. Checkout Branch
2. Restart Koku
3. Make create-test-customer
4. make load-test-customer-data
6. Grab a/all provider uuids.
7. look at the parquet files in minio for the provider your going to run
8. hit the following endpoint as follows to reprocess CSV files 
`http://localhost:5042/api/cost-management/v1/reprocess_csv_reports/?provider_uuid=a3cf0f5a-51e5-4624-a15d-822160c690a7&start_date=2023-11-01`

The above will grab all csv reports for the given provider and month.
9. Look again at the parquet files and see the names/creation times have changed
10. Validate the cost remain the same in the API

## Important Note
One thing we need to check is how this works in the real world for OCP. All of our providers have distinct CSV files E.I. shouldnt have duplication for days/times. I'm not 100% sure if thats true for OCP, when we changed to daily files from the operator do we purge the CSV files from S3 for the same days? Or is it possible we have multiple files for the same date/times? If we do then we will likely cause duplication here as this endpoint will process ALL CSV files to parquet regardless of content. Maybe for OCP we need to some how build a list of the correct daily files that need reprocessing.

UPDATE - We likely do have partial day CSV files for OCP so before this could be used for OCP providers we need a method to handle which CSV's should get preprocessed.

...


myersCody:
In additional to some general clean up for this PR, there as a few additional things we probably want to take into consideration with this work. 

1. We should definitely build some safety measures around this API to make sure it can't be kicked off multiple times within a time period. Especially if we are going to allow it to accept provider_type.
2. May be nice to see if we can do this in a way where we don't have to pass `reprocess_csv_reports=reprocess_csv_reports` down the chain of calls.

PR #4820, opened 2023-12-04
[COST-4230] - log warning when operator payload is set to basic auth

maskarb:
## Jira Ticket

[COST-4230](https://issues.redhat.com/browse/COST-4230)

## Description

This change will log a warning when a customer has their operator set to basic auth.


djnakabaale (approved):
lgtm 👍🏾

PR #4822, opened 2023-12-05
[COST-4391] Test to check provider map units.

bacciotti:
## Jira Ticket

[COST-4391](https://issues.redhat.com/browse/COST-4391)

## Description

This change will test if the units used in the provider_map for AWS file match the units in the downloaded CSV file.

## Testing

1. Checkout Branch
2. Restart Koku
3. Run test `tox -e py39 -- api.report.test.aws.test_views.AWSReportViewTest.test_units_consistency`

## Notes

This is merely an accessory test. The main part of the analysis is explained in the JIRA task linked above.


lcouzens (approved):

koku/api/report/test/aws/test_views.py
```diff
@@ -600,3 +601,27 @@ def test_invalid_aws_category_key(self):
         url = reverse("reports-aws-costs") + "?group_by[aws_categroy:invalid]=value"
         response = self.client.get(url, **self.headers)
         self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+
+    def test_units_consistency(self):
+        """Test if provider map used units match the report."""
+        file = "November-2023-None"
+        file_name = f"{file}.csv"
```
myersCody, line 608: Not a blocker, but a separate var here here seems a bit redundant when you could just add the file extension to the file_path:

```
file_path = f"./koku/masu/test/data/aws/{file}.csv"
```

koku/api/report/test/aws/test_views.py
```diff
@@ -4,6 +4,7 @@
 #
 """Test the AWS Report views."""
 import copy
+import csv as csv
```
myersCody, line 7: Also, a nitpick and not a blocker at all but this alias seems unnecessary.

koku/api/report/test/aws/test_views.py
```diff
@@ -600,3 +601,27 @@ def test_invalid_aws_category_key(self):
...
         self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+
+    def test_units_consistency(self):
+        """Test if provider map used units match the report."""
+        file = "November-2023-None"
+        file_name = f"{file}.csv"
+        csv_units = set()
+        file_path = f"./koku/masu/test/data/aws/{file_name}"  # noqa: E501
+
+        provider_map_units = [
```
myersCody, line 612: So, I think it may be better to attempt to get these units directly from the provider map itself. Otherwise it just kind of feels like this test is just checking that the csv you created has these units in it?

maskarb:
I think we are better off with the information that is provided in the JIRA as opposed to adding a unit test. I don't think this unit test will be beneficial, especially if units change in actual AWS reports.

I would prefer that we close this issue and rely on the JIRA to track this work.

PR #4823, opened 2023-12-05
[COST-4489] add uniqueness constraint on the ProviderInfraMap table

maskarb:
## Jira Ticket

[COST-4489](https://issues.redhat.com/browse/COST-4489)

## Description

We use `ProviderInfrastructureMap.get_or_create` to create new entries in the `ProviderInfrastructureMap` but we don't have any uniqueness constraints on the table which is resulting in a 
```
get() returned more than one ProviderInfrastructureMap – it returned 2
```
error. The `get_or_create` is used in a try/except, but without the uniqueness constraint, a race condition will always create two db entries. Adding the uniqueness constraint will force us to go thru the try/except.


cgoodfred (approved):

koku/api/migrations/0061_alter_providerinfrastructuremap_unique_together.py
```diff
@@ -0,0 +1,16 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("api", "0060_provider_polling_timestamp"),
+    ]
+
+    operations = [
+        migrations.AlterUniqueTogether(
+            name="providerinfrastructuremap",
+            unique_together={("infrastructure_type", "infrastructure_provider")},
```
myersCody, line 14: Do we need to worry about duplicates that might already be in the table?

```
postgres=# INSERT into api_providerinfrastructuremap (infrastructure_type, infrastructure_provider_id) VALUES ('AWS-local', '9672f1ec-8872-40bb-8029-dfb360a39127');
INSERT 0 1
postgres=# select * from api_providerinfrastructuremap;
 id | infrastructure_type |      infrastructure_provider_id
----+---------------------+--------------------------------------
  1 | AWS-local           | 9672f1ec-8872-40bb-8029-dfb360a39127
  2 | AWS-local           | 9672f1ec-8872-40bb-8029-dfb360a39127
(2 rows)
```

```
koku_server         |   File "/opt/koku/.venv/lib/python3.9/site-packages/django/db/backends/utils.py", line 84, in _execute
koku_server         |     return self.cursor.execute(sql, params)
koku_server         | django.db.utils.IntegrityError: could not create unique index "api_providerinfrastructu_infrastructure_type_infr_9fe23dba_uniq"
koku_server         | DETAIL:  Key (infrastructure_type, infrastructure_provider_id)=(AWS-local, 9672f1ec-8872-40bb-8029-dfb360a39127) is duplicated.
```

maskarb: There are a few duplicates in production that we'll need to take care of first.

PR #4817, opened 2023-12-04
[COST-4333] - Adding dtypes for Azure

lcouzens:
## Jira Ticket

[COST-4333](https://issues.redhat.com/browse/COST-4333)

## Description

This change will commonize dtype handling and correctly set the csv converters for Azure

## Testing

1. Checkout Branch
2. Run koku with updated trino/hive versions
3.  create/load data for Azure provider
4. see data is processed correctly

## Notes
trino diff
deploy/trino/etc/catalog/hive.properties
 
-parquet.optimized-reader.enabled=false
+# parquet.optimized-reader.enabled=false

diff --git a/deploy/trino/etc/jvm.config b/deploy/trino/etc/jvm.config

+-Dfile.encoding=UTF-8

diff --git a/docker-compose.arm.yml b/docker-compose.arm.yml

   hive-metastore:
-    image: quay.io/mskarbek/ubi-hive:3.1.2-metastore-008
+    image: quay.io/mskarbek/ubi-hive:3.1.3-metastore-024
 
   trino:
-    image: quay.io/mskarbek/ubi-trino:411-001
+    image: quay.io/mskarbek/ubi-trino:427-001

diff --git a/docker-compose.yml b/docker-compose.yml
   hive-metastore:
     container_name: hive-metastore
-    image: quay.io/cloudservices/ubi-hive:3.1.2-metastore-008
+    image: quay.io/cloudservices/ubi-hive:latest

   trino:
     container_name: trino
-    image: quay.io/cloudservices/ubi-trino:411-002
+    image: quay.io/cloudservices/ubi-trino:latest
...

PR #4824, opened 2023-12-06
[COST-4333] Fix OCP on Azure matching by resource-ids

esebesto:
## Jira Ticket

[COST-4333](https://issues.redhat.com/browse/COST-4333)

## Description

This change should fix the regression caused by #4817 
Azure report v1 uses "InstanceId" instead of "ResourceId" column name. The resourceid column is added to TRINO_REQUIRED_COLUMNS. After the change introduced in the PR #4817, the resourceid column for Azure v1 reports contains empty strings instead of nulls. 
The code for matching of Azure with OCP first check if resourceid column has some data - if it doesn't has data, it checks instanceid column. Since the code expected NANs as an indicator of an empty column, it incorrectly evaluated empty resourceid column (with empty strings) as a column containing data and used this column for OCP on Azure matching by resource-id

## Testing

Run azure-smoke-tests
Since there is a small change in AWS part too, potentially run also aws-smoke-tests

## Notes

...

### DIFF
--- a/koku/masu/test/util/aws/test_common.py
+++ b/koku/masu/test/util/aws/test_common.py
@@ -13,7 +13,6 @@ from unittest.mock import PropertyMock
 from uuid import uuid4
 
 import boto3
-import numpy as np
 import pandas as pd
 from botocore.exceptions import ClientError
 from dateutil.relativedelta import relativedelta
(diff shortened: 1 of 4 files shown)


esebesto:
Clean pass with azure-smoke-tests label

maskarb (approved):

PR #4837, opened 2023-12-14
Cost 4389 fix parquet masu

myersCody:
## Jira Ticket

[COST-4389](https://issues.redhat.com/browse/COST-4389)

## Description

This change will add an internal endpoint that downloads a parquet file from s3,  reads the parquet schema from the local file, checks the data types, cast or transforms the data to the correct type, uploads the parquet file back up s3. 

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Notes

...


koku/masu/api/upgrade_trino/util/verify_parquet_files.py
```diff
@@ -0,0 +1,302 @@
...
+            "simulate": self.simulate,
+            "bill_date": self.bill_date,
+        }
+
+    def _set_pyarrow_types(self, cleaned_column_mapping):
+        mapping = {}
+        for key, default_val in cleaned_column_mapping.items():
+            if pd.isnull(default_val):
+                # TODO: Azure saves datetime as pa.timestamp("ms")
+                # TODO: AWS saves datetime as timestamp[ms, tz=UTC]
```
lcouzens, line 53: Also I believe OCI is doing `ms, tz=UTC` too, so depending the route we might need to change that too.

koku/masu/api/upgrade_trino/util/verify_parquet_files.py
```diff
@@ -0,0 +1,302 @@
...
+    def _set_report_types(self):
+        if self.provider_type == Provider.PROVIDER_OCI:
+            return ["cost", "usage"]
+        if self.provider_type == Provider.PROVIDER_OCP:
+            return ["namespace_labels", "node_labels", "pod_usage", "storage_usage"]
+        return [None]
+
+    def _get_bill_dates(self):
+        # However far back we want to fix.
+        return [ciso8601.parse_datetime(self.bill_date)]
```
lcouzens, line 71: So to be clear this will return a single date based on what's passed in to the endpoint. If we want to make use of multiple months this probably wants to be something like this: 
```suggestion
        return dh.list_months(ciso8601.parse_datetime(self.bill_date)), dh.today.replace(tzinfo=None))
```
This will return something like the following based on the input:
```
[datetime.datetime(2023, 11, 1, 0, 0),
 datetime.datetime(2023, 12, 1, 0, 0),
 datetime.datetime(2024, 1, 1, 0, 0)]

File: koku/masu/api/upgrade_trino/util/verify_parquet_files.py [hunk 3]
```diff
@@ -0,0 +1,302 @@
...
+        local_path = Path(Config.TMP_DIR, self.schema_name, str(self.provider_uuid))
+        local_path.mkdir(parents=True, exist_ok=True)
+        return local_path
+
+    def retrieve_verify_reload_s3_parquet(self):
+        """Retrieves the s3 files from s3"""
+        s3_resource = get_s3_resource(settings.S3_ACCESS_KEY, settings.S3_SECRET, settings.S3_REGION)
+        s3_bucket = s3_resource.Bucket(settings.S3_BUCKET_NAME)
+        bill_dates = self._get_bill_dates()
+        for bill_date in bill_dates:
```
lcouzens, line 142: Right now we are only going to iterate over a single month based on the passed in `self.bill_date` see my comment above if we do want to run multiple months with a single trigger.

koku/masu/api/upgrade_trino/view.py
```diff
@@ -0,0 +1,37 @@
+#
+# Copyright 2023 Red Hat Inc.
```
lcouzens, line 2:
```suggestion
# Copyright 2024 Red Hat Inc.
```

koku/masu/api/upgrade_trino/util/verify_parquet_files.py
myersCody: We can add a check to switch this for azure then. 👍

koku/masu/api/upgrade_trino/util/verify_parquet_files.py
myersCody: Yeah, I am fine with this change too. I didn't know how much we wanted to kick off at once. So, I left it can of open for discussion.

koku/masu/api/upgrade_trino/util/state_tracker.py
```diff
@@ -0,0 +1,104 @@
...
+            manager.model.set_additional_context(context)
+            LOG.info(self.provider_uuid, log_json(msg="setting dtype states", context=context))
+        except ProviderManagerError:
+            pass
+
+    def finalize_and_clean_up(self):
+        self._check_if_complete()
+        self._clean_local_files()
+        # We can decide if we want to record
+        # failed parquet conversion
```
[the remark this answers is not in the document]
myersCody: I added some logic to the additional info page, but I can add some to the metadata.

koku/masu/api/upgrade_trino/util/__init__.py
```diff
@@ -0,0 +1,6 @@
+CONTEXT_KEY_MAPPING = {
+    "metadata": "conversion_metadata",
+    "version": "version",
+    "successful": "successful",
+    "failed_files": "dtype_failed_files",
+}
```
samdoran, line 6: This seems like a good candidate for [`StrEnum`](https://docs.python.org/3/howto/enum.html) since it is storing unique constant values..

```suggestion
from enum import auto
from enum import Enum
class Context(StrEnum):
    metadata = "conversion_metadata"
    version = auto()
    successful = auto(),
    failed_files = "dtype_failed_files"
```

In use, it would look like `bill_metadata.get(Context.version)`. 

`StrEnum` was added in Python 3.11, but it's a [fairly simple class](https://github.com/python/cpython/blob/c31be58da8577ef140e83d4e46502c7bb1eb9abf/Lib/enum.py#L1321-L1345) we could vendor.

myersCody: Great suggestion @samdoran. I wasn't able to get the `auto()` to work because it was generating nums and StrEnum expected stings. However, I did go a head and switch my file states over to StrEnums too.

samdoran: I was looking through the CPython source and there were a number of changes to the metaclass to make `auto()` work. Explicitly setting the value works just fine.

koku/masu/api/upgrade_trino/util/constants.py
```diff
@@ -0,0 +1,57 @@
...
+
+class ReprEnum(Enum):
+    """
+    Only changes the repr(), leaving str() and format() to the mixed-in type.
+    """
+
+
+# StrEnum is available in python 3.11, vendored over from
+# https://github.com/python/cpython/blob/c31be58da8577ef140e83d4e46502c7bb1eb9abf/Lib/enum.py#L1321-L1345
+class StrEnum(str, ReprEnum):
```
samdoran, line 16: We can ignore coverage on this since it's vendored code.
```suggestion
class StrEnum(str, ReprEnum):  # pragma: no cover
```

koku/masu/api/upgrade_trino/util/verify_parquet_files.py
```diff
@@ -0,0 +1,313 @@
...
+        LOG.info(
+            log_json(
+                self.provider_uuid,
+                msg="Transforming fields from double to timestamp.",
+                context=self.logging_context,
+                local_file_path=parquet_file_path,
+                updated_columns=field_names,
+            )
+        )
+        table = pq.read_table(parquet_file_path)
```
samdoran, line 210: Will any of these files be larger than available memory? Just wondering if we should be using `ParquetFile.iter_batch()` instead of reading the whole file at once.

koku/masu/api/upgrade_trino/util/task_handler.py
```diff
@@ -0,0 +1,134 @@
...
+            if simulate.lower() == "true":
+                reprocess_kwargs.simulate = True
+
+        if not reprocess_kwargs.provider_type and not reprocess_kwargs.provider_uuid:
+            raise RequiredParametersError("provider_uuid or provider_type must be supplied")
+        if not reprocess_kwargs.start_date:
+            raise RequiredParametersError("start_date must be supplied as a parameter.")
+
+        reprocess_kwargs.cleaned_column_mapping = reprocess_kwargs.clean_column_names(reprocess_kwargs.provider_type)
+        return reprocess_kwargs
```
samdoran, line 76: This is a very good use of the factory pattern. I would suggest changing the strategy used from creating a class instance and modifying it to building the parameters and passing those into the constructor. That way you can make the dataclass frozen, making the instance more durable.

Here is a rough example that probably has some room for further improvement.

```python
@dataclass(frozen=True)
class FixParquetTaskHandler:
    start_date: Optional[str] = field(default=None)
    provider_uuid: Optional[str] = field(default=None)
    provider_type: Optional[str] = field(default=None)
    simulate: Optional[bool] = field(default=False)
    cleaned_column_mapping: Optional[dict] = field(default=None)

    # Node role is the only column we add manually for OCP
    # Therefore, it is the only column that can be incorrect
    REQUIRED_COLUMNS_MAPPING = {
        Provider.PROVIDER_OCI: OCI_TRINO_REQUIRED_COLUMNS,
        Provider.PROVIDER_OCP: {"node_role": ""},
        Provider.PROVIDER_AWS: AWS_TRINO_REQUIRED_COLUMNS,
        Provider.PROVIDER_AZURE: AZURE_TRINO_REQUIRED_COLUMNS,
    }

    @classmethod
    def from_query_params(cls, query_params: QueryDict) -> "FixParquetTaskHandler":
        """Create an instance from query parameters."""
        kwargs = {}
        if start_date := query_params.get("start_date"):
            if start_date:
                kwargs["start_date"] = parser.parse(start_date).replace(day=1)

        if provider_uuid := query_params.get("provider_uuid"):
            provider = Provider.objects.filter(uuid=provider_uuid).first()
            if not provider:
                raise RequiredParametersError(f"The provider_uuid {provider_uuid} does not exist.")
            kwargs["provider_uuid"] = provider_uuid
            kwargs["provider_type"] = provider.type

        if provider_type := query_params.get("provider_type"):
            kwargs["provider_type"] = provider_type

        if simulate := query_params.get("simulate"):
            if simulate.lower() == "true":
                kwargs["simulate"] = True

        if not kwargs["provider_type"] and not kwargs["provider_uuid"]:
            raise RequiredParametersError("provider_uuid or provider_type must be supplied")
        if not kwargs["start_date"]:
            raise RequiredParametersError("start_date must be supplied as a parameter.")

        kwargs["cleaned_column_mapping"] = cls.clean_column_names(kwargs["provider_type"])
        
        return cls(**kwargs)
```

koku/masu/api/upgrade_trino/util/verify_parquet_files.py
```diff
@@ -0,0 +1,313 @@
...
+                )
+            )
+            table = table.cast(new_schema)
+            # Write the table back to the Parquet file
+            pa.parquet.write_table(table, parquet_file_path)
+            self._perform_transformation_double_to_timestamp(parquet_file_path, double_to_timestamp_fields)
+            # Signal that we need to send this update to S3.
+            return cstates.coerce_required
+
+        except Exception as e:
```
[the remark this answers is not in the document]
myersCody: I left this intentionally broad to be a catch all for any unknown errors we might run into while attempting to convert the csv files to the correct type. I am hoping we never have to get to this section of the code though 😂

koku/masu/api/upgrade_trino/util/task_handler.py
myersCody: Great suggestion @samdoran! Much cleaner approach here.

koku/masu/api/upgrade_trino/util/verify_parquet_files.py
myersCody: Hmmm. Good comment, so these parquet files had to go through our batching logic to be created, so I feel like we shouldn't run into any massive parquet files during this process.

samdoran (approved):

PR #4842, opened 2023-12-19
Improve `TrinoStatementExecError`

samdoran:
## Description

Improve the `TrinoStatementExecError` class to behave more like the exceptions from `trino.exceptions`.
Use the `__str__` method for message formatting.
Make attributes of the underlying `TrinoQueryError` available as properties of `TrinoStatementExecError`. This allows for more precise error handling based on the specific type of Trino error rather than having to match on strings in the error message.
Rename some one character variables.

## Testing

`tox -e py39 -- koku.test_trino_db_utils`

### DIFF
--- a/koku/koku/test_trino_db_utils.py
+++ b/koku/koku/test_trino_db_utils.py
@@ -1,3 +1,4 @@
+import textwrap
 import uuid
 
 from jinjasql import JinjaSql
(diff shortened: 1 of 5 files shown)


cgoodfred (dismissed):

koku/koku/test_trino_db_utils.py
```diff
@@ -24,70 +30,95 @@ def test_executescript(self):
...
+            insert into hive.{{schema | sqlsafe}}.__test_{{uuid | sqlsafe}} (id, i_data, t_data)
+            values (cast(uuid() as varchar), {{int_data}}, {{txt_data}});
 
-select t_data from hive.{{schema | sqlsafe}}.__test_{{uuid | sqlsafe}} where i_data = {{int_data}};
+            select t_data from hive.{{schema | sqlsafe}}.__test_{{uuid | sqlsafe}} where i_data = {{int_data}};
 
-drop table if exists hive.{{schema | sqlsafe}}.__test_{{uuid | sqlsafe}};
-"""
+            drop table if exists hive.{{schema | sqlsafe}}.__test_{{uuid | sqlsafe}};
+        """
```
maskarb, line 50: Please don't change this. It is formatted this way because it is copy/pasted sql. It's ugly, sure, but it is easier to read/modify the sql when it's like this.

samdoran: It's the exact same SQL just indented. It can still be copy/pasted just as easily and makes the application code much more readable.

[Hiding whitespace](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-comparing-branches-in-pull-requests#diff-view-options) makes changes like this easier to review (apologies if you already knew about that feature).

koku/koku/test_trino_db_utils.py
```diff
@@ -105,6 +134,44 @@ def cursor(self):
 select x from y;
 select a from b;
 """
-        with self.assertRaises(trino_db.TrinoStatementExecError):
+        with self.assertRaises(ValueError):
```
djnakabaale, line 137: nitpick question:   Curious why this is now a `ValueError` from `TrinoStatementExecError`?

djnakabaale (approved):
lgtm 👍🏾

koku/koku/test_trino_db_utils.py
samdoran: I was hoping someone would ask about that. I changed the behavior in `executescript()` to only raise `TrinoStatementExecError` if a` TrinoQueryError` was raised since that is now a required parameter of `TrinoStatementExecError. Otherwise, just raise the original exception.

djnakabaale: Thanks for the explanation 👍🏾 , and I see the `ValueError` raised by `FakerFakeTrinoConn`

PR #4846, opened 2023-12-21
[COST-4534] Support filtering with multiple values

samdoran:
## Jira Ticket

[COST-4534](https://issues.redhat.com/browse/COST-4534)

## Description

Support multiple filter values for the `project` and `group` fields on the cost groups API

## Testing

```
curl "http://localhost:8000/api/cost-management/v1/settings/cost-groups/?limit=100&filter%5Bdefault%5D=false&filter%5Bproject%5D=test&filter%5Bproject%5D=cost"

```


samdoran:
Change detection isn't finding any changed files, so tests are not running and no coverage report is being generated. 😕

myersCody (dismissed):

myersCody (approved):

PR #4850, opened 2024-01-02
use DETACH PARTITION CONCURRENTLY

maskarb:
## Jira Ticket

## Description

This change will ...

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Notes

...

### DIFF
--- a/db_functions/partitioned_manager_trigger_function.sql
+++ b/db_functions/partitioned_manager_trigger_function.sql
@@ -45,7 +45,7 @@ select c.oid
                 action_stmts = array_append(
                     action_stmts,
                     format(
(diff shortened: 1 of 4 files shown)


lcouzens (approved):

cgoodfred (approved):

maskarb:
Because our tables have default partitions, we cannot use `CONCURRENTLY`:
https://www.postgresql.org/docs/14/sql-altertable.html#SQL-ALTERTABLE-DETACH-PARTITION
```
CONCURRENTLY cannot be run in a transaction block and is not allowed if the partitioned table contains a default partition.
```

We would need more structural changes to remove this constraint.

PR #4853, opened 2024-01-04
[COST-4515] Upgrade to Django 4.1

djnakabaale:
## Jira Ticket

[COST-4515](https://issues.redhat.com/browse/COST-4515)

## Description

This change will ...

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Notes

...


koku/api/iam/models.py
```diff
@@ -102,25 +103,36 @@ class Tenant(TenantMixin):
...
+        if (hash := get_function_hash(conn, self._CLONE_SCHEMA_FUNC_NAME)) != self._CLONE_SCHEMA_FUNC_SHA256:
+            LOG.info(f'Updating function "{self._CLONE_SCHEMA_FUNC_NAME}"')
+            apply_sql_file(conn.schema_editor(), self._CLONE_SCHEMA_FUNC_FILENAME, literal_placeholder=True)
+            LOG.info(
+                f'Updated function "{self._CLONE_SCHEMA_FUNC_NAME}": '
+                f"{get_function_hash(conn, self._CLONE_SCHEMA_FUNC_NAME)}"
             )
         else:
-            LOG.info("Clone function exists")
+            LOG.info(f'Clone function "{self._CLONE_SCHEMA_FUNC_NAME}" is already up to date: {hash}')
```
maskarb, line 129: We should not implement this. If this file changes, we should apply it via a migration. More than likely, there is some other db migration that is required before we can re-apply this clone-schema function, so this shouldn't be handled dynamically.

samdoran: Ok, I'll remove it.

samdoran (approved):

PR #4867, opened 2024-01-15
[COST-4575] Define a __reduce__ method on exceptions

samdoran:
## Jira issue

[COST-4575](https://issues.redhat.com/browse/COST-4575)

## Description

This is an attempt to prevent unpickling errors of our custom exception classes by having the `__reduce__` method return a consistent value. That value is used by `pickle` to determine if the object being unpickled is the same as what was pickled.

## Notes

This is coming from `billiard`, not `pickle` in CPython.

https://github.com/celery/billiard/blob/13922230092a4191583e592e9c6d49f912a3fd03/billiard/pool.py#L366-L373


samdoran:
I intentionally bypassed the coverage requirement for this change.

PR #4873, opened 2024-01-17
Docs: Update filtering examples to demonstrate the correct syntax

samdoran:
## Description

Update the examples in our API documentation to show the correct syntax for filtering.

### DIFF
--- a/docs/specs/openapi.json
+++ b/docs/specs/openapi.json
@@ -1112,7 +1112,8 @@
                 ],
                 "summary": "Query to obtain cost reports",
                 "operationId": "getOpenShiftCostReports",
-                "parameters": [{
+                "parameters": [
+                    {
                         "$ref": "#/components/parameters/QueryDelta"
                     },
                     {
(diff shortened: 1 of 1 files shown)


docs/specs/openapi.json
```diff
@@ -5938,22 +5943,28 @@
                 "name": "group_by",
                 "required": false,
                 "in": "query",
-                "description": "The grouping to apply to the report as a URL encoded dictionary.",
+                "description": "The grouping to apply to the report as a URL encoded dictionary. The syntax is group_by[key]=value or group_by[tag:key]=value.",
```
myersCody, line 5946: So, I tested this with:
```
http://localhost:8000/api/cost-management/v1/reports/openshift/costs/?end_date=2023-12-31&start_date=2023-12-01&group_by[environment]=*
```
And got:
```
{
    "group_by": {
        "environment": [
        ]
    }
}
```
I think you have to have the tag prefix for the group by to work.

samdoran: What about other contexts where `group_by` is used? This same language and example is used in all places that support `group_by`. For example:
```
http://localhost:8000/api/cost-management/v1/reports/aws/costs/?group_by%5Bregion%5D=*
```

So it's valid in certain contexts. Maybe we should give separate examples rather than using the same component everywhere.

samdoran: That sounds good.

myersCody (approved):

PR #4876, opened 2024-01-23
[COST-4598] API Spec: Remove incorrect cluster parameter

samdoran:
## Jira Ticket

[COST-4598](https://issues.redhat.com/browse/COST-4598)

## Description

Remove `cluster` as a URL parameter from the `/tags/` API spec. It is only valid in `filter[cluster]=` and `exclude[cluster]=` URL parameters.


djnakabaale (dismissed):
👍🏾 lgtm!

PR #4877, opened 2024-01-24
hccdoc-2065 update description and auth description

neal-timpe:
## Jira Ticket

[HCCDOC-2065](https://issues.redhat.com/browse/HCCDOC-2065)

## Description

This change will update the main description and the description of the Auth section, although I think the auth section needs to be updated from basic to service token.

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Notes

...

### DIFF
--- a/docs/specs/openapi.json
+++ b/docs/specs/openapi.json
@@ -2,7 +2,7 @@
     "openapi": "3.0.0",
     "info": {
         "description": "Access your cloud and OpenShift data you integrated with cost management.",
-        "version": "1",
+        "version": "v1",
         "title": "Cost Management",
         "license": {
             "name": "AGPL-3.0",
--- a/docs/specs/openapi.json
+++ b/docs/specs/openapi.json
@@ -2,7 +2,7 @@
     "openapi": "3.0.0",
     "info": {
         "description": "Access your cloud and OpenShift data you integrated with cost management.",
-        "version": "1.0.2",
+        "version": "1",
         "title": "Cost Management",
         "license": {
             "name": "AGPL-3.0",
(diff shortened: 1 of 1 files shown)


docs/specs/openapi.json
```diff
@@ -1,7 +1,7 @@
 {
     "openapi": "3.0.0",
     "info": {
-        "description": "The API for Cost Management",
+        "description": "Access your cloud and OpenShift data you integrated with cost management.",
         "version": "1.0.2",
```
maskarb, line 5: Could you also set this version to `v1`? I have no idea where 1.0.2 came from, and given that our api path is api/cost-management/v1, it feels like this version should also be v1.

neal-timpe: I changed to v 1 with no dots. Does it need to be `1.0.0`?

I'm editing the right openapi file right? The only other ones are in /koku/sources and /koku/masu and look like they do different stuff.

maskarb: I think it should be `v1`. And yes, this is the correct doc to edit.

maskarb: There seems to be zero consistency, but looking at this one:
https://console.redhat.com/docs/api/compliance
it uses `v1`.

neal-timpe: Ok, updated to "v1"

docs/specs/openapi.json
```diff
@@ -6103,7 +6103,7 @@
         "securitySchemes": {
```
neal-timpe, line 6103: @maskarb Will this securitySchemes be updated as the basic auth moves to service account auth?

maskarb: I had not thought about that at all, but it will probably need to, yes.

maskarb (approved):

PR #4897, opened 2024-02-06
[COST-4589] Correctly get end of month data for Azure Metering

cgoodfred:
## Jira Ticket

[COST-4589](https://issues.redhat.com/browse/COST-4589)

## Description

This change will use a case statement to determine if the end of month processing should occur. If we are processing data for this resource and more than 2 days have passed, we expect that data to be updated. This addresses the end of month boundary and another scenario where a resource is turned off as we would never process any usage on the final day of the resource running.

This PR also cleans up a lot of SQL into files per provider and utilizes jinja templating better, it should make the flow simpler to follow as its not as much building SQL in strings.

## Testing
Either fake the date on your system to be the second day of the month or change this value to cross a month boundary: 
```
self.creation_processing_time = self.creation_processing_time - timedelta(days=30)
```
1. Checkout main
2. Restart Koku
3. Ingest Azure data that corresponds to the subs matching and let processing complete
4.  Verify that the final day of a month is not present. So looking at January data, I do not see a record in the csv in the minio bucket that has a `date` of `2024-01-31`.
Sample minio path: 
```
koku-bucket/org1234567/Azure/source=c3df4296-e9f3-4a43-b411-8c7bd525d4dd/date=2024-02-06
```
5. Delete the source created or create a new one
6. Checkout this branch
7. Restart Koku
8. ingest Azure data that again corresponds to subs matching and let processing complete
9. Verify that the final day of a month is now present. Looking at January, you should now see a record in the csv that corresponds to a `date` of `2024-01-31`

## Notes

...


lcouzens (approved):
Coolio!

esebesto (approved):

PR #4899, opened 2024-02-07
temporarily stop using codecov token

maskarb:
The codecov token has stopped working. This PR removes it from our workflow since we technically don't need it for open source projects anyway.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -273,7 +273,7 @@ jobs:
           file: ./coverage.xml
           flags: unittests
           name: Python-${{ matrix.python-version}}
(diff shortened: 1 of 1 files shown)


samdoran:
The funny thing about this is Codecov [updated their docs last year](https://github.com/codecov/codecov-action/pull/955/files#r1164770745) explicitly recommending a token. In the past, they recommended _against_ using a token for public projects since secrets are not available to forks, which breaks coverage uploads for PRs from forks.

Seems like the new codecov CLI tool requires a token. They have probably made some server side changes around this, which broke the use of an explicit token in the unified uploader (which is what the v3 action uses).

djnakabaale (approved):

PR #4902, opened 2024-02-08
[COST-4194] replace jinjasql with jinjasql2

maskarb:
## Jira Ticket

[COST-4194](https://issues.redhat.com/browse/COST-4194)

## Description

This change will replace jinjasql with jinjasql2 and upgrade Jinja (and all other dependencies).

The default `param_style` was changed in jinjasql2, so this PR explicitly defines it to maintain previous functionality.


koku/koku/trino_database.py
```diff
@@ -137,16 +137,8 @@ def executescript(trino_conn, sqlscript, *, params=None, preprocessor=None):
...
-                try:
-                    stmt, s_params = preprocessor(p_stmt, params)
-                except Exception as exc:
-                    LOG.warning(
-                        f"Preprocessor Error ({exc.__class__.__name__}) : {exc}{os.linesep}"
-                        f"Statement template : {p_stmt}{os.linesep}"
-                        f"Parameters : {params}"
-                    )
-                    exc_type = exc.__class__.__name__
-                    raise PreprocessStatementError(f"{exc_type} :: {exc}") from exc
```
maskarb, line 149: This try/except is here because jinjasql would fail to convert an iterable into the correct SQL format. jinjasql2 solved this issue, so the try/except was removed.

koku/masu/api/enabled_tags.py
```diff
@@ -111,7 +111,7 @@ def post(self, request):
                 msg = f"Disabled tags for schema: {schema_name}."
                 LOG.info(msg)
             elif action.lower() == "remove_stale":
-                jinja_sql = JinjaSql()
+                jinja_sql = JinjaSql(param_style="format")
```
samdoran, line 114: Did the default for `param_style` change in this fork?

samdoran (approved):

koku/masu/api/enabled_tags.py
samdoran: Oh, looks like it did. They should update the docs.

PR #4904, opened 2024-02-09
[COST-4392] Converting gigibyte to gigabyte and fixing units labels

bacciotti:
## Jira Ticket

[COST-4392](https://issues.redhat.com/browse/COST-4392)

## Description

This change will:
- Change unit labels from "gibibyte month" to "GB-Mo", and "hours" to "Hrs" to keep the standard;
- Multiply "usage_amount_in_pricing_units" column by 1.074 to convert correctly.

## Testing

1. Checkout Branch
2. Restart Koku
3. Hit endpoint v1/reports/gcp/storage/
    1. You should see the "GB-Mo" and "Hrs" labels


koku/masu/database/trino_sql/reporting_gcpcostentrylineitem_daily_summary.sql
```diff
@@ -37,7 +37,11 @@ SELECT uuid() as uuid,
     nullif(location_region, '') as region,
     json_extract_scalar(json_parse(system_labels), '$["compute.googleapis.com/machine_spec"]') as instance_type,
     max(usage_pricing_unit) as unit,
-    cast(sum(usage_amount_in_pricing_units) AS decimal(24,9)) as usage_amount,
+    CASE
+        WHEN max(usage_pricing_unit) = 'gibibyte month'
+        THEN cast(sum(usage_amount_in_pricing_units) * 1.074 AS decimal(24,9)) -- Convert to gigabyte
```
myersCody, line 42: We may not want to round here, since this will be at a daily aggregation. It could add up over time, we may want to use the exact conversion which would be: `1.073741824`

myersCody (changes requested):
Great job on the case statements, but this will need a few more changes before it is ready to go into main. Let me know if you have any questions or would like to walk through pieces of the review together.

bacciotti:
@myersCody I applied your suggestions. Thank you.

Moreover, I needed to bypass some tests that stopped working, I think we should review this (take a look at the diff, please).

koku/masu/database/trino_sql/gcp/openshift/reporting_ocpgcpcostlineitem_daily_summary.sql
```diff
@@ -303,8 +303,22 @@ SELECT gcp.uuid as gcp_uuid,
...
+        THEN 'GB'
+        WHEN max(usage_pricing_unit) = 'gibibyte month'
+        THEN 'GB-Mo'
+        WHEN max(usage_pricing_unit) = 'gibibyte hour'
+        THEN 'GB-Hours'
+    ELSE max(usage_pricing_unit)
+    END as unit,
+    CASE
+        WHEN max(usage_pricing_unit) = 'gibibyte month' OR max(usage_pricing_unit) = 'gibibyte' OR max(usage_pricing_unit) = 'gibibyte hour'
+        THEN cast(sum(usage_amount_in_pricing_units) * 1.073741824 AS decimal(24,9)) -- Convert to gigabyte
```
myersCody, line 319: So, one thing we want to pay special attention to here is how we are calculating the previous values. Notice how original SQL is using a `max` here:
```
cast(max(gcp.usage_amount_in_pricing_units) AS decimal(24,9)) as usage_amount,
```
But you changed it to a sum to match what you did in the gcp cost entry table:
```
cast(sum(gcp.usage_amount_in_pricing_units) AS decimal(24,9)) as usage_amount,
```
That could result in an unintended bloating of the number:
```
 usage_unit |  original_amount   |    usage_amount_in     |      usage_amount
------------+--------------------+------------------------+------------------------
 GB-Mo      |      244.725077280 |       173704.559606529 |       173704.559606529
 Hrs        |        0.997722222 |          266.195118511 |          266.195118511
 GB         | 92993765.703750000 | 140059886352.765350000 | 140059886352.765350000
 GB-Hours   |        0.335236414 |          130.238963714 |          130.238963714
 ```
 My guess is that the `gcp_openshift_daily` is already an aggregated value, and we might be doing some double aggregation.

**Review by myersCody — DISMISSED** [review]

File: koku/masu/database/trino_sql/reporting_gcpcostentrylineitem_daily_summary.sql [hunk 3]
```diff
@@ -36,8 +36,22 @@ SELECT uuid() as uuid,
     date(usage_start_time) as usage_end,
     nullif(location_region, '') as region,
     json_extract_scalar(json_parse(system_labels), '$["compute.googleapis.com/machine_spec"]') as instance_type,
-    max(usage_pricing_unit) as unit,
-    cast(sum(usage_amount_in_pricing_units) AS decimal(24,9)) as usage_amount,
+    CASE
+        WHEN max(usage_pricing_unit) = 'hour'
+        THEN 'Hrs'
+        WHEN max(usage_pricing_unit) = 'gibibyte'
```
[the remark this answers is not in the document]
bacciotti: Thanks, suggestions taken.

koku/masu/database/trino_sql/gcp/openshift/reporting_ocpgcpcostlineitem_daily_summary.sql
bacciotti: Makes sense, but do you suggest switching back to **max**? Because in the main branch, it is currently using **max**.

myersCody: Yes! I recommend switching back to max for the ocp on gcp flow.

myersCody: @bacciotti I figured out a better way to explain this! I set up a meeting for next week to talk through our daily tables & Trino managed tables. I want to talk through how we create the parquet files for the daily cloud tables.  I think it will help explain why we use max here instead of sum.

koku/masu/database/trino_sql/gcp/openshift/reporting_ocpgcpcostlineitem_daily_summary_resource_id.sql
```diff
@@ -770,8 +770,18 @@ SELECT uuid(),
     sku_id,
     sku_alias,
     region,
-    unit,
-    usage_amount,
+    CASE unit
```
myersCody, line 773: wouldn't you need to `max(unit)`

lcouzens (dismissed):
Need to remove the IQE image marker

lcouzens (dismissed):
Just drop the custom IQE image :)

koku/api/report/test/gcp/openshift/test_ocp_gcp_query_handler.py
```diff
@@ -1368,7 +1368,8 @@ def test_exclude_functionality(self):
                     overall_output = handler.execute_query()
                     overall_total = handler.query_sum.get("cost", {}).get("total", {}).get("value")
                     opt_dict = overall_output.get("data", [{}])[0]
-                    opt_dict = opt_dict.get(f"{exclude_opt}s")[0]
+                    if opt_dict.get(f"{exclude_opt}s"):
+                        opt_dict = opt_dict.get(f"{exclude_opt}s")[0]
                     opt_value = opt_dict.get(exclude_opt)
                     self.assertIsNotNone(opt_dict)
```
samdoran, line 1374: This test looks off. It seems it should be asserting that `opt_value` is not `None` since that is used further on in this test. If `opt_dict` were `None` a `TypeError` would be raised, which you fixed here.

Here's an alternative solution.
```
python
                    opt_dict = opt_dict.get(f"{exclude_opt}s", [{}])[0]
                    opt_value = opt_dict.get(exclude_opt)
                    self.assertIsNotNone(opt_value)
```

koku/api/report/test/gcp/test_gcp_query_handler.py
```diff
@@ -1274,12 +1274,14 @@ def test_query_table(self):
                 GCPCostSummaryByServiceP,
             ),
             (
-                "?filter[service]=Virtual%20Network,VPN,DNS,Traffic%20Manager,ExpressRoute,Load%20Balancer,Application%20Gateway",  # noqa: E501
+                "?filter[service]=Virtual%20Network,VPN,DNS,Traffic%20Manager,ExpressRoute,Load%20Balancer,\
+                Application%20Gateway",
```
samdoran, line 1278: This isn't quite what you want. This preserves the leading white space, so what you end up with is a string without the newline but a lot of white space in the middle:

```
'?filter[service]=Virtual%20Network,VPN,DNS,Traffic%20Manager,ExpressRoute,Load%20Balancer,                Application%20Gateway'
```

I'm surprised the test still passes 🤨.

Instead of an [escape sequence](https://docs.python.org/3/reference/lexical_analysis.html#escape-sequences), use [string literal concatenation](https://docs.python.org/3/reference/lexical_analysis.html#string-concatenation).


```suggestion
                (
                    "?filter[service]=Virtual%20Network,VPN,DNS,Traffic%20Manager,ExpressRouteLoad%20Balancer"
                    "Application%20Gateway"
                ),
```

Note the lack of `,` at the end of each line. This is a [parenthesized expression](https://docs.python.org/3/reference/expressions.html#parenthesized-forms). Adding commas between the elements would make this a tuple of strings, which is not what you want.

koku/api/report/test/gcp/test_gcp_query_handler.py
```diff
@@ -1539,7 +1543,8 @@ def test_exclude_functionality(self):
                     overall_output = handler.execute_query()
                     overall_total = handler.query_sum.get("cost", {}).get("total", {}).get("value")
                     opt_dict = overall_output.get("data", [{}])[0]
-                    opt_dict = opt_dict.get(f"{exclude_opt}s")[0]
+                    if opt_dict.get(f"{exclude_opt}s"):
+                        opt_dict = opt_dict.get(f"{exclude_opt}s")[0]
```
samdoran, line 1547: Same suggestion as above.

koku/api/report/test/gcp/test_gcp_query_handler.py
```diff
@@ -1466,7 +1468,9 @@ def test_query_storage_with_totals(self):
                         self.assertNotEqual(cost_total, Decimal(0))
                         self.assertIsInstance(value.get("usage", {}).get("value"), Decimal)
                     service_checked = True
-        self.assertTrue(service_checked)
+
+        if service_checked:
+            self.assertTrue(service_checked)
```
samdoran, line 1473: What is this test asserting? This changes it to "at least one data_item was checked" to silently passing if no data items were checked. Is that correct?

koku/masu/database/trino_sql/gcp/openshift/reporting_ocpgcpcostlineitem_daily_summary_resource_id.sql
```diff
@@ -281,7 +291,7 @@ SELECT cast(uuid() as varchar),
     max(nullif(gcp.sku_id, '')) as sku_id,
     max(nullif(gcp.sku_description, '')) as sku_alias,
     gcp.location_region as region,
-    max(gcp.usage_pricing_unit) as unit,
+    max(gcp.usage_pricing_unit) as uni4,
```
samdoran, line 294:
```suggestion
    max(gcp.usage_pricing_unit) as unit,
```

samdoran (changes requested):

koku/api/report/test/gcp/openshift/test_ocp_gcp_query_handler.py
bacciotti: Makes sense. Let's try this approach.

bacciotti:
The smokes with the new changes passed.

koku/api/report/test/gcp/test_gcp_query_handler.py
```diff
@@ -1274,12 +1274,16 @@ def test_query_table(self):
...
+                    "?filter[service]=Virtual%20Network,VPN,DNS,Traffic%20Manager,ExpressRouteLoad%20Balancer"
+                    "Application%20Gateway"
+                ),
                 GCPCostView,
                 GCPCostSummaryByServiceP,
             ),
             (
-                "?filter[service]=Virtual%20Network,VPN,DNS,Traffic%20Manager,ExpressRoute,Load%20Balancer,Application%20Gateway&group_by[account]=*",  # noqa: E501
+                "?filter[service]=Virtual%20Network,VPN,DNS,Traffic%20Manager,ExpressRoute,Load%20Balancer,\
+                Application%20Gateway&group_by[account]=*",
```
samdoran, line 1286: This also needs to use implicit string concatenation.

ci/functions.sh
```diff
@@ -132,7 +132,7 @@ function run_smoke_tests_stage() {
     echo "Running E2E tests with IQE:"
     echo "IQE_MARKER_EXPRESSION: '$IQE_MARKER_EXPRESSION'"
     echo "IQE_FILTER_EXPRESSION: '$IQE_FILTER_EXPRESSION'"
-
+
```
samdoran, line 135: This is causing sanity test failures.

koku/api/report/test/gcp/test_gcp_query_handler.py
```diff
@@ -1274,12 +1274,16 @@ def test_query_table(self):
...
+                    "Application%20Gateway"
+                ),
                 GCPCostView,
                 GCPCostSummaryByServiceP,
             ),
             (
-                "?filter[service]=Virtual%20Network,VPN,DNS,Traffic%20Manager,ExpressRoute,Load%20Balancer,Application%20Gateway&group_by[account]=*",  # noqa: E501
+                "?filter[service]=Virtual%20Network,VPN,DNS,Traffic%20Manager,ExpressRoute,Load%20Balancer,"
+                "Application%20Gateway&group_by[account]=*",
                 GCPCostView,
```
samdoran, line 1287: This is a simpler form of doing implicit string concatenation, but it may look like a bug to a future reader. Leave a comment stating this is intentional, or wrap it in an expression as above.

samdoran (approved):

PR #4910, opened 2024-02-12
pin pandas < 2.2

maskarb:
## Description

Pandas 2.2 may have introduced a memory leak. This pins Pandas <2.2 until a fix is released.


esebesto (approved):

Pipfile
```diff
@@ -43,7 +43,7 @@ kombu = "<5.3"  # https://issues.redhat.com/browse/COST-3997
 msrestazure = "*"
 numpy = {version = "*", markers = "platform_machine == 'aarch64' or platform_machine == 'arm64' or platform_machine == 'x86_64'"}
 oci = "*"
-pandas = "*"
+pandas = "<2.2"
```
samdoran, line 46: Can we add a comment about why this constraint is needed so we'll know when it's ok to remove it?

PR #4917, opened 2024-02-15
[COST-4394] Inserting the 'TB_MS' clause in WHEN

bacciotti:
## Jira Ticket

[COST-4394](https://issues.redhat.com/browse/COST-4394)

## Description

This change will convert the "TB_MS" (Terabyte per millisecond) unit to "GB-Mo" on the WHEN clause to keep the standard.

## Testing

1. Checkout Branch
2. Restart Koku
3. Check thought Postgres with `SELECT DISTINCT(unit) FROM <tenant>.reporting_ocicostentrylineitem_daily_summary GROUP BY distinct(unit);`
4. Check if there is no "TB_MS" field **or**
5. Run `OCIReportDBAccessorTest.test_tb_ms_unit` unit test.


koku/masu/test/database/test_oci_report_db_accessor.py
```diff
@@ -161,3 +163,19 @@ def test_table_properties(self):
...
+        with open(file_path) as file:
+            csv = file.readlines()
+            for line in csv:
+                if "TB_MS" in line:
+                    url = reverse("reports-oci-storage")
+                    response = self.client.get(url, **self.headers)
+                    if response.status_code == status.HTTP_200_OK:
+                        unit = response.data["meta"]["total"]["usage"]["units"]
+                        self.assertNotEqual(unit, "TB_MS")
+                        self.assertEqual(unit, "GB-Mo")
```
maskarb, line 181: This test file and that API response are completely unrelated. Unfortunately, we don't run the trino-sql anywhere in unit tests, so I would remove this test and coordinate with QE to ensure there is a test case for this.

PR #4929, opened 2024-02-21
Update redis cache settings

samdoran:
## Jira Ticket

[COST-2231](https://issues.redhat.com/browse/COST-2231)

## Description

Ignore `ConnectionError` and retry instead of failing immediately. This was the configuration we had before with `django_redis` and the `IGNORE_EXCEPTIONS` setting.

With the move to Django’s `RedisCache`, it just passes through the `OPTIONS` directly to the backend library, which in this case is `redis-py`.

`redis-py` actually provides finer grained controls around handling and retries. It’s possible to select which errors are ignore and which retry strategy is used.

I went chose to ignore only `ConnectionError` and retry with exponential backoff. We are only seeing `ConnectionError` currently. This can be changed as needed.

I also cleaned up other redis settings so that values are only defined in one place and made the long integers easier to read.

## Testing

1. Checkout Branch
2. Restart Koku
3. Hit endpoint
4. Things should work


koku/koku/settings.py
```diff
@@ -238,28 +243,22 @@
...
-                "health_check_interval": REDIS_HEALTH_CHECK_INTERVAL,
-                "retry_on_timeout": REDIS_RETRY_ON_TIMEOUT,
-            },
+            "OPTIONS": REDIS_CONNECTION_POOL_KWARGS,
         },
         "worker": {
             "BACKEND": "django.core.cache.backends.db.DatabaseCache",
             "LOCATION": "worker_cache_table",
-            "TIMEOUT": 86400,  # 24 hours
+            "TIMEOUT": 86_400,  # 24 hours
```
myersCody, line 261: Is this underscore intentional here?

samdoran: Yup. Makes long integers more readable.

myersCody (approved):

PR #4933, opened 2024-02-21
[COST-4571] Mechanism to delay summary for settings

myersCody:
## Jira Ticket

[COST-4571](https://issues.redhat.com/browse/COST-4571)

## Description

This change will add in a delay mechanism for kicking off re-summary tasks whenever the settings for tag mapping have been updated. 

## Testing
1. Add a tag key mapping
2. Check the new model

```
select * from delayed_celery_tasks;
-[ RECORD 1 ]-----+-----------------------------------------------------------------------------------------------------------------------------
id                | 12
task_name         | masu.processor.tasks.update_summary_tables
task_args         | ["org1234567"]
task_kwargs       | {"start_date": "2024-02-01 00:00:00+00:00", "provider_type": "AWS", "provider_uuid": "0e4725de-73e1-45bf-a967-fbff70a08277"}
timeout_timestamp | 2024-02-28 15:13:22.531483+00
provider_uuid     | 0e4725de-73e1-45bf-a967-fbff70a08277
queue_name        | summary
metadata          | {}
-[ RECORD 2 ]-----+-----------------------------------------------------------------------------------------------------------------------------
id                | 13
task_name         | masu.processor.tasks.update_summary_tables
task_args         | ["org1234567"]
task_kwargs       | {"start_date": "2024-02-01 00:00:00+00:00", "provider_type": "OCI", "provider_uuid": "b4763e4f-bb38-4c2b-a61c-de86599ab74f"}
timeout_timestamp | 2024-02-28 15:15:04.312772+00
provider_uuid     | b4763e4f-bb38-4c2b-a61c-de86599ab74f
queue_name        | summary
metadata          | {}
```
4. Use django to delete the ids, and watch the tasks start. 

## Notes

...


koku/reporting_common/migrations/0039_delayedcelerytasks.py
```diff
@@ -0,0 +1,28 @@
+# Generated by Django 3.2.23 on 2024-01-18 20:44
+from django.db import migrations
```
myersCody, line 2: I plan to move these migrations to a separate PR.

bacciotti (dismissed):

koku/api/settings/tags/mapping/query_handler.py
```diff
@@ -25,3 +35,24 @@ def format_tag_mapping_relationship(original_response):
...
+@dataclass
+class AddChildQueryHandler:
+    serializer: AddChildSerializer
+    children_rows: List[EnabledTagKeys] = field(default_factory=list)
+    parent_row: EnabledTagKeys = field(init=False)
+
+    def __post_init__(self):
+        """Collects the enabled tag keys rows for the parent & children."""
+        self.parent_row = EnabledTagKeys.objects.get(uuid=self.serializer.data.get("parent"))
+        self.children_rows = list(EnabledTagKeys.objects.filter(uuid__in=self.serializer.data.get("children")))
```
maskarb, line 49: What do we gain by using `dataclass` here?

koku/api/settings/tags/mapping/utils.py
```diff
@@ -0,0 +1,63 @@
...
+from reporting.provider.ocp.models import OCPTagsValues
+from reporting.provider.ocp.models import OCPUsageReportPeriod
+
+
+@dataclass
+class FindTagKeyProviders:
+    """This class given a list of enabled tag key rows will find the
+    provider_uuids associationed with those rows."""
+
+    enabled_rows: List[EnabledTagKeys]
```
maskarb, line 28: Same question as above, why is this a `dataclass`?

koku/koku/cache.py
```diff
@@ -34,6 +34,7 @@ class KokuCacheError(Exception):
 OPENSHIFT_GCP_CACHE_PREFIX = "openshift-gcp-view"
 OPENSHIFT_ALL_CACHE_PREFIX = "openshift-all-view"
 SOURCES_CACHE_PREFIX = "sources"
+DELAYED_RESUMMARY_PREFIX = "delayed_resummary"
```
maskarb, line 37: What's this for?

koku/masu/config.py
```diff
@@ -74,6 +75,9 @@ class Config:
     # Limit the number of enabled tags:
     ENABLED_TAG_LIMIT = ENVIRONMENT.int("TAG_ENABLED_LIMIT", default=DEFAULT_ENABLED_TAG_LIMIT)
 
+    # Wait period for the the delay summary queue.
+    DELAYED_RESUMMARY_TIME = ENVIRONMENT.int("DELAYED_RESUMMARY_TIME", default=DEFAULT_ENABLED_TAG_LIMIT)
```
maskarb, line 79: I understand why this was put into this file, but I think we should stop adding new things to this file and just rely on the settings file. It can be a bit confusing to track config variables thruout our code because they come from different places.

All of our celery config lives in [settings.py](https://github.com/project-koku/koku/blob/cb49aa39c0bf96706e458cbfea29c705a51c1d33/koku/koku/settings.py#L552-L565) anyway, so it would make sense to put this there too.

I would move this to koku.settings, and then use it wherever it is needed like: 
```
settings.py:
DELAYED_RESUMMARY_TIME = ENVIRONMENT.int("DELAYED_RESUMMARY_TIME", default=3600)

where used:
from django.conf import settings
settings.DELAYED_RESUMMARY_TIME
```

koku/reporting_common/models.py
```diff
@@ -141,3 +151,84 @@ class Meta:
...
+        self.task_args = json.dumps(args)
+
+    def get_task_args(self):
+        return json.loads(self.task_args)
+
+    def set_task_kwargs(self, kwargs):
+        self.task_kwargs = json.dumps(kwargs)
+
+    def get_task_kwargs(self):
+        return json.loads(self.task_kwargs)
```
maskarb, line 180: Are these methods necessary if the task_args and task_kwargs fields are JSONFields instead? I see json.load/dump in several places around these fields, so it seems like they should be JSONFields.

koku/reporting_common/models.py
```diff
@@ -141,3 +151,84 @@ class Meta:
...
+
+    def set_timeout(self, timeout_seconds):
+        now = timezone.now()
+        self.timeout_timestamp = now + timezone.timedelta(seconds=timeout_seconds)
+
+    @classmethod
+    def remove_expired_records(cls):
+        now = timezone.now()
+        expired_records = cls.objects.filter(timeout_timestamp__lt=now)
+        expired_records.delete()
```
maskarb, line 190: I think this is a fine way to do this, however, this logic could just live in the task itself:
```
@celery_app.task(name="masu.celery.tasks.delayed_summary_polling", queue=GET_REPORT_FILES_QUEUE)
def delayed_summary_polling(*args, **kwargs):
    """Removes the expired records starting the delayed celery tasks."""
    DelayedCeleryTasks.objects.filter(timeout_timestamp__lt=timezone.now()).delete()
```

Is there another need for this classmethod?

maskarb (commented):
Looks good so far! Some initial feedback:

koku/api/settings/utils.py
```diff
@@ -20,13 +21,42 @@
...
+from masu.processor.tasks import UPDATE_SUMMARY_TABLES_QUEUE
+from masu.processor.tasks import UPDATE_SUMMARY_TABLES_QUEUE_XL
+from masu.processor.tasks import UPDATE_SUMMARY_TABLES_TASK
 from reporting.user_settings.models import UserSettings
+from reporting_common.models import DelayedCeleryTasks
 
 """Utilities for Settings."""
 
 
+def delayed_summarize_current_month(schema_name: str, provider_uuids: list, provider_type: str):
```
maskarb, line 37: This func definition seems out of place here. Is it possible to put this next to the model and the pre-delete method?

koku/reporting_common/models.py
myersCody: This may be personal style, but I preferred having the logic for removing expired records encapsulated within the class. I thought it made it easy how to use the model without having to search the codebase. However, it isn't strictly necessary if you feel strongly that it should be encapsulated within the celery task instead.

koku/api/settings/tags/mapping/query_handler.py
myersCody: Long story short, this used to be combined with the other data class and I decided to pull the FindTagKeyProviders out at the time. Thanks for pointing it out, I think I found a cleaner alternative.

koku/reporting_common/models.py
myersCody: Let me find out... 🤔

myersCody: You are correct! Awesome suggestion @maskarb Thank you.

maskarb (commented):
A few more comments:

koku/reporting_common/models.py
maskarb: Ok, that's fine to leave this as a class method. Maybe the name should be updated though? Does `remove_expired_records` tell someone what is expected to happen when that function is called? I'd argue that it is a bit unclear about what this function does other than deleting records that are expired (that doesn't tell me that the actual task will be triggered)

Maybe something like "trigger_delayed_tasks"? I also think `delayed_summary_polling` is unclear too. Perhaps these could be named the same?

Docstrings would also be good to add.

myersCody: You make a very valid point, and I like your name way better. I wrote the class as a POC, so the function at the time was not actually triggering celery tasks. I think your name does a way better job of explaining the role it plays during data processing. 

Ill push up a change for the docstring & name change. Thanks @maskarb

koku/koku/celery.py
```diff
@@ -206,6 +206,15 @@ def validate_cron_expression(expresssion):
     "schedule": crontab(0, 0, day_of_month="15"),
 }
 
+# Specify the frequency for checking delayed summary tasks
+DELAYED_TASK_POLLING_MINUTES = ENVIRONMENT.get_value("DELAYED_TASK_POLLING_MINUTES", default="30")
+trigger_delayed_tasks_schedule = crontab(minute=f"*/{DELAYED_TASK_POLLING_MINUTES}")
+app.conf.beat_schedule["source_status_beat"] = {
```
maskarb, line 212:
```suggestion
app.conf.beat_schedule["delayed_tasks_trigger"] = {
```

maskarb: This key needs to be unique.

maskarb (approved):

PR #4941, opened 2024-02-27
[COST-4714] - explicitly check for Tenant existence in KokuTenantSchemaExistsMiddleware

maskarb:
## Jira Ticket

[COST-4714](https://issues.redhat.com/browse/COST-4714)

## Description

This change will update `KokuTenantSchemaExistsMiddleware` to explicitly check if the Tenant does or does not exist if the request has raised a ProgrammingError. This PR also simplifies some of the code around getting the Tenant from the db, and some unit tests.

## Testing (requires iqe)

on this branch:
1. start with fresh db, and create test customers (we do not need to ingest data):
    ```
    $ make docker-down
    $ make delete-db
    $ make docker-up-min-no-build (if db doesn't start, CRTL+C, and try again)
    $ make create-test-customer
    ```
1. visit http://localhost:8000/api/cost-management/v1/sources/ and see created sources
1. update the [cost_local_user](https://gitlab.cee.redhat.com/insights-qe/iqe-cost-management-plugin/-/blob/f76ff59e632e6c177cca4f079699fb93c04bf8db/iqe_cost_management/conf/cost_management.default.yaml#L113-129) so that it does not match the local dev user (just add a `1` to the org/account/user info`)
1. `$ ENV_FOR_DYNACONF=local DYNACONF_IQE_VAULT_LOADER_ENABLED=false DYNACONF_IQE_VAULT_IGNORE_PATH_ERRORS=True iqe tests plugin cost_management -k test_api_pre_tenant_create_check --pdb` passes

now, mimic a ProgrammingError, keeping the DB from the previous steps:
1. `$ git checkout 69b5510166f355d470134fbaa0acd0238282f9e7`
3. `$ git checkout remove-unused-middleware -- koku/koku/middleware.py`
4. `make delete-redis-cache`
5. visit http://localhost:8000/api/cost-management/v1/sources/ and see 500 status with a ProgrammingError (may need to refresh the page)
6. `$ ENV_FOR_DYNACONF=local DYNACONF_IQE_VAULT_LOADER_ENABLED=false DYNACONF_IQE_VAULT_IGNORE_PATH_ERRORS=True iqe tests plugin cost_management -k test_api_pre_tenant_create_check --pdb` still passes

What all of this is showing is that:
1. a non-existent Tenant is not seeing a 500 response when they access any API endpoints
2. existent Tenants, when there IS something wrong with the db, DOES see a 500 response <<< this is what we have been missing with some of our failed migrations.

## Notes

...


koku/koku/test_middleware.py
```diff
@@ -34,10 +35,25 @@
...
+from koku.middleware import KokuTenantSchemaExistsMiddleware
 from koku.middleware import RequestTimingMiddleware
 from koku.test_rbac import mocked_requests_get_500_text
 
 
+class KokuTenantSchemaExistsMiddlewareTest(IamTestCase):
+    def test_process_exception_empty_response(self):
+        """Assert that an exception raised for a non-existent Tenant returns empty result set."""
+        mock_request = MagicMock()
+        result = KokuTenantSchemaExistsMiddleware(Mock()).process_exception(mock_request, Exception)
```
samdoran, line 47: An exception instance needs to be passed in, not the class itself.
```suggestion
        result = KokuTenantSchemaExistsMiddleware(Mock()).process_exception(mock_request, Exception())
```

koku/koku/test_middleware.py
```diff
@@ -34,10 +35,25 @@
...
+        self.assertIsInstance(result, JsonResponse)
+
+    def test_process_exception_not_empty_response(self):
+        """Assert that an exception raised for a valid tenant does not return emtpy result set."""
+        mock_request = self.request_context["request"]
+        result = KokuTenantSchemaExistsMiddleware(Mock()).process_exception(mock_request, Exception)
```
samdoran, line 53:
```suggestion
        result = KokuTenantSchemaExistsMiddleware(Mock()).process_exception(mock_request, Exception())
```

koku/koku/test_middleware.py
samdoran: It doesn't affect the test since the code under test is no longe doing an `isintance` check, but it is the correct thing to do.

koku/koku/middleware.py
```diff
@@ -107,7 +106,11 @@ class KokuTenantSchemaExistsMiddleware(MiddlewareMixin):
     """A middleware to check if schema exists for Tenant."""
 
     def process_exception(self, request, exception):
-        if isinstance(exception, (Tenant.DoesNotExist, ProgrammingError)):
+        # double check if the Tenant exists
+        schema_name = request.user.customer.schema_name
+        try:
+            Tenant.objects.get(schema_name=schema_name)
+        except Tenant.DoesNotExist:
```
samdoran, line 113: Do we need to do any special handling of ProgrammingError? It looks like we [removed some important error handling code](https://github.com/project-koku/koku/commit/e06699cee3076a3cf55d6d9b79f34aa0e48b7fc8#diff-36209c7bd69cd5bf23b1960dc4297f06047d910797c7b954cc0dcc14421b706aL116) when settings were deprecated. Do we need to pull some of that back in minus the setting specific parts?

koku/koku/middleware.py
```diff
@@ -187,17 +190,15 @@ def _check_user_has_access(self, request):
...
-            msg = f"User {username} does not have permissions for Cost Management."
+            msg = f"User {request.user.username} does not have permissions for Cost Management."
             LOG.warning(msg)
             # For /user-access we do not want to raise the exception since the API will
             # return a false boolean response that the platfrom frontend code is expecting.
             if request.path != reverse("user-access"):
                 raise PermissionDenied(msg)
 
-    def _get_or_create_tenant(self, request):
+    def _get_tenant(self, request):
```
samdoran, line 201: This is a bit of a misnomer since this method does create a tenant in some circumstances. If it makes sense to change the name, the docstring needs to be updated as well.

maskarb: Can you name all the instances where we would be creating the public schema?

koku/koku/middleware.py
maskarb: What functionality would we need to pull back in?

samdoran: I don't know, really. I was just looking at the code that was removed a while back and wondering if that used to prevent us from swallowing the `ProgrammingError`. This is what was removed.

```python
           if (
                settings.ROOT_URLCONF == "koku.urls"
                and request.path in reverse("settings")
                and (not schema_exists(request.tenant.schema_name) or request.tenant.schema_name == "public")
            ):

                doc_link = generate_doc_link("/")

                err_page = {
                    "name": "middleware.settings.err",
                    "component": "error-state",
                    "errorTitle": "Configuration Error",
                    "errorDescription": f"Before adding settings you must create a Source for Cost Management. "
                    f"<br /><span><a href={doc_link}>[Learn more]</a></span>",
                }

                return JsonResponse(
                    [{"fields": [err_page], "formProps": {"showFormControls": False}}], safe=False, status=200
                )
```

Is any of that still useful or is what you have currently better overall? Trying to reason through that compound condinitonal statement is not fun.

maskarb: No, that bit was required when we were using the platform settings page. Since that was moved, this bit of code was no longer needed.

The issue of swallowing the ProgrammingErrors still existed even with this code in place.

lcouzens (approved):

PR #4952, opened 2024-02-29
[COST-4715] add glitchtip config

maskarb:
## Jira Ticket

[COST-4715](https://issues.redhat.com/browse/COST-4715)

## Description

This change will update our deploy yaml to prepare to switch to Glitchtip. The `API_SENTRY_DSN_KEY`, `LISTENER_SENTRY_DSN_KEY`, `CELERY_SENTRY_DSN_KEY`, and `SOURCE_SENTRY_DSN_KEY` keys are required for Sentry (we have these as individual "projects" in Sentry), but these will be consolidated into env specific projects in Glitchtip.

This PR maintains the defaults for production, so we can push this all the way to production before we need to update any variables in app-interface.


lcouzens (approved):
Nice!

PR #4955, opened 2024-03-03
[COST-4728] Log the AWS region

samdoran:
## Jira Ticket

[COST-4728](https://issues.redhat.com/browse/COST-4728)

## Description

Log the AWS region name. Looking into a customer issue that seems to indicate the region information is missing from the source even though it set correctly in the database.


lcouzens (approved):

aperea86:
cordial health; My name is Alex Perea, I am from Colombia and I urgently wish to have contact with you. I have a series of aerohive devices which I want to be able to use outside of the controller, but I have encountered many barriers since I do not handle the issue with these devices, I want to be able to put them into operation in AP mode and deliver a pair of Vlans. I have tried to manage them from ssh but to no avail. I hope I can get some paid help. Thank you

PR #4965, opened 2024-03-11
[COST-4730] Fixing pagination to paginate by parent instead of by children

bacciotti:
## Jira Ticket

[COST-4730](https://issues.redhat.com/browse/COST-4730)

## Description

This change will change the behavior of `v1/settings/tags/mappings/` pagination. 

**Current behavior**
The `limit` parameter is being applied to the children exclusively. In other words, when we query `v1/settings/tags/mappings/?limit=3`, it will return a maximum of 3 children, regardless of the total number of parents present in the result set.

**Right/expected behavior**
It should paginate by parents. This means that when querying `v1/settings/tags/mappings/?limit=3`, it should retrieve 3 parents along with all of their respective children.

## Testing

1. Checkout Branch
2. Restart Koku
3. Insert `n` tag mappings
4. Hit endpoint `v1/settings/tags/mappings/?limit=n`
    1. Verify if `n` corresponds to the number of parents; if it does, then it's correct.


koku/api/settings/tags/mapping/view.py
```diff
@@ -85,9 +85,10 @@ class SettingsTagMappingView(generics.GenericAPIView):
     def get(self, request: Request, **kwargs):
         filtered_qset = self.filter_queryset(self.get_queryset())
         serializer = self.serializer_class(filtered_qset, many=True)
-        paginator = ListPaginator(serializer.data, request)
+        formatted_response = format_tag_mapping_relationship(serializer)
```
samdoran, line 88: Ther are no type hints on `format_tag_mapping_relationship` so I'm confused. This looks like the caller went from passing in a `Response` object to a `Serializer` object even though the parameter name is `original_response`. That seems to work since they both have a `data` property/attribute, but it may make the interface of `format_tag_mapping_relationship` more clear if it accepts a MutableMapping and returns a MutableMapping.
```suggestion
        formatted_response = format_tag_mapping_relationship(serializer.data)
```

```python
def format_tag_mapping_relationship(data: dict[str: dict[str:str]]) -> dict[str: dict[str:str]]:
   ....
```

Even better would be to use a Pydantic model or dataclass to enforce the type and shape of the data.

samdoran: Here's a rough concept of what storing this data in objects could look like.
```python
import dataclasses

@dataclasses.dataclass(frozen=True)
class TagKey:
    uuid: str
    key: str
    source_type: str
    

@dataclasses.dataclass(frozen=True)
class Relationship:
    parent: TagKey
    child: TagKey
    children: list[TagKey] = dataclasses.field(default_factory=list)
```

koku/api/settings/tags/mapping/view.py
```diff
@@ -85,9 +85,10 @@ class SettingsTagMappingView(generics.GenericAPIView):
...
+        paginator = ListPaginator(formatted_response.data, request)
```
samdoran, line 89: Since only the data is used from the `Response` object here, it seems like there's little benefit to having `format_tag_mapping_relationship` return a `Response` object. That object is used in the `get` method, but I think it would make `format_tag_mapping_relationship`'s function more clear if it took in data, transformerd it, and returned data rather an `Response`. The caller could decide what to do with the data after it is returned.

bacciotti:
@samdoran I pushed 2 commits: the first one is using explicit data typing in the function. The last one is using Dataclasses (for that I had to create a new model.py file).

PR #4966, opened 2024-03-11
[COST-4751] GCP Scrub disabled tags in trino.

myersCody:
## Jira Ticket

[COST-4751](https://issues.redhat.com/browse/COST-4751)

## Description

This change will transition logic to remove disabled keys to to Trino. 

## Testing

1. Checkout Branch
2. Check on the GCP tags:
```
http://localhost:8000/api/cost-management/v1/settings/tags/?provider_type=GCP
```
3. They may all be disabled, if so enable them:
```
http://localhost:8000/api/cost-management/v1/settings/tags/enable/

{
    "ids": ["6862b0ca-7111-4285-b3d5-60bdff889e03", "da382acd-6b09-4b4c-a6db-19b83e0b61dd"]
}
```
4. Resummarize the source
```
http://127.0.0.1:5042/api/cost-management/v1/report_data/?start_date=2024-03-01&schema=org1234567&provider_uuid=b895aaaa-ddd8-4550-8b81-76a218d8c43d
```
5. Check Postgresql to make sure only enable tags appear. 
```
postgres=# SELECT distinct(tags) from reporting_gcpcostentrylineitem_daily_summary;
                    tags
--------------------------------------------
 {}
 {"test_storage_key": "test_storage_label"}
 {"vm_key_proj2": "vm_label_proj2"}
```

## Notes

...


koku/masu/database/trino_sql/reporting_gcpcostentrylineitem_daily_summary.sql
```diff
@@ -61,6 +74,6 @@ GROUP BY billing_account_id,
     date(usage_end_time),
     location_region,
     json_extract_scalar(json_parse(system_labels), '$["compute.googleapis.com/machine_spec"]'),
-    labels,
+    16, -- matches column num of tag's map_filter
```
djnakabaale, line 77: Couple questions for clarity:

1. It seems like this is being used as positional reference for tags column. Do we know if the tags column will not frequently change from being 16th? 
2. Also why positional reference vs the alias `tags` ?

myersCody: Trino doesn't like it when I reference the alias and decides to not execute the query. So I followed [what others did](https://github.com/project-koku/koku/blob/main/koku/masu/database/trino_sql/reporting_ocpusagelineitem_daily_summary.sql#L310) and used the positional reference. 

There is some maintaince, but it seems to be a requirement for the query to function.

djnakabaale: Oh I see, thanks for the clarification.

djnakabaale (approved):
lgtm 👍🏾

PR #4968, opened 2024-03-11
[COST-3761] Add models for OCP node networking costs

samdoran:
## Jira Ticket

[COST-3761](https://issues.redhat.com/browse/COST-3761)
[COST-4740](https://issues.redhat.com/browse/COST-4740)

## Description

Add models for storing networking costs summary, by project, and by node.

## Testing

```
make run-migrations
```

Look at the created tables and verify they are partitioned tables
```
# Connect to the PostgreSQL server
\d org1234567.reporting_ocp_network_summary_p
\d org1234567.reporting_ocp_network_summary_by_node_p
\d org1234567.reporting_ocp_network_summary_by_project_p
```

<details>
  <summary>Example table</summary>

```
                Partitioned table "org1234567.reporting_ocp_network_summary_by_project_p"
              Column               |           Type           | Collation | Nullable |      Default
-----------------------------------+--------------------------+-----------+----------+--------------------
 id                                | uuid                     |           | not null | uuid_generate_v4()
 cluster_alias                     | text                     |           |          |
 cluster_id                        | text                     |           | not null |
 cost_model_cpu_cost               | numeric(33,15)           |           |          |
 cost_model_memory_cost            | numeric(33,15)           |           |          |
 cost_model_rate_type              | text                     |           |          |
 cost_model_volume_cost            | numeric(33,15)           |           |          |
 data_source                       | character varying(64)    |           |          |
 distributed_cost                  | numeric(33,15)           |           |          |
 infrastructure_data_in_gigabytes  | numeric(33,15)           |           |          |
 infrastructure_data_out_gigabytes | numeric(33,15)           |           |          |
 infrastructure_markup_cost        | numeric(33,15)           |           |          |
 infrastructure_raw_cost           | numeric(33,15)           |           |          |
 namespace                         | character varying(253)   |           |          |
 raw_currency                      | text                     |           |          |
 resource_count                    | integer                  |           |          |
 resource_ids                      | character varying(256)[] |           |          |
 usage_end                         | date                     |           | not null |
 usage_start                       | date                     |           | not null |
 cost_category_id                  | integer                  |           |          |
 source_uuid                       | uuid                     |           |          |
Partition key: RANGE (usage_start)
Indexes:
    "reporting_ocp_network_summary_by_project_p_pk" PRIMARY KEY, btree (usage_start, id)
    "ocp_net_summ_proj_namespace" btree (namespace)
    "ocp_net_summ_proj_usage_start" btree (usage_start)
    "reporting_ocp_network_summ_cost_category_id_e2e13dc9" btree (cost_category_id)
    "reporting_ocp_network_summary_by_project_p_source_uuid_df915ee2" btree (source_uuid)
Foreign-key constraints:
    "reporting_ocp_networ_cost_category_id_e2e13dc9_fk_reporting" FOREIGN KEY (cost_category_id) REFERENCES org1234567.reporting_ocp_cost_category(id) DEFERRABLE INITIALLY DEFERRED
    "reporting_ocp_networ_source_uuid_df915ee2_fk_reporting" FOREIGN KEY (source_uuid) REFERENCES org1234567.reporting_tenant_api_provider(uuid) DEFERRABLE INITIALLY DEFERRED
```
</details>

Verify partition tables entries were created correctly.
```
# Connect to the PostgreSQL server
SELECT * FROM org1234567.partitioned_tables WHERE table_name LIKE 'reporting_ocp_network_summary%';
SELECT * FROM template0.partitioned_tables WHERE table_name LIKE 'reporting_ocp_network_summary%';
```


koku/reporting/provider/ocp/models.py
```diff
@@ -268,6 +268,125 @@ class Meta:
...
+    cluster_id = models.TextField()
+    # Cost model fields need to exist even though they are not relevant to this model
+    cost_model_cpu_cost = models.DecimalField(max_digits=33, decimal_places=15, null=True)
+    cost_model_memory_cost = models.DecimalField(max_digits=33, decimal_places=15, null=True)
+    cost_model_rate_type = models.TextField(null=True)
+    cost_model_volume_cost = models.DecimalField(max_digits=33, decimal_places=15, null=True)
+    data_source = models.CharField(max_length=64, null=True)
+    distributed_cost = models.DecimalField(max_digits=33, decimal_places=15, null=True)
+    infrastructure_data_in = models.DecimalField(max_digits=33, decimal_places=15, null=True)
+    infrastructure_data_out = models.DecimalField(max_digits=33, decimal_places=15, null=True)
```
cgoodfred, line 294: These models should be specific aggregations/subsets of the `OCPUsageLineItemDailySummary` table, we can't add fields here that don't exist on the base table because we wont have the data to populate from.

Also when a query for the future API endpoint cannot be resolved by the summary tables it will fall back to the base table (assuming we match the behavior of all the other APIs), so if a field is present here but not there, we can't fall back appropriately.

koku/reporting/provider/ocp/models.py
```diff
@@ -268,6 +268,125 @@ class Meta:
...
+    infrastructure_markup_cost = models.DecimalField(max_digits=33, decimal_places=15, null=True)
+    infrastructure_raw_cost = models.DecimalField(max_digits=33, decimal_places=15, null=True)
+    raw_currency = models.TextField(null=True)
+    resource_count = models.IntegerField(null=True)
+    unit = models.CharField(max_length=63, null=True)
```
cgoodfred, line 299: Same as above, the `unit` field does not exist on the base table, we either need to add the `unit` to the base table and track it very specifically for networking bytes or remove the `unit` field and track the unit in the field name like we do with other fields in the OCP tables.

samdoran: How do we track the unit in the field name? Do we extract the units from the fileld name? That seems fragile.

koku/reporting/provider/ocp/models.py
samdoran: That's an unfortunate relationship that isn't very apparent. It really seems like these fields make sense on this model. I guess we need to add them to `OCPUsageLineItemDailySummary`.

koku/reporting/provider/ocp/models.py
cgoodfred: Well, if you look at a lot of the other fields in the OCP tables you'll see fields like `pod_effective_usage_cpu_core_hours` and `persistentvolumeclaim_capacity_gigabyte_months` which is what I meant by the unit is in the field name. It requires knowing and converting units before values are inserted into the field and then the unit for that field should be constant. It is definitely more fragile than tracking a unit, but its an alternative to adding a unit field to the tables if you wanted to go that route.

cgoodfred:
Overall I think the migrations look good and have what we need. The only thing is I think we should add is the alter column SQL block for each table like we did here [(link)](https://github.com/project-koku/koku/pull/4741/files#diff-18f3e416089723155e2f15070f3e656c33e5b549845d1ebe88cad8fda18373beR94-R97). This just makes it to where the ID is a generated UUID if we don't generate one on record insert and I think thats a fine default to have in there.
```
migrations.RunSQL(
            "alter table reporting_ocp_pod_summary_by_node_p alter column id set default uuid_generate_v4();",
            reverse_sql="select 1;",
        ),
```

koku/reporting/migrations/0319_node_network_costs.py
```diff
@@ -0,0 +1,199 @@
...
+                    models.Index(fields=["usage_start"], name="ocp_net_summ_node_usage_start"),
+                    models.Index(fields=["node"], name="ocp_net_summ_node_node"),
+                ],
+            },
+        ),
+        migrations.RunSQL(
+            "ALTER TABLE reporting_ocp_network_summary_by_node_p ALTER COLUMN ID SET DEFAULT uuid_generate_v4();",
+            reverse_sql="SELECT 1;",
+        ),
+    ]
```
cgoodfred, line 199: We need to include the unset code on the end of the migration like we run the set code at the top.
```suggestion
        ),
        migrations.RunPython(code=unset_pg_extended_mode, reverse_code=set_pg_extended_mode),
    ]
```

samdoran: Done.

cgoodfred (approved):

PR #4981, opened 2024-03-19
Do not hard code S3 endpoint

samdoran:
## Description

Use the `S3_ENDPOINT` environment variable in `load_test_customer_data.sh` to allow use of of a remote host running minio.

## Testing

```
make load-test-customer-data
```


dev/scripts/load_test_customer_data.sh
```diff
@@ -278,8 +278,8 @@ build_aws_data() {
   render_yaml_files "${_yaml_files[@]}"
 
   log-info "Building OpenShift on ${_source_name} report data..."
-  nise_report ocp --static-report-file "$YAML_PATH/ocp_on_aws/rendered_ocp_static_data.yml" --ocp-cluster-id my-ocp-cluster-1 --minio-upload http://localhost:9000 --daily-reports --payload-name "$_ocp_payload"
-  # nise_report ocp --static-report-file "$YAML_PATH/ocp_on_aws/rendered_ocp_static_data.yml" --ocp-cluster-id my-ocp-cluster-1 --minio-upload http://localhost:9000 --payload-name "$_ocp_payload"
+  nise_report ocp --static-report-file "$YAML_PATH/ocp_on_aws/rendered_ocp_static_data.yml" --ocp-cluster-id my-ocp-cluster-1 --minio-upload "$S3_ENDPOINT" --daily-reports --payload-name "$_ocp_payload"
```
maskarb, line 281: I think this needs to change. Locally I have `$S3_ENDPOINT` defined as `http://koku-minio:9000` because that's what the containers running in docker need to use. But outside of those containers (e.g. this script) this endpoint needs to be localhost. That's probably why this script is written the way it is.

I don't mind making it a variable, but it should probably be different than `$S3_ENDPOINT`

samdoran: That's a good point. I put entries in `/etc/hosts` awhile ago to allow name resolution to work from the host but I don't know if anyone else does that. We would have to update the development setup docs in order to use `S3_ENDPOINT`. That's cleaner in that it's just using the same variable everywhere.

Another option would be to use a different variable and fallback to `localhost`. That's less disrutptive but introduces yet another environment variable, and it would only be used by this script.

PR #4976, opened 2024-03-14
[COST-4772] Tag mapping multiple filter functionality

myersCody:
## Jira Ticket

[COST-4772](https://issues.redhat.com/browse/COST-4772)

## Description

This change will add multiple filter functionality to tag mapping endpoint.

## Testing

1. Create some tag mappings

```
http://localhost:8000/api/cost-management/v1/settings/tags/mappings/child/add/
```

```
{
    "parent": "49829a25-13e8-442e-bb6a-aa5449c4714b",
    "children": ["f6a04487-9287-4292-8f44-434d9ab3e65a"]
}
```

2. Check if you can do multiple filters.

*Tag Mapping*

Parent Key
```
http://localhost:8000/api/cost-management/v1/settings/tags/mappings/?filter[parent]=i&filter[parent]=p
```
Example Return:
```
{
    "meta": {
        "count": 3,
        "limit": 3,
        "offset": 0
    },
    "links": {
        "first": "/api/cost-management/v1/settings/tags/mappings/?filter%5Bparent%5D=i&filter%5Bparent%5D=p&limit=3&offset=0",
        "next": null,
        "previous": null,
        "last": "/api/cost-management/v1/settings/tags/mappings/?filter%5Bparent%5D=i&filter%5Bparent%5D=p&limit=3&offset=0"
    },
    "data": [
        {
            "parent": {
                "uuid": "23af1240-03f4-4a07-9b53-351836380ae2",
                "key": "Mapping",
                "source_type": "AWS",
                "children": [
                    {
                        "uuid": "97982dc7-e412-494a-b995-2acf95aab118",
                        "key": "dashed-key-on-aws",
                        "source_type": "AWS"
                    },
                    {
                        "uuid": "7a40eee0-f188-408c-92dc-dc672b6618cb",
                        "key": "environment",
                        "source_type": "AWS"
                    }
                ]
            }
        },
        {
            "parent": {
                "uuid": "49829a25-13e8-442e-bb6a-aa5449c4714b",
                "key": "app",
                "source_type": "AWS",
                "children": [
                    {
                        "uuid": "f6a04487-9287-4292-8f44-434d9ab3e65a",
                        "key": "CoM_RedHat_Rhel_varianT",
                        "source_type": "AWS"
                    }
                ]
            }
        }
    ]
}
```

Child Key
```
http://localhost:8000/api/cost-management/v1/settings/tags/mappings/?filter[child]=k&filter[child]=v
```
Example Return:
```
{
    "meta": {
[...]

PR #4983, opened 2024-03-19
[COST-4786] Remove ast.literal_eval from view

myersCody:
## Jira Ticket

[COST-4786](https://issues.redhat.com/browse/COST-4786)

## Description

This change will use a separate function to convert the string of a list to list of values. 

## Testing

You should still be able to multiple filter for source_type on tag mapping.

```
http://localhost:8000/api/cost-management/v1/settings/tags/mappings/?filter[source_type]=AWS&filter[source_type]=OCP
```
Return:
```
{
    "meta": {
        "count": 1,
        "limit": 1,
        "offset": 0
    },
    "links": {
        "first": "/api/cost-management/v1/settings/tags/mappings/?filter%5Bsource_type%5D=AWS&limit=1&offset=0",
        "next": null,
        "previous": null,
        "last": "/api/cost-management/v1/settings/tags/mappings/?filter%5Bsource_type%5D=AWS&limit=1&offset=0"
    },
    "data": [
        {
            "parent": {
                "uuid": "d9374eef-0ce2-424c-84a4-0e5b94127452",
                "key": "app",
                "source_type": "OCP"
            },
            "children": [
                {
                    "uuid": "03d2bd07-1874-4109-a34c-b664f2837cd9",
                    "key": "com_redhat_rhel_usage",
                    "source_type": "AWS"
                }
            ]
        }
    ]
}
```

## Notes

...

### DIFF
--- a/koku/api/settings/tags/mapping/view.py
+++ b/koku/api/settings/tags/mapping/view.py
@@ -2,7 +2,6 @@
 # Copyright 2024 Red Hat Inc.
 # SPDX-License-Identifier: Apache-2.0
 #
-import ast
 import dataclasses
 
 import django_filters
@@ -35,6 +34,15 @@ from reporting.provider.all.models import EnabledTagKeys
 from reporting.provider.all.models import TagMapping
 
 
+def parse_list(string_to_parse):
+    """Parse a list to a string."""
+    if "[" not in string_to_parse:
+        return [string_to_parse]
+    split_values = string_to_parse.strip("[]").split(",")
+    parse_list = [value.strip().strip("'\"") for value in split_values]
+    return parse_list
+
+
 class CostModelAnnotationMixin:
     def get_annotated_queryset(self):
         when_conditions = []
@@ -55,10 +63,7 @@ class SettingsTagMappingFilter(SettingsFilter):
         default_ordering = ["parent"]
 
     def filter_by_source_type(self, queryset, name, value):
-        try:
-            value = ast.literal_eval(value)
-        except ValueError:
-            value = [value]
+        value = parse_list(value)
         return queryset.filter(Q(parent__provider_type__in=value) | Q(child__provider_type__in=value))
 
 
--- a/koku/api/settings/tags/mapping/view.py
+++ b/koku/api/settings/tags/mapping/view.py
@@ -34,12 +34,13 @@ from reporting.provider.all.models import EnabledTagKeys
 from reporting.provider.all.models import TagMapping
 
 
-def parse_list(string_to_parse):
+def parse_list(string_to_parse: str) -> list[str]:
     """Parse a list to a string."""
-    if "[" not in string_to_parse:
+    if not isinstance(string_to_parse, str):
+        raise TypeError(f"Expected str but got {type(string_to_parse)}")
         return [string_to_parse]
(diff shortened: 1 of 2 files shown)


koku/api/settings/tags/mapping/view.py
```diff
@@ -35,6 +34,15 @@
 from reporting.provider.all.models import TagMapping
 
 
+def parse_list(string_to_parse):
```
myersCody, line 37: Wrote as a separate function for now because I will use it in the multi filter PR I have open. This is just a quick fix to close out COST-4786.

samdoran, line 37: This is converting a string to a list. The current name makes me think it is operating on a list. Type hints would make it clearer.

```suggestion
def parse_list(string_to_parse: str) -> list[str]:
```

koku/api/settings/tags/mapping/view.py
```diff
@@ -35,6 +34,15 @@
...
+    """Parse a list to a string."""
+    if "[" not in string_to_parse:
```
samdoran, line 39: It's a good idea to start with input validation.
```suggestion
    if not isinstance(string_to_parse, str):
        raise TypeError(f"Expected str but got {type(string_to_parse)}")
```

Would it be better to check that the string starts with a left bracket? A contains check seems overly greedy.

koku/api/settings/tags/mapping/view.py
```diff
@@ -35,6 +34,15 @@
...
+        return [string_to_parse]
+    split_values = string_to_parse.strip("[]").split(",")
+    parse_list = [value.strip().strip("'\"") for value in split_values]
```
samdoran, line 42: Also strip leading/trailing spaces.
```suggestion
    parse_list = [value.strip().strip("'\" ") for value in split_values]
```

koku/api/settings/test/tags/mappings/test_view.py
```diff
@@ -338,3 +339,8 @@ def test_order_by_fake_value(self):
         url = url + "?order_by[parent]=FAKE"
         response = self.client.get(url, **self.headers)
         self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+
+    def test_parse_list(self):
+        """Test that we can parse a list."""
+        self.assertEqual(["A", "B"], parse_list("['A', 'B']"))
+        self.assertEqual(["A"], parse_list("A"))
```
samdoran, line 346: Add some scenarios for unexpected data to make sure it does the right thing such as `'some[long],str[ing]'` or passing in unexpected data types.

myersCody: @samdoran In our current implementation I don't think those kind combination would be possible, since the value is being built into the list by the serializer. Do you mind if I update testing in the multi filter PR instead so that I can get this one in? I have a feeling some of your suggestions may make this function obsolete, but would like to remove the ast sooner than latter.

samdoran: Yup, that's fine. I was wondering if the serializer would prevent a ceratain amount of bad data from getting here.

myersCody:
closed in favor of https://github.com/project-koku/koku/pull/4976

PR #4995, opened 2024-03-22
[COST-4802] Add conversion field to subs kafka message

cgoodfred:
## Jira Ticket

[COST-4802](https://issues.redhat.com/browse/COST-4802)

## Description

This change will add a conversion field to the subs kafka message and set it to True as that is currently the only data we process.

## Testing

1. For local testing, either spin up kafka or change this `self.send_kafka_message(msg)` to a print message to see the printed message
2. Create and load the aws test customer data
3. See the message body contains a `conversion=true` field

Smokes should fail until they are updated to expected the new field.

## Release Notes
- [x] proposed release note

```markdown
* [COST-4802](https://issues.redhat.com/browse/COST-4802) Add conversion field to kafka message for Subscriptions
```


chambridge (approved):
LGTM

PR #5006, opened 2024-03-28
[COST-3617] Fixing invoice_month overwriting

lcouzens:
## Jira Ticket

[COST-3617](https://issues.redhat.com/browse/COST-3617)

## Description

This change will address the issue of overwriting the summary date range per invoice month (GCP).

## Testing

1. Checkout Branch
2. Restart Koku
3. Run unit tests
  a. `tox -e py39 -- masu.test.external.downloader.gcp_local.test_gcp_local_report_downloader.GCPLocalReportDownloaderTest`
  b. `tox -e py39 -- masu.test.api.test_report_data.ReportDataTests`
  c. `tox -e py39 -- masu.test.util.gcp.test_common.TestGCPUtils`

## Release Notes
- [ ] Fix overwriting on summary date range per invoice month (GCP)

```markdown
* [COST-3617](https://issues.redhat.com/browse/COST-3617) Fix overwriting on summary date range per invoice month (GCP)
```


koku/masu/database/trino_sql/reporting_gcpcostentrylineitem_daily_summary.sql
```diff
@@ -61,9 +61,6 @@ FROM hive.{{schema | sqlsafe}}.{{table | sqlsafe}}
 CROSS JOIN
     cte_pg_enabled_keys as pek
 WHERE source = '{{source_uuid | sqlsafe}}'
-    AND year = '{{year | sqlsafe}}'
-    AND month = '{{month | sqlsafe}}'
-    AND invoice_month = '{{year | sqlsafe}}{{month | sqlsafe}}'
```
cgoodfred, line 66: Why are we removing these from the SQL here? `year` and `month` are partitions on the data and should not be removed even if we remove `invoice_month`

koku/masu/external/downloader/gcp_local/gcp_local_report_downloader.py
```diff
@@ -30,52 +26,6 @@ class GCPReportDownloaderError(Exception):
     pass
 
 
-def create_daily_archives(tracing_id, account, provider_uuid, filename, filepath, manifest_id, start_date, context):
-    """
-    Create daily CSVs from incoming report and archive to S3.
-
-    Args:
```
cgoodfred, line 37: Why is this function being removed from the local gcp downloader?

koku/masu/processor/parquet/parquet_report_processor.py
```diff
@@ -487,6 +487,10 @@ def convert_to_parquet(self):  # noqa: C901
                     )
                 )
                 raise ParquetReportProcessorError(msg)
+        if self.create_table and not self.trino_table_exists.get(self.report_type):
+            self.create_parquet_table(parquet_base_filename)
+        if self.provider_type not in (Provider.PROVIDER_AZURE):
+            self.create_parquet_table(parquet_base_filename, daily=True)
```
cgoodfred, line 493: What changed to make this addition now necessary?

koku/masu/processor/parquet/parquet_report_processor.py
```diff
@@ -566,8 +570,6 @@ def convert_csv_to_parquet(self, csv_filename: os.PathLike):  # noqa: C901
                     )
                 )
                 self.post_processor.finalize_post_processing()
-            if self.create_table and not self.trino_table_exists.get(self.report_type):
-                self.create_parquet_table(parquet_filepath)
```
cgoodfred, line 570: Why does gcp invoice month impact creating parquet tables here at all?

koku/masu/database/gcp_report_db_accessor.py
```diff
@@ -134,8 +132,6 @@ def populate_line_item_daily_summary_table_trino(
             "schema": self.schema,
             "table": TRINO_LINE_ITEM_TABLE,
             "source_uuid": source_uuid,
-            "year": invoice_month_date.strftime("%Y"),
-            "month": invoice_month_date.strftime("%m"),
```
cgoodfred, line 138: We still need a year and a month to update here as these are how the data is partitioned.

cgoodfred (changes requested):
I have more to review and look over as this is a reasonably big change, but there are a few questions to look at initially and some changes that I don't think we can make around year and month being removed from the SQL.

koku/masu/external/downloader/gcp_local/gcp_local_report_downloader.py
lcouzens: This was me cleaning up the local downloader and moving to the standard non local one, just reducing some code. I forgot I had this change tbh :sweat_smile: not a terrible idea if nothing breaks since its only local but we can drop it out if we want :+1:

koku/masu/processor/parquet/parquet_report_processor.py
lcouzens: This is regarding the sync partitions piece we discussed. `create_parquet_table` is what triggers the sync partitions and we wanted to run this after all processed files, this was an attempt to do that outside the for loop.

koku/masu/processor/parquet/parquet_report_processor.py
lcouzens: same as above, basically moving outside that for loop.

maskarb:
@bacciotti We need to stop running smoke tests on this PR and the comments from @cgoodfred need to be addressed. `year` and `month` must be included in this trino sql because those are necessary for partitioning.

koku/masu/processor/parquet/parquet_report_processor.py
maskarb: Is moving this code related to the invoice_month piece? Maybe that should be done in a different PR.

koku/masu/database/trino_sql/reporting_gcpcostentrylineitem_daily_summary.sql
lcouzens: This might not be accurate and obviously needs testing, I just chucked this together quickly. :smile:  That said since we are not using invoice month I think we want to select and insert the data from both partitions at the same time as we no longer trigger this per month. The selected and inserted data still contains the invoice month so it _should_ still land in the correct places, but I could still be overlooking/missing something.

koku/masu/processor/parquet/parquet_report_processor.py
lcouzens: Yes its related, the partitions wont get synced correctly otherwise. :disappointed:

lcouzens: I guess we could do it as a pre-req if we wanted, but right now it blocks this at least.

lcouzens: Right it was not fully tested, I just threw this together as a start for Lucas. :wink: 

As for the partition sync piece, this was what we discussed on the triage the other week. If you process GCP files cross month bounds we only sync the first months partition rather then once for both months. That lead to us missing the second partition since it wasn't synced. The net of that discussion was to move it outside the loop so we only sync once at the end. It's possible I just missed another change when I pushed this up though because I definitely tested GCP and thought I fixed that exception. :disappointed:

lcouzens:
Also @bacciotti the testing steps on this PR should not be run unit tests :sweat_smile: As Corey mentioned in a comment too, we do need to make sure ingest works correctly for all providers locally before running smokes in this PR. :+1:

bacciotti:
After several tests, I've changed Trino's query again to re-insert the month and year. At first, the local data was ingested correctly. @lcouzens

PR #5019, opened 2024-04-04
[COST-4877] Report correct PostgreSQL version

samdoran:
## Jira Ticket

[COST-4877](https://issues.redhat.com/browse/COST-4877)

## Description

The version string digits changed after PostgreSQL 9. Update the version string parsing to work with both version number schemas.

Change `get_pg_engine_version` method to return a string instead of list of ints. This makes it easier for the caller.

Use a simpler SQL command to get the version string from the database
Remove unecessary use of a global
Add more test cases

## Testing

```
tox -- masu.test.api.test_db_performance.py
```

## Release Notes
- [x] proposed release note

```markdown
* Report correct PostgreSQL version on metrics dashboard (#5019)
```


koku/masu/test/api/test_db_performance.py
```diff
@@ -72,10 +71,27 @@ def test_del_closes_connection(self):
...
         """Test that the db engine version can be retrieved."""
-        with DBPerformanceStats("KOKU", CONFIGURATOR) as dbp:
-            ver = dbp.get_pg_engine_version()
-            self.assertEqual(ver, SERVER_VERSION)
-            self.assertTrue(all(isinstance(v, int) for v in ver))
+        # FIXME: Rewrite using parametrize when pytest is available
+        testcases = (
+            ("090600", "9.6.0"),
+            ("090100", "9.1.0"),
+            ("092600", "9.26.0"),
```
maskarb, line 78: What about 9.0.23?

koku/masu/api/db_performance/db_performance.py
```diff
@@ -140,21 +138,33 @@ def get_pg_settings(self, setting_names=None):
...
+        # The second part of the version number was only used in PastgreSQL < 10
+        # https://www.postgresql.org/support/versioning/
+        # https://www.postgresql.org/docs/release/
+        #
+        # Examples:
+        #   9.0.0, 9.1.2
+        #   10.1, 10.2
+        #
+        major = parts[0]
+        if major < 10:
```
maskarb, line 162: Will `major` ever be less than 10? The oldest supported version of postgres is 12, so we do not need to support the major.minor.patch versions of postgres.

samdoran: Nope, it won't be.

koku/masu/api/db_performance/db_performance.py
```diff
@@ -140,21 +138,33 @@ def get_pg_settings(self, setting_names=None):
...
- where name = 'server_version_num';
-"""
-            res = self._execute(sql, None).fetchone()
-            SERVER_VERSION.extend(res.values())
-
-        return SERVER_VERSION
+    def get_pg_engine_version(self) -> str:
+        sql = "SHOW server_version_num;"
+        res = self._execute(sql, None).fetchone()
+        server_version_num = res["server_version_num"]
```
maskarb, line 144: Instead of the `server_version_num`, we could just use the `server_version`. Then we don't need to do any parsing of the result at all.
```
postgres=# SHOW server_version;
         server_version         
--------------------------------
 14.9 (Debian 14.9-1.pgdg120+1)
```

samdoran: I looked at that but didn't know if we wanted the extra information at the end.

koku/masu/api/db_performance/db_performance.py
samdoran: But I figured you wanted to add better support for 9.x and earlier based on the previous comment. I didn't initially test much for versions < 10.

koku/masu/api/db_performance/db_performance.py
samdoran: I wasn't sure how consistent that output would be between versions. But we could go with a naive use of that. It would simplify things considerably.

maskarb: If you look here:
https://internal.console.redhat.com/api/cost-management/v1/db-performance/db-settings/
you can see the `server_version` which is just 14.9, so I doubt the extra stuff is even there.

samdoran: Interesting. I'm less sure about the output of `server_version` but I can give that a try.

samdoran: I switched to using `server_version`. I just made up some inputs for the test.

koku/masu/test/api/test_db_performance.py
```diff
@@ -170,13 +186,17 @@ def test_explain(self):
...
+                DBPerformanceStats("KOKU", CONFIGURATOR) as dbp,
+            ):
+                res = dbp.explain_sql(bad_sql)
 
         expected = [
             {
-                "query_plan": "Result  (cost=0.00..0.01 rows=1 width=4)\n  Output: 1\nQuery Identifier: 1147616880456321454",  # noqa: E501
+                "query_plan": (
+                    "Result  (cost=0.00..0.01 rows=1 width=4)\n  " "Output: 1\nQuery Identifier: 1147616880456321454"
+                ),
```
maskarb, line 199:
```suggestion
                "query_plan": (
                    "Result  (cost=0.00..0.01 rows=1 width=4)\n  Output: 1\nQuery Identifier: 1147616880456321454"
                ),
```

maskarb (approved):

lcouzens:
@samdoran can we link this to a JIRA please.

PR #5031, opened 2024-04-11
[COST-861] Tokenizing error - skip bad rows when reading csv

maskarb:
## Jira Ticket

[COST-861](https://issues.redhat.com/browse/COST-861)

## Description

This change will add `on_bad_lines="warn"` to read_csv for OCP reports.

## Testing

2. send attached payload to masu on main (it is a payload for the my-test-cluster-2 OCP source (ocp-on-azure))
```
$ curl -F 'file1=@payload.tar.gz'  http://localhost:5042/api/cost-management/v1/ingest_ocp_payload/
```
3. See the payload ingest failure:
```
masu_server  | [2024-04-15 14:30:49,370] ERROR None 23 File /tmp/tmphseyl2po/insights_local/my-ocp-cluster-2/20240401-20240501/2b29eb8a-cac1-4dd8-a985-494e61601618_openshift_report.2.csv could not be parsed. Reason: Error tokenizing data. C error: Expected 6 fields in line 3, saw 9
```
4. checkout this branch and ensure masu reloads.
5. Resend payload again, and see ingest in masu on this branch:
```
2024-04-15 14:33:23,113] INFO None 3971 {'message': 'Successfully extracted OCP for my-ocp-cluster-2/20240401-20240501', 'tracing_id': '2b29eb8a-cac1-4dd8-a985-494e61601618', 'account': 'no_account', 'org_id': 'no_org_id', 'request_id': '054a480b15de4dac9cf709e161daa026', 'cluster_id': 'my-ocp-cluster-2', 'manifest_uuid': '2b29eb8a-cac1-4dd8-a985-494e61601618', 'provider_type': 'OCP', 'schema': 'org1234567'}
masu_server  | Skipping line 3: expected 6 fields, saw 9
masu_server  | 
```

## Release Notes
- [x] proposed release note

```markdown
* [COST-861](https://issues.redhat.com/browse/COST-861) Tokenizing error - skip bad rows when reading csv
```

## Notes:

This payload contains a bad row in the `2b29eb8a-cac1-4dd8-a985-494e61601618_openshift_report.2.csv` file:

[payload.tar.gz](https://github.com/project-koku/koku/files/14980499/payload.tar.gz)


koku/masu/external/kafka_msg_handler.py
```diff
@@ -70,18 +70,21 @@ class KafkaMsgHandlerError(Exception):
...
-    Split local file into daily content.
-    """
-    daily_files = []
-
+def get_data_frame(file_path: os.PathLike):
+    """read csv file into dataframe"""
     try:
-        data_frame = pd.read_csv(file_path, dtype=pd.StringDtype(storage="pyarrow"))
+        return pd.read_csv(file_path, dtype=pd.StringDtype(storage="pyarrow"), on_bad_lines="warn")
     except Exception as error:
```
samdoran, line 77: This should be more targeted unless there is a very good reason for catching all exception types. Catching `pandas.errors.ParserError` and maybe `UnicodeDecodeError` would be better.

koku/masu/external/kafka_msg_handler.py
```diff
@@ -70,18 +70,21 @@ class KafkaMsgHandlerError(Exception):
...
         LOG.error(f"File {file_path} could not be parsed. Reason: {str(error)}")
         raise error
 
+
+def divide_csv_daily(file_path: os.PathLike, manifest_id: int):
+    """
+    Split local file into daily content.
+    """
+    data_frame = get_data_frame(file_path)
```
samdoran, line 86: Does this need to be a separate function? It's great to keep `divide_csv_daily` focused, but I don't know if `get_data_frame` is doing enough to justiify its existence.

koku/masu/test/external/test_kafka_msg_handler.py
```diff
@@ -941,6 +941,19 @@ def test_divide_csv_daily(self):
                     for expected_item in expected:
                         self.assertIn(expected_item, daily_files)
 
+    def test_get_data_frame(self):
+        """Test the divide_csv_daily method."""
```
samdoran, line 945: This test should be broken into two distinct tests: one that asserts it returns as expected with valid data, the other that it behaves as intended with invalid data.

The way it reads right now is not immediatebly obvious what the test is asserting. It's a "no news is good news" test instead of explicitly spelling out pass/fail criteria.

samdoran: Also, the docstring needs to be updated or, if the test codes is obvious enough, omitted.

koku/masu/external/kafka_msg_handler.py
maskarb: This exception is not the focus of this PR. If you feel it should be more targeted, then let's raise a tech-debt issue to address it.

koku/masu/external/kafka_msg_handler.py
samdoran: Excellent. That's a great reason.

samdoran:
What happens if we get a file that has many invalid lines and a substatial portion of the file is ignored without error? We would see a lot of warnings in the logs (one per invalid line in my testing), but would the final data be trustworthy? This may not be a concern, but it's something I was wondering.

lcouzens (dismissed):

samdoran:
> It's not so much that the data would be not trustworthy, there would simply be gaps in the data.

Thanks for clarifying. That's what I thought but wanted to double check.

lcouzens (approved):
LGTM

PR #5045, opened 2024-04-16
Use multi-arch container images for development

samdoran:
## Description

Instead of pulling different images from different registries and maintaining a separate Docker Compose file, use images that contain both architectures.

## Testing

1. Checkout Branch
2. `make docker-up-min`
3. Verify the `trino` and `hive-metastore` images match your system architecture

```
docker ps --format 'table {{ .ID }}\t{{ .Names }}\t{{ .Image }}\t{{ .State }}\t{{ .Label "architecture" }}\t{{ .Ports }}'
```
## Release Notes
- [x] proposed release note

```markdown
* Use multi-arch container images for local development #5045
```


docker-compose.yml
```diff
@@ -645,7 +645,7 @@ services:
 
   trino:
     container_name: trino
-    image: quay.io/cloudservices/ubi-trino:411-002
+    image: quay.io/samdoran/ubi-trino:411-002
```
samdoran, line 648: `443-002` is also available for testing but it requires the changes in #4717.

bacciotti:
I've tested and it worked. The output matches my arch: aarch64. Also, everything is up and running. LGTM.

bacciotti (approved):

PR #5046, opened 2024-04-17
[COST-4940] ros_report_shipper - Log report urls

saltgen:
## Jira Ticket

[COST-4940](https://issues.redhat.com/browse/COST-4940)

## Description

ROS is planning to create Kibana dashboard to display csv files that it may receive from Cost.

This simple change would make it easier to track such csvs.

## Testing

Since this change is quite trivial I have not tested the same locally

### DIFF
--- a/koku/masu/external/ros_report_shipper.py
+++ b/koku/masu/external/ros_report_shipper.py
@@ -115,7 +115,10 @@ class ROSReportShipper:
             return
 
         kafka_msg = self.build_ros_msg(report_urls, upload_keys)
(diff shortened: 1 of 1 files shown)


saltgen:
> @saltgen Why are we adding this log message to cost management instead of the ROS backend?

@maskarb I assumed it would be beneficial to log the urls as and when those are being sent i.e. at the source

maskarb:
What information do you _need_ to log? I'm still not following the logic around adding this in cost management instead of ROS. Does the ROS backend not log the kafka message and contents that it processes?

PR #5051, opened 2024-04-18
[COST-3761] API for node network costs returning example data

samdoran:
## Jira Ticket

[COST-3761](https://issues.redhat.com/browse/COST-3761)

## Description

Add a new API at `reports/openshift/network/` that returns example data for UI development purposes.

Access is gated by a feature flag.

## Testing

Create and enable the feature flag `cost-management.backend.feature-cost-3761-node-network`

```
curl http://localhost:8000/api/cost-management/v1/reports/openshift/network/
```

## Release Notes
- [ ] proposed release note

```markdown
* [COST-3761](https://issues.redhat.com/browse/COST-3761) New API for node network costs (dev only)
```


cgoodfred (approved):
We might want to add additional days or make this generated dynamically since its for UI testing, but we can go forward with this to start and add any additional data as needed.

PR #5066, opened 2024-04-24
[COST-4912] Calculate node based unattributed storage (Azure)

myersCody:
## Jira Ticket

[COST-3912](https://issues.redhat.com/browse/COST-3912)

## Description

This change will calculate the node based unattributed storage for Azure. 

## Testing

1. Checkout Branch
2. Create the test customer
3. Update the unleash flag:
```
def is_feature_unattributed_storage_enabled(account):
    """Should unattributed storage feature be enabled."""
    unleash_flag = "cost-management.backend.unattributed_storage"
    account = convert_account(account)
    context = {"schema": account}
    # TODO: Turn on fallback when QE is ready to start testing
    # UNLEASH_CLIENT.is_enabled(unleash_flag, context, fallback_development_true)
    # return UNLEASH_CLIENT.is_enabled(unleash_flag, context)
    return True
```
4. `make create-test-customer`
5. `make load-test-customer-data test_source=Azure`
6. Ensure that the `Storage unattributed` namespace was populated:
```
select sum(infrastructure_raw_cost), resource_id from reporting_ocpusagelineitem_daily_summary where namespace='Storage unattributed' group by resource_id;
```
```
        sum        |               resource_id
-------------------+-----------------------------------------
 4.236326764000000 | pv-123-claimless
 2.337303951000000 | azure-cloud-prefix-pvc-partial-matching
 1.554439789000000 | disk-id-1234567
 ```

The claimless will be be 100% unattributed cost; however, we can check our other resource to see that the cost was split:
```
select namespace, sum(infrastructure_raw_cost) from reporting_ocpusagelineitem_daily_summary where resource_id='azure-cloud-prefix-pvc-partial-matching' group by namespace;
```
```
      namespace       |        sum
----------------------+-------------------
 news-site            | 0.001427496000000
 Storage unattributed | 2.337303951000000
 ```
 7. Undo the changes to the unleash flag:
 ```
git checkout -- koku/masu/processor/__init__.py
 ```
 This will set it back to being False.
 8. Resummarize the Azure source
 9. Recheck the database to see if unattributed storage was added:
 ```
 select count(*) from reporting_ocpusagelineitem_daily_summary where namespace='Storage unattributed';
 ```
 ```
  count
-------
     0
(1 row)
```
 
 
## Release Notes
- [x] proposed release note

- https://gist.github.com/myersCody/b300fea8db309e55d50a7bf020a8e4b8

### CODE REVIEW

File: koku/masu/database/trino_sql/reporting_ocpazurecostlineitem_daily_summary.sql [hunk 1]
```diff
@@ -599,6 +600,112 @@ JOIN cte_rankings as r
...
+    AND coalesce(nullif(azure.resourceid, ''), azure.instanceid) LIKE '%%Microsoft.Compute/disks/%%'
+    AND filtered.usage_date IS NULL
+GROUP BY coalesce(azure.date, azure.usagedatetime),
+    split_part(coalesce(nullif(azure.resourceid, ''), azure.instanceid), '/', 9),
+    coalesce(nullif(azure.servicename, ''), azure.metercategory),
+    coalesce(nullif(azure.subscriptionid, ''), azure.subscriptionguid),
+    azure.subscriptionname,
+    azure.resourcelocation,
+    coalesce(nullif(azure.billingcurrencycode, ''), nullif(azure.currency, ''), azure.billingcurrency),
+    azure.tags;
```
maskarb, line 707:
```suggestion
    azure.tags
;
```

koku/masu/database/sql/reporting_ocpazure_ocp_infrastructure_back_populate.sql
```diff
@@ -103,3 +103,106 @@ INSERT INTO {{schema | sqlsafe}}.reporting_ocpusagelineitem_daily_summary (
...
+    ocp_azure.resource_id,
+    CASE WHEN ocp_azure.data_source = 'Pod'
+        THEN ocp_azure.pod_labels
+        ELSE '{}'::jsonb
+    END as pod_labels,
+    CASE WHEN ocp_azure.data_source = 'Storage'
+        THEN ocp_azure.pod_labels
+        ELSE '{}'::jsonb
+    END as volume_labels,
+    NULL as source_uuid,
```
maskarb, line 170: This doesn't seem correct. How can there be no source_uuid?

myersCody: So, in the previous back populate query we use the cluster_id to get the the openshift source uuid.
```
    JOIN {{schema | sqlsafe}}.reporting_ocpusagereportperiod AS rp
        ON ocp_azure.cluster_id = rp.cluster_id
            AND DATE_TRUNC('month', ocp_azure.usage_start)::date  = date(rp.report_period_start)
```

However, since the cost are `unattributed` we can not associate these cloud costs to a cluster. So, I don't know which OCP source uuid to use here.

koku/masu/database/sql/reporting_ocpazure_ocp_infrastructure_back_populate.sql
```diff
@@ -103,3 +103,106 @@ INSERT INTO {{schema | sqlsafe}}.reporting_ocpusagelineitem_daily_summary (
...
+    AND ocp_azure.namespace = 'Storage unattributed'
+GROUP BY ocp_azure.report_period_id,
+    ocp_azure.usage_start,
+    ocp_azure.cluster_id,
+    ocp_azure.cluster_alias,
+    ocp_azure.namespace,
+    ocp_azure.data_source,
+    ocp_azure.node,
+    ocp_azure.persistentvolumeclaim,
+    ocp_azure.resource_id,
```
maskarb, line 206: Why do we need to group by resource_id here? Won't this always appear as a single line item as `storage unattributed`? Would the API ever show the individual components of the storage unattributed?

koku/masu/database/sql/reporting_ocpazure_ocp_infrastructure_back_populate.sql
maskarb: I guess if this cost is not associated with a specific cluster, then there is no source-uuid. Right? How do we ensure all `Storage unattributed` are correctly cleaned up? If we delete an Azure source, is this cost removed?

koku/masu/database/sql/reporting_ocpazure_ocp_infrastructure_back_populate.sql
maskarb: Specifically, does the ocp-on-azure endpoint group by resource-id?

koku/masu/database/sql/reporting_ocpazure_ocp_infrastructure_back_populate.sql
```diff
@@ -103,3 +103,106 @@ INSERT INTO {{schema | sqlsafe}}.reporting_ocpusagelineitem_daily_summary (
...
+    ocp_azure.pod_labels
```
maskarb, line 207: For most of these group bys, aren't they mostly null? Do we need to group by `cluster_id`, `cluster_alias`, `namespace`, `data_source`, `node`, `persistentvolumeclaim`, and `pod_labels`?

koku/masu/database/sql/reporting_ocpazure_ocp_infrastructure_back_populate.sql
myersCody: 🤷 If you believe we should remove it I can look into it.

koku/masu/database/sql/reporting_ocpazure_ocp_infrastructure_back_populate.sql
myersCody: I think this is a good use case to double check. I checked that if we resummarize the data it doesn't double, but we should double check the clean up.

koku/masu/database/sql/reporting_ocpazure_ocp_infrastructure_back_populate.sql
myersCody: I just used the previous insert as the skeleton for this insert. We could drop these group bys.

maskarb: Are these values expected to be `null`? If so, I'm not sure it makes sense to aggregate or group by them. If they are _not_ expected to be null, then a group by is fine with me.

koku/masu/database/sql/reporting_ocpazure_ocp_infrastructure_back_populate.sql
cgoodfred: I'm not fully sure this matches my understanding of what we are trying to identify. Aren't we looking for the `node-attached` storage? Shouldn't we be able to attribute these costs to a particular node but not broken down to the namespace level?

koku/masu/database/trino_sql/reporting_ocpazurecostlineitem_daily_summary.sql
```diff
@@ -599,6 +600,113 @@ JOIN cte_rankings as r
...
+    ON date(coalesce(azure.date, azure.usagedatetime)) = filtered.usage_date
+    AND coalesce(nullif(azure.resourceid, ''), azure.instanceid) = filtered.resource_id
+WHERE azure.source = {{azure_source_uuid}}
+    AND azure.year = {{year}}
+    AND azure.month = {{month}}
+    AND coalesce(azure.date, azure.usagedatetime) >= TIMESTAMP '{{start_date | sqlsafe}}'
+    AND coalesce(azure.date, azure.usagedatetime) < date_add('day', 1, TIMESTAMP '{{end_date | sqlsafe}}')
+    AND coalesce(nullif(azure.servicename, ''), azure.metercategory) LIKE '%Storage%'
+    AND coalesce(nullif(azure.resourceid, ''), azure.instanceid) LIKE '%%Microsoft.Compute/disks/%%'
+    AND filtered.usage_date IS NULL
```
cgoodfred, line 699: Maybe I don't understand exactly what data were looking for, but isn't this just Azure storage data for a virtual machine that had no relationship to the OpenShift clusters we have information for? Does it make sense to show cost information from resources that could be completely unrelated to OpenShift?

koku/masu/database/trino_sql/reporting_ocpazurecostlineitem_daily_summary.sql
```diff
@@ -599,6 +600,113 @@ JOIN cte_rankings as r
...
+    sum(cast(coalesce(nullif(azure.quantity, 0), azure.usagequantity) as decimal(24, 9))) as usage_quantity,
+    max(CASE
+        WHEN split_part(azure.unitofmeasure, ' ', 2) = 'Hours'
+            THEN  'Hrs'
+        WHEN split_part(azure.unitofmeasure, ' ', 2) = 'GB/Month'
+            THEN  'GB-Mo'
+        WHEN split_part(azure.unitofmeasure, ' ', 2) != ''
+            THEN  split_part(azure.unitofmeasure, ' ', 2)
+        ELSE azure.unitofmeasure
+    END) as unit_of_measure,
```
cgoodfred, line 679: Does storage data have all of these different units and do we need to do any unit conversion on the quantity?

koku/masu/database/trino_sql/reporting_ocpazurecostlineitem_daily_summary.sql
myersCody: From my understand the point of unattributed cost is track the cloud node attached storage that can not be associated with an Openshift cluster.

koku/masu/database/sql/reporting_ocpazure_ocp_infrastructure_back_populate.sql
myersCody: Part of the requirements for Unattributed storage was to track unprovisioned storage, and the design doc uses `(i. e. EBS volume exists but no PV exists yet` as an example. I took this to mean the Azure  managed disks that have been created, but have not been attached to a node or utilized by Persistent Volumes (PVs) in the cluster.

Feel free to correct me if you feel I am working off of the wrong definition for unprovisioned storage; However, I feel like this SQL does capture what I think to be unprovisioned node attached storage.

koku/masu/database/trino_sql/reporting_ocpazurecostlineitem_daily_summary.sql
maskarb: @myersCody what does `cloud node attached storage` mean?

myersCody: The node-attached storage in a cluster typically refers to the disks that are mounted to VMs serving as nodes within a cluster. Each cloud has their own tooling for creating these disks, for azure this is Managed Disks. I took unattributed "node attached" storage to mean what portion of the Managed disks cost reported in the Azure bill is not not correlated with an OpenShift cluster.

koku/masu/database/sql/reporting_ocpazure_ocp_infrastructure_back_populate.sql
myersCody: *Recap:*
This review comment prompted a lot of great discussions that lead down a path of us updating and refining how we define Unattributed node based storage. These details have been captured and added to the design doc.

koku/masu/database/trino_sql/reporting_ocpazurecostlineitem_daily_summary.sql
myersCody: Recap:
This review comment prompted a lot of great discussions that lead down a path of us updating and refining how we define Unattributed node based storage. These details have been captured and added to the design doc.

myersCody:
Successful Azure Smokes: https://ci.ext.devshift.net/job/koku-pipeline-pr-check-main/2468/

myersCody:
Successful [Azure Smoke](https://ci.ext.devshift.net/job/koku-pipeline-pr-check-main/2821/) Run

koku/masu/database/trino_sql/reporting_ocpazurecostlineitem_daily_summary.sql
```diff
@@ -135,6 +135,19 @@ CREATE TABLE IF NOT EXISTS hive.{{schema | sqlsafe}}.reporting_ocpazurecostlinei
 ) WITH(format = 'PARQUET', partitioned_by=ARRAY['azure_source', 'ocp_source', 'year', 'month', 'day'])
 ;
 
+-- Now create our proper table if it does not exist
+CREATE TABLE IF NOT EXISTS hive.{{schema | sqlsafe}}.azure_openshift_disk_capacities_temp
```
maskarb, line 139: This table needs to added to the trino management command file:
https://github.com/project-koku/koku/blob/main/koku/masu/management/commands/migrate_trino_tables.py#L53

koku/masu/database/trino_sql/reporting_ocpazurecostlineitem_daily_summary.sql
```diff
@@ -285,7 +298,322 @@ GROUP BY coalesce(azure.date, azure.usagedatetime),
...
+        cast(NULL as double) as pod_limit_cpu_core_hours,
+        cast(NULL as double) as pod_usage_memory_gigabyte_hours,
+        cast(NULL as double) as pod_request_memory_gigabyte_hours,
+        cast(NULL as double) as pod_effective_usage_memory_gigabyte_hours,
+        cast(NULL as double) as node_capacity_cpu_core_hours,
+        cast(NULL as double) as node_capacity_memory_gigabyte_hours,
+        cast(NULL as double) as cluster_capacity_cpu_core_hours,
+        cast(NULL as double) as cluster_capacity_memory_gigabyte_hours,
+        cast(NULL as varchar) as pod_labels,
+        cast(NULL as varchar) as volume_labels,
```
maskarb, line 582: (suggestion) I think we should just remove these and and remove them from the insert statement. It's very difficult to follow this sql because of how much extraneous stuff is here.

Also, casting a NULL to something just returns a NULL, so I'm not really sure this is necessary anyway.

koku/masu/database/trino_sql/reporting_ocpazurecostlineitem_daily_summary.sql
```diff
@@ -285,7 +298,322 @@ GROUP BY coalesce(azure.date, azure.usagedatetime),
...
+    max(nullif(azure.service_name, '')) as service_name,
+    max(azure.instance_type) as instance_type,
+    max(azure.subscription_guid) as subscription_guid,
+    max(azure.subscription_name) as subscription_name,
+    max(nullif(azure.resource_location, '')) as resource_location,
+    'GB-Mo' as unit_of_measure, -- Has to have this unit to show up on ocp on cloud storage endpoint
+    max(cast(azure.usage_quantity as decimal(24,9))) as usage_quantity,
+    max(azure.currency) as currency,
+    max(persistentvolumeclaim_capacity_gigabyte) / max(az_disk.capacity) * max(cast(azure.pretax_cost as decimal(24,9)))  as pretax_cost,
+    (max(persistentvolumeclaim_capacity_gigabyte) / max(az_disk.capacity) * max(cast(azure.pretax_cost as decimal(24,9)))) * cast({{markup}} as decimal(24,9)) as markup_cost, -- pretax_cost x markup = markup_cost
```
maskarb, line 426: (question) why is this cast as a decimal? We're inserting into `reporting_ocpazurecostlineitem_project_daily_summary_temp` which defines these fields as `double`, so I'm not sure this cast here is necessary.

koku/masu/database/trino_sql/reporting_ocpazurecostlineitem_daily_summary.sql
```diff
@@ -285,7 +298,322 @@ GROUP BY coalesce(azure.date, azure.usagedatetime),
...
+    cast(NULL as double) as pod_limit_cpu_core_hours,
+    cast(NULL as double) as pod_usage_memory_gigabyte_hours,
+    cast(NULL as double) as pod_request_memory_gigabyte_hours,
+    cast(NULL as double) as pod_effective_usage_memory_gigabyte_hours,
+    cast(NULL as double) as node_capacity_cpu_core_hours,
+    cast(NULL as double) as node_capacity_memory_gigabyte_hours,
+    cast(NULL as double) as cluster_capacity_cpu_core_hours,
+    cast(NULL as double) as cluster_capacity_memory_gigabyte_hours,
+    cast(NULL as varchar) as pod_labels,
+    cast(NULL as varchar) as volume_labels,
```
maskarb, line 441: (suggestion) I think we should just remove these NULL selects and remove them from the insert statement. It's very difficult to follow this sql because of how much extraneous stuff is here.

Also, casting a NULL to something just returns a NULL, so I'm not really sure this is necessary anyway.

koku/masu/database/trino_sql/reporting_ocpazurecostlineitem_daily_summary.sql
```diff
@@ -285,7 +298,322 @@ GROUP BY coalesce(azure.date, azure.usagedatetime),
...
+)
+SELECT azure.uuid as azure_uuid,
+    max(ocp.cluster_id) as cluster_id,
+    max(ocp.cluster_alias) as cluster_alias,
+    'Storage' as data_source,
+    CASE
+        WHEN max(persistentvolumeclaim) = ''
+            THEN 'Storage unattributed'
+        ELSE max(namespace)
+    END as namespace,
```
maskarb, line 409: (question) What is this CASE statement for? When would the `persistentvolumeclaim` be an empty string? Is this when a persistentvolume exists but has no claim?

koku/masu/database/trino_sql/reporting_ocpazurecostlineitem_daily_summary.sql
```diff
@@ -285,7 +298,322 @@ GROUP BY coalesce(azure.date, azure.usagedatetime),
...
+    {{ocp_source_uuid}} as ocp_source,
+    {{year}} as year,
+    {{month}} as month
+FROM azure_line_items as azure
+JOIN postgres.public.reporting_common_diskcapacity as az_disk_capacity
+    ON azure.metername LIKE '%' || az_disk_capacity.product_substring || ' %' -- space here is important to avoid partial matching
+    AND az_disk_capacity.provider_type = 'Azure'
+JOIN cte_ocp_filtered_resources as ocp_filtered
+    ON split_part(coalesce(nullif(azure.resourceid, ''), azure.instanceid), '/', 9) = ocp_filtered.azure_partial_resource_id
+WHERE coalesce(azure.date, azure.usagedatetime) >= TIMESTAMP '{{start_date | sqlsafe}}'
```
maskarb, line 343: (issue, blocking) This WHERE clause needs the partitioning filters:
```
WHERE ...
        AND azure.year = {{year}}
        AND azure.month = {{month}}
```

maskarb (commented):
some questions and comments:

koku/masu/database/trino_sql/reporting_ocpazurecostlineitem_daily_summary.sql
myersCody: Yeah, this is for when a persistentvolume has no claim it should be counted as a unattributed storage.

koku/masu/database/trino_sql/reporting_ocpazurecostlineitem_daily_summary.sql
myersCody: I just followed the syntax used in the rest of the sql file.

koku/masu/database/trino_sql/reporting_ocpazurecostlineitem_daily_summary.sql
myersCody: The file has been updated!

koku/masu/database/trino_sql/reporting_ocpazurecostlineitem_daily_summary.sql
myersCody: Conditions added! Great catch!

koku/masu/database/trino_sql/reporting_ocpazurecostlineitem_daily_summary.sql
myersCody: Good point! I cleaned up all the sql files casting nulls here. I did replace this one with a case statement, I realized we should be saving tags for the rows that are not unattributed storage.

koku/masu/database/trino_sql/reporting_ocpazurecostlineitem_daily_summary.sql
maskarb: Let's not add these casts in this PR because they aren't doing anything beneficial, and it's just extra work for trino to do.

koku/masu/database/trino_sql/reporting_ocpazurecostlineitem_daily_summary.sql
```diff
@@ -285,7 +298,272 @@ GROUP BY coalesce(azure.date, azure.usagedatetime),
...
+SELECT azure.uuid as azure_uuid,
+    max(ocp.cluster_id) as cluster_id,
+    max(ocp.cluster_alias) as cluster_alias,
+    'Storage' as data_source,
+    CASE
+        WHEN max(persistentvolumeclaim) = ''
+            THEN 'Storage unattributed'
+        ELSE max(namespace)
+    END as namespace,
+    max(ocp.node) as node,
```
maskarb, line 399: (question) What is the result of `max(ocp.node)` here for claimless PVs? Below (in lines 461-564), when inserting for `Storage unattributed`, the node is null, but here it seems like the node would be not null. Is that accurate?

myersCody: The node column is null in the ocp daily summary table for the claimless persistent volume. So, it would be null here too:
```
|       node        | persistentvolumeclaim |  persistentvolume   | storageclass |       resource_id        |
| NULL              |                       | pv-volume_claimless | sc_disk_one  | abc_pv-volume_claimless  |
```

maskarb (approved):

PR #5099, opened 2024-05-10
[COST-5028] guard against report_status not existing

maskarb:
## Jira Ticket

[COST-5028](https://issues.redhat.com/browse/COST-5028)

## Description

This change will prevent task failures when `report_status` does not exist for `OCPCloudParquetReportProcessor`.

## Testing

starting on main:
1. load test data (make load-test-customer-data)
2. trigger ocp-on-cloud resummary:
```
http://localhost:5042/api/cost-management/v1/report/process/openshift_on_cloud/?schema=org1234567&provider_uuid={any cloud source uuid}&start_date=2024-04-01
```
3. in worker logs, see task failure:
```
koku-worker-1  | [2024-05-10 17:18:40,119] ERROR 5b2f96c5-43f6-451e-8a02-29902e6bfc15 1160 Task failed: 'OCPCloudParquetReportProcessor' object has no attribute 'report_status'
```
4. checkout branch and ensure worker reloads
5. re-hit this endpoint:
```
http://localhost:5042/api/cost-management/v1/report/process/openshift_on_cloud/?schema=org1234567&provider_uuid={any cloud source uuid}&start_date=2024-04-01
```
6. see the task complete successfully.

## Release Notes
- [x] proposed release note

```markdown
* [COST-5028](https://issues.redhat.com/browse/COST-5028) guard against non-existent report_status
```

### DIFF
--- a/koku/masu/processor/parquet/parquet_report_processor.py
+++ b/koku/masu/processor/parquet/parquet_report_processor.py
@@ -79,14 +79,12 @@ class ParquetReportProcessor:
         ingress_reports_uuid=None,
     ):
         """initialize report processor."""
-        if context is None:
-            context = {}
         self._schema_name = schema_name
         self._provider_uuid = provider_uuid
         self._report_file = Path(report_path)
         self.provider_type = provider_type
         self._manifest_id = manifest_id
-        self._context = context
+        self._context = context or {}
         self.start_date = self._context.get("start_date")
         self.invoice_month_date = None
         if invoice_month := self._context.get("invoice_month"):
@@ -98,10 +96,6 @@ class ParquetReportProcessor:
 
         self.split_files = [Path(file) for file in self._context.get("split_files") or []]
         self.ocp_files_to_process: dict[str, dict[str, str]] = self._context.get("ocp_files_to_process")
-        if self.manifest_id:
-            self.report_status = CostUsageReportStatus.objects.get(
-                report_name=Path(self._report_file).name, manifest_id=self.manifest_id
-            )
 
     @property
     def schema_name(self):
@@ -148,6 +142,14 @@ class ParquetReportProcessor:
     def manifest_id(self):
         """The manifest id."""
         return self._manifest_id
+
+    @property
+    def report_status(self):
+        if self.manifest_id:
+            return CostUsageReportStatus.objects.get(
+                report_name=Path(self._report_file).name, manifest_id=self.manifest_id
+            )
+        return None
 
     @property
     def report_file(self):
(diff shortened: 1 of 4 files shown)


djnakabaale (dismissed):
lgtm 👍🏾

koku/masu/processor/parquet/parquet_report_processor.py
```diff
@@ -79,14 +79,12 @@ def __init__(
...
         """initialize report processor."""
-        if context is None:
-            context = {}
         self._schema_name = schema_name
         self._provider_uuid = provider_uuid
         self._report_file = Path(report_path)
         self.provider_type = provider_type
         self._manifest_id = manifest_id
-        self._context = context
+        self._context = context or {}
```
samdoran, line 87: This is probably fine, but any "falsy" value will now result in `self._context` being initilaized as `{}`.
```suggestion
        self._context = context if context is not None else {}
```

koku/masu/processor/parquet/parquet_report_processor.py
```diff
@@ -149,6 +143,14 @@ def manifest_id(self):
         """The manifest id."""
         return self._manifest_id
 
+    @property
+    def report_status(self):
+        if self.manifest_id:
+            return CostUsageReportStatus.objects.get(
+                report_name=Path(self._report_file).name, manifest_id=self.manifest_id
+            )
```
samdoran, line 151: Could this value be cached on the instance? Assuming `self._report_name = None` is added to the `__init__`:
```suggestion
    def report_status(self):
        if self.manifest_id:
            if self._report_name is None:
                self._report_name = CostUsageReportStatus.objects.get(
                    report_name=Path(self._report_file).name, manifest_id=self.manifest_id
                )
        return self._report_name
```

koku/masu/processor/parquet/parquet_report_processor.py
```diff
@@ -582,7 +584,10 @@ def convert_csv_to_parquet(self, csv_filename: os.PathLike):  # noqa: C901
                 ),
                 exc_info=err,
             )
-            self.report_status.update_status(CombinedChoices.FAILED)
+            if self.report_status:
+                # internal masu endpoints may result in this being None,
+                # so guard this in case there is no status to update
+                self.report_status.update_status(CombinedChoices.FAILED)
```
samdoran, line 590: This looks like potentially two database calls, unless the result is cached as described above. If caching the value in an attribute isn't desireable, the walrus operator would keep this at one database call.
```suggestion
            if report_status := self.report_status:
                # internal masu endpoints may result in this being None,
                # so guard this in case there is no status to update
                report_status.update_status(CombinedChoices.FAILED)
```

koku/masu/processor/parquet/parquet_report_processor.py
maskarb: I don't really see the difference here. Context should be a dict, not some other falsy value.

koku/masu/processor/parquet/parquet_report_processor.py
maskarb: I switched this to a cached_property instead.

samdoran (approved):

PR #5103, opened 2024-05-13
[COST-5024] Avoid custom JSON parsing for GCP credits field

samdoran:
## Jira Ticket

[COST-5024](https://issues.redhat.com/browse/COST-5024)

## Description

Do not modify a valid JSON string in the GCP credits field. Fall back to custom conversion if the string is invalid JSON. Finally, warn if there is still an error.

## Testing

`tox -e py39 -- masu.test.util.gcp.test_gcp_post_processor`

## Release Notes
- [x] proposed release note

```markdown
* [COST-5024](https://issues.redhat.com/browse/COST-5024) Avoid use of custom JSON parsing for GCP credits if possible (#5103)
```


koku/masu/util/gcp/gcp_post_processor.py
```diff
@@ -124,9 +135,11 @@ def _generate_daily_data(self, data_frame):
 
         # this parses the credits column into just the dollar amount so we can sum it up for daily rollups
         rollup_frame = data_frame.copy()
+        # FIXME: Change the credits data to be a dictionary to avoid deserializing
         rollup_frame["credits"] = rollup_frame["credits"].apply(json.loads)
         rollup_frame["daily_credits"] = 0.0
         for i, credit_dict in enumerate(rollup_frame["credits"]):
+            # FIXME: Remove chained assignment
             rollup_frame["daily_credits"][i] = credit_dict.get("amount", 0.0)
```
maskarb, line 143: Instead of adding a FIXME, we could just fix this:
```suggestion
        rollup_frame["daily_credits"] = rollup_frame["credits"].apply(lambda x: x.get("amount", 0.0))
```
```

myersCody (approved):

PR #5135, opened 2024-05-28
[COST-5059] - Change source status beat cadence

lcouzens:
## Jira Ticket

[COST-5059](https://issues.redhat.com/browse/COST-5059)

## Description

This change will change our source status check cadence to once a day rather than running once an hour.

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Release Notes
- [ ] proposed release note

```markdown
* [COST-5059](https://issues.redhat.com/browse/COST-5059) Reduce cadence of source status checks
```


koku/koku/celery.py
```diff
@@ -85,10 +85,10 @@ def readiness_check(self):
         self._write_response(ProbeResponse(status, msg))
 
 
-def validate_cron_expression(expresssion):
+def validate_cron_expression(expresssion, default):
```
samdoran, line 88: This could be an optional parameter using the current default. Returning the literal string`default` violates the contract of this function, which is to validate a crontab or return a valid fallback value.
```suggestion
def validate_cron_expression(expresssion, default="0 4,14 * * *"):
```

samdoran: The default/fallback value should also be validated.

koku/koku/celery.py
```diff
@@ -175,8 +175,10 @@ def validate_cron_expression(expresssion):
 }
 
 # Specify the frequency for pushing source status.
-SOURCE_STATUS_FREQUENCY_MINUTES = ENVIRONMENT.get_value("SOURCE_STATUS_FREQUENCY_MINUTES", default="30")
-source_status_schedule = crontab(minute=f"*/{SOURCE_STATUS_FREQUENCY_MINUTES}")
+status_expression = "0 3 * * *"
+SOURCE_STATUS_SCHEDULE = ENVIRONMENT.get_value("SOURCE_STATUS_SCHEDULE", default=status_expression)
+SOURCE_STATUS_SCHEDULE = validate_cron_expression(SOURCE_STATUS_SCHEDULE, default=status_expression)
```
samdoran, line 180: Readability could be improved here by only assignming `SOURCE_STATUS_SCHEDULE` a validated value. There is also the possibility of an invalid value being passed to `crontab` since the fallback value passed to `validate_cron_expression` is not validated.

A cleaner flow would be:
1. Get the value from the environment or use a local default
2. Validate that value and assign it to `SOURCE_STATUS_SCHEDULE`

```suggestion
fallback = "0 3 * * *"
status_expression = ENVIRONMENT.get_value("SOURCE_STATUS_SCHEDULE", default=fallback)
SOURCE_STATUS_SCHEDULE = validate_cron_expression(status_expression, fallback)
```

myersCody (approved):

PR #5136, opened 2024-05-29
[COST-4966] - Removing unleash logic for savings plan cost

lcouzens:
## Jira Ticket

[COST-4966](https://issues.redhat.com/browse/COST-4966)

## Description

This change will remove the unleash logic for enabling customers savings plan cost over unblended for OCP on AWS. 
This means all customers will use savings plan for calculations if available else fall back to unblended cost. 

## Testing

1. Checkout Branch
2. Restart Koku
3. Add OCP on AWS providers (with savings plan cost)
4. See openshift back populate sql populated using savings plan cost

Additionally if the cloud provider has no savings plan costs we fall back to unblended

## Release Notes
- [ x] proposed release note

```markdown
* [COST-4966](https://issues.redhat.com/browse/COST-4966) Enable Savings plan costs for OCP on AWS calculations
```

Note:
Do not release till after the 1st of June


koku/masu/test/database/test_aws_report_db_accessor.py
```diff
@@ -328,17 +324,13 @@ def test_back_populate_ocp_infrastructure_costs(self, mock_unleash, mock_execute
...
 
         sql_params = {
             "schema": self.schema,
             "start_date": start_date,
             "end_date": end_date,
             "report_period_id": report_period_id,
-            "is_savingsplan_cost": is_savingsplan_cost,
         }
 
         accessor.prepare_query.assert_called_with(sql, sql_params)
```
cgoodfred, line 336: We can just remove this whole half of the test since the only thing that changed was the savings plan flag.

cgoodfred (approved):

PR #5148, opened 2024-06-05
[COST-4981] reset local unleash

maskarb:
## Jira Ticket

[COST-4981](https://issues.redhat.com/browse/COST-4981)

## Description

Closes #5068

This change will:
* hard reset unleash locally:
  * remove everything in the flags.json (this is all deprecated anyway, so we should not rely on this method to seed the db)
  * create 2 "template" flags: `cost-management.backend.schema-flag-template` and `cost-management.backend.source-uuid-flag-template`
  * these flags can be cloned to create a real flag for local testing
* move the `.unleash` dir to `dev/containers/unleash`
  * update the docker-compose to use this new directory
* with unleash v5, we need to use a personal access token (PAT) to use the API. Unfortunately, you need to have a PAT in the db in order to use the API to create new PAT. To get around this, I've written a python script that directly seeds the unleash db with a PAT (`dev/scripts/setup_unleash.py`). This script also seeds the template flags.
* update unleash version to match the insights instance
* several new env variables:
  * the old admin unleash token is renamed `UNLEASH_ADMIN_TOKEN`
  * the new `UNLEASH_TOKEN` is scoped to the `development` environment
  * `UNLEASH_PAT` is the user PAT used to seed the db and can be used with any API calls wanted

## Testing

1. `make docker-down delete db`
2. `make docker-up-db`
3. see logs:
```
$ make docker-up-db
/Users/mskarbek/.rd/bin/docker compose up -d db
...
[+] Running 1/0
 ✔ Container koku-db  Running                                                                                                                                                                                                                                       0.0s 
 PostgreSQL is available!
/Users/mskarbek/.rd/bin/docker compose up -d unleash
...
[+] Running 2/0
 ✔ Container koku-db       Running                                                                                                                                                                                                                                  0.0s 
 ✔ Container koku-unleash  Running                                                                                                                                                                                                                                  0.0s 
dev/scripts/setup_unleash.py <<<<<< all of this is new
waiting for unleash to become ready
unleash is ready
adding unleash personal access token to db
[('user:6188b62f2f59348f3c195b66983147111682f4bb78a3f7ed9626bd84', 'admin', 1, datetime.datetime(3000, 1, 1, 0, 0, tzinfo=datetime.timezone.utc), datetime.datetime(2024, 6, 5, 15, 16, 14, 82000, tzinfo=datetime.timezone.utc), datetime.datetime(2024, 6, 5, 15, 13, 17, 56355, tzinfo=datetime.timezone.utc), 1)]
creating template unleash flags
200
...
```
4. log into the unleash instance and see that there exists a PAT:
http://localhost:4242/profile/personal-api-tokens
5. whilst in unleash, see that there are only 2 unleash flags (mentioned above)

## Release Notes
- [x] proposed release note

```markdown
* [COST-####](https://issues.redhat.com/browse/COST-####) refresh unleash locally
```


docker-compose.yml
```diff
@@ -140,7 +140,7 @@ services:
       - 5001:9000
     volumes:
       - './koku:/koku/koku'
-      - './.unleash:/koku/.unleash'
+      - './dev/containers/unleash:/koku/.unleash'
```
samdoran, line 143: Are the contents of this directory used my masu, the worker, the listener, the subs-worker, or the sources-client?

dev/scripts/setup_unleash.py
```diff
@@ -0,0 +1,179 @@
...
+
+def create_personal_access_token():
+    """Retrieve data from the vendors table"""
+    print("adding unleash personal access token to db")
+    dbinfo = {
+        "database": "unleash",
+        "user": os.getenv("DATABASE_USER") or "postgres",
+        "password": os.getenv("DATABASE_PASSWORD") or "postgres",
+        "port": os.getenv("POSTGRES_SQL_SERVICE_PORT") or 15432,
+        "host": os.getenv("POSTGRES_SQL_SERVICE_HOST") or "localhost",
```
samdoran, line 139:
```suggestion
        "user": os.getenv("DATABASE_USER", "postgres"),
        "password": os.getenv("DATABASE_PASSWORD", "postgres"),
        "port": os.getenv("POSTGRES_SQL_SERVICE_PORT", 15432),
        "host": os.getenv("POSTGRES_SQL_SERVICE_HOST", "localhost"),
```

dev/scripts/setup_unleash.py
```diff
@@ -0,0 +1,179 @@
...
+                        not exists (
+                            select secret from personal_access_tokens where secret = '{unleash_pat}'
+                        )
+                    ;
+                """
+                )
+                cur.execute("select * from personal_access_tokens ;")
+                print(cur.fetchall())
+    except (Exception, psycopg2.DatabaseError) as error:
+        print(error)
```
samdoran, line 161:
```suggestion
        sys.exit(error)
```

dev/scripts/setup_unleash.py
```diff
@@ -0,0 +1,179 @@
...
+def create_feature_flags():
+    print("creating template unleash flags")
+    response = requests.post(
+        "http://localhost:4242/api/admin/features-batch/import",
+        json=data,
+        headers={
+            "Authorization": f"{unleash_pat}",
+        },
+    )
+    print(response.status_code)
```
samdoran, line 173: Maybe a `raise_for_status` or test the status code and provide a human readable message.

.env.example
```diff
@@ -21,7 +21,9 @@ DEVELOPMENT=True
 
 PROMETHEUS_MULTIPROC_DIR='/tmp'
 CURRENCY_URL=https://open.er-api.com/v6/latest/USD
-UNLEASH_TOKEN='*:*.dbffffc83b1f92eeaf133a7eb878d4c58231acc159b5e1478ce53cfc'
+UNLEASH_ADMIN_TOKEN='*:*.dbffffc83b1f92eeaf133a7eb878d4c58231acc159b5e1478ce53cfc'
```
bacciotti, line 24: Quick question: why do we need the old token here?

docker-compose.yml
```diff
@@ -573,7 +574,7 @@ services:
     ports:
       - 4242:4242
     volumes:
-      - './.unleash:/.unleash'
+      - './dev/containers/unleash:/.unleash'
```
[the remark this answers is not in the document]
maskarb: I left this here so that we explicitly drop whatever is in the db when it is started.

dev/scripts/setup_unleash.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+if __name__ == "__main__":
+    wait_unleash_ready()
```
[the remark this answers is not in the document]
maskarb: What would the healthcheck be for unleash?

dev/scripts/setup_unleash.py
```diff
@@ -0,0 +1,179 @@
...
+
+import psycopg2
+import requests
+from requests.adapters import HTTPAdapter
+from requests.adapters import Retry
+
+
+unleash_pat = os.getenv("UNLEASH_PAT") or "user:6188b62f2f59348f3c195b66983147111682f4bb78a3f7ed9626bd84"
+assert isinstance(unleash_pat, str)
+data = {
```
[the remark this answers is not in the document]
maskarb: Why make it more complicated than it needs to be?

docker-compose.yml
maskarb: Yes, the unleash cache is put into this directory which is used by the other containers.

.env.example
maskarb: No real reason other than "just in case".

dev/scripts/setup_unleash.py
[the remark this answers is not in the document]
maskarb: Why make it more complicated than it needs to be? I'm fine with leaving it defined as a dict here because this is the only thing that will use this data.

.env.example
samdoran: It's needed for doing certain tasks with the Unleash API. It's good to have if we need it.

dev/scripts/setup_unleash.py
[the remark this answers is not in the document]
samdoran: The `unleash` image only has `busybox`, so something like this.
```yaml
healthcheck:
  test:
    - "CMD"
    - "wget"
    - "-qO-"
    - "http://localhost:4242/health"
```

That still needs some refinement.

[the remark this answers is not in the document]
maskarb: I don't know how to leverage a healthcheck here. Even when adding one, the make command just moves on, healthy or not. The DB healthcheck is good though, so I added that. I'm just not sure we can do away with the `wait_unleash_ready` from this script.

cgoodfred (approved):

PR #5149, opened 2024-06-06
Override start date for cost model updates

lcouzens:
## Jira Ticket

[COST-5099](https://issues.redhat.com/browse/COST-5099)

## Description

We are currently failing to update cost model distribution costs because of the dates used for these tasks. Before this change we would only process cost model updates for the processing range of the payload sent to cost management via the operator. The issue here is we likely dont have Cloud infra cost data for that processing range yet. This means we dont have any costs to distribute, this also means the previous days we do have costs to distribute we dont calculate because they are outside our processing range.

This change updates our cost model task logic to always run the entire months calculations. Meaning we will always calculate all changes to distribution that occur from any cloud billing data no matter when it is processed.

## Testing

1. Checkout Branch
2. Restart Koku
3. Ingest OCP on AWS data which includes a cost model to distribute unallocated costs with an end date prior to current day.
4. Ingest a second OCP payload for current day
5. So cost model update run for single day and no distributed costs updated

To save time Eva put together a dandy QE test branch up for this:
1. Checkout test branch: esebesto/reproducer_distribution_issue
2. Run the test `test_api_reproduce_infra_raw_cost_distribution_issue`
3. This will stop once data is ingested (test data should show the '2024-06-04' will NOT be distributing data correctly since OCP has not run a cost model update yet)
4. Continue test `c` and it will ingest a single days data for you.
6. See logs show cost model update was run for full month range
7. API now shows distributed cost as expected.

By all means try the above on main and this branch to compare.

## Release Notes
- [ ] proposed release note

```markdown
* [COST-5099](https://issues.redhat.com/browse/COST-5099) Change cost model update task to always run full month calculations
```


koku/masu/processor/tasks.py
```diff
@@ -873,6 +873,9 @@ def update_cost_model_costs(
         None
 
     """
+    # Override cost model start date to calculate costs for full month
+    LOG.info("overriding cost model start date to process full month")
+    start_date = DateHelper().month_start(start_date)
```
myersCody, line 878: So, I think with this approach we loose the ability to update cost model costs for previous months through our masu endpoint.

Maybe we can check to see if the date is within the current month and overwrite it to this month start?

myersCody: Ignore me, I can't read. `month_start` != `this_month_start`

myersCody (dismissed):

myersCody (dismissed):

lcouzens:
Unrelated smoke failure, so all good here!

myersCody (approved):

PR #5154, opened 2024-06-06
Remove suffix from masu trino endpoint

myersCody:
If you include a `;` in the trino query endpoint, you get a 500 error. This remove the suffix `;` to prevent the 500 error.


koku/masu/api/trino.py
```diff
@@ -32,6 +32,7 @@ def trino_query(request):
         data = request.data
         query = data.get("query")
         schema_name = data.get("schema")
+        query = query.removesuffix(";")
```
samdoran, line 35: Move this after the `if query is None` check.

myersCody: Great catch @samdoran

samdoran (approved):

koku/masu/api/trino.py
```diff
@@ -36,6 +36,7 @@ def trino_query(request):
         if query is None:
             errmsg = "Must provide a query key to run."
             return Response({"Error": errmsg}, status=status.HTTP_400_BAD_REQUEST)
+        query = query.removesuffix(";")
```
maskarb, line 39:
```suggestion
        query = query.strip().removesuffix(";")
```
We could add the `.strip()` here too in case someone adds whitespace at the end of a query, like:
```
"query": "SELECT schema_name FROM information_schema.schemata; "
```

maskarb (approved):

PR #5166, opened 2024-06-14
[COST-5141] Add command to remove expired hive partitions

myersCody:
## Jira Ticket

[COST-5141](https://issues.redhat.com/browse/COST-5141)

## Description

This change will add a trino management command for dropping expired data.

## Testing

1. Populate dummy data:
https://gist.github.com/myersCody/d08a166e4e52b764223e02ead47ba3bb

2. Run management command:
```
 DJANGO_READ_DOT_ENV_FILE=True koku/manage.py migrate_trino_tables --remove-expired-partitions reporting_ocpusagelineitem_daily_summary
```
**NOTE:** If you get an error here you may need to set your `TRINO_HOST` to localhost.

Example output
```
[2024-06-17 12:18:24,712] INFO None 39346 running against the following schemas: ['no_ocp_table', 'ocp_table_2', 'ocp_table_two', 'org1234567']
[2024-06-17 12:18:24,712] INFO None 39346 ** dropping expired partitions from table for no_ocp_table
[2024-06-17 12:18:24,753] INFO None 39346 reporting_ocpusagelineitem_daily_summary does not exist for no_ocp_table
[2024-06-17 12:18:24,753] INFO None 39346 ** dropping expired partitions from table for ocp_table_2
[2024-06-17 12:18:24,791] INFO None 39346 reporting_ocpusagelineitem_daily_summary does not exist for ocp_table_2
[2024-06-17 12:18:24,791] INFO None 39346 ** dropping expired partitions from table for ocp_table_two
[2024-06-17 12:18:24,826] INFO None 39346 Current date time is 2024-06-17 00:00:00+00:00
[2024-06-17 12:18:24,826] INFO None 39346 Report data expiration is 2024-01-01 for a 5 month retention policy
[2024-06-17 12:18:24,826] INFO None 39346 Finding expired partitions for ocp_table_two reporting_ocpusagelineitem_daily_summary
[2024-06-17 12:18:24,864] INFO None 39346 Found 6
[2024-06-17 12:18:24,864] INFO None 39346 Removing partition for 123 2021-5
[2024-06-17 12:18:25,176] INFO None 39346 DELETE PARTITION result: [[None]]
[2024-06-17 12:18:25,176] INFO None 39346 Removing partition for 123 2022-1
[2024-06-17 12:18:25,644] INFO None 39346 DELETE PARTITION result: [[None]]
[2024-06-17 12:18:25,644] INFO None 39346 Removing partition for 123 2023-2
[...]

PR #5171, opened 2024-06-17
[COST-5141] vaccum expired trino partitions

myersCody:
## Jira Ticket

[COST-5141](https://issues.redhat.com/browse/COST-5141) 

## Description

This change will add an internal masu endpoint to trigger expired trino partitions logic.

## Testing
Same instructions as (https://github.com/project-koku/koku/pull/5166) except use the masu endpoint instead of the management command to clear the expired partitions

## Release Notes
- [x] proposed release note

```markdown
* [COST-5141](https://issues.redhat.com/browse/COST-5141) Automate removing expired partitions
```


koku/masu/processor/ocp/ocp_report_db_cleaner.py
```diff
@@ -147,3 +148,27 @@ def purge_expired_report_data_by_date(self, expired_date, simulate=False):
...
+                else:
+                    LOG.info("No expired partitions")
+                    return
+                if simulate:
+                    for partition in results:
+                        LOG.info(f"partition_info: {partition}")
+                else:
+                    for result in results:
+                        year, month, source_value = result
+                        accessor.delete_hive_partition_by_month(table, source_value, year, month, source_column)
```
maskarb, line 174: Total nitpick, but this if else isn't totally necessary:
```suggestion
                for partition in results:
                    LOG.info(f"partition_info: {partition}")
                    if not simulate:
                        year, month, source_value = partition
                        accessor.delete_hive_partition_by_month(table, source_value, year, month, source_column)
```

koku/masu/processor/orchestrator.py
```diff
@@ -542,3 +543,32 @@ def remove_expired_report_data(self, simulate=False):
...
+        Removes expired trino partitions for each account.
+        """
+        async_results = []
+        schemas = set(
+            Provider.objects.filter(type=Provider.PROVIDER_OCP)
+            .values_list("customer__schema_name", flat=True)
+            .distinct()
+        )
+        # <QuerySet ['org1234567', 'org1234567', 'org1234567', 'org1234567']>
+        # distinct is not removing duplicates from this list, so using a set to reduce instead
```
maskarb, line 558: The reason this query is not working correctly is because Django is ordering by the api_provider.name which adds an extra select statement:
```
SELECT DISTINCT "api_customer"."schema_name", "api_provider"."name" FROM "api_provider" LEFT OUTER JOIN "api_customer" ON ("api_provider"."customer_id" = "api_customer"."id") WHERE "api_provider"."type" = 'OCP' ORDER BY "api_provider"."name" ASC LIMIT 21; args=('OCP',)
```

That ordering is coming from [here](https://github.com/project-koku/koku/blob/main/koku/api/provider/models.py#L108).

To get around this, you can add an `order_by()` to your query, which basically tells us to order by nothing:
```
            Provider.objects.order_by().filter(type=Provider.PROVIDER_OCP)
            .values_list("customer__schema_name", flat=True)
            .distinct()

```
which results in the follow sql and result:
```
>>> Provider.objects.order_by().filter(type=Provider.PROVIDER_OCP).values_list("customer__schema_name", flat=True).distinct()

DEBUG:django.db.backends:(0.040) SELECT DISTINCT "api_customer"."schema_name" FROM "api_provider" LEFT OUTER JOIN "api_customer" ON ("api_provider"."customer_id" = "api_customer"."id") WHERE "api_provider"."type" = 'OCP' LIMIT 21; args=('OCP',); alias=default
<QuerySet ['org1234567']>
>>> 
```

koku/masu/processor/tasks.py
```diff
@@ -310,6 +311,30 @@ def remove_expired_data(schema_name, provider, simulate, provider_uuid=None, que
     _remove_expired_data(schema_name, provider, simulate, provider_uuid)
 
 
+@celery_app.task(name="masu.processor.tasks.remove_expired_trino_partitions", queue=DEFAULT)
+def remove_expired_trino_partitions(schema_name, provider_type, simulate, provider_uuid=None, queue_name=None):
```
maskarb, line 315: Why is `queue_name` an argument? It is not used.

koku/masu/processor/_tasks/remove_expired.py
```diff
@@ -34,3 +35,36 @@ def _remove_expired_data(schema_name, provider, simulate, provider_uuid=None):
     if removed_data:
         status_msg = "Expired Data" if simulate else "Removed Data"
         LOG.info(log_json(msg=status_msg, removed_data=removed_data, context=context))
+    # We could extend the logic below here, or keep it as a separate celery task
```
maskarb, line 38: Let's just keep it separate.

koku/masu/api/expired_data.py
```diff
@@ -36,3 +36,23 @@ def expired_data(request):
...
+
+@never_cache
+@api_view(http_method_names=["GET", "DELETE"])
+@permission_classes((AllowAny,))
+@renderer_classes(tuple(api_settings.DEFAULT_RENDERER_CLASSES))
+def expired_trino_partitions(request):
+    """Return expired data."""
+    simulate = True
+    if request.method == "DELETE" and Config.DEBUG:
+        simulate = False
```
maskarb, line 49: Is this intentional? This endpoint can't be used in production.

maskarb (commented):
some questions and comments

koku/masu/api/expired_data.py
myersCody: Yes, I wanted to keep the logic consistent with what we do for `expired_data` endpoint.

koku/masu/processor/_tasks/remove_expired.py
myersCody: whoops forgot to remove this comment 🙇

koku/masu/processor/orchestrator.py
myersCody: Awesome! Thanks for digging into this one for me Michael. 🙇

maskarb (dismissed):
lgtm!

koku/masu/api/expired_data.py
maskarb: sounds good!

koku/masu/processor/ocp/ocp_report_db_cleaner.py
```diff
@@ -147,3 +148,25 @@ def purge_expired_report_data_by_date(self, expired_date, simulate=False):
                 LOG.info(log_json(msg="deleted table partitions", count=del_count, schema=self._schema))
 
         return removed_items
+
+    def purge_expired_trino_partitions(self, expired_date, provider_uuid=None, simulate=False):
+        """Removes expired trino partitions."""
+        LOG.debug(f"purge_expired_trino_partitions: {expired_date}, {provider_uuid}, {simulate}")
+        if expired_date is None and provider_uuid is None:
```
samdoran, line 155: If one of `expired_date` or `provider_uuid` is required, then `expired_date` cannot be a required parameter. I'm also assuming it's ok to provide both values and that `None` is not valid.
```suggestion
    def purge_expired_trino_partitions(self, expired_date=None, provider_uuid=None, simulate=False):
        """Removes expired trino partitions."""
        LOG.debug(f"purge_expired_trino_partitions: {expired_date}, {provider_uuid}, {simulate}")
        require_one_of = (expired_date, provider_uuid)
        if all(value is None for value in require_one_of):
```

If you want to check if a parameter was provided in addition to its value, you can use a sentinel object instead of `None` as the default value.

samdoran: Where is `provider_uuid` used?

koku/masu/database/ocp_report_db_accessor.py
```diff
@@ -283,6 +283,26 @@ def delete_hive_partitions_by_source(self, table, partition_column, provider_uui
...
+    SELECT year as year,
+        month as month,
+        day as day,
+        cast(date_parse(concat(year, '-', month, '-', day), '%Y-%m-%d') as date) as partition_date,
+        {source_column} as source
+    FROM  "{table}$partitions"
+) as partitions
+WHERE partitions.partition_date < DATE '{date_str}'
+GROUP BY partitions.year, partitions.month, partitions.source
+"""
```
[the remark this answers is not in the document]
myersCody: So, I did look into this and tried to utilize the built in `sql_params` available in the `_execute_trino_raw_sql_query`. However, when the params where passed to the EXECUTE IMMEDIATE query, it does not properly replace the `?` with the parameters.
```
EXECUTE IMMEDIATE '
SELECT partitions.year, partitions.month, partitions.source
FROM (
    SELECT year as year,
        month as month,
        day as day,
        cast(date_parse(concat(year, ''-'', month, ''-'', day), ''%Y-%m-%d'') as date) as partition_date,
        ? as source
    FROM  "?$partitions"
) as partitions
WHERE partitions.partition_date < DATE ''?''
GROUP BY partitions.year, partitions.month, partitions.source' USING 'source','reporting_ocpusagelineitem_daily_summary','2024-03-01'
```
Error: ` Incorrect number of parameters: expected 1 but found 3`
I did attempt to adjust my query to get something functional. I discovered a fix for the date. However, I am still running into issues trying to replace `"{table}$partitions"` within Trino itself. I even tried passing the entire from clause in as param to work around the double quotes. However, Trino does not seem happy with `?` in a from clause. 

```
trino:org1234567> EXECUTE IMMEDIATE '
               -> SELECT partitions.year, partitions.month, partitions.source
               -> FROM (
               ->     SELECT year as year,
               ->         month as month,
               ->         day as day,
               ->         cast(date_parse(concat(year, ''-'', month, ''-'', day), ''%Y-%m-%d'') as date) as partition_date,
               ->         ? as source
               ->     FROM  ?
               -> ) as partitions
               -> WHERE partitions.partition_date < CAST(? as DATE)
               -> GROUP BY partitions.year, partitions.month, partitions.source' USING 'source','"reporting_ocpusagelineitem_daily_summary$partitions"','2024-03-01';
```
I believe the solution to be to not use the `qmark` param style when preparing the query. However, that  would involve allowing me to change the default preprocessor for jinja queries in the `report_db_accessor_base.py`. At this point I decided maybe we are are starting to reach outside the scope of this issue. 

My vote is to create a tech debt issue for using parameters for all of our db accessor. I took a look and it appears we use f-strings for all of db_accessor files including the base. I think we should consider refactoring these as a whole.

koku/masu/processor/ocp/ocp_report_db_cleaner.py
myersCody: Good call out Sam, I went ahead and removed that param. 🙇

koku/masu/database/ocp_report_db_accessor.py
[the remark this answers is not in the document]
samdoran: It's definintely not something that is easy to address, especially for all cases. A separate issue to address that is warranted. I want to call it out so we don't get accustomed to using string interpolation for SQL queries because it will bite us eventually.

koku/reporting/models.py
```diff
@@ -176,3 +177,31 @@
     "gcp_openshift_daily_resource_matched_temp": "ocp_source",
     "gcp_openshift_daily_tag_matched_temp": "ocp_source",
 }
+
+# These are cleaned during expired_data flow
+EXPIRE_MANAGED_TABLES = {
```
samdoran, line 182: These could be [MappingProxyType](https://docs.python.org/3/library/types.html#types.MappingProxyType) in order to make them immutable.

samdoran (approved):

samdoran (approved):

PR #5177, opened 2024-06-19
[COST-4852] feat: adding an exception to handle TrinoErrors.

bacciotti:
## Jira Ticket

[COST-4852](https://issues.redhat.com/browse/COST-4852)

## Description

This change will handle the **NoSuchKey** error on Trino.

## Testing

1. Checkout Branch
2. Restart Koku
3. Try to create and load sample data.

## Release Notes
- [ ] proposed release note

```markdown
* [COST-4852](https://issues.redhat.com/browse/COST-4852) Handling some erros during Trino SQL execution.
```


lcouzens (commented):
It might just be me miss-understanding, but I don't fully grasp the changes. To me it looks like we added a new exception for `TrinoExternalError` which would already get collected by the generic `Exception` no? 

Then within that new exception we have additional retry logic?? Is this necessary since we are already rerunning the queries until the retry count is 0 in the generic exception anyway?

The real change here I guess are these lines:
`if "NoSuchKey" in str(ex) and attempts_left > 0:`
            `LOG.warning(log_json(msg="file not found, retrying", log_ref=log_ref, context=ctx), exc_info=ex)`

To which, would it not make more sense to just put this inside the generic exception?
For example:
`   except Exception as ex:`
         `   if attempts_left == 0:`
              `  # Additional if changing the log type based on "NoSuchKey" in str(ex)`

bacciotti:
Well, I think I didn't get where in the current code and in your suggestion the query would be re-run. We haven't a new call to **_execute_trino_raw_sql_query_with_description** function here:

```
        except Exception as ex:
            if attempts_left == 0:
                LOG.error(log_json(msg="failed trino sql execution", log_ref=log_ref, context=ctx), exc_info=ex)
            raise ex
```

That is, when is it going to be called again in this scope (if there are attempts left)?

I mean, unless I explicitly put a new call to **_execute_trino_raw_sql_query_with_description** recursively, it wouldn't call while there are attempts left.

PR #5188, opened 2024-06-25
[COST-5198] - split read traffic to read replica db using nginx proxy

chambridge:
## Jira Ticket

[COST-5198](https://issues.redhat.com/browse/COST-5198)

## Description

This change will ...

* update nginx with HTTP method routing
* switch koku-api to koku-api-writes
* duplicate koku-api to koku-api-reads add a optional mounted secret for the read replica
* update clowder configurator to read from read replica secret if mounted and enabled via ENV var

## Testing

1. Checkout the EE deployment
2. After the IQE tests have completed (you can look at the PR check), note that API request that went to the read or write api deployment
3. All test should pass as the API should default to use the clowder DB out of the box

## Release Notes
- [ ] proposed release note

```markdown
* [COST-5198](https://issues.redhat.com/browse/COST-5198) split read traffic to read replica db using nginx proxy
```


koku/koku/configurator.py
```diff
@@ -500,26 +502,46 @@ def get_object_store_region(requested_name: str = ""):
     @staticmethod
     def get_database_name():
         """Obtain database name."""
+        use_readreplica = ClowderConfigurator._use_read_replica()
+        if use_readreplica:
+            with open("/etc/db/readreplica/db_name") as file:
+                return file.read().rstrip()
```
samdoran, line 508: `pathlib` makes reading the contents of a file a bit more succinct. I'm not sure we need the local variable unless it's there to increase legibility.
```suggestion
        if ClowderConfigurator._use_read_replica():
            try:
                return pathlib.Path("/etc/db/readreplica/db_name").read_text().rstrip()
            except FileNotFoundError:
                # maybe log something here
```

koku/koku/configurator.py
```diff
@@ -534,6 +556,18 @@ def get_database_ca_file():
...
+    def _use_read_replica():
+        read_replica_file_list = [
+            "/etc/db/readreplica/db_host",
+            "/etc/db/readreplica/db_port",
+            "/etc/db/readreplica/db_name",
+            "/etc/db/readreplica/db_user",
+            "/etc/db/readreplica/db_password",
+        ]
+        use_read_replica = ENVIRONMENT.bool("USE_READREPLICA", default=False)
+        return use_read_replica and all(list(map(os.path.isfile, read_replica_file_list)))
```
samdoran, line 569: `all` accepts an iterable, so there's no need to unwind the iterable from `map` into a list.
```suggestion
        return use_read_replica and all(map(os.path.isfile, read_replica_file_list))
```

A generator expression can be used instead of `map()`.
```suggestion
        return use_read_replica and all(pathilb.Path(file).is_file() for file in read_replica_file_list)
```

koku/koku/configurator.py
```diff
@@ -534,6 +556,18 @@ def get_database_ca_file():
             return LoadedConfig.rds_ca()
         return None
 
+    @staticmethod
+    def _use_read_replica():
```
samdoran, line 560:
```suggestion
    def _use_read_replica() -> bool:
```

koku/koku/configurator.py
chambridge: updates made.

koku/koku/configurator.py
chambridge: updates made.

koku/koku/configurator.py
chambridge: updates made.

lcouzens (approved):

djnakabaale (approved):
lgtm 👍🏾

PR #5194, opened 2024-06-28
[COST-5213] - fix S3 prepare

lcouzens:
## Jira Ticket

[COST-5213](https://issues.redhat.com/browse/COST-5213)

## Description

This change will revert the default of `s3_parquet_cleared` flag. Then during daily processing we should set this flag accordingly.

The problem before is each worker was flipping this flag back and forth instead of holding a constant for all running workers. Meaning as we processed more files for a single provider this filtering/collection of files for deletion would take longer and longer while ultimately deleting nothing. 

## Testing

1. Checkout Branch
2. Restart Koku with multiple workers
3. Load AWS/Azure data with finalised bills which have multiple report files (X2)
4. Watch the logs and see ` removed s3 files and marked manifest s3_parquet_cleared` is only called once per month. 

## Release Notes
- [ ] proposed release note

```markdown
* [COST-5213](https://issues.redhat.com/browse/COST-5213) Improve collection for delete from S3
```


koku/masu/util/aws/common.py
```diff
@@ -871,6 +870,8 @@ def clear_s3_files(
                     exc_info=err,
                 )
     delete_s3_objects(request_id, to_delete, context)
+    manifest = ReportManifestDBAccessor().get_manifest_by_id(manifest_id)
+    ReportManifestDBAccessor().mark_s3_parquet_cleared(manifest)
```
samdoran, line 874: Do we need to create two `ReportManifestDBAccessor` instances? These methods look like they could be class methods. Otherwise, create an instance once instead of avoiding the overhead of creating the instance twice.

lcouzens: Yeh that was a mistake for sure. Thanks.

koku/masu/util/aws/common.py
```diff
@@ -871,6 +870,9 @@ def clear_s3_files(
                     exc_info=err,
                 )
     delete_s3_objects(request_id, to_delete, context)
+    manifest_accessor = ReportManifestDBAccessor()
+    manifest = manifest_accessor.get_manifest_by_id(manifest_id)
+    manifest_accessor.mark_s3_parquet_cleared(manifest)
```
myersCody, line 875: Can we add a comment to highlight that this is the only reason we are not deleting all parquet files later on in the process.

cgoodfred (approved):

myersCody (approved):

PR #5196, opened 2024-06-28
[COST-5214] Remove directive to make the build work with docker

samdoran:
## Jira Ticket

[COST-5214](https://issues.redhat.com/browse/COST-5214)

## Description

There is a podman bug that needs to be resolved which required adding this out of stage directive. Unfortunately that broke the image build for docker, which we use heavily.

This is a terrible hack until the bug in podman is resolved.

https://github.com/containers/podman/issues/23046

## Testing

1. Checkout Branch
2. `make docker-build`

## Release Notes
- [x] proposed release note

```markdown
* [COST-5214](https://issues.redhat.com/browse/COST-5214) Fix the container image for local development
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -340,7 +340,14 @@ _koku-wait:
      done
 
 docker-build:
+    # HACK: Remove the ARG TARGETARCH directive to make the build work with docker
+    #       until https://github.com/containers/podman/issues/23046 is resolved.
+	@tail -n +3 Dockerfile > Dockerfile.temp
+	@mv Dockerfile.temp Dockerfile
 	$(DOCKER_COMPOSE) build koku-base
+	@echo "ARG TARGETARCH\n" | cat - Dockerfile > Dockerfile.temp
+	@mv Dockerfile.temp Dockerfile
+
 
 docker-up: docker-build
 	$(DOCKER_COMPOSE) up -d --scale koku-worker=$(scale)
--- a/Makefile
+++ b/Makefile
@@ -340,13 +340,8 @@ _koku-wait:
      done
 
 docker-build:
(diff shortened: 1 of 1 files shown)


samdoran:
What's very confusing about this is that adding the global `ARG TARGETARCH` declaration fixed the build for `podman`, but it breaks it for `docker`.

As best I can tell, `docker` [sets `TARGETARCH` automatically](https://docs.docker.com/reference/dockerfile/#automatic-platform-args-in-the-global-scope). By adding `ARG TARGETARCH` it sets the value to nothing, which is what breaks the docker build.

Changing the directive to `ARG TARGETARCH=arm64` allows it work for both docker and podman, but hard cording the architecture defeats the purpose of using the variable in the first place.

Makefile
```diff
@@ -340,7 +340,14 @@ _koku-wait:
      done
 
 docker-build:
+    # HACK: Remove the ARG TARGETARCH directive to make the build work with docker
+    #       until https://github.com/containers/podman/issues/23046 is resolved.
+	@tail -n +3 Dockerfile > Dockerfile.temp
+	@mv Dockerfile.temp Dockerfile
 	$(DOCKER_COMPOSE) build koku-base
+	@echo "ARG TARGETARCH\n" | cat - Dockerfile > Dockerfile.temp
+	@mv Dockerfile.temp Dockerfile
```
maskarb, line 349: could we just update the command to this instead?
```
$(DOCKER_COMPOSE) build --build-arg TARGETARCH=$(shell uname -m) koku-base
```

maskarb (dismissed):

myersCody (approved):

PR #5210, opened 2024-07-09
[COST-4852] Inserting attempts loop to Trino queries

bacciotti:
## Jira Ticket

[COST-4852](https://issues.redhat.com/browse/COST-4852)

## Description

This change will implement an attempts' logic on Trino queries to avoid some Exceptions raisings.
We are removing the loop from the delete function and inserting in the functions that execute insertion queries.
I've set the attempts to three just for Azure for now. Let's monitor the Exceptions.

## Testing

1. Checkout Branch
2. Restart Koku
3. Test local ingestion, specially with Azure
4. Run `tox -e py39 -- hcs.test.database.test_report_db_accessor.TestHCSReportDBAccessor `and  `tox -e py39 -- masu.test.database.test_azure_report_db_accessor.AzureReportDBAccessorTest`

## Release Notes
- [ ] Improving some Trino queries logic to avoid Exceptions.

```markdown
* [COST-4582](https://issues.redhat.com/browse/COST-4582) Improving some Trino queries logic to avoid Exceptions.
```


koku/koku/trino_database.py
```diff
@@ -138,20 +137,23 @@ def executescript(trino_conn, sqlscript, *, params=None, preprocessor=None):
...
+                    break
+                except TrinoQueryError as trino_exc:
+                    trino_statement_error = TrinoStatementExecError(
+                        statement=stmt, statement_number=stmt_num, sql_params=s_params, trino_error=trino_exc
+                    )
+                    LOG.warning(f"{trino_statement_error!s}")
+                    raise trino_statement_error from trino_exc
+                except Exception as exc:
+                    LOG.warning(str(exc))
+                    raise
```
lcouzens, line 157: I'm not sure the attempt_left has any retry logic here, so we are just going to raise on the first exception no?

koku/masu/database/report_db_accessor_base.py
```diff
@@ -133,21 +132,33 @@ def _execute_trino_raw_sql_query_with_description(
...
-            results = trino_cur.fetchall()
-            description = trino_cur.description
-        except Exception as ex:
-            if attempts_left == 0:
-                LOG.error(log_json(msg="failed trino sql execution", log_ref=log_ref, context=ctx), exc_info=ex)
-            raise ex
-        running_time = time.time() - t1
-        LOG.info(log_json(msg="executed trino sql", log_ref=log_ref, running_time=running_time, context=ctx))
-        return results, description
+        attempts_left = max(attempts_left, 1)
```
lcouzens, line 135: Would it be better to just set the default to 3 attempts in the definition so any trino query will always get 3 attempts?

koku/masu/database/report_db_accessor_base.py
```diff
@@ -133,21 +132,33 @@ def _execute_trino_raw_sql_query_with_description(
...
+            try:
+                trino_cur = trino_conn.cursor()
+                trino_cur.execute(sql, bind_params)
+                results = trino_cur.fetchall()
+                description = trino_cur.description
+                running_time = time.time() - t1
+                LOG.info(log_json(msg="executed trino sql", log_ref=log_ref, running_time=running_time, context=ctx))
+                return results, description
+            except Exception as ex:
+                if attempts_left > 0:
```
lcouzens, line 150: Are you sure this works as expected? I could be wrong but I don't see any change to the original attempts_left value so we will never get to 0 and stop retrying? I think your missing something like `attempts_left=(attempts_left - 1) - i` somewhere.
```
attempts = 3
for i in range(attempts):
    if attempts > 0:
        print(True)
    print(i)

Result:
True
0
True
1
True
2

**Review by lcouzens — CHANGES_REQUESTED** [review]

File: koku/koku/trino_database.py [hunk 4]
```diff
@@ -107,7 +106,7 @@ def connect(**connect_args):
     return trino.dbapi.connect(**trino_connect_args)
 
 
-def executescript(trino_conn, sqlscript, *, params=None, preprocessor=None):
+def executescript(trino_conn, sqlscript, *, params=None, preprocessor=None, attempts_left=0):
```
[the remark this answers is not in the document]
bacciotti: I've created a `HIVE_EXECUTE_QUERY_RETRIES` variable to handle that. Thanks.

koku/koku/trino_database.py
bacciotti: Yes it wasn't right. I've put the right logic now.

koku/masu/database/report_db_accessor_base.py
bacciotti: It wasn't correct. I've fixed it! Thanks.

lcouzens (dismissed):

koku/koku/trino_database.py
```diff
@@ -149,6 +148,15 @@ def executescript(trino_conn, sqlscript, *, params=None, preprocessor=None):
                 cur.execute(stmt, params=s_params)
                 results = cur.fetchall()
             except TrinoQueryError as trino_exc:
+                if "NoSuchKey" in str(trino_exc) and trino_external_error_retries > 0:
```
samdoran, line 151: Rather than using the `str` of the exception object, use [one of the properties](https://github.com/trinodb/trino-python-client/blob/master/trino/exceptions.py#L83-L114) for a more exact comparison.

koku/koku/trino_database.py
```diff
@@ -149,6 +148,15 @@ def executescript(trino_conn, sqlscript, *, params=None, preprocessor=None):
...
+                    LOG.debug("TrinoExternalError Exception, retrying...")
+                    return executescript(
+                        trino_conn=trino_conn,
+                        sqlscript=sqlscript,
+                        params=params,
+                        preprocessor=preprocessor,
+                        trino_external_error_retries=trino_external_error_retries - 1,
+                    )
```
samdoran, line 159: This should probably include delay and/or backoff. Adding some jitter may also be helpful to prevent multiple processes from retrying at the same time.

koku/masu/database/report_db_accessor_base.py
```diff
@@ -153,6 +164,27 @@ def _execute_trino_raw_sql_query_with_description(
         LOG.info(log_json(msg="executed trino sql", log_ref=log_ref, running_time=running_time, context=ctx))
         return results, description
 
+    def _handle_trino_external_error(
```
samdoran, line 167: This would be better as a [decorator](https://realpython.com/primer-on-python-decorators/). That way we would not need separate but mostly similar implementations and we avoid any potential issues that arise from recursive calls.

[Here is a pretty good example](https://github.com/pyprogrammerblog/python-retry/blob/master/python_retry/retry.py#L9).

koku/hcs/test/database/test_report_db_accessor.py
```diff
@@ -84,3 +88,62 @@ def test_data_hcs_customer(self, mock_dba_query, mock_fh_writer, mock_fh):
...
             self.assertIn("acquiring marketplace data", _logs.output[0])
             self.assertIn("data found", _logs.output[1])
+
+    def test_handle_trino_external_error_with_no_such_key(self):
+        """Test handle_trino_external_error with NoSuchKey error."""
+        accessor = ReportDBAccessorBase(schema="test_schema")
+
+        with patch.object(accessor, "_execute_trino_raw_sql_query_with_description") as mock_retry, patch(
+            "masu.database.report_db_accessor_base.LOG"
+        ) as mock_log:
```
samdoran, line 98: As of Python 3.10, `with` statements can be broken across multiple lines using `()` for better readability..
```suggestion
        with (
            patch.object(accessor, "_execute_trino_raw_sql_query_with_description") as mock_retry,
            patch("masu.database.report_db_accessor_base.LOG") as mock_log
        ):
```

koku/masu/database/report_db_accessor_base.py
```diff
@@ -153,6 +164,27 @@ def _execute_trino_raw_sql_query_with_description(
...
+        self, ex, sql, sql_params, context, log_ref, attempts_left, trino_external_error_retries, conn_params, ctx
```
samdoran, line 168: An interface with this many parameters should require the use of kwargs. It's too easy to get parameters out of order causing confusing errors.

koku/hcs/test/database/test_report_db_accessor.py
```diff
@@ -84,3 +88,62 @@ def test_data_hcs_customer(self, mock_dba_query, mock_fh_writer, mock_fh):
...
+            mock_log.error.assert_not_called()
+
+    def test_handle_trino_external_error_without_no_such_key(self):
+        """Test handle_trino_external_error without NoSuchKey error."""
+        accessor = ReportDBAccessorBase(schema="test_schema")
+        error_instance = TrinoExternalError({"error": "Trino Error"})
+
+        with patch.object(accessor, "_execute_trino_raw_sql_query_with_description") as mock_retry, patch(
+            "masu.database.report_db_accessor_base.LOG"
+        ) as mock_log, self.assertRaises(TrinoExternalError):
```
samdoran, line 114:
```suggestion
        with (
            patch.object(accessor, "_execute_trino_raw_sql_query_with_description") as mock_retry,
            patch("masu.database.report_db_accessor_base.LOG") as mock_log
        ):
            self.assertRaises(TrinoExternalError):
```

koku/koku/test_trino_db_utils.py
```diff
@@ -146,6 +146,46 @@ def cursor(self):
             conn = FakerFakeTrinoConn()
             executescript(conn, sqlscript)
 
+    def test_retry_logic_on_no_such_key_error(self):
+        class FakerFakeTrinoCur(FakeTrinoCur):
+            execute_calls = 0
+
+            def execute(self, *args, **kwargs):
+                self.__class__.execute_calls += 1
+                if self.__class__.execute_calls == 1:
```
samdoran, line 155: This should be an instance attribute, not a class attribute.
```suggestion
        class FakerFakeTrinoCur(FakeTrinoCur):
            def __init__(self, *args, **kwargs):
                self.execute_calls = 0

            def execute(self, *args, **kwargs):
                self.execute_calls += 1
                if self.execute_calls == 1:
```

samdoran (changes requested):

koku/masu/database/report_db_accessor_base.py
maskarb: To add to this, `context` and `ctx` should represent the same thing, so we really shouldn't be passing both of these to a function. Admittedly, this `extract_context_from_sql_params` is not implemented very well and is probably overly complicated, but still, we should just pick one or the other to log instead of both.

koku/masu/database/report_db_accessor_base.py
bacciotti: I've created a decorator. Thanks!

koku/masu/database/report_db_accessor_base.py
```diff
@@ -23,10 +24,45 @@
...
+                )
+                kwargs["trino_external_error_retries"] -= 1
+                return wrapper(self, *args, **kwargs)
+            LOG.error(
+                log_json(msg="failed trino sql execution: TrinoExternalError", log_ref=log_ref, context=ctx),
+                exc_info=ex,
+            )
+            raise ex
+
+    return wrapper
```
samdoran, line 63: This decorator needs to work with function or a method so it needs two nested functions. The number of retries should be a parameter to the retry function, not extraceted from the callable being wrapped.
```suggestion
def retry_query(
    original_callable: t.Callable = None,
    *,
    retries: int = 3,
    retry_on: tuple[Exception, ...],
    attribute_to_check: str = "message",
    max_wait: int = 30,
):
    def _retry_query(callable: t.Callable):
        @functools.wraps(callable)
        def wrapper(func, *args, **kwargs):
            sql_params = kwargs.get("sql_params", {})
            log_ref = kwargs.get("log_ref", "Trino query")
            context = ReportDBAccessorBase.extract_context_from_sql_params(sql_params)

            for attempt in range(retries + 1):
                try:
                    return func(*args, **kwargs)
                except retry_on as ex:
                    if attempt < retries and "NoSuchKey" in getattr(ex, attribute_to_check ,""):
                        LOG.warning(
                            log_json(msg="TrinoExternalError Exception, retrying...", log_ref=log_ref, context=context),
                            exc_info=ex,
                        )
                        backoff = min(2**attempt, max_wait)
                        jitter = random.uniform(0, 1)
                        delay = backoff + jitter
                        time.sleep(delay)
                        continue

                    LOG.error(
                        log_json(msg="failed trino sql execution: TrinoExternalError", log_ref=log_ref, context=context),
                        exc_info=ex,
                    )
                    raise ex

        return wrapper

    if original_callable:
        return _retry_query(original_callable)

    return _retry_query
```

samdoran: The only thing I don't really like about this above suggestion is it's tightly coupled to the callable only for logging purposes. Maybe the logging could be changed so that this decorator could be used more broadly without the tight coupling to the kwargs of the callable.

koku/masu/database/report_db_accessor_base.py
```diff
@@ -118,8 +154,17 @@ def _execute_trino_raw_sql_query(self, sql, *, sql_params=None, context=None, lo
         )
         return results
 
+    @handle_trino_external_error
```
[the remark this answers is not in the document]
samdoran: The name doesn't really matter but if the decorator accepts a tuple of exceptions to retry on as suggested above, the name should be changed to reflect that it is not limited to a specific exception type.

koku/masu/database/report_db_accessor_base.py
```diff
@@ -118,8 +154,17 @@ def _execute_trino_raw_sql_query(self, sql, *, sql_params=None, context=None, lo
...
-        self, sql, *, sql_params=None, context=None, log_ref="Trino query", attempts_left=0, conn_params=None
+        self,
+        sql,
+        *,
+        sql_params=None,
+        context=None,
+        log_ref="Trino query",
+        attempts_left=0,
+        conn_params=None,
+        trino_external_error_retries=3,
```
samdoran, line 167: The try except statement later on in this method also needs to be reworked since that functionality is handled by the decorator.

Anything that calls this method with `attempts_left` no longer needs to. The calling code should be updated.

samdoran, line 167: These should be a parameters to the decorator, not this function.
```suggestion
        conn_params=None,
```

koku/common/sql_retries.py
```diff
@@ -0,0 +1,72 @@
...
+
+def log_json(msg, log_ref, context):
+    return f"{msg} - {log_ref} - {context}"
+
+
+def retry(
+    original_callable: t.Callable = None,
+    *,
+    retries: int = 3,
+    retry_on: tuple[Exception, ...],
```
samdoran, line 22: This actually needs to be a subclass of Exception.
```suggestion
    retry_on: tuple[t.Type[Exception], ...],
```

samdoran (commented):
The decorator should probaby go in `koku.common` that way the import statement is `from koku.common import retry`. I don't think it justifies its own module at this time.

koku/masu/database/report_db_accessor_base.py
```diff
@@ -114,12 +114,19 @@ def _execute_raw_sql_query(self, table, sql, bind_params=None, operation="UPDATE
     def _execute_trino_raw_sql_query(self, sql, *, sql_params=None, context=None, log_ref=None, attempts_left=0):
         """Execute a single trino query returning only the fetchall results"""
         results, _ = self._execute_trino_raw_sql_query_with_description(
-            sql, sql_params=sql_params, context=context, log_ref=log_ref, attempts_left=attempts_left
+            sql, sql_params=sql_params, context=context, log_ref=log_ref
         )
         return results
 
+    @retry(retry_on=(Exception,), context_extractor=extract_context_from_sql_params)
```
samdoran, line 121: The defaults can be used here. I don't think we need to specify a context extractor. Just use `ReportDBAccessorBase.extract_context_from_sql_params` directly in the decorator for now. 
```suggestion
    @retry
```

koku/koku/trino_database.py
```diff
@@ -107,6 +108,7 @@ def connect(**connect_args):
     return trino.dbapi.connect(**trino_connect_args)
 
 
+@retry(retry_on=(Exception,))
```
samdoran, line 111:
```suggestion
@retry(retry_on=(TrinoQueryError,))
```

koku/common/sql_retries.py
samdoran: Set the default to just retry on any excption.
```suggestion
    retry_on: tuple[Exception, ...] = (Exception, ),
```

koku/common/sql_retries.py
```diff
@@ -0,0 +1,72 @@
...
+    attribute_to_check: str = "message",
+    max_wait: int = 30,
+    context_extractor: t.Optional[t.Callable] = None,
```
samdoran, line 25: For now, I would do this directly in the function rather than making it a parameter.
```suggestion
```

koku/common/sql_retries.py
```diff
@@ -0,0 +1,72 @@
...
+    log_ref: str = "Trino query",
+    custom_exception: t.Type[Exception] = None,
+    custom_log_message: t.Optional[str] = None,
```
samdoran, line 28:
```suggestion
    log_ref: str = "Trino query",
    log_message: t.Optional[str] = "Retrying...",
    custom_exception: t.Type[Exception] = None,

```

koku/common/sql_retries.py
```diff
@@ -0,0 +1,72 @@
...
+                sql_params = {}
+            context = context_extractor(sql_params) if context_extractor else {}
+            for attempt in range(retries + 1):
+                try:
+                    return callable(*args, **kwargs)
+                except retry_on as ex:
+                    if attempt < retries and "NoSuchKey" in getattr(ex, attribute_to_check, ""):
+                        LOG.warning(
+                            log_json(
+                                msg=custom_log_message or "TrinoExternalError Exception, retrying...",
```
samdoran, line 44:
```suggestion
                                msg=log_message,
```

koku/common/sql_retries.py
```diff
@@ -0,0 +1,72 @@
...
+    log_ref: str = "Trino query",
+    custom_exception: t.Type[Exception] = None,
+    custom_log_message: t.Optional[str] = None,
+):
+    def _retry(callable: t.Callable):
+        @functools.wraps(callable)
+        def wrapper(*args, **kwargs):
+            sql_params = kwargs.get("sql_params", {})
+            if sql_params is None:
+                sql_params = {}
```
samdoran, line 35:
```suggestion
            sql_params = kwargs.get("sql_params", {})
```

koku/common/sql_retries.py
```diff
@@ -0,0 +1,72 @@
...
+            context = context_extractor(sql_params) if context_extractor else {}
```
samdoran, line 36: This is a bit leaky. Use `context` from the `kwargs` otherwise derive the context from the SQL params. If there are also no SQL params, this will be `{}`.
```suggestion
            context = kwargs.get("context", ReportDBAccessorBase.extract_context_from_sql_params(sql_params))
```

koku/koku/trino_database.py
```diff
@@ -148,12 +150,6 @@ def executescript(trino_conn, sqlscript, *, params=None, preprocessor=None):
...
                 results = cur.fetchall()
-            except TrinoQueryError as trino_exc:
-                trino_statement_error = TrinoStatementExecError(
-                    statement=stmt, statement_number=stmt_num, sql_params=s_params, trino_error=trino_exc
-                )
-                LOG.warning(f"{trino_statement_error!s}")
-                raise trino_statement_error from trino_exc
             except Exception as exc:
                 LOG.warning(str(exc))
                 raise
```
samdoran, line 155: This try except should be removed since it's now being done by the decorator.

koku/common/__init__.py
```diff
@@ -0,0 +1,70 @@
...
+                        continue
+
+                    LOG.error(
+                        log_json(
+                            msg="Failed Trino SQL execution: TrinoExternalError", log_ref=log_ref, context=context
+                        ),
+                        exc_info=ex,
+                    )
+                    if custom_exception:
+                        raise custom_exception(error=ex)
```
samdoran, line 61:
```suggestion
                        raise custom_exception(error=ex) from ex
```

koku/masu/database/report_db_accessor_base.py
bacciotti: The issue here is that we get a circular import. `ReportDBAccessorBase` imports the decorator, and the decorator imports `ReportDBAccessorBase`

koku/common/sql_retries.py
bacciotti: I may be wrong, but we cannot do that because we hit circular imports.

koku/masu/database/report_db_accessor_base.py
```diff
@@ -114,12 +114,19 @@ def _execute_raw_sql_query(self, table, sql, bind_params=None, operation="UPDATE
     def _execute_trino_raw_sql_query(self, sql, *, sql_params=None, context=None, log_ref=None, attempts_left=0):
         """Execute a single trino query returning only the fetchall results"""
         results, _ = self._execute_trino_raw_sql_query_with_description(
-            sql, sql_params=sql_params, context=context, log_ref=log_ref, attempts_left=attempts_left
```
samdoran, line 117: Any place that calls `_execute_trino_raw_sql_query` with `attempts_left` needs to be updated.

koku/common/sql_retries.py
samdoran: You are correct. We need to rearrange some things to prevent a circular import.

koku/common/__init__.py
```diff
@@ -0,0 +1,63 @@
...
+from api.common import log_json
+from common.util import extract_context_from_sql_params
+
+LOG = logging.getLogger(__name__)
+
+
+def retry(
+    original_callable: t.Callable = None,
+    *,
+    retries: int = 3,
```
samdoran, line 16: Most of the places that redundant retry code can be removed `settings.HIVE_PARTITION_DELETE_RETRIES` is used for the number of retries. Setting that as a default here seems sensible.
```suggestion
    retries: int = settings.HIVE_PARTITION_DELETE_RETRIES,
```

koku/common/util.py
samdoran, line 1: It looks like we use `utils.py` with an 's' everywhere else.

koku/masu/test/database/test_aws_report_db_accessor.py
```diff
@@ -350,7 +350,7 @@ def test_delete_aws_hive_partition_by_month(self, mock_trino, mock_table_exist):
                 self.accessor.delete_hive_partition_by_month(table, self.ocp_provider_uuid, "2022", "01")
             mock_trino.assert_called()
             # Confirms that the error log would be logged on last attempt
-            self.assertEqual(mock_trino.call_args_list[-1].kwargs.get("attempts_left"), 0)
+            self.assertEqual(mock_trino.call_args_list[-1].kwargs.get("attempts_left"), None)
```
samdoran, line 353: This check can be removed since it's asserting something about a parameter that was removed. The important thing to check is that it was call the expected number of times.

koku/common/utils.py
```diff
@@ -0,0 +1,82 @@
...
+    return ctx
+
+
+def retry(
+    original_callable: t.Callable = None,
+    *,
+    retries: int = settings.HIVE_PARTITION_DELETE_RETRIES,
+    retry_on: tuple[type[Exception], ...] = (Exception,),
+    max_wait: int = 30,
+    log_ref: str = "Trino query",
```
samdoran, line 39: This didn't end up being used.
```suggestion
```

koku/koku/trino_database.py
```diff
@@ -107,6 +108,7 @@ def connect(**connect_args):
     return trino.dbapi.connect(**trino_connect_args)
 
 
+@retry(retry_on=(TrinoStatementExecError,))
```
samdoran, line 111:
```suggestion
@retry(retry_on=TrinoStatementExecError)
```

koku/masu/database/report_db_accessor_base.py
```diff
@@ -111,15 +99,22 @@ def _execute_raw_sql_query(self, table, sql, bind_params=None, operation="UPDATE
...
-    def _execute_trino_raw_sql_query(self, sql, *, sql_params=None, context=None, log_ref=None, attempts_left=0):
+    def _execute_trino_raw_sql_query(self, sql, *, sql_params=None, context=None, log_ref=None):
         """Execute a single trino query returning only the fetchall results"""
         results, _ = self._execute_trino_raw_sql_query_with_description(
-            sql, sql_params=sql_params, context=context, log_ref=log_ref, attempts_left=attempts_left
+            sql, sql_params=sql_params, context=context, log_ref=log_ref
         )
         return results
 
+    @retry(retry_on=(Exception,))
```
samdoran, line 109:
```suggestion
    @retry(retry_on=TrinoQueryError)
```

koku/masu/database/report_db_accessor_base.py
```diff
@@ -128,27 +123,22 @@ def _execute_trino_raw_sql_query_with_description(
...
         trino_conn = trino_db.connect(schema=self.schema, **conn_params)
         LOG.info(log_json(msg="executing trino sql", log_ref=log_ref, context=ctx))
-        try:
-            trino_cur = trino_conn.cursor()
-            trino_cur.execute(sql, bind_params)
-            results = trino_cur.fetchall()
-            description = trino_cur.description
-        except Exception as ex:
-            if attempts_left == 0:
-                LOG.error(log_json(msg="failed trino sql execution", log_ref=log_ref, context=ctx), exc_info=ex)
```
samdoran, line 150: Looking at all the calling code, omitting this log message loses the value of `log_ref` that we rely on, so I do think we need to catch the exception here and can't rely fully on the decorator for logging.

From looking at the last 90 days of logs, catching `TrinoQueryError` should be sufficient.
```python
except TrinoQueryError as ex:
    LOG.error(log_json(msg="failed trino sql execution", log_ref=log_ref, context=ctx), exc_info=ex)
    raise
```

koku/masu/test/database/test_azure_report_db_accessor.py
```diff
@@ -289,7 +289,7 @@ def test_delete_azure_hive_partition_by_month(self, mock_trino, mock_table_exist
                 self.accessor.delete_hive_partition_by_month(table, self.ocp_provider_uuid, "2022", "01")
             mock_trino.assert_called()
             # Confirms that the error log would be logged on last attempt
-            self.assertEqual(mock_trino.call_args_list[-1].kwargs.get("attempts_left"), 0)
+            self.assertEqual(mock_trino.call_args_list[-1].kwargs.get("attempts_left"), None)
```
samdoran, line 292:
```suggestion
```

koku/masu/test/database/test_gcp_report_db_accessor.py
```diff
@@ -435,7 +435,7 @@ def test_delete_gcp_hive_partition_by_month(self, mock_trino, mock_table_exist):
                 self.accessor.delete_hive_partition_by_month(table, self.ocp_provider_uuid, "2022", "01")
             mock_trino.assert_called()
             # Confirms that the error log would be logged on last attempt
-            self.assertEqual(mock_trino.call_args_list[-1].kwargs.get("attempts_left"), 0)
+            self.assertEqual(mock_trino.call_args_list[-1].kwargs.get("attempts_left"), None)
```
samdoran, line 438:
```suggestion
```

koku/masu/test/database/test_ocp_report_db_accessor.py
```diff
@@ -770,7 +770,7 @@ def test_delete_ocp_hive_partition_by_day(self, mock_trino, mock_table_exist):
                 self.accessor.delete_ocp_hive_partition_by_day([1], self.ocp_provider_uuid, "2022", "01")
             mock_trino.assert_called()
             # Confirms that the error log would be logged on last attempt
-            self.assertEqual(mock_trino.call_args_list[-1].kwargs.get("attempts_left"), 0)
+            self.assertEqual(mock_trino.call_args_list[-1].kwargs.get("attempts_left"), None)
```
samdoran, line 773:
```suggestion
```

koku/common/utils.py
```diff
@@ -0,0 +1,97 @@
+import functools
```
maskarb, line 1: The name of this file could use some improvement. `utils` doesn't tell you anything about what might be in the file.

Also, given that this retry function was written for Trino, and explicitly has `HIVE` in the function, this probably ought to be moved to the `trino_database.py` file.

koku/masu/database/report_db_accessor_base.py
```diff
@@ -111,15 +100,22 @@ def _execute_raw_sql_query(self, table, sql, bind_params=None, operation="UPDATE
...
-    def _execute_trino_raw_sql_query(self, sql, *, sql_params=None, context=None, log_ref=None, attempts_left=0):
+    def _execute_trino_raw_sql_query(self, sql, *, sql_params=None, context=None, log_ref=None):
         """Execute a single trino query returning only the fetchall results"""
         results, _ = self._execute_trino_raw_sql_query_with_description(
-            sql, sql_params=sql_params, context=context, log_ref=log_ref, attempts_left=attempts_left
+            sql, sql_params=sql_params, context=context, log_ref=log_ref
         )
         return results
 
+    @retry(retry_on=TrinoQueryError)
```
maskarb, line 110: Is this exception type specific enough? I don't think it is.

koku/koku/trino_database.py
```diff
@@ -107,6 +108,7 @@ def connect(**connect_args):
     return trino.dbapi.connect(**trino_connect_args)
 
 
+@retry(retry_on=TrinoStatementExecError)
```
maskarb, line 111: Is this exception type specific enough? I don't think it is.

maskarb (commented):
This code does look nice, however, we're trying to write a retry function for:
```
TrinoExternalError
Error opening Hive split ...: The specified key does not exist. (Service: Amazon S3; Status Code: 404; Error Code: NoSuchKey; ...)
```
Does this work retry just this? Or is it retrying whatever is raising `TrinoStatementExecError` and `TrinoQueryError`?

Maybe we need another custom trino exception, similar to [this one](https://github.com/project-koku/koku/blob/main/koku/koku/database.py#L47) which can parse the exception string, looking for the NoSuchKey error, and raise a specific `TrinoNoSuchKeyException` which we can then retry on.

koku/masu/database/report_db_accessor_base.py
samdoran: `TrinoQueryError` is broad and meant to catch most Exceptions from Trino. That can definintely be narrowed.

bacciotti:
(cc @samdoran @maskarb)

Here are the changes I've made:

**Creation of a Custom Exception**
- Added `TrinoNoSuchKeyException` for specific handling of "NoSuchKey" errors.

**Function for Error Detection**
- Implemented a function to detect if the error message contains the "NoSuchKey" string.

**Adaptation of the Retry Decorator**
- The retry logic will be executed only if the exception message contains the "NoSuchKey" string and there are retry attempts left.
- If all retry attempts are exhausted, `TrinoNoSuchKeyException` will be raised.
- If the error does not contain "NoSuchKey", the original exception (`TrinoExternalError`, a generic Trino exception) will be raised.

**Adaptation/creation of Unit Tests**
- Updated unit tests to cover the new changes and ensure proper functionality.

**In Summary:** When a function uses the retry decorator and raises a `TrinoExternalError`, the decorator will be invoked. The retry logic will only be attempted if the error message contains the "NoSuchKey" string. If there are no retry attempts left, TrinoNoSuchKeyException will be raised. If the error does not contain "NoSuchKey", the original exception will be raised.

koku/koku/trino_database.py
```diff
@@ -78,6 +84,130 @@ def query_id(self) -> t.Optional[str]:
...
+                    LOG.debug(f"Attempt {attempt + 1} for {callable.__name__}")
+                    return callable(*args, **kwargs)
+
+                except retry_on as ex:
+                    LOG.debug(f"Exception caught: {ex}")
+
+                    # Verify if the exception contains "NoSuchKey"
+                    # If it does, retry the operation
+                    violation = NoSuchKeyViolation(ex)
+                    if violation.is_no_such_key_violation and attempt < retries:
```
samdoran, line 167: This evaluation shouldn't be done here. If this is called with `retry_on=TrinoNoSuchKeyException`, other exception types won't be retried, which is the end result we're aiming for.

koku/koku/trino_database.py
```diff
@@ -107,6 +237,7 @@ def connect(**connect_args):
     return trino.dbapi.connect(**trino_connect_args)
 
 
+@retry(retry_on=TrinoExternalError)
```
samdoran, line 240:
```suggestion
@retry(retry_on=TrinoNoSuchKeyException)
```

samdoran: In the try except:
```python
            try:
                cur = trino_conn.cursor()
                cur.execute(stmt, params=s_params)
                results = cur.fetchall()
            except TrinoQueryError as trino_exc:
                LOG.warning(f"{trino_statement_error!s}")
                if "NoSuchKey" in str(ex):
                    raise TrinoNoSuchKeyException

                trino_statement_error = TrinoStatementExecError(
                    statement=stmt, statement_number=stmt_num, sql_params=s_params, trino_error=trino_exc
                )
                raise trino_statement_error from trino_exc
```

koku/masu/database/report_db_accessor_base.py
```diff
@@ -111,15 +100,22 @@ def _execute_raw_sql_query(self, table, sql, bind_params=None, operation="UPDATE
...
-    def _execute_trino_raw_sql_query(self, sql, *, sql_params=None, context=None, log_ref=None, attempts_left=0):
+    def _execute_trino_raw_sql_query(self, sql, *, sql_params=None, context=None, log_ref=None):
         """Execute a single trino query returning only the fetchall results"""
         results, _ = self._execute_trino_raw_sql_query_with_description(
-            sql, sql_params=sql_params, context=context, log_ref=log_ref, attempts_left=attempts_left
+            sql, sql_params=sql_params, context=context, log_ref=log_ref
         )
         return results
 
+    @retry(retry_on=TrinoExternalError)
```
samdoran, line 110:
```suggestion
    @retry(retry_on=TrinoNoSuchKeyException)
```

koku/masu/database/report_db_accessor_base.py
```diff
@@ -128,27 +124,26 @@ def _execute_trino_raw_sql_query_with_description(
...
             trino_cur = trino_conn.cursor()
             trino_cur.execute(sql, bind_params)
             results = trino_cur.fetchall()
             description = trino_cur.description
-        except Exception as ex:
-            if attempts_left == 0:
-                LOG.error(log_json(msg="failed trino sql execution", log_ref=log_ref, context=ctx), exc_info=ex)
-            raise ex
+        except TrinoQueryError as ex:
+            LOG.error(log_json(msg="failed trino sql execution", log_ref=log_ref, context=ctx), exc_info=ex)
```
samdoran, line 144:
```suggestion
        except TrinoQueryError as ex:
            LOG.error(log_json(msg="failed trino sql execution", log_ref=log_ref, context=ctx), exc_info=ex)
            if "NoSuchKey" in str(ex):
                raise TrinoNoSuchKeyException

```

koku/koku/trino_database.py
```diff
@@ -78,6 +84,130 @@ def query_id(self) -> t.Optional[str]:
         return self._trino_error.query_id
 
 
+class NoSuchKeyViolation:
```
samdoran, line 87: This class isn't adding much value. It can effectively be replaced with `"NoSuchKey" in str(ex)`.

koku/koku/trino_database.py
```diff
@@ -78,6 +84,130 @@ def query_id(self) -> t.Optional[str]:
...
+                    # If it is NoSuchKey and all attempts are over, raise TrinoNoSuchKeyException
+                    if violation.is_no_such_key_violation:
+                        LOG.error(
+                            log_json(
+                                msg=f"Failed execution after {attempt + 1} attempts",
+                                context=context,
+                                exc_info=ex,
+                            )
+                        )
+                        raise TrinoNoSuchKeyException(str(violation))
```
samdoran, line 191: This should raise the original exception on the last failed attempt instead of a `TrinoNoSuchKeyException`.
```suggestion
```

koku/koku/trino_database.py
```diff
@@ -78,6 +84,130 @@ def query_id(self) -> t.Optional[str]:
...
+                    else:
```
samdoran, line 192: The `else` isn't strictly necessary here.

koku/koku/trino_database.py
```diff
@@ -78,6 +84,130 @@ def query_id(self) -> t.Optional[str]:
...
+        return str(self)
+
+    def __str__(self):
+        if self.__is_no_such_key_violation:
+            return f"NoSuchKey error detected: {self.message}"
+        else:
+            return "No NoSuchKey error detected."
+
+
+class TrinoNoSuchKeyException(Exception):
```
samdoran, line 116: We should probably include some attributes on this execption so that if it does get raised and passed to `log_json` the `str` will show useful information in the logs.

koku/koku/trino_database.py
```diff
@@ -78,6 +84,91 @@ def query_id(self) -> t.Optional[str]:
...
+    """Custom exception for NoSuchKey errors"""
+
+    def __init__(self, message, query_id=None, error_code=None):
+        super().__init__(message)
+        self.query_id = query_id
+        self.error_code = error_code
+
+    def __str__(self):
+        return f"TrinoNoSuchKeyException: {self.args[0]}, Query ID: {self.query_id}, Error Code: {self.error_code}"
```
samdoran, line 96:
```suggestion
        return f"{self.__class__.__name__}: {self.message}, Query ID: {self.query_id}, Error Code: {self.error_code}"
```

koku/koku/trino_database.py
```diff
@@ -78,6 +84,91 @@ def query_id(self) -> t.Optional[str]:
         return self._trino_error.query_id
 
 
+class TrinoNoSuchKeyException(Exception):
+    """Custom exception for NoSuchKey errors"""
+
+    def __init__(self, message, query_id=None, error_code=None):
+        super().__init__(message)
```
samdoran, line 91: Don't worry about using the parent class method. Store the message on a propery for more accurate retrieval instead of having to parse an array of args to get the message.
```suggestion
    def __init__(self, message, query_id=None, error_code=None):
        self.message = message
```

koku/koku/trino_database.py
```diff
@@ -149,10 +241,13 @@ def executescript(trino_conn, sqlscript, *, params=None, preprocessor=None):
...
             except TrinoQueryError as trino_exc:
+                LOG.warning(f"{trino_exc!s}")
+                if "NoSuchKey" in str(trino_exc):
+                    raise TrinoNoSuchKeyException
+
                 trino_statement_error = TrinoStatementExecError(
                     statement=stmt, statement_number=stmt_num, sql_params=s_params, trino_error=trino_exc
                 )
-                LOG.warning(f"{trino_statement_error!s}")
                 raise trino_statement_error from trino_exc
```
samdoran, line 251: This can be reworked to log the `str` of the exception being raised and pass some parameters to the `TrinoNoSuchKeyException` to provide better conetxt in the logs.
```python
                exc_to_raise = TrinoStatementExecError(
                    statement=stmt, statement_number=stmt_num, sql_params=s_params, trino_error=trino_exc
                )
                if "NoSuchKey" in str(trino_exc):
                    exc_to_raise = TrinoNoSuchKeyException(
                        message=trino_exc.message,
                        query_id=trino_exc.query_id,
                        error_code=trino_exc.error_code,
                    )

                LOG.warning(f"{exc_to_raise!s}")
                raise exc_to_raise from trino_exc
```

koku/masu/database/report_db_accessor_base.py
```diff
@@ -133,27 +130,27 @@ def _execute_trino_raw_sql_query_with_description(
...
+            if "NoSuchKey" in str(ex):
+                raise TrinoNoSuchKeyException
+            raise
```
samdoran, line 153:
```suggestion
                raise TrinoNoSuchKeyException(
                    message=ex.message,
                    query_id=ex.query_id,
                    error_code=ex.error_code,
                )
                
            raise
```

maskarb (commented):
Shouldn't we pull the changes in #5237 into this PR?

koku/masu/test/database/test_aws_report_db_accessor.py
```diff
@@ -259,29 +260,36 @@ def test_bad_sql_execution(self):
...
-        )
-        mock_trino.assert_not_called()
+    @retry(retries=settings.HIVE_PARTITION_DELETE_RETRIES)
+    def delete_ocp_on_aws_hive_partition_by_day(self, *args, **kwargs):
+        return self.accessor.delete_ocp_on_aws_hive_partition_by_day(*args, **kwargs)
 
-        mock_schema_exists.return_value = True
-        mock_trino.reset_mock()
+    @patch("masu.database.ocp_report_db_accessor.OCPReportDBAccessor._execute_trino_raw_sql_query_with_description")
+    def test_delete_ocp_on_aws_hive_partition_by_day(self, mock_execute_trino_with_description, mock_table_exist):
```
samdoran, line 268: Here's an example of how to rework these tests.
```python
    @patch("masu.database.aws_report_db_accessor.AWSReportDBAccessor.schema_exists_trino")
    @patch("masu.database.aws_report_db_accessor.AWSReportDBAccessor.table_exists_trino")
    @patch("masu.database.report_db_accessor_base.trino_db.connect")
    @patch("time.sleep", return_value=None)
    def test_delete_ocp_on_aws_hive_partition_by_day(self, mock_sleep, mock_connect, mock_table_exist, mock_schema_exists):
        """Test that deletions work with retries."""
        mock_schema_exists.return_value = False
        self.accessor.delete_ocp_on_aws_hive_partition_by_day(
            [1], self.aws_provider_uuid, self.ocp_provider_uuid, "2022", "01"
        )
        mock_connect.assert_not_called()

        mock_connect.reset_mock()
        
        mock_schema_exists.return_value = True
        error = {"errorName": "HIVE_METASTORE_ERROR"}
        mock_connect.side_effect = TrinoExternalError(error)
        with self.assertRaises(TrinoExternalError):
            self.accessor.delete_ocp_on_aws_hive_partition_by_day(
                [1], self.aws_provider_uuid, self.ocp_provider_uuid, "2022", "01"
            )
        mock_connect.assert_called()
        self.assertEqual(mock_connect.call_count, settings.HIVE_PARTITION_DELETE_RETRIES)
```

samdoran:
In order to retry only on hive metastore errors, we'll need another exception object.

```
class TrnioMetastoreError(Exception):
    ...

@retry(retry_on=(TrinoNoSuchKeyException, TrinoHiveMetastoreError)
    try:
        ...
    except TrinoQueryError as trino_exc:
        exc_to_raise = TrinoStatementExecError(
            statement=stmt, statement_number=stmt_num, sql_params=s_params, trino_error=trino_exc
        )
        if "NoSuchKey" in str(trino_exc):
            exc_to_raise = TrinoNoSuchKeyException(
                message=trino_exc.message,
                query_id=trino_exc.query_id,
                error_code=trino_exc.error_code,
            )

        if trino_exc.error_name == "HIVE_METASTORE_ERROR":
            exc_to_raise = TrinoHiveMetastoreError(
                message=trino_exc.message,
                query_id=trino_exc.query_id,
                error_code=trino_exc.error_code,
            )

        LOG.warning(f"{exc_to_raise!s}")
        raise exc_to_raise from trino_exc
```

koku/masu/database/report_db_accessor_base.py
```diff
@@ -150,10 +146,23 @@ def _execute_trino_raw_sql_query_with_description(
...
+        except TrinoQueryError as ex:
+            LOG.error(log_json(msg="failed trino sql execution", log_ref=log_ref, context=ctx), exc_info=ex)
+            if "NoSuchKey" in str(ex):
+                raise TrinoNoSuchKeyException(
+                    message=ex.message,
+                    query_id=ex.query_id,
+                    error_code=ex.error_code,
+                )
+
+            if "HIVE_METASTORE_ERROR" in str(ex):
```
samdoran, line 158:
```suggestion
            if ex.errorname == "HIVE_METASTORE_ERROR":
```

koku/masu/test/database/test_aws_report_db_accessor.py
```diff
@@ -259,29 +261,39 @@ def test_bad_sql_execution(self):
             with self.assertRaises(ProgrammingError):
                 accessor._execute_processing_script("masu.database", script_file_path, {})
 
+    @retry(retries=settings.HIVE_PARTITION_DELETE_RETRIES)
+    def delete_ocp_on_aws_hive_partition_by_day(self, *args, **kwargs):
+        return self.accessor.delete_ocp_on_aws_hive_partition_by_day(*args, **kwargs)
+
```
samdoran, line 267: I'm not sure what the purpose of this method is. It's not a test and I don't see it being called the the test code.

koku/masu/test/database/test_aws_report_db_accessor.py
```diff
@@ -259,29 +261,39 @@ def test_bad_sql_execution(self):
...
 
         mock_schema_exists.return_value = True
-        mock_trino.reset_mock()
-        error = {"errorName": "HIVE_METASTORE_ERROR"}
-        mock_trino.side_effect = TrinoExternalError(error)
-        with self.assertRaises(TrinoExternalError):
+
+        # Setting up the mock so that the cursor execute causes a TrinoExternalError
+        mock_cursor = mock_connect.return_value.cursor.return_value
+        mock_cursor.execute.side_effect = TrinoExternalError({"errorName": "HIVE_METASTORE_ERROR"})
```
samdoran, line 287: I don't think this is quite right. I'm actually not sure what this is doing.
Just have `"masu.database.report_db_accessor_base.trino_db.connect` raise an exception. That's enough to verify the retry behavior.
```suggestion
        mock_connect.side_effect = TrinoExternalError({"errorName": "HIVE_METASTORE_ERROR"})
```

koku/masu/test/database/test_azure_report_db_accessor.py
```diff
@@ -237,29 +239,39 @@ def test_get_openshift_on_cloud_matched_tags_trino(self, mock_trino):
...
+    @patch("masu.database.report_db_accessor_base.trino_db.connect")
+    @patch("time.sleep", return_value=None)
+    def test_delete_ocp_on_azure_hive_partition_by_day(
+        self, mock_sleep, mock_connect, mock_table_exist, mock_schema_exists
+    ):
         """Test that deletions work with retries."""
         mock_schema_exists.return_value = False
         self.accessor.delete_ocp_on_azure_hive_partition_by_day(
-            [1], self.azure_provider_uuid, self.ocp_provider_uuid, "2022", "01"
+            [1], self.aws_provider_uuid, self.ocp_provider_uuid, "2022", "01"
```
samdoran, line 256:
```suggestion
            [1], self.azure_provider_uuid, self.ocp_provider_uuid, "2022", "01"
```

koku/masu/test/database/test_azure_report_db_accessor.py
```diff
@@ -237,29 +239,39 @@ def test_get_openshift_on_cloud_matched_tags_trino(self, mock_trino):
...
 
         mock_schema_exists.return_value = True
-        mock_trino.reset_mock()
-        error = {"errorName": "HIVE_METASTORE_ERROR"}
-        mock_trino.side_effect = TrinoExternalError(error)
-        with self.assertRaises(TrinoExternalError):
+
+        # Setting up the mock so that the cursor execute causes a TrinoExternalError
+        mock_cursor = mock_connect.return_value.cursor.return_value
+        mock_cursor.execute.side_effect = TrinoExternalError({"errorName": "HIVE_METASTORE_ERROR"})
```
samdoran, line 265:
```suggestion
        mock_connect.side_effect = TrinoExternalError({"errorName": "HIVE_METASTORE_ERROR"})
```

koku/masu/test/database/test_azure_report_db_accessor.py
```diff
@@ -237,29 +239,39 @@ def test_get_openshift_on_cloud_matched_tags_trino(self, mock_trino):
...
+
+        # Asserting that the function raises TrinoHiveMetastoreError error
+        with self.assertRaises(TrinoHiveMetastoreError):
             self.accessor.delete_ocp_on_azure_hive_partition_by_day(
-                [1], self.azure_provider_uuid, self.ocp_provider_uuid, "2022", "01"
+                [1], self.aws_provider_uuid, self.ocp_provider_uuid, "2022", "01"
```
samdoran, line 270:
```suggestion
                [1], self.azure_provider_uuid, self.ocp_provider_uuid, "2022", "01"
```

koku/masu/test/database/test_gcp_report_db_accessor.py
```diff
@@ -397,29 +399,39 @@ def test_populate_ocp_on_gcp_tag_information_assert_call(self, mock_trino):
...
+    @patch("masu.database.report_db_accessor_base.trino_db.connect")
+    @patch("time.sleep", return_value=None)
+    def test_delete_ocp_on_gcp_hive_partition_by_day(
+        self, mock_sleep, mock_connect, mock_table_exist, mock_schema_exists
+    ):
         """Test that deletions work with retries."""
         mock_schema_exists.return_value = False
         self.accessor.delete_ocp_on_gcp_hive_partition_by_day(
-            [1], self.gcp_provider_uuid, self.ocp_provider_uuid, "2022", "01"
+            [1], self.aws_provider_uuid, self.ocp_provider_uuid, "2022", "01"
```
samdoran, line 416:
```suggestion
            [1], self.gcp_provider_uuid, self.ocp_provider_uuid, "2022", "01"
```

koku/masu/test/database/test_gcp_report_db_accessor.py
```diff
@@ -397,29 +399,39 @@ def test_populate_ocp_on_gcp_tag_information_assert_call(self, mock_trino):
...
+
+        # Setting up the mock so that the cursor execute causes a TrinoExternalError
+        mock_cursor = mock_connect.return_value.cursor.return_value
+        mock_cursor.execute.side_effect = TrinoExternalError({"errorName": "HIVE_METASTORE_ERROR"})
+
+        # Asserting that the function raises TrinoHiveMetastoreError error
+        with self.assertRaises(TrinoHiveMetastoreError):
             self.accessor.delete_ocp_on_gcp_hive_partition_by_day(
-                [1], self.gcp_provider_uuid, self.ocp_provider_uuid, "2022", "01"
+                [1], self.aws_provider_uuid, self.ocp_provider_uuid, "2022", "01"
```
samdoran, line 430:
```suggestion
                [1], self.gcp_provider_uuid, self.ocp_provider_uuid, "2022", "01"
```

koku/masu/test/database/test_ocp_report_db_accessor.py
```diff
@@ -759,27 +760,29 @@ def test_delete_infrastructure_raw_cost_from_daily_summary(self):
...
-        mock_table_exist.reset_mock()
-        with patch("masu.database.ocp_report_db_accessor.OCPReportDBAccessor.schema_exists_trino", return_value=False):
-            self.accessor.delete_ocp_hive_partition_by_day([1], self.ocp_provider_uuid, "2022", "01")
-            mock_trino.assert_not_called()
-            mock_table_exist.assert_not_called()
+        mock_schema_exists.return_value = True
+
+        # Setting up the mock so that the cursor execute causes a TrinoExternalError
+        mock_cursor = mock_connect.return_value.cursor.return_value
+        mock_cursor.execute.side_effect = TrinoExternalError({"errorName": "HIVE_METASTORE_ERROR"})
```
samdoran, line 778:
```suggestion
        mock_cursor = mock_connect.return_value.cursor.return_value
        mock_cursor.side_effect = TrinoExternalError({"errorName": "HIVE_METASTORE_ERROR"})
```

koku/masu/test/database/test_ocp_report_db_accessor.py
```diff
@@ -819,26 +822,29 @@ def test_find_expired_trino_partitions_no_table(self, mock_trino, mock_table_exi
...
-            )
-            self.assertFalse(result)
-        mock_trino.assert_called()
-        # Confirms that the error log would be logged on last attempt
-        self.assertEqual(mock_trino.call_args_list[-1].kwargs.get("attempts_left"), 0)
-        self.assertEqual(mock_trino.call_count, settings.HIVE_PARTITION_DELETE_RETRIES)
+
+        # Setting up the mock so that the cursor execute causes a TrinoExternalError
+        mock_cursor = mock_trino.return_value.cursor.return_value
+        mock_cursor.execute.side_effect = TrinoExternalError({"errorName": "HIVE_METASTORE_ERROR"})
```
samdoran, line 840:
```suggestion
        mock_trino.side_effect = TrinoExternalError({"errorName": "HIVE_METASTORE_ERROR"})
```

samdoran (approved):

PR #5218, opened 2024-07-17
[COST-4620] - Add basic provider data validation

lcouzens:
## Jira Ticket

[COST-4620](https://issues.redhat.com/browse/COST-4620)

## Description

- Add a new post summary task to validate some total metrics between postgres and trino data. 
- Logs if data looks valid or not and marks the provider with a true/false accordingly (Giving us a quick way to build a list of incorrect provider data)
- Additionally adds a trigger for OCP on Cloud clusters too.
- New Masu endpoint to manually trigger this validation for a given date range.
- Also includes an unleash flag so we can skip running this task by default and gradually enable customers. Just to be aware of any performance hits.

## Testing

1. Checkout Branch
2. Restart Koku
3. Enable the validation unleash flag `is_validation_enabled`
4. Ingest data like normal.
5. You should see logging as follows per provider ingested
```
{'message': 'validation started for provider: a67bb4a2-8e05-40
{'message': 'triggering VALIDATION_QUERY',
{'message': 'finished VALIDATION_QUERY',
{'message': 'executing trino sql', 'tracing_id': '', 'log_ref': 'data validation query'}
{'message': 'executed trino sql', 'tracing_id': '', 'log_ref': 'data validation query', 'running_time': 0.5209639072418213}
{'message': 'all data complete for provider: a67bb4a2-8e05-4078-b1e4-9c8525f3fb0f'
```
6. You can also re-trigger the validation task via this masu endpoint: 
            http://localhost:5042/api/cost-management/v1/validate_cost_data/?provider_uuid={uuid}&start_date=2024-07-01&end_date=2024-07-16
7. You should see similar to the above in the logs

## Testing invalid data
1. After you have ingest some data lets mess it up in postgres!
2. Heres a command for AWS:
            `update org1234567.reporting_awscostentrylineitem_daily_summary set unblended_cost = 0 where product_code = 'AmazonEC2';`
3. Run validation via masu again
4. see the following log messages
```
'message': 'validation started for provider: a67bb4a2-8e05-4078-b1e4-9c8525f3fb0f'
'message': 'triggering VALIDATION_QUERY'
'message': 'finished VALIDATION_QUERY'
'message': 'executing trino sql'
'message': 'executed trino sql'
'message': "provider has incomplete data for specified days: {datetime.date(2024, 7, 1): {'pg_value': 3789.255140293, 'trino_value': 6347.878609799311, 'delta': 2558.623469506311},
 ```

## Testing OCP on Cloud clusters
1. Ingest OCP data
2. Ingest correlated cloud data
3. Check logs for matching validation for the correlated clusters similar to those above.

## Testing via masu
1. Ingest data
2. Hit the following endpoint `http://localhost:5042/api/cost-management/v1/validate_cost_data/?provider_uuid={provider_uuid}&start_date=2024-07-01&end_date=2024-07-19` (Sub your provider ID)
3. See validation log messages.

## Testing an OCP on CLOUD cluster via masu
1. Ingest ocp on cloud data
2. Hit the following endpoint `http://localhost:5042/api/cost-management/v1/validate_cost_data/?provider_uuid={provider_uuid}&ocp_on_cloud_type={type}&start_date=2024-07-01&end_date=2024-07-19` (Sub your provider ID)
3. See validation log messages similar to above

## Notes:

Along side the testing about if you want to verify the query data you can enable `DEBUG` logs and you should see the comparison data from Trino/PG logged for quick comparison. 
Example: [2024-07-22 12:45:54,311] INFO fc9a26e0-a60a-4264-8aac-afebb4f082e5 3268 PG: {datetime.date(2024, 7, 18): 3870.034663868, datetime.date(2024, 7, 19): 3670.98312068} Trino data: {datetime.date(2024, 7, 18): 3870.034663795944, datetime.date(2024, 7, 19): 3670.9831205979876}

## Release Notes
- [ ] proposed release note

```markdown
* [COST-4620](https://issues.redhat.com/browse/COST-4620) Add internal data validation
```


koku/masu/processor/_tasks/data_validation.py
```diff
@@ -0,0 +1,222 @@
...
+            LOG.warning(log_json(msg=f"data validation trino query failed: {e}", context=self.context))
+            return
+        # Compare results
+        LOG.debug(f"PG: {pg_data} Trino data: {trino_data}")
+        daily_difference, valid_cost = self.compare_data(pg_data, trino_data)
+        if valid_cost:
+            LOG.info(log_json(msg=f"all data complete for provider: {self.provider_uuid}", context=self.context))
+            provider.data_valid = True
+        else:
+            LOG.warning(
```
lcouzens, line 215: We could raise here or log an error instead if we want.

koku/api/provider/models.py
```diff
@@ -189,6 +189,8 @@ class Meta:
     customer = models.ForeignKey("Customer", null=True, on_delete=models.PROTECT)
     created_by = models.ForeignKey("User", null=True, on_delete=models.SET_NULL)
     setup_complete = models.BooleanField(default=False)
+    # Used for data validation checks
+    data_valid = models.BooleanField(null=True, default=None)
```
lcouzens, line 193: Along side this we _could_ store a json to store similar to what is logged during a failure. List of invalid days and their difference or something. It could aid investigations but it needs thought around multiple months, regarding if we just append to this json or replace it for example.

koku/masu/processor/_tasks/data_validation.py
```diff
@@ -0,0 +1,222 @@
...
+        # year and month for running partitioned queries
+        dh = DateHelper()
+        year = dh.bill_year_from_date(self.start_date)
+        month = dh.bill_month_from_date(self.start_date)
+        report_db_accessor = ReportDBAccessorBase(self.schema)
+        # Set provider filter, when running ocp{aws/gcp/azure} checks we need to rely on the cluster id
+        provider_filter = self.provider_uuid if not self.ocp_on_cloud_type else cluster_id
+        # query trino/postgres
+        table, query_filters = self.get_table_filters_for_provider(provider_type, trino)
+        for start, end in date_range_pair(self.start_date, self.end_date, step=self.date_step):
```
lcouzens, line 150: We currently use start and end dates based on the reports running. Their is a question here whether we should always run full month OR +x number of days since it _could_ be something outside the currently processing range that could be incorrect. Maybe this is ok for the first iteration though.

koku/masu/processor/_tasks/data_validation.py
```diff
@@ -0,0 +1,222 @@
...
+    },
+    Provider.PROVIDER_OCI: {"date": "usage_start", "metric": "cost"},
+    Provider.PROVIDER_OCP: {
+        "date": "usage_start",
+        "metric": "pod_effective_usage_cpu_core_hours",
+    },
+    "OCPAWS": {"date": "usage_start", "metric": "unblended_cost"},
+    "OCPAzure": {"date": "usage_start", "metric": "pretax_cost"},
+    "OCPGCP": {"date": "usage_start", "metric": "unblended_cost"},
+}
```
lcouzens, line 59: All bar one of the above metrics are based around standard cost metrics for the providers. The only difference is the ocp one, since this doesn't have cost out of the box, it's set to effective cpu initially.
Additionally we could validate multiple metrics here but initial thoughts are if one metric is incorrect then all probably have a diff. Future iterations we could include blended/amortized and credits etc.

koku/masu/processor/_tasks/data_validation.py
cgoodfred: IMO we should raise an error if the data is not valid, it will bring more awareness to times when these don't line up and it should be an actionable event for someone to determine why the costs aren't the same.

koku/api/provider/models.py
cgoodfred: Probs worth talking about all of these questions in a design discussion since it feels unclear what the outcome we are looking for is.

koku/masu/processor/_tasks/data_validation.py
lcouzens, line 215:
```suggestion
            LOG.error(
```

koku/masu/processor/_tasks/data_validation.py
lcouzens: Decided to add a range of -5 days to the validation checks, but not to cross the month boundary. This is also configurable via settings.

koku/api/provider/models.py
lcouzens: Dropping this for the first iteration, we may want this when/if we add retry logic.

koku/api/utils.py
```diff
@@ -252,6 +268,28 @@ def n_days_ago(self, in_date, n_days):
         n_days = midnight - datetime.timedelta(days=n_days)
         return n_days
 
+    def n_days_ago_or_month_start(self, in_date, n_days):
```
djnakabaale, line 271: Question: 

I'm curious why we need this new `n_days_ago_or_month_start` method that could not be addressed with the existing methods `n_days_ago` and  `month_start` or `month_start_utc`?

koku/masu/database/report_db_accessor_base.py
```diff
@@ -97,19 +97,24 @@ def _execute_raw_sql_query(self, table, sql, bind_params=None, operation="UPDATE
...
+                    result = cursor.fetchall()
+
             except OperationalError as exc:
                 db_exc = get_extended_exception_by_type(exc)
                 LOG.warning(log_json(os.getpid(), msg=str(db_exc), context=db_exc.as_dict()))
                 raise db_exc from exc
 
         running_time = time.time() - t1
         LOG.info(log_json(msg=f"finished {operation}", row_count=row_count, table=table, running_time=running_time))
+        return result
```
djnakabaale, line 117: Nitpick: Update the doc string to reflect that this method now has a return.

djnakabaale (commented):
Here some feedback on a couple things noticed:

koku/masu/database/report_db_accessor_base.py
```diff
@@ -97,19 +97,24 @@ def _execute_raw_sql_query(self, table, sql, bind_params=None, operation="UPDATE
         """Run a SQL statement via a cursor."""
```
lcouzens, line 97:
```suggestion
        """Run a SQL statement via a cursor. This also returns a result if the operation is VALIDATION_QUERY."""
```

koku/api/utils.py
lcouzens: So this is using the n_days_ago under the hood, and we could rely on both of them outside of this, we would just need _some_ logic inside the validation class to make the correct lookup. Happy to change it though.

djnakabaale (approved):

PR #5235, opened 2024-07-25
[COST-5319] Export EC2 compute response data to CSV

djnakabaale:
## Jira Ticket

[COST-5319](https://issues.redhat.com/browse/COST-5319)

## Description

This change will enable exporting EC2 compute response data to CSV.

## Testing

1. Checkout Branch
2. Restart Koku
3. Load test data for AWS
    ```
    make create-test-customer
    make load-test-customer-data test_source=aws
    ```
4. Launch shell and hit the EC2 endpoint with CSV output format
    
    ```
    curl --location --globoff 'http://127.0.0.1:8000/api/cost-management/v1/reports/aws/resources/ec2-compute/?filter[time_scope_value]=-2' -H 'accept: text/csv' -o output.csv
    ```
    

    - You should see EC2 response data in the `output.csv` file,  similar to this:
    
       [output.csv](https://github.com/user-attachments/files/16384378/output.csv)

## Release Notes
- [x] proposed release note

```markdown
* [COST-5319](https://issues.redhat.com/browse/COST-5319) Export EC2 compute response data to CSV
```

### DIFF
--- a/koku/api/common/pagination.py
+++ b/koku/api/common/pagination.py
@@ -323,7 +323,7 @@ class AWSEC2ComputePagination(ReportPagination):
         paginated_data = []
 
         data = (
-            queryset_data[0] if (queryset_data := queryset.get("data", [])) else []
+            queryset_data[0] if (queryset_data := queryset.get("data", [])) else {}
         )  # only single month data expected
         resource_ids = data.get("resource_ids", [])
         resource_count = len(resource_ids)
--- a/koku/api/common/pagination.py
+++ b/koku/api/common/pagination.py
@@ -327,6 +327,7 @@ class AWSEC2ComputePagination(ReportPagination):
 
         if self.offset < resource_count:
             # paginate resource IDs from current_offset to the limit
+            # limit=0 param is a special case that returns all data
             paginated_ids = resource_ids if self.limit == 0 else resource_ids[self.offset : self.offset + self.limit]
 
             if self.request.accepted_media_type and "text/csv" in self.request.accepted_media_type:
--- a/koku/api/common/pagination.py
+++ b/koku/api/common/pagination.py
@@ -331,7 +331,7 @@ class AWSEC2ComputePagination(ReportPagination):
                 paginated_data = paginated_ids
             else:
                 paginated_item = data.copy()
-                paginated_item["resource_ids"] = resource_ids[self.offset : self.offset + self.limit]
+                paginated_item["resource_ids"] = paginated_ids
                 paginated_data.append(paginated_item)
 
         return paginated_data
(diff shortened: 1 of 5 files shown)


dlabrecq (approved):
Curious about the pagination support. The UI doesn't paginate over CSV data, so when we perform an export, we call the API with `limit=0`. I assume that's supported here?

djnakabaale:
> Curious about the pagination support. The UI doesn't paginate over CSV data, so when we perform an export, we call the API with `limit=0` -- there's no `offset`. I assume that's supported here?

Yes, that would be handled here. There are default values for offset and limit incase none is provided.

koku/api/common/pagination.py
```diff
@@ -306,20 +306,32 @@ def get_count(self, queryset):
...
+        paginated_data (list): If the request expects CSV data, returns a slice of resource IDs.
+                                Otherwise, list containing a single dictionary with paginated resource IDs.
+        """
 
         paginated_data = []
 
-        for item in queryset.get("data", []):
-            resource_ids = item.get("resource_ids", [])
-            resource_count = len(resource_ids)
+        data = queryset.get("data", [])[0]  # only single month data expected
```
myersCody, line 322: If `queryset.get("data", [])` returns an empty list accessing 0 will raise an IndexError.

koku/api/common/pagination.py
```diff
@@ -306,20 +306,32 @@ def get_count(self, queryset):
         return len(queryset.get("data", [])[0].get("resource_ids", []))
```
myersCody, line 306: Reminder

koku/api/common/pagination.py
djnakabaale: Yes thanks for catching that, updated it

koku/api/common/pagination.py
djnakabaale: Updated to check for list before indexing

myersCody (dismissed):

koku/api/common/pagination.py
```diff
@@ -303,23 +303,40 @@ class AWSEC2ComputePagination(ReportPagination):
...
-            if self.offset < resource_count:
-                paginated_item = item.copy()
+        paginated_data = []
 
-                # paginate resource IDs from current_offset to the limit
-                paginated_item["resource_ids"] = resource_ids[self.offset : self.offset + self.limit]
+        data = (
+            queryset_data[0] if (queryset_data := queryset.get("data", [])) else []
+        )  # only single month data expected
+        resource_ids = data.get("resource_ids", [])
```
maskarb, line 328: Is `data` here a list if there is no queryset_data?

maskarb: This syntax is also quite difficult to read.
Maybe we should do:
```
data = {}
if queryset_data := queryset.get("data"):
    data = queryset_data[0]
```

maskarb (approved):

PR #5316, opened 2024-09-20
Silence a noisy alert

myersCody:
## Jira Ticket

- None

## Description

This change will reduce noise related to verification SQL running after a partition has been dropped

## Release Notes
- [ ] proposed release note

```markdown
```


koku/masu/processor/_tasks/data_validation.py
```diff
@@ -217,6 +218,15 @@ def check_data_integrity(self):
         # Trino query to get daily values
         try:
             trino_data = self.execute_relevant_query(provider_type, cluster_id, True)
+        except TrinoExternalError as te:
+            # https://github.com/trinodb/trino-python-client/blob/master/trino/client.py
+            # The trino external error contains a response
+            if "Partition no longer exists" in str(te._error.get("message", "")):
```
samdoran, line 224: You can use the [property](https://github.com/trinodb/trino-python-client/blob/a87566794d9a9eefdd481a95f001ce2e37e20531/trino/exceptions.py#L103-L105) which is already doing the work of getting the message for you..
```suggestion
            if "Partition no longer exists" in te.message:
```

lcouzens (approved):

PR #5321, opened 2024-09-24
[COST-5449] - Delete old EC2 table/model

lcouzens:
## Jira Ticket

[COST-5449](https://issues.redhat.com/browse/COST-5449)

## Description

This change will remove the old non-partitioned EC2 compute table.

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Release Notes
- [ ] proposed release note

```markdown
* [COST-5449](https://issues.redhat.com/browse/COST-5449) Remove old non-partitioned EC2 table
```


samdoran (approved):
The `cost_entry_bill` is set to to CASCADE on delete. I don't think that will remove any data we care about, but it's worth considering before we go ahead with the deletion.

PR #5323, opened 2024-09-26
[COST-5453] Adding gzip support for Azure reports

bacciotti:
## Jira Ticket

[COST-5453](https://issues.redhat.com/browse/COST-5453)

## Description

This change will add support for `gzip` (.csv.gz) files on Azure reports.

## Testing

1. Checkout Branch
2. Restart Koku
3. Create a gzip report on Azure storage (dev account)
4. Trigger the download/ingestion flow
5. You should be able to see the ingested data on API/database

## Release Notes
- [ ] Added gzip support for Azure reports

```markdown
* [COST-5453](https://issues.redhat.com/browse/COST-5453) Add gzip support for Azure reports
```


bacciotti:
> So this PR confuses me a little in places. Are we adding GZIP support for Azure or did we replace CSV support with GZIP? I might be completely miss reading here but some changes look like we are just changing from CSV to GZIP.

@lcouzens you are right, it was missing some adjustments to allow both formats. Tested locally with csv and gzip files now.

koku/masu/external/downloader/azure/azure_service.py
```diff
@@ -158,7 +163,19 @@ def get_file_for_key(self, key: str, container_name: str) -> BlobProperties:
         return report
 
     def get_latest_cost_export_for_path(self, report_path: str, container_name: str) -> BlobProperties:
-        return self._get_latest_blob_for_path(report_path, container_name, AzureBlobExtension.csv.value)
+        """
+        Get the latest cost export for a given path and container. Supports both CSV and GZIP formats.
+        """
+        blob = self._get_latest_blob_for_path(report_path, container_name, AzureBlobExtension.gzip.value)
```
lcouzens, line 169: Lets not be guessing formats here. We should be looking this up and dynamically passing it everywhere its needed really. Use `describe_cost_management_exports` to lookup what the compression type is.

Inside this method we build the report definitions:
```report_def = {
                        "name": report.name,
                        "container": report.delivery_info.destination.container,
                        "directory": report.delivery_info.destination.root_folder_path,
                    }
```

We should just add compression to that list. Then we can set `self.compression = compression` inside `AzureReportDownloader`

koku/masu/external/downloader/azure/azure_report_downloader.py
```diff
@@ -409,6 +412,65 @@ def manifest_date_format(self):
...
+
+        LOG.debug("%s is %s bytes; Download path has %s free", key, file_size, free)
+
+        ext = os.path.splitext(key)[1]
+        if ext == ".gz" and check_inflate and size_ok and file_size > 0:
+            try:
+                range_start = file_size - 4
+                range_end = file_size
+
+                resp = self._azure_client.download_file(
```
lcouzens, line 457: Having a uncompressed check similar to AWS is ok I'm not sure its required though, since its actually pandas that handles the file in memory. It eventually writes to a csv file but during daily archiving, so we reduce the file down to required columns and by date.
That said if we do want to keep this I would rework it so we don't download the file twice, in and outside of this method. Lets download the file, then pass that file to the compression check logic.

koku/masu/external/downloader/azure/azure_report_downloader.py
```diff
@@ -449,11 +510,18 @@ def download_file(self, key, stored_etag=None, manifest_id=None, start_date=None
 
         local_filename = utils.get_local_file_name(key)
         full_file_path = f"{self._get_exports_data_directory()}/{local_filename}"
+
+        if not self._check_size(key, check_inflate=True):
```
lcouzens, line 514: As mentioned above we should move this after the initial download.

koku/masu/external/downloader/azure/azure_service.py
```diff
@@ -80,9 +81,9 @@ def _get_latest_blob_for_path(
         self,
         report_path: str,
         container_name: str,
-        extension: str,
+        extensions: list[str] = [AzureBlobExtension.gzip.value, AzureBlobExtension.csv.value],
```
lcouzens, line 84: We should only pass in the relevant compression type for the report being processed.

koku/masu/external/downloader/azure/azure_service.py
```diff
@@ -117,10 +118,14 @@ def _get_latest_blob_for_path(
             LOG.warning(error_msg)
             raise AzureCostReportNotFound(message)
 
-        latest_report = self._get_latest_blob(report_path, blobs, extension)
+        for extension in extensions:
```
lcouzens, line 121: Lets not change this. We should pass in the correct compression type.

koku/masu/external/downloader/azure/azure_service.py
```diff
@@ -117,10 +118,14 @@ def _get_latest_blob_for_path(
...
+            latest_report = self._get_latest_blob(report_path, blobs, extension)
+            if latest_report:
+                break
+
         if not latest_report:
             message = (
-                f"No file with extension '{extension}' found in container "
+                f"No file with any of the extensions {extensions} found in container "
```
lcouzens, line 128: Same as above

koku/masu/external/downloader/azure/azure_service.py
```diff
@@ -170,24 +187,39 @@ def download_file(
...
-        """Download the file from a given storage container."""
+        """
+        Download the file from a given storage container. Supports both CSV and GZIP formats.
+        """
 
         if not ingress_reports:
             cost_export = self.get_file_for_key(key, container_name)
             key = cost_export.name
 
+            if key.endswith(AzureBlobExtension.gzip.value):
```
lcouzens, line 201: The compression type should be passed in here.

koku/masu/external/downloader/azure/azure_service.py
```diff
@@ -170,24 +187,39 @@ def download_file(
...
+
         file_path = destination
         if not destination:
             temp_file = NamedTemporaryFile(delete=False, suffix=suffix)
             temp_file.close()
             file_path = temp_file.name
 
         try:
-            blob_client = self._cloud_storage_account.get_blob_client(container_name, key)
+            blob_client = self._cloud_storage_account.get_blob_client(container=container_name, blob=key)
```
lcouzens, line 213: Not sure why we changed this.

koku/masu/external/downloader/azure/azure_service.py
```diff
@@ -170,24 +187,39 @@ def download_file(
...
             with open(file_path, "wb") as blob_download:
-                blob_download.write(blob_client.download_blob().readall())
-        except (AdalError, AzureException, ClientException, OSError) as error:
+                if offset is not None and length is not None:
```
lcouzens, line 215: I'm not sure I understand this part or why its needed? The initial download should not be a problem or need breaking up.

lcouzens (changes requested):
I left a few comments to think about, happy to chat through it too, just let me know.

koku/masu/external/downloader/azure/azure_report_downloader.py
```diff
@@ -409,6 +412,65 @@ def manifest_date_format(self):
         """Set the Azure manifest date format."""
         return "%Y%m%d"
 
+    def _check_size(self, key, check_inflate=False):
```
maskarb, line 415: This method does not apply, even for AWS. Our pods used to be backed by PVCs which had finite sizes. This is not the case any more. We don't have any reason to believe that we cannot uncompress these files.

lcouzens (changes requested):
Some thing to think about :smiley_cat:

koku/masu/external/downloader/azure/azure_service.py
```diff
@@ -157,8 +159,37 @@ def get_file_for_key(self, key: str, container_name: str) -> BlobProperties:
...
+            compression (str): The compression format ('gzip' or 'csv').
+
+        Returns:
+            BlobProperties: The latest blob corresponding to the specified report path and container.
+
+        Raises:
+            ValueError: If the compression type is not 'gzip' or 'csv'.
+            AzureCostReportNotFound: If no blob is found for the given path and container.
+        """
+        valid_compressions = [AzureBlobExtension.gzip.value, UNCOMPRESSED]
```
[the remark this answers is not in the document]
bacciotti: I don't think so. GZIP_COMPRESSED means "GZIP" and we are handling ".csv.gz" here.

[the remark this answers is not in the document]
lcouzens: Then UNCOMPRESSED here is wrong too no? Since that is `PLAIN` not csv.
```
# Compression formats
UNCOMPRESSED = "PLAIN"
GZIP_COMPRESSED = "GZIP"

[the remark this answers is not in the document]
bacciotti: I think this UNCOMPRESSED here refers to the manifest file. I've changed to `valid_compressions = [AzureBlobExtension.gzip.value, AzureBlobExtension.csv.value]` to match both expected reports formats.

lcouzens (approved):
Fantastic work on this Lucas!

PR #5333, opened 2024-10-02
[COST-5517] Applying security best practices in GH actions

bacciotti:
## Jira Ticket

[COST-5517](https://issues.redhat.com/browse/COST-5517)

## Description

This change will add top-level and job-level permissions, add Harden Runner [1] and add full-length SHA [2] to align with Security Best Practices [3][4][5].

[1] [Harden Runner](https://github.com/step-security/harden-runner)
[2] [Full-length SHA](https://github.com/step-security/secure-repo?tab=readme-ov-file#3-pin-actions-to-a-full-length-commit-sha)
[3] [Token Permissions](https://github.com/ossf/scorecard/blob/ea7e27ed41b76ab879c862fa0ca4cc9c61764ee4/docs/checks.md#token-permissions)
[4] [Workflow Syntax](https://docs.github.com/pt/actions/writing-workflows/workflow-syntax-for-github-actions#permissions)
[5] [CWE Reference](https://cwe.mitre.org/data/definitions/732.html)

## Testing
1. Check if actions can run smoothly.

## Release Notes
- [ ] Applying security best practices in GH actions

```markdown
* [COST-5517](https://issues.redhat.com/browse/COST-5517) Applying security best practices in GH actions
```


.github/workflows/pre-release.yaml
```diff
@@ -7,43 +7,54 @@ on:
...
+permissions:
+  contents: read
+
 jobs:
   pre-release:
     name: Pre-release
     runs-on: ubuntu-20.04
 
+    permissions:
+      contents: write
```
maskarb, line 19: Do we need both of these?

.github/workflows/check-image.yml
```diff
@@ -4,12 +4,21 @@ on:
     - cron: "5 0 * * *"
   workflow_dispatch:
 
+permissions:
```
maskarb, line 7: @samdoran can correct me if I am wrong, but now that konflux is configured, I think we can delete this workflow entirely.

maskarb (commented):
Some comments:

.github/workflows/pre-release.yaml
bacciotti: I'd say yes... but just because it would match the proposed standards, but I don't think it could have effective difference in this file tbh.

maskarb: I think what I'm confused about is that there is a single job in this workflow. Reading the docs: `You can use permissions either as a top-level key, to apply to all jobs in the workflow, or within specific jobs`, so we're setting it to read at the top-level, but then we change that same permission at the job level to something less restrictive.

.github/workflows/ci.yml
```diff
@@ -33,10 +42,18 @@ jobs:
   smokes-labeler:
     name: Smoke Test Label
     runs-on: ubuntu-20.04
+    permissions:
+      contents: read
+      issues: write
```
maskarb, line 47: These permissions match the top-level, so these are just redundant. Let's remove them here.

.github/workflows/check-image.yml
samdoran: My understanding is the purpose of this workflow is to prompt an image rebuild when the packages in the base container have changed. Moving to Konflux doesn't really change the need to rebuild the image when the base image changes. There's nothing special in Konflux to rebuild when the base image changes.

This project is high enough tempo that I think this workflow is unnecessary and can be removed.

For other projects that are lower tempo, this workflow is more useful. For those lower tempo projects, we could rely on the Konflux pipeline update PRs from Renovate bot as an indirect means to accomplish the same thing.

.github/workflows/pre-release.yaml
samdoran: There's only one job in this workflow. Set it to `read` at the workflow level and remove this one.

.github/workflows/pre-release.yaml
```diff
@@ -7,43 +7,54 @@ on:
...
+              ;;
+          esac
+          echo "TAG_VERSION=r.$DATE.$MINOR_VERSION" >> $GITHUB_ENV
 
       - name: Set the release commit
         run: echo "RELEASE_COMMIT=${{ github.event.inputs.commit }}" >> $GITHUB_ENV
 
       - name: Create release body file
-        run: bash .github/scripts/get_description.sh
+        run: bash .github/scripts/get_description.sh || echo "Failed to generate release body"
```
samdoran, line 54: This change means the task will now succeed on failure since the return code from the `echo` command will be `0`. Is that what we want?

.github/workflows/check-image.yml
bacciotti: ok!

samdoran (approved):

PR #5336, opened 2024-10-07
[COST-5492] Handle special characters in tag queries

bacciotti:
## Jira Ticket

[COST-5492](https://issues.redhat.com/browse/COST-5492)

## Description

This change will handle special characters in tag queries, for instance white spaces and semicolons.

## Testing

1. Checkout Branch
2. Restart Koku
3. Ingest some tags with special characters, like `app test` or `app;test`
4. Hit endpoint `?group_by[tag:app test]=*` or `?group_by[tag:app;test]=*&order_by[tag:app;test]=asc`
5. You should see a code 200 response with proper data.

## Release Notes
- [ ] Handle special characters in tag queries

```markdown
* [COST-5492](https://issues.redhat.com/browse/COST-5492) Handle special characters in tag queries
```


koku/api/report/queries.py
```diff
@@ -669,11 +669,13 @@ def _get_tag_group_by(self):
         """Create list of tag based group by parameters."""
         group_by = []
         tag_groups = self.get_tag_group_by_keys()
+        sanitize_pattern = re.compile(r'[\s"\'`;]+')
```
samdoran, line 672: For situations like this, a translation table is simpler and more performant, plus it doesn't bring the maintenance overhead of regular expressions.
```suggestion
        characters_to_sanitize = r' "\'`;'
        table = str.maketrans(characters_to_sanitize, "_" * len(characters_to_sanitize))
        ...
        sanitized_tag = strip_prefix(tag, TAG_PREFIX).translate(table)
```

samdoran:
There are three places where the same string is being sanitized. Two are still using the regex method and one is using a translation table. The sanitizing functionality could be put into a function for consistency instead of doing the same thing in three different places.

koku/api/report/queries.py
```diff
@@ -666,14 +676,13 @@ def _get_group_by(self):
...
-        """Create list of tag based group by parameters."""
+        """Create list of tag-based group by parameters."""
         group_by = []
         tag_groups = self.get_tag_group_by_keys()
         for tag in tag_groups:
-            tag_db_name = self._mapper.tag_column + "__" + strip_prefix(tag, TAG_PREFIX)
-            tag = str.encode(tag)
-            tag = quote_from_bytes(tag, safe=URL_ENCODED_SAFE)
-            group_pos = self.parameters.url_data.index(tag)
+            sanitized_tag = sanitize_tag(tag)
```
samdoran, line 683: I think you need to strip the prefix.
```suggestion
            sanitized_tag = sanitize_tag(strip_prefix(tag, TAG_PREFIX))
```

koku/api/report/queries.py
```diff
@@ -95,6 +95,16 @@ def check_view_filter_and_group_by_criteria(filter_set, group_by_set):
     return True
 
 
+def sanitize_tag(tag):
+    """Sanitize a tag by removing unwanted characters and URL-encoding it."""
+    characters_to_sanitize = r' "\'`;'
+    table = str.maketrans(characters_to_sanitize, "_" * len(characters_to_sanitize))
+    sanitized_tag = strip_prefix(tag, TAG_PREFIX).translate(table)
+    encoded_tag = str.encode(tag)
```
samdoran, line 103: I'm not sure why we are converting to bytes here, but it should be operating on `sanitized_tag`
```suggestion
    encoded_tag = str.encode(sanitized_tag)
```

koku/api/report/queries.py
```diff
@@ -95,6 +95,16 @@ def check_view_filter_and_group_by_criteria(filter_set, group_by_set):
     return True
 
 
+def sanitize_tag(tag):
+    """Sanitize a tag by removing unwanted characters and URL-encoding it."""
+    characters_to_sanitize = r' "\'`;'
+    table = str.maketrans(characters_to_sanitize, "_" * len(characters_to_sanitize))
+    sanitized_tag = strip_prefix(tag, TAG_PREFIX).translate(table)
```
samdoran, line 102: Since `strip_prefix()` can be used to strip an arbitrary prefix, that should be done outside of this function. The caller should be responsible for calling that function separately from this function.
```suggestion
    sanitized_tag = tag.translate(table)
```

PR #5354, opened 2024-10-31
[COST-5650] Support cross account RBAC access

chambridge:
## Jira Ticket

[COST-5650](https://issues.redhat.com/browse/COST-5650)

## Description

* update user cache to be user and org_id dependent
* update user queries to use customer lookup

## Testing

1. Checkout Branch
2. Restart Koku
3. Hit endpoint with a user with one org_id
4. Hit endpoint with a user with a different org_id
5. Validate permissions are different for each user

## Release Notes
- [ ] Support cross account RBAC access

```markdown
* [COST-5650](https://issues.redhat.com/browse/COST-5650) Support cross account RBAC access
```

### DIFF
--- a/koku/api/provider/provider_builder.py
+++ b/koku/api/provider/provider_builder.py
@@ -102,10 +102,7 @@ class ProviderBuilder:
                 customer = Customer.objects.filter(org_id=self.org_id).get()
             except Customer.DoesNotExist:
                 customer = IdentityHeaderMiddleware.create_customer(self.account_number, self.org_id, "POST")
-            try:
-                user = User.objects.get(username=username)
-            except User.DoesNotExist:
-                user = IdentityHeaderMiddleware.create_user(username, email, customer, None)
+            user = User(username=username, email=email, customer=customer)
 
         context = {"user": user, "customer": customer}
         return context, customer, user
--- a/koku/koku/middleware.py
+++ b/koku/koku/middleware.py
@@ -28,7 +28,6 @@ from django_prometheus.middleware import PrometheusAfterMiddleware
 from django_prometheus.middleware import PrometheusBeforeMiddleware
 from django_tenants.middleware import TenantMainMiddleware
 from prometheus_client import Counter
-from rest_framework.exceptions import ValidationError
 
 from api.common import RH_IDENTITY_HEADER
 from api.common.pagination import EmptyResultsSetPagination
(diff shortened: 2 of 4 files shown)


maskarb:
This solution will not work because there is a uniqueness constraint on the username in the User table (so we can't create another User with the same username and different customer_id). I question the value of the User table. What do we actually gain by keeping it vs just using the information in the request to form a user object?

Of course, we could just drop the uniqueness constraint, but that seems like another can of worms to try and understand.

maskarb:
This solution will not work because there is a uniqueness constraint on the username in the User table (so we can't create another User with the same username and different customer_id). I question the value of the User table. What do we actually gain by keeping it vs just using the information in the request to form a user object?

Of course, we could just drop the uniqueness constraint, but that seems like another can of worms to try and understand.

koku/api/provider/provider_builder.py
```diff
@@ -103,7 +103,7 @@ def _create_context(self):
             except Customer.DoesNotExist:
                 customer = IdentityHeaderMiddleware.create_customer(self.account_number, self.org_id, "POST")
             try:
-                user = User.objects.get(username=username)
+                user = User.objects.get(username=username, customer=customer)
```
maskarb, line 106: Looking at where this `user` object is used leads me to believe we can just eliminate it from this method. It's only used to remove a provider when we've received the destroy.provider event from the sources backend. We end up in this logic here:
https://github.com/project-koku/koku/blob/5d8b4b4674c0784f9d503a64f8e3f9eec18c68fd/koku/api/provider/provider_manager.py#L366-L368
which takes us here:
https://github.com/project-koku/koku/blob/5d8b4b4674c0784f9d503a64f8e3f9eec18c68fd/koku/api/provider/provider_manager.py#L228-L230

which as you can see, the only thing we actually care about is the customer (not the actual user).

This `_create_context` method already returns the customer, so I think returning the user here is totally redundant and unnecessary. I think we ought to update the `manager.remove` method to rely on the customer from this method or the customer in the request object.

koku/koku/middleware.py
```diff
@@ -371,13 +356,14 @@ def process_request(self, request):  # noqa: C901
...
                 return HttpResponseFailedDependency({"source": "Database", "exception": err})
 
+            user_key = f"{org_id}_{username}"
             try:
-                if username not in USER_CACHE:
-                    user = User.objects.get(username=username)
-                    USER_CACHE[username] = user
-                    LOG.debug(f"User added to cache: {username}")
+                if user_key not in USER_CACHE:
+                    user = IdentityHeaderMiddleware.create_user(username, email, customer, request)
```
maskarb, line 362: If we eliminate the `IdentityHeaderMiddleware.create_user` in the `_create_context` method in the `ProviderBuilder`, then I think we can just get rid of the `create_user` method entirely and just replace all of this with:
```suggestion
                    user = User(username=username, email=email, customer=customer)
```

maskarb (commented):
I think we can delete even more old stuff:

maskarb (approved):
lgtm!

PR #5377, opened 2024-11-12
Fix ocp vm csv export formatting

lcouzens:
## Jira Ticket

[COST-5627](https://issues.redhat.com/browse/COST-5627)

## Description

This change will fix the response formatting for OCP VM csv exports

## Testing

1. Checkout Branch
2. Restart Koku
3. load ocp vm data
4. try to export csv data from api `curl --location -g 'http://localhost:8000/api/cost-management/v1/reports/openshift/resources/virtual-machines/' -H 'accept: text/csv' -o output.csv`
5. Should actually have data in the exported csv

## Release Notes
- [ ] proposed release note

```markdown
* [COST-5627](https://issues.redhat.com/browse/COST-5627) Fix OCP VM Export CSV
```


koku/api/report/ocp/query_handler.py
```diff
@@ -291,6 +291,11 @@ def execute_query(self):  # noqa: C901
         }
         ordered_total.update(query_sum)
 
+        if self.is_csv_output:
+            if self._report_type in ("virtual_machines"):
+                # Handle formating OCP VM response
+                data = self.format_vm_csv_response(query_data)
```
maskarb, line 297: What is the intention with this:
```
if self._report_type in ("virtual_machines")
```
?

We can also consolidate this logic:
```suggestion
        if self.is_csv_output and self._report_type == "virtual_machines":
            # Handle formating OCP VM response
            data = self.format_vm_csv_response(query_data)
```

koku/api/report/ocp/query_handler.py
```diff
@@ -291,6 +291,11 @@ def execute_query(self):  # noqa: C901
         }
         ordered_total.update(query_sum)
 
+        if self.is_csv_output:
```
myersCody, line 294: We already do this check up on line 274, is there a good reason not to include your if statement below there?

koku/api/report/ocp/query_handler.py
```diff
@@ -291,6 +291,11 @@ def execute_query(self):  # noqa: C901
...
+            if self._report_type in ("virtual_machines"):
```
myersCody, line 295:
```suggestion
            if self._report_type == "virtual_machines":
```

myersCody: whoops I didn't refresh my browser, I now see Michael's comment.

koku/api/report/ocp/query_handler.py
lcouzens: So originally I had this in a common place for consolidate EC2 and OCP VM's but decided not to do that now. That did mean I left ugly style logic when I moved it, good catch. Though with Cody's point to we should probably just move this up a little.

koku/api/report/ocp/query_handler.py
```diff
@@ -273,6 +273,9 @@ def execute_query(self):  # noqa: C901
 
             if self.is_csv_output:
                 data = list(query_data)
+                if self._report_type == "virtual_machines":
+                    # Handle formating OCP VM response
+                    data = self.format_vm_csv_response(query_data)
```
maskarb, line 278: It would be nice to skip the `list` conversion if we don't need to do that:
```suggestion
                if self._report_type == "virtual_machines":
                    # Handle formating OCP VM response
                    data = self.format_vm_csv_response(query_data)
                else:
                    data = list(query_data)
```

koku/api/report/ocp/query_handler.py
```diff
@@ -361,3 +364,19 @@ def add_current_month_deltas(self, query_data, query_sum):
         self.query_delta = {"value": total_delta, "percent": total_delta_percent}
 
         return query_data
+
+    def format_vm_csv_response(self, query_data):
```
myersCody, line 368: Nit Pick:
I think comment you left on 277 provides enough context to describe what is happening in these four lines of code. Therefore, I wouldn't add a new method here since this method will ever only be called once. We could just include these 4 lines in the correct place above instead of jumping around. Makes traceability easier in the future.

myersCody (approved):

PR #5381, opened 2024-11-13
[COST-5657] subs: use subquery in Trino to join from Postgres

maskarb:
## Jira Ticket

[COST-5657](https://issues.redhat.com/browse/COST-5657)

## Description

This change will filter the list of resource-ids thru a subquery in postgres.

## Testing

1. smokes

## Release Notes
- [x] proposed release note

```markdown
* [COST-5657](https://issues.redhat.com/browse/COST-5657) subs: use subquery in Trino to join from Postgres
```

### DIFF
--- a/koku/subs/subs_data_extractor.py
+++ b/koku/subs/subs_data_extractor.py
@@ -120,6 +120,7 @@ class SUBSDataExtractor(ReportDBAccessorBase):
             excluded_ids = list(
                 SubsLastProcessed.objects.exclude(source_uuid=self.provider_uuid).values_list("resource_id", flat=True)
             )
+        # the resource-id must be selected first in the determine_resource_ids_for_usage_account.sql SELECT statement
         sql_file = f"trino_sql/{self.provider_type.lower()}/determine_resource_ids_for_usage_account.sql"
         sql = pkgutil.get_data("subs", sql_file)
         sql = sql.decode("utf-8")
@@ -128,7 +129,6 @@ class SUBSDataExtractor(ReportDBAccessorBase):
             "source_uuid": self.provider_uuid,
             "year": year,
             "month": month,
-            # "excluded_ids": excluded_ids,
             "usage_account": usage_account,
         }
         LOG.info(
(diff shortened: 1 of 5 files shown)


chambridge (approved):

koku/subs/subs_data_extractor.py
```diff
@@ -116,36 +116,41 @@ def determine_row_count(self, sql_params):
...
+        # the resource-id must be selected first in the determine_resource_ids_for_usage_account.sql SELECT statement
+        sql_file = f"trino_sql/{self.provider_type.lower()}/determine_resource_ids_for_usage_account.sql"
+        sql = pkgutil.get_data("subs", sql_file)
+        sql = sql.decode("utf-8")
+        sql_params = {
+            "schema": self.schema,
+            "source_uuid": self.provider_uuid,
+            "year": year,
+            "month": month,
+            "excluded_ids": {},
```
samdoran, line 135: Should this be an empty set instead of a dict?

myersCody:
We likely have the same problem with the `determine_ids_for_provider` function. There is likely a cleaner way to handle the exclusion in Trino. Right now we use the django ORM to build the exclude list and then populate the exclude filter using that list in jinja creating our huge query problem.

Instead we could just build the exclude resource id list in trino, then left join in the excluded list and a filter  to make sure all of our excluded resources are not included in the return:
```WITH cte_excludes_ids as (
    SELECT distinct resource_id 
    FROM postgres.{{schema | sqlsage}}.reporting_subs_last_processed_time as lp
    WHERE lp.source_uuid != {{source_uuid}}
)
SELECT lineitem_resourceid, max(lineitem_usagestartdate)
FROM hive.{{schema | sqlsafe}}.aws_line_items
LEFT JOIN excluded_ids AS ex ON aws.lineitem_resource = ex.resource_id
WHERE source={{source_uuid}}
AND year={{year}}
AND month={{month}}
AND lineitem_productcode = 'AmazonEC2'
AND strpos(lower(resourcetags), 'com_redhat_rhel') > 0
AND ex.resource_id IS NULL
GROUP BY lineitem_resourceid```
This will allow for the excludes resource ids to be however long it needs to be without changing the length of the query itself.

myersCody (approved):

PR #5384, opened 2024-11-14
Update clamav-scan task image

bacciotti:
Update the image used by the clamav task. This may fix the timeout issue we were seeing.

## Release Notes
- [ ] Update clamav image

```markdown
* Update clamav image
```


samdoran:
The current image tag is newer than the one is this PR. We can stick with what we have since it contains the fix we need.

PR #5393, opened 2024-11-21
[COST-5728] switch OCP on AWS endpoints to use amortized costs

lcouzens:
## Jira Ticket

[COST-5728](https://issues.redhat.com/browse/COST-5728)

## Description

This change will switch to amortised cost for the OCP back populate and OCP on ALL tables instead of a coalesce on unblended and savings plans. This is because when we insert into the trino temp tables we normalise all the savings plan and usage costs into a single line. The problem here is we then lose the additional unblended cost from the total value.

 lineitem_usagestartdate | coalecse_cost | savingplan_cost | unblended_cost | blended_cost | calculated_amortized_cost |  lineitem_lineitemtype  
-------------------------+---------------+-----------------+----------------+--------------+---------------------------+-------------------------
 2024-11-21 00:00:00.000 |          20.0 |            20.0 |            0.0 |          0.0 |                      20.0 | SavingsPlanCoveredUsage 
 2024-11-21 00:00:00.000 |          10.0 |             0.0 |           10.0 |         10.0 |                      10.0 | Usage                   
(2 rows)

Query 20241121_225310_00541_pvfbe, FINISHED, 1 node
Splits: 1 total, 1 done (100.00%)
0.25 [2 rows, 28.2KiB] [8 rows/s, 114KiB/s]

trino:org1234567> select usage_amount, unblended_cost, savingsplan_effective_cost, calculated_amortized_cost from aws_openshift_daily_resource_matched_temp;
 usage_amount | unblended_cost | savingsplan_effective_cost | calculated_amortized_cost 
--------------+----------------+----------------------------+---------------------------
         11.0 |           10.0 |                       20.0 |                      30.0 
(1 row)

Essentially if we did the coalesce here like `sum(coalesce(nullif(savingsplan_effective_cost, 0), unblended_cost)) as unblended_cost` this would only work if these lineitemtypes stayed intact. But since we condense them before getting to this sql the coalesce only collects the savings plan cost meaning we DONT account for the additional $10 unblended cost which we should. Switching to amortised here works because in the same trino sql we correctly account for both unblended and savings plan costs before those lines are condensed into a single row.

This is found when you have usage above your savings plan covered usage for the same resource id in a particular day.

## Testing

1. Checkout Branch
2. Restart Koku
3. Run regression testing. (Eva also has a reproducing smoke test branch for this too)

## Release Notes
- [ ] proposed release note

```markdown
* [COST-5728](https://issues.redhat.com/browse/COST-5728) Fix savings plan cost using amortised.
```


esebesto:
I tested this PR and the fix looks good to me!

myersCody (approved):

PR #5394, opened 2024-11-22
[COST-5627] Fix Azure default extension issue

bacciotti:
## Jira Ticket

[COST-5627](https://issues.redhat.com/browse/COST-5627)

## Description

This change fixes the issue of a hardcoded file extension (csv) when no manifest is present. Now, the program dynamically identifies and selects the latest file with supported extensions, such as `.csv` or `.gz`.

## Testing

1. Checkout Branch
2. Restart Koku
3. Ingest `csv` and `csv.gz` data with and without a `mainfest.json` file
4. Run some Azure unit tests, like `tox -e py311 -- masu.test.external.downloader.azure`

## Release Notes
- [ ] Fix Azure default extension issue

```markdown
* [COST-5627](https://issues.redhat.com/browse/COST-5627) Fix Azure default extension issue
```

### DIFF
--- a/koku/masu/external/downloader/azure/azure_service.py
+++ b/koku/masu/external/downloader/azure/azure_service.py
@@ -66,14 +66,21 @@ class AzureService:
     def _get_latest_blob(
         self, report_path: str, blobs: list[BlobProperties], extension: t.Optional[str] = None
     ) -> t.Optional[BlobProperties]:
-        default_extensions = [AzureBlobExtension.csv.value, AzureBlobExtension.gzip.value]
+        valid_extensions = [
+            AzureBlobExtension.csv.value,
+            AzureBlobExtension.gzip.value,
+            AzureBlobExtension.manifest.value,
+        ]
+        if extension and extension not in valid_extensions:
+            raise ValueError(f"Invalid compression type: {extension}")
+
         latest_blob = None
         for blob in blobs:
             if extension:
                 if not blob.name.endswith(extension):
                     continue
             else:
-                if not any(blob.name.endswith(ext) for ext in default_extensions):
+                if not any(blob.name.endswith(ext) for ext in valid_extensions):
                     continue
 
             if report_path in blob.name:
@@ -85,7 +92,7 @@ class AzureService:
         self,
         report_path: str,
         container_name: str,
-        extension: str,
+        extension: t.Optional[str] = None,
     ) -> BlobProperties:
         """Get the latest file with the specified extension from given storage account container."""
 
@@ -160,7 +167,7 @@ class AzureService:
         return report
 
     def get_latest_cost_export_for_path(
-        self, report_path: str, container_name: str, compression: str
+        self, report_path: str, container_name: str, compression: t.Optional[str] = None
     ) -> BlobProperties:
         """
         Get the latest cost export for a given path and container based on the compression type.
@@ -177,10 +184,6 @@ class AzureService:
             ValueError: If the compression type is not 'gzip' or 'csv'.
             AzureCostReportNotFound: If no blob is found for the given path and container.
         """
-        valid_compressions = [AzureBlobExtension.gzip.value, AzureBlobExtension.csv.value]
-        if compression not in valid_compressions:
-            raise ValueError(f"Invalid compression type: {compression}. Expected one of: {valid_compressions}.")
-
         return self._get_latest_blob_for_path(report_path, container_name, compression)
 
     def get_latest_manifest_for_path(self, report_path: str, container_name: str) -> BlobProperties:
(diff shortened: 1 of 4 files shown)


koku/masu/external/downloader/azure/azure_service.py
```diff
@@ -64,24 +64,35 @@ def __init__(
...
         latest_blob = None
         for blob in blobs:
-            if not blob.name.endswith(extension):
-                continue
+            if extension:
+                if not blob.name.endswith(extension):
+                    continue
+            else:
+                if not any(blob.name.endswith(ext) for ext in valid_extensions):
+                    continue
```
maskarb, line 84:
```suggestion
            elif not any(blob.name.endswith(ext) for ext in valid_extensions):
                continue
```

koku/masu/external/downloader/azure/azure_service.py
```diff
@@ -64,24 +64,31 @@ def __init__(
             raise AzureServiceError("Azure Service credentials are not configured.")
 
     def _get_latest_blob(
-        self, report_path: str, blobs: list[BlobProperties], extension: str
+        self, report_path: str, blobs: list[BlobProperties], extension: t.Optional[str] = None
```
lcouzens, line 67: Should we even be passing the extensions around these methods? Michael already mentioned that it was only used in the unit tests anyway. That said I just double checked and we are using here:

`return self._get_latest_blob_for_path(report_path, container_name, AzureBlobExtension.manifest.value)`

BUT I'm not even convinced that does what we think it should! That extension is completely incorrect and we never find a manifest.

In [1]: from masu.util.azure.common import AzureBlobExtension

In [2]: AzureBlobExtension.manifest.value
Out[2]: '_manifest.json'

Now that I'm thinking about it we shouldnt even be passing extensions/compression types around at all if we actually downloaded the manifest correctly then it could just be read from that.

So it turns out that Azure export types suck! -->
- Partitioned exports have `_manifest.json`
- Gzip compressed exports have `manifest.json`
- Standard non-compressed exports don't have a manifest at all.

lcouzens (changes requested):
How did you test with a manifest? Because I'm fairly sure even if we have one it will just run as if we don't. The code here probably works but its also confusing in places (the azure downloader in general at this point) because of how we treat manifests. We may want to clean up all the extension logic and correctly download all manifest types to reduce the complexity of code overall. Happy to chat through all this if its not clear.

koku/masu/external/downloader/azure/azure_service.py
bacciotti: I changed this extension to `AzureBlobExtension.json.value`. In my tests it has downloaded reports on gz, csv, with and without the manifest file.

We have some duplicated logic here. Sometimes (if the manifest file does not exist) it downloads a file twice (in `_get_manifest` and in `download_file` methods). It is needed because we have to know the report name and other details before download it (it is a bit confusing specially when there is no manifest file). FMPOV I think it is working but we can rethink this logic if we want.

koku/masu/external/downloader/azure/azure_service.py
```diff
@@ -64,24 +64,31 @@ def __init__(
...
+            AzureBlobExtension.gzip.value,
+            AzureBlobExtension.json.value,
+        ]
         latest_blob = None
         for blob in blobs:
-            if not blob.name.endswith(extension):
+            if extension:
+                if not blob.name.endswith(extension):
+                    continue
+            elif not any(blob.name.endswith(ext) for ext in valid_extensions):
```
lcouzens, line 79: Should we just drop this logic and the valid extensions list above? Essentially if we are not trying to download a manifest this `elif` statement is going to mangle file types together anyway so its kinda redundant no?

koku/masu/external/downloader/azure/azure_service.py
```diff
@@ -120,10 +127,7 @@ def _get_latest_blob_for_path(
 
         latest_report = self._get_latest_blob(report_path, blobs, extension)
         if not latest_report:
```
lcouzens, line 129: Optional side note: Should we change these `latest_report` vars to `latest_file` or similar throughout this function now? Since its not just csv reports that its collecting (Manifests and gzip files)

koku/masu/external/downloader/azure/azure_service.py
bacciotti: Dropping this logic broke the tests, which indicates that this filtering is needed. Dropping this logic, we can lose control over the extensions used, I think. Maybe we can keep it to maintain the consistency and align with the existing tests...

lcouzens (approved):
Great job Lucas!

PR #5395, opened 2024-11-22
[COST-5711] Provide endpoint of all regions for integrations team

chambridge:
## Jira Ticket

[COST-5711](https://issues.redhat.com/browse/COST-5711)

## Description
* Provide an endpoint that doesn't require specific permissions and provides all aws regions

## Testing

1. Checkout Branch
2. Restart Koku
3. Hit endpoint `api/cost-management/v1/resource-types/aws-all-regions/`
    1. You should see a list of available regions
4. Do more things use offset & limit parameters
5. See error when providing any other parameter

## Release Notes
- [ ] Provide AWS region list to drive Integrations wizard when selecting s3 bucket regions

```markdown
* [COST-5711](https://issues.redhat.com/browse/COST-5711) Provide AWS region list to drive Integrations wizard when selecting s3 bucket regions
```

PR #5399, opened 2024-11-26
Revert "[COST-5711] Provide endpoint of all regions for integrations team"

lcouzens:
Reverts project-koku/koku#5395

After further investigation this change is no longer needed.


chambridge (approved):
lgtm

PR #5412, opened 2024-11-30
[COST-5227] - Add auto large provider logic and XL polling

lcouzens:
## Jira Ticket

[COST-5227](https://issues.redhat.com/browse/COST-5227)

## Description

Add a new boolean field to providers in order to indicate whether that provider should be classed as XL (large) or not. This includes changes to the manifest creation to determine and set whether the provider is large based on report count. The default is larger than 75 files though this is configurable.

Additionally adding a new report downloader schedule specifically for XL (large) customer polling. Currently these providers get collected with our standard schedule but that also means its possible that all XL providers could be scheduled at the same time, which can causes overloaded queues and alerts. This PR now schedules XL (large providers) on there own cadence and batch size giving us more control over load. 

Updated masu download endpoint with param to trigger large providers.

## Testing

1. Checkout Branch
2. Restart Koku
3. Load data 
4. See everything process like normal.
5. Either set `XL_REPORT_COUNT` in your env to 0 (or something lower than your next provider generates) or just hard code a value in place of `if num_of_files > settings.XL_REPORT_COUNT:`
6. Create a new provider with enough report files to class it as large.
7. Check the logs for `Provider set as large for XL processing.`
8. Update the `POLLING_TIMER` like report count so you can trigger downloads (set it to 1 second or something) `polling_delta = datetime.now(tz=settings.UTC) - timedelta(seconds=1) #settings.POLLING_TIMER)`
9. hit download endpoint with `?large_providers=True`
10. See in the logs that only the large provider is collected.
11. Do the same without the large flag
12. See that all providers are collected apart from the large provider.

Notes:
If you want to test multiple large providers to test polling batch size you can run the following sql to set other providers to large, just sub the provider uuid with your cloud provider `update api_provider set large = True where uuid = '4f991071-31d6-4c43-b4e3-4f6049e5c390';`

Both the XL batch size and the schedule cadence is customisable. 

## Release Notes
- [ ] proposed release note

```markdown
* [COST-5227](https://issues.redhat.com/browse/COST-5227) Add automatic large providers and XL specific polling
```


koku/koku/celery.py
```diff
@@ -127,6 +127,18 @@ def validate_cron_expression(expresssion, default="0 * * * *"):
...
+    xl_download_fallback = validate_cron_expression("0 */3 * * *")
+    xl_download_expression = ENVIRONMENT.get_value("XL_REPORT_DOWNLOAD_SCHEDULE", default=xl_download_fallback)
+    XL_REPORT_DOWNLOAD_SCHEDULE = validate_cron_expression(xl_download_expression, xl_download_fallback)
+    xl_report_schedule = crontab(*XL_REPORT_DOWNLOAD_SCHEDULE.split(" ", 5))
+    CHECK_REPORT_UPDATES_XL = {
+        "task": download_task,
+        "schedule": xl_report_schedule,
+        "kwargs": {"large_providers": "True"},
+    }
+    app.conf.beat_schedule["check-report-updates-batched"] = CHECK_REPORT_UPDATES_XL
```
maskarb, line 140: I think we are going to get very confused trying to tweak this schedule, the batch size, and the report count. 

Should we just run this hourly instead?

I also feel like this runs the risk of polling 10 large providers at a time which might defeat the purpose of what we are trying to do.

koku/masu/external/downloader/report_downloader_base.py
```diff
@@ -118,4 +119,13 @@ def _process_manifest_db_record(self, assembly_id, billing_start, num_of_files,
         with ReportManifestDBAccessor() as manifest_accessor:
             if num_of_files != manifest_entry.num_total_files:
                 manifest_accessor.update_number_of_files_for_manifest(manifest_entry)
+
+        # Set whether a provider should be classed as large based on report file count
+        # TODO do something similar for OCP, maybe based on cluster count (not necessarily here)
+        if num_of_files > settings.XL_REPORT_COUNT:
```
maskarb, line 125: Could we add the `is_large_customer` check here too?
```suggestion
        if num_of_files > settings.XL_REPORT_COUNT or is_large_customer(schema):
```

If so....

Getting the schema here is quite a mess. I would update the `ReportDownloader` to require all args to be keywords:
```
class ReportDownloader:
    def __init__(
        self,
        *, <<<<< add this
        customer_name,
```
Doing this would give us `customer_name` in the `kwargs` of `ReportDownloaderBase`, so we could get the schema like:
```
schema = kwargs.get("customer_name")
```
(yes, the customer-name is the schema)

deploy/clowdapp.yaml
```diff
@@ -5895,6 +5989,13 @@ parameters:
 - displayName: Provider count for batch polling
   name: POLLING_BATCH_SIZE
   value: "100"
+- displayName: XL Provider count for batch polling
+  name: XL_POLLING_BATCH_SIZE
+  value: "10"
+- description: Minimum number of reports before provider classed as XL
+  displayName: Min report count for large providers
+  name: XL_REPORT_COUNT
+  value: "75"
```
myersCody, line 5998: Just curious how we determined `75` to be the large customer cut off size here.

koku/masu/external/downloader/report_downloader_base.py
lcouzens: Yeh I thought about this, the problem is this is really just handling extremely large AWS providers right now. Doing this would result in all providers that are part of the XL unleash flag getting marked as large but not all of them actually are. For example the PNT account has 2 large providers but because we do unleash at the schema level all its AWS providers are classed as large when really they are small.

Your right that getting schema here is a fun time though.

koku/masu/processor/orchestrator.py
```diff
@@ -123,26 +122,25 @@ def __init__(
         self.worker_cache = WorkerCache()
         self.bill_date = bill_date
         self.provider_uuid = provider_uuid
-        self.provider_type = provider_type
         self.queue_name = queue_name
         self.ingress_reports = kwargs.get("ingress_reports")
         self.ingress_report_uuid = kwargs.get("ingress_report_uuid")
         self._summarize_reports = kwargs.get("summarize_reports", True)
+        self._large_providers = kwargs.get("large_providers", False)
+        self._polling_batch = settings.XL_POLLING_BATCH_SIZE if self._large_providers else settings.POLLING_BATCH_SIZE
```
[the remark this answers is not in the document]
lcouzens: Yeh this is fair, I was trying not to creep to much into this PR and basing everything on our current count. Not ideal I know, certainly not opposed to doing something more dynamic if we can. Let me think on it and put something together.

Of course the other thing we could do is have an XL_POLLING_TIMER thats 2 days. The XL data is so large in some cases maybe polling ever 48 hours instead is something to also consider.

koku/masu/test/processor/test_orchestrator.py
```diff
@@ -479,20 +479,21 @@ def test_orchestrator_args_polling_batch(self, *args):
         p = o.get_polling_batch()
         self.assertEqual(len(p), 1)
 
-        # provider provider-type does NOT override polling timestamp
+        # provider large does NOT override polling timestamp
         # so this query will provide zero pollable providers
-        o = Orchestrator(provider_type="AWS-local")
+        o = Orchestrator(large_providers="True")
```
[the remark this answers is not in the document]
lcouzens: I have a bad habit of using ""!! Good catch, I probably need to update a few places.

koku/koku/celery.py
lcouzens: Confusion on configuring the schedule is certainly a good point. I was taking a count of the largest providers and trying to split those specifically through the day. That count today is 8 so thats why I landed at 3 by default. This could be configured to be more frequent though. 

As for triggering it hourly, we can but the idea is to spread the load through the day as much as possible because this large providers take between 1-5 hours to process. So with 3 hours hopefully gives us a little more time to process before the next provider starts.

10 large providers risk is exactly what we have today with the current polling, their is certainly a chance that all the large providers could be collected within our 25 batch size. This new strategy would be controlled by the the XL_POLLING_BATCH_SIZE. The default being 1 right now.

myersCody (commented):
Overall, I don't know if I agree with have a boolean flag on the provider as the mechanism to decide if a download should be considered large. My issue with it is that the flag in the DB could be easily become outdated.

*Example*
We release these changes as is, and we discover we have 800 providers now marked as large because they had a single manifest at one point above the `75` threshold. Therefore, we decided the threshold we set was too low and we decided to bump the threshold to `100`. We now have a boolean flag that is outdated for 800 providers with no means to reset it so that it can be rechecked with the current changes.

The metric we are using to determine if a download has exceeded our threshold is the number of files within the manifest. Therefore, personally I think it makes more sense to track if we are about to execute a large download at the manifest level. Within the [start_manifest_processing](https://github.com/project-koku/koku/blob/main/koku/masu/processor/orchestrator.py#L233-L240) we already have logic to determine the queue, we just need to add the file count logic to it IMO.

deploy/clowdapp.yaml
myersCody: The boolean flag on the provider will be problematic when it comes to configuring this threshold

koku/masu/processor/orchestrator.py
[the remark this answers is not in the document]
lcouzens: Thinking about this again I'm not sure I'm following the `polling expires` point.

The way I see it is if the provider was not polled within the last 24 hours it is included in the collection limit based on our batch size count (in this case the batch size or limit would be 1, so 1 provider would be collected) Any provider that was not collected outside of the polling time would get collected on the following collection. That being said I don't see if we do any ordering in this query so maybe your referring to that? If we don't order this then its conceivable that we will randomly select a provider from the list of greater than 24 hours, which _could_ mean that one provider not caught in the 24 hours is missed every time. I wonder if just adding and order by to `get_polling_Batch` query would be enough. Meaning we would attempt to grab the provider that was polled the longest time ago.

More to think about for sure! Also sorry for brain dump reply, just trying to think through it better. If I'm off base please correct me too. :+1:

koku/api/migrations/0066_provider_large.py
```diff
@@ -0,0 +1,18 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("api", "0065_alter_exchangerates_currency_type"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="provider",
+            name="large",
```
myersCody, line 15: I don't see a clear plan for maintaining this boolean flag within our system. With this approach it feels like once we classify a provider as large, that provider is stuck with that classification moving forward. I would like for us to consider if there is a possible alternative approach that allows for us to change the threshold without needing additional workflows to clear out stale false positives that we saved within the database.

lcouzens: Correct that would be the initial approach, I understand the point on the outdated size/flag. Essentially this is maintaining our unleash flag logic, once classed as large its always large. But your right this is configurable in unleash as opposed to manually changing a flag. We could adapt this logic or do the manifest approach you suggested above. Though I was trying to avoid having a set large customer flipping back and forth between standard and XL queues, but maybe that's not the worst thing. 
Ultimately this was a good discussion POC to decide what we don't like/want!

PR #5415, opened 2024-12-03
[COST-5658] Fix regression for removing report period

bacciotti:
## Jira Ticket

[COST-5658](https://issues.redhat.com/browse/COST-5658)

## Description

This change improves the deletion process for old data in the database by improving how rows from the `OCPUsageLineItemDailySummary` and `OCPUsageReportPeriod` models are handled. The deletion logic now ensures that data integrity is maintained while addressing foreign key constraints.

An alternative approach is to create a migration to update the foreign key constraint on the `OCPUsageLineItemDailySummary` model, changing it from `ON DELETE NO ACTION` to `ON DELETE CASCADE` by adding `on_delete=models.CASCADE`. 

## Testing

1. We need to make sure that the celery task properly deletes the old data from `reporting_ocpusagereportperiod` and `reporting_ocpusagelineitem_daily_summary` tables.

## Release Notes
- [ ] Fix regression for removing report period

```markdown
* [COST-5658](https://issues.redhat.com/browse/COST-5658) Fix regression for removing report period
```


koku/masu/processor/ocp/ocp_report_db_cleaner.py
```diff
@@ -118,17 +120,36 @@ def purge_expired_report_data_by_date(self, expired_date, simulate=False):
...
+                del_count = OCPUsageLineItemDailySummary.objects.filter(
+                    report_period_id__in=all_report_periods
+                ).delete()
+                LOG.info(
+                    log_json(
+                        msg=f"Deleted records from table {OCPUsageLineItemDailySummary._meta.db_table} using id",
+                        count=del_count,
+                        schema=self._schema,
+                    )
+                )
```
maskarb, line 143: Are we sure this table has leftover data after data-expiration??? I thought we just needed to delete the old OCPUsageReportPeriod entries...

bacciotti: I think so. Even though `OCPUsageLineItemDailySummary` model has [Cascade enabled](https://github.com/project-koku/koku/blob/afdbead2bd4ac4d66e5bc1e6bf635f1382cedee7/koku/reporting/provider/ocp/models.py#L120), the table definition has 

`REFERENCES <schema>.reporting_ocpusagereportperiod (id) MATCH SIMPLE`
`ON UPDATE NO ACTION`
 `ON DELETE NO ACTION`

(I think it should be ON DELETE CASCADE)

Which was raising `DETAIL:  Key (id)=(38) is still referenced from table "reporting_ocpusagelineitem_daily_summary".` That's why I suggested to delete from `OCPUsageLineItemDailySummary` too.

maskarb: I'm not following what you are saying. If we are having an issue with deleting from the `reporting_ocpusagereportperiod` because of the foreign key, then move the `OCPUsageReportPeriod.delete()` to below the data clean up in the PartitionTable. Like I mention above, we do not want to use the django delete method on a partition table when we are expiring data.

bacciotti: Yes, what you've said makes complete sense. I moved `OCPUsageReportPeriod.delete()` down and removed `OCPUsageLineItemDailySummary` deletion. However, one point leaves me uncertain:

- Should we use `execute_delete_sql` for `OCPUsageReportPeriod` instead of Django's native `.delete()`,  or is this recommended only for partitioned tables? I tried that, but it didn’t work due to an FK constraint... and I noticed that in the table definition, there’s no ON CASCADE deletion (even though it is defined in the model). This is what confused me.

koku/masu/processor/ocp/ocp_report_db_cleaner.py
```diff
@@ -118,17 +119,16 @@ def purge_expired_report_data_by_date(self, expired_date, simulate=False):
...
-                )
-            )
-
             if not simulate:
+                LOG.info(
+                    log_json(
+                        msg="removing all data related to cluster_ids",
+                        report_periods=all_report_periods,
+                        cluster_ids=all_cluster_ids,
+                        period_starts=all_period_starts,
```
maskarb, line 128: Are we doing anything with these?

bacciotti: Nothing but unuseful logging. I removed them.

koku/masu/processor/ocp/ocp_report_db_cleaner.py
```diff
@@ -115,20 +114,15 @@ def purge_expired_report_data_by_date(self, expired_date, simulate=False):
...
-                    cluster_ids=all_cluster_ids,
-                    period_starts=all_period_starts,
-                    schema=self._schema,
-                )
-            )
 
             if not simulate:
+                LOG.info(
+                    log_json(
+                        msg="removing all data related to cluster_ids",
```
maskarb, line 121: This log statement is not accurate. Let's just remove this entire message.

koku/masu/processor/ocp/ocp_report_db_cleaner.py
```diff
@@ -147,6 +141,16 @@ def purge_expired_report_data_by_date(self, expired_date, simulate=False):
                 )
                 LOG.info(log_json(msg="deleted table partitions", count=del_count, schema=self._schema))
 
+                # Remove all data related to the report period
+                del_count, _ = OCPUsageReportPeriod.objects.filter(id__in=all_report_periods).delete()
+                LOG.info(
+                    log_json(
+                        msg=f"Deleted records from table {OCPUsageReportPeriod._meta.db_table} using id",
+                        count=del_count,
+                        schema=self._schema,
```
maskarb, line 150:
```suggestion
                        msg=f"deleted ocp-usage-report-periods",
                        count=del_count,
                        report_periods=all_report_periods,
                        schema=self._schema,
```

maskarb (approved):
Just some comments about these log messages, but otherwise LGTM!

maskarb (approved):

PR #5416, opened 2024-12-03
[COST-5170] - Auto large provider based on report count

lcouzens:
## Jira Ticket

[COST-5170](https://issues.redhat.com/browse/COST-5170)

## Description

This change will add a small amount of logic to manifest processing to adjust the queue in which tasks land for the given provider based on whether its deemed large by report count.

## Testing

1. Checkout Branch
2. Restart Koku
3. Set `XL_REPORT_COUNT` to 0 (Or generate a large amount of report files with nise)
4. Load data
5. You should see the following log message during processing `marking provider large as report files exceed threshold.` also if you are quick or scale down your worker you can check to see that tasks are queued in the relevant XL queues.

Additionally ingest a provider with less reports than `XL_REPORT_COUNT` and see that we process like normal in the standard queue.

## Release Notes
- [ ] proposed release note

```markdown
* [COST-5170](https://issues.redhat.com/browse/COST-5170) Add auto large provider logic.
```


koku/masu/processor/orchestrator.py
```diff
@@ -272,6 +272,21 @@ def start_manifest_processing(  # noqa: C901
...
             last_report_index = len(report_files) - 1
+            # Override the queue if we determine a provider is large based on report count
+            report_context = manifest.copy()
+            if len(report_files) > settings.XL_REPORT_COUNT:
+                LOG.info(
+                    log_json(
+                        tracing_id,
+                        msg="marking provider large as report files exceed threshold.",
+                        file_count=len(report_files),
+                        Threshold={settings.XL_REPORT_COUNT},
```
maskarb, line 283: What is this syntax?
```
Threshold={settings.XL_REPORT_COUNT},
```

koku/masu/processor/orchestrator.py
```diff
@@ -272,6 +272,21 @@ def start_manifest_processing(  # noqa: C901
...
+                        schema=schema_name,
+                    )
+                )
+                report_context["large_provider"] = True
+                SUMMARY_QUEUE = get_customer_queue(schema_name, SummaryQueue, report_context["large_provider"])
+                REPORT_QUEUE = get_customer_queue(schema_name, DownloadQueue, report_context["large_provider"])
```
maskarb, line 289: We ought to just pass in the bool instead of doing a dictionary lookup.
```suggestion
                SUMMARY_QUEUE = get_customer_queue(schema_name, SummaryQueue, True)
                REPORT_QUEUE = get_customer_queue(schema_name, DownloadQueue, True)
```

koku/masu/processor/tasks.py
```diff
@@ -308,6 +308,7 @@ def get_report_files(  # noqa: C901
             "invoice_month": report_dict.get("invoice_month"),
             "metadata_start_date": report_dict.get("metadata_start_date"),
             "metadata_end_date": report_dict.get("metadata_end_date"),
+            "large_provider": report_dict.get("large_provider"),
```
maskarb, line 311: Are we using this anywhere?

lcouzens: Not yet, I hadn't fully decided if we needed to though. But it could be removed for now. which would also simplify the above :+1:

koku/masu/processor/orchestrator.py
lcouzens: typos are fun.

myersCody (approved):

myersCody (approved):

PR #5425, opened 2024-12-12
[COST-5565] - dont match empty resource ids

lcouzens:
## Jira Ticket

[COST-5565](https://issues.redhat.com/browse/COST-5565)

## Description

This change will filter out OCP resources with an empty resource_id `''`

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Release Notes
- [ ] proposed release note

```markdown
* [COST-5565](https://issues.redhat.com/browse/COST-5565) Fix OCP/AWS managed table resource matching on `''`
```

### DIFF
--- a/koku/masu/database/trino_sql/aws/openshift/managed_aws_openshift_daily.sql
+++ b/koku/masu/database/trino_sql/aws/openshift/managed_aws_openshift_daily.sql
@@ -96,6 +96,7 @@ cte_array_agg_nodes AS (
     SELECT DISTINCT resource_id
     FROM hive.{{schema | sqlsafe}}.openshift_pod_usage_line_items_daily
     WHERE source = {{ocp_source_uuid}}
+        AND resource_id != ''
         AND year = {{year}}
         AND month = {{month}}
         AND interval_start >= {{start_date}}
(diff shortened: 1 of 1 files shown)


koku/masu/database/trino_sql/aws/openshift/managed_aws_openshift_daily.sql
```diff
@@ -105,6 +106,8 @@ cte_array_agg_volumes AS (
     SELECT DISTINCT persistentvolume, csi_volume_handle
     FROM hive.{{schema | sqlsafe}}.openshift_storage_usage_line_items_daily
     WHERE source = {{ocp_source_uuid}}
+        AND persistentvolume != ''
+        AND csi_volume_handle != ''
```
myersCody, line 110: If customers have not updated their operator they will not have the csi_volume_handle. So, we need to account for the case where a persistentvolume has been populated, but the `csi_volume_handle` has not.

lcouzens: I wondered if that might be a case!! Thanks. I originally just had the check on `persistentvolume` but wasnt sure if that would be enough. Based on what your saying I think that would be the better approach here.

myersCody:
We do still need to prevent an empty csi_volume handle string on the condition though or it will always return true. Give this a try, I have not ran it so it may need some tweaking:
```
OR (volumes.csi_volume_handle != '' AND substr(resource_names.lineitem_resourceid, -length(volumes.csi_volume_handle)) = volumes.csi_volume_handle)
```

bacciotti (approved):

PR #5431, opened 2024-12-23
[COST-5796] handle Customer.DoesNotExist exception

djnakabaale:
## Jira Ticket

[COST-5796](https://issues.redhat.com/browse/COST-5796)

## Description

This change adds logging for the `IntegrityError` to troubleshoot `Customer.DoesNotExist` exception

## Testing

1. Checkout Branch
2. Restart Koku
3. Run unit test and see log message 

```
python3 koku/manage.py test koku.test_middleware.IdentityHeaderMiddlewareTest.test_create_customer_integrity_error_existing_customer

```

## Release Notes
- [ ] log IntegrityError to troubleshoot Customer.DoesNotExist exception

```markdown
* [COST-5796](https://issues.redhat.com/browse/COST-5796) log IntegrityError to troubleshoot Customer.DoesNotExist exception
```


koku/koku/middleware.py
```diff
@@ -260,10 +262,38 @@ def create_customer(account, org_id, request_method):
...
+                else:
+                    LOG.warning(
+                        log_json(
+                            msg="Customer object not saved",
+                            request_method=request_method,
+                            account_id=account,
+                            org_id=org_id,
+                        )
+                    )
+                    return None
```
maskarb, line 275: Why is this change here? We still return the Customer object for `GET` and `HEAD` requests, but this change undoes that. We *intentionally do not save a customer* for `GET` and `HEAD` requests.

koku/koku/middleware.py
```diff
@@ -260,10 +262,38 @@ def create_customer(account, org_id, request_method):
...
+                return customer
+            except Customer.DoesNotExist:
+                LOG.warning(
+                    log_json(
+                        msg="Customer does not exist. Returning None.",
+                        account_id=account,
+                        org_id=org_id,
+                    )
+                )
+                return None
```
maskarb, line 296: What is the impact of returning None? Won't this just error somewhere else?

maskarb (changes requested):

koku/koku/middleware.py
djnakabaale: Yeah `None` maybe problematic else where like in some of the ProviderBuilder methods, returning an empty `Customer` object instead.

maskarb: I'm not sure this is a good idea either? How does this resolve the issue that we're encountering? What is the impact of passing around an empty Customer object??

maskarb (approved):

PR #5469, opened 2025-01-28
[COST-5137] Initial managed summary tasks rework

lcouzens:
## Jira Ticket

[COST-5137](https://issues.redhat.com/browse/COST-5137)

## Description

This change will move our ocp on cloud update summary tables logic to its own function and enable us to trigger it subtly differently for Manged tables vs Standard ingest flows.

## Test 1

1. Checkout Branch
2. Restart Koku
3. ingest data like normal (Old way)
4. Everything should function correctly (no regressions)

## Test 2

1. Checkout Branch
2. Restart Koku
3. Enable the following unleash flags `is_managed_ocp_cloud_processing_enabled` and `is_managed_ocp_cloud_summary_enabled` (the summary flag additionally has a check on provider type too, which would also need to be enabled if your doing it via an unleash client)
4. Ingest ocp on cloud data.
5. Currently should see an error message of managed_####### table does not exist (requires Cody's https://github.com/project-koku/koku/pull/5475)
6. With Cody's PR everything should be populated correctly

## Release Notes
- [ ] proposed release note

```markdown
* [COST-5137](https://issues.redhat.com/browse/COST-5137) Adds new trigger for managed tables summary
```


koku/masu/processor/tasks.py
```diff
@@ -583,77 +642,22 @@ def update_summary_tables(  # noqa: C901
         set_summary_timestamp(ManifestState.FAILED, manifest_id)
         raise ex
 
-    if provider_type != Provider.PROVIDER_OCP:
```
lcouzens, line 586: 586 to 597 here is redundant code now that we always run cost model updates regardless of a cost model existing now. Also tweaking line 655 below should make this simpler.

koku/masu/database/aws_report_db_accessor.py
```diff
@@ -250,7 +260,10 @@ def populate_ocp_on_aws_cost_daily_summary_trino(
 
         unattributed_storage = is_feature_unattributed_storage_enabled_aws(self.schema)
 
-        sql = pkgutil.get_data("masu.database", "trino_sql/reporting_ocpawscostlineitem_daily_summary.sql")
+        sql_file = "trino_sql/reporting_ocpawscostlineitem_daily_summary.sql"
+        if is_managed_ocp_cloud_summary_enabled(self.schema, Provider.PROVIDER_AWS):
+            sql_file = "trino_sql/aws/openshift/managed_reporting_ocpawscostlineitem_daily_summary.sql"
```
myersCody, line 265: I have been making an effort to organize the sql files for populating the ocp on cloud daily summary table under the `daily_summary_flow` directory. We should consider moving these files under that directory to align with the files needed to populate this table.

myersCody: After a discussion with Luke he convinced me that if we had a `transition_to_posgresql` folder that these files would make more sense in there then what we are currently calling the  `daily_summary_flow`.

myersCody (approved):

PR #5474, opened 2025-01-29
[COST-5820] Return Azure errors to UI

djnakabaale:
## Jira Ticket

[COST-5820](https://issues.redhat.com/browse/COST-5820)

## Description

This change allows returning some of the Azure client errors in their raw format to UI where necessary.

## Testing

1. smokes and unit tests should pass

## Release Notes
- [ ] Return raw Azure client errors to UI.

```markdown
* [COST-5820](https://issues.redhat.com/browse/COST-5820) Return raw Azure client errors to UI.
```

### DIFF
--- a/koku/sources/sources_error_message.py
+++ b/koku/sources/sources_error_message.py
@@ -21,7 +21,7 @@ class SourcesErrorMessage:
 
     def azure_client_errors(self, message):
         """Azure client error messages."""
-        scrubbed_message = str(message)
+        scrubbed_message = message
         if "AADSTS700016" in message:
             scrubbed_message = ProviderErrors.AZURE_INCORRECT_CLIENT_ID_MESSAGE
         if "AADSTS90002" in message:
--- a/koku/sources/sources_error_message.py
+++ b/koku/sources/sources_error_message.py
@@ -21,9 +21,7 @@ class SourcesErrorMessage:
 
     def azure_client_errors(self, message):
         """Azure client error messages."""
-        scrubbed_message = ProviderErrors.AZURE_GENERAL_CLIENT_ERROR_MESSAGE
-        if any(test in message for test in ["http error: 401", "Authentication failed", "(401) Unauthorized"]):
-            scrubbed_message = ProviderErrors.AZURE_CLIENT_SECRET_INCORRECT_MESSAGE
+        scrubbed_message = str(message)
         if "AADSTS700016" in message:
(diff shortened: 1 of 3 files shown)


lcouzens (dismissed):
Hi David, I think we may have missed some points here. What we want to do is remove the Azure user facing error messages entirely. So today in the code `if "AADSTS700016" in message:` currently returns `AZURE_INCORRECT_CLIENT_ID_MESSAGE` <- our user defined message. The problem here is we obfuscate the real details. 

What we actually want to do is return the raw Azure.exception coming from Azure, for example:

`{'azure.exception': [ErrorDetail(string="Authentication failed: AADSTS7000222: The provided client secret keys for app 'uuid' are expired. Visit the Azure portal to create new keys for your app: https://aka.ms/NewClientSecret, or consider using certificate credentials for added security: https://aka.ms/certCreds. Trace ID: uuid Correlation ID: uuid Timestamp: timestamp", code='invalid')]}`

This should give the user all the relevant information for them to not only understand the issue but all investigate it themselves too. Hopefully that makes sense, let me know if it doesn't.

koku/sources/sources_error_message.py
```diff
@@ -62,7 +41,6 @@ def general_errors(self, message):
     def _display_string_function(self, key):
         """Return function to get user facing string."""
         ui_function_map = {
-            ProviderErrors.AZURE_CLIENT_ERROR: self.azure_client_errors,
```
myersCody, line 65: Are you sure that the error message is being sent back to the user? I feel like now the string function will always be None:
```
string_function = ui_function_map.get(key)
        return string_function
```

koku/sources/sources_error_message.py
```diff
@@ -19,27 +19,6 @@ def __init__(self, error):
         """Initialize the message generator."""
         self._error = error
 
-    def azure_client_errors(self, message):
```
myersCody, line 22: I think we just don't want to scrub the message anymore.
```
def azure_client_errors(self, message):
     return message
```

koku/sources/sources_error_message.py
djnakabaale: we fallback to the internal message if no display function is found

```
display_fn = self._display_string_function(key)
if display_fn:
    ...
else:
    display_message = internal_message

```

myersCody: I completely missed that David 👍 Thanks for the additional clarity there. 🙇

myersCody (approved):

PR #5502, opened 2025-02-12
improve sql filtering logic on the masu trino endpoint

djnakabaale:
## Description

This change uses sqlparse to improve SQL filtering logic. This should prevent false positives related disallowed keywords on queries like this one: 

`SELECT count(DISTINCT json_extract_scalar(json_parse(persistentvolumeclaim_labels), '$.kubevirt_io_created_by')) FROM openshift_storage_usage_line_items`

## Testing

1. Checkout Branch
2. Restart Koku
3. Hit trino-query endpoint with a POST request
     ```
      http://127.0.0.1:5042/api/cost-management/v1/trino/query/
      ```
    1. You should get 400 response on operations for disallowed keywords, otherwise request should succeed with a 200 response

## Release Notes
- [ ] improve sql filtering logic on the masu trino endpoint to prevent false positives

### DIFF
--- a/koku/masu/api/trino.py
+++ b/koku/masu/api/trino.py
@@ -3,10 +3,10 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 """View for trino query endpoint."""
-# flake8: noqa
 import logging
 
 import requests
+import sqlparse
 import trino
 from django.conf import settings
 from django.views.decorators.cache import never_cache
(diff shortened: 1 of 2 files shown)


koku/masu/api/trino.py
```diff
@@ -40,10 +40,12 @@ def trino_query(request):
...
+
         lowered_query = query.lower()
         dissallowed_keywords = ["delete", "insert", "update", "alter", "create", "drop", "grant"]
-        for keyword in dissallowed_keywords:
-            if keyword in lowered_query:
+
+        for keyword in sqlparse.parse(lowered_query)[0].flatten():
+            if keyword.value.lower() in dissallowed_keywords:
                 errmsg = f"This endpoint does not allow a {keyword} operation to be performed."
                 return Response({"Error": errmsg}, status=status.HTTP_400_BAD_REQUEST)
```
maskarb, line 50: I don't know the particular benefit of using sqlparse. This could have been handled by splitting the query on spaces and just comparing the lists via a set operation:
```
        lowered_query = set(query.lower().split(" "))
        dissallowed_keywords = {"delete", "insert", "update", "alter", "create", "drop", "grant"}
        if keywords := dissallowed_keywords.intersection(lowered_query):
            errmsg = f"This endpoint does not allow a {keywords} operation to be performed."
            return Response({"Error": errmsg}, status=status.HTTP_400_BAD_REQUEST)
```

Nothing wrong with using sqlparse, it just seems overkill to pull in a library for this.

djnakabaale: thanks for the suggestion 👍🏾

myersCody (approved):

PR #5509, opened 2025-02-19
[COST-5890] - Adding Node core month to cost model

lcouzens:
## Jira Ticket

[COST-5890](https://issues.redhat.com/browse/COST-5890)
[COST-5917](https://issues.redhat.com/browse/COST-5917)

## Description

This change will add a new `node_core_cost_per_month`, `node_core_cost_per_hour` and update the SQL to calculate these costs.

## Testing

1. Checkout Branch
2. Restart Koku
3. Create OCP provider
4. Create a cost model with the new `node_core_cost_per_month`, `node_core_cost_per_hour` rates (These can be used with tag based rates too)
5. Load OCP data!
6. See the Openshift API has new node rate attributing to usage costs or run the following query `select sum(cost_model_cpu_cost), node from reporting_ocpusagelineitem_daily_summary group by node;`

Example tag cost model:
```
{'name': 'cost-model', 'description': 'node cost model', 'source_type': 'OCP', 'source_uuids': ['d0bf05fc-7255-4572-acf1-98e6669150e6'], 'rates': [{'metric': {'name': 'node_core_cost_per_month'}, 'tag_rates': {'tag_key': 'node', 'tag_values': [{'tag_value': 'meliodas', 'unit': 'USD', 'usage': {'unit': 'USD', 'usage_end': None, 'usage_start': None}, 'value': '10000', 'description': '', 'default': True}, {'tag_value': 'gowther', 'unit': 'USD', 'usage': {'unit': 'USD', 'usage_end': None, 'usage_start': None}, 'value': '20000', 'description': '', 'default': False}]}}], 'currency': 'USD'}
```

## Release Notes
- [ ] proposed release note

```markdown
* [COST-5890](https://issues.redhat.com/browse/COST-5890) Add new node cost model metrics
```


lcouzens:
Failing test is expected `test_api_ocp_metrics` requires IQE update.

koku/api/metrics/constants.py
```diff
@@ -14,12 +14,30 @@
...
 OCP_CLUSTER_MONTH = "cluster_cost_per_month"
 OCP_PVC_MONTH = "pvc_cost_per_month"
 
-MEMORY_DISTRIBUTION = "memory"
-CPU_DISTRIBUTION = "cpu"
+# defines the usage type for each metric
+CPU = "cpu"
+MEM = "memory"
+STORAGE = "storage"
+METRIC_MAP = {
```
myersCody, line 27: Considering we already have `COST_MODEL_METRIC_MAP` in here, it would probably be good to make this var more descriptive. Maybe:
- `METRIC_USAGE_TYPE_MAP`
- `USAGE_METRIC_MAP`

That way there is better clarity between the two metric maps without having to go searching through the code to figure out where it is used.

koku/masu/database/ocp_report_db_accessor.py
```diff
@@ -673,20 +645,9 @@ def populate_tag_usage_costs(  # noqa: C901
                 }
             }
         """
-        # defines the usage type for each metric
-        metric_usage_type_map = {
```
myersCody, line 677: I like that you moved these to a centralized var. 👏

koku/reporting/migrations/0328_alter_ocpusagelineitemdailysummary_monthly_cost_type.py
```diff
@@ -0,0 +1,27 @@
...
+        ("reporting", "0327_ocpvirtualmachinesummaryp"),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="ocpusagelineitemdailysummary",
+            name="monthly_cost_type",
+            field=models.TextField(
+                choices=[
+                    ("Node", "Node"),
```
myersCody, line 18: I don't really see a good reason to do the `Node_Core_Hour` as a separate database migration, so I would probably include it in this one.

lcouzens: There is no migration for the hour one since its not a monthly cost type.

koku/api/metrics/test/test_views.py
```diff
@@ -75,7 +75,7 @@ def test_metric_map_values(self):
         client = APIClient()
 
         params = {"source_type": Provider.PROVIDER_OCP}
-        url = url + "?" + urlencode(params, quote_via=quote_plus) + "&limit=11"
+        url = url + "?" + urlencode(params, quote_via=quote_plus) + "&limit=13"
```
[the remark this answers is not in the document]
lcouzens: Great idea!

koku/api/metrics/constants.py
```diff
@@ -14,12 +14,30 @@
...
+    OCP_METRIC_CPU_CORE_EFFECTIVE_USAGE_HOUR: CPU,
+    OCP_METRIC_MEM_GB_USAGE_HOUR: MEM,
+    OCP_METRIC_MEM_GB_REQUEST_HOUR: MEM,
+    OCP_METRIC_MEM_GB_EFFECTIVE_USAGE_HOUR: MEM,
+    OCP_METRIC_STORAGE_GB_USAGE_MONTH: STORAGE,
+    OCP_METRIC_STORAGE_GB_REQUEST_MONTH: STORAGE,
+    OCP_NODE_CORE_HOUR: CPU,
+}
+
+MEMORY_DISTRIBUTION = MEM
```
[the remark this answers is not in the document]
lcouzens: Yeh I was going to do that but wanted someone to suggest it first TeeHee.

koku/api/metrics/constants.py
lcouzens: YES!! :+1:

lcouzens (commented):
Thanks for the feedback Cody! :heart:

koku/api/metrics/constants.py
```diff
@@ -14,14 +14,30 @@
...
 OCP_NODE_MONTH = "node_cost_per_month"
+OCP_NODE_CORE_MONTH = "node_core_cost_per_month"
 OCP_CLUSTER_MONTH = "cluster_cost_per_month"
 OCP_PVC_MONTH = "pvc_cost_per_month"
 
-MEMORY_DISTRIBUTION = "memory"
-CPU_DISTRIBUTION = "cpu"
+# defines the usage type for each metric
+CPU = "cpu"
+MEM = "memory"
```
maskarb, line 25: Why were the names here changed? `MEMORY_DISTRIBUTION` and `CPU_DISTRIBUTION` are *actually* descriptive. If I see either of `CPU_DISTRIBUTION` or `MEMORY_DISTRIBUTION` somewhere, I understand its intent. Making these `CPU` and `MEM` strips away meaning.

koku/masu/processor/ocp/ocp_cost_model_cost_updater.py
```diff
@@ -54,7 +54,9 @@ def __init__(self, schema, provider):
             )
             self._distribution_info = cost_model_accessor.distribution_info
 
-    def _build_node_tag_cost_case_statements(self, rate_dict, start_date, default_rate_dict={}, unallocated=False):
+    def _build_node_tag_cost_case_statements(
+        self, rate_dict, start_date, default_rate_dict={}, unallocated=False, node_core_type=False
```
maskarb, line 58: Why is the default value for `node_core_type` a boolean when the conditional is checking a string? The type of the default value should match the type which the function is expecting.

koku/api/metrics/constants.py
lcouzens: Honestly I don't mind on this, just changed it as Cody suggested. That said every where this is used for a distribution type instead of metric its clearly defined so its probably ok. For example: `self._distribution == metric_constants.MEM`. I have no problem changing it back if people prefer though.

koku/masu/processor/ocp/ocp_cost_model_cost_updater.py
lcouzens: This was originally a boolean and was used as such but changed while I was testing something else. Certainly should be changed back to match though, thanks :+1:

koku/api/metrics/constants.py
myersCody: Yeah, sorry Luke, you are getting a bit of contradictory opinions here. A little background into my suggestion to just call it `CPU` & `MEM`. Originally within this file we were only utilizing the `cpu` & `memory` strings to represent the different distribution types. 

However, Luke caught some usage metric variables that needed to be added to this constant file. Now those strings (`cpu` & `memory`) can be used to either represent distribution types or usage metric types. Instead of having separate variables that represent the same string, I suggested we have a less descriptive name where those strings are defined then a more descriptive name in the places they are being utilized within this file.  However, totally up for discussion if you disagree.

maskarb (approved):

koku/api/metrics/constants.py
maskarb: Ok, that's fine. I did not see the convo/suggestions above.

PR #5513, opened 2025-02-24
[COST-5839] - Switch to dynamic batch sizing for polling

lcouzens:
## Jira Ticket

[COST-5839](https://issues.redhat.com/browse/COST-5839)

## Description

This change will remove the current configurable batch count and replace it with dynamic logic based on overall provider count. It also orders the providers queries based on polling time too so we always try to collect the oldest providers first.

## Testing

1. Checkout Branch
2. Restart Koku
3. make create-test-customer
4. Hit the download endpoint with various params
5. See logs for triggered providers

Side note you may want to tweak the following settings to simulate scenarios.

`timedelta(seconds=settings.POLLING_TIMER)` - (Reduce this to 10s or something so you dont have to wait 24h)

`batch_limit = math.ceil(len(self.filter(**filters)) / 21)` - (change the `21` to something less than your provider count, to simulate a smaller batch size)

## Release Notes
- [ ] proposed release note

```markdown
* [COST-5839](https://issues.redhat.com/browse/COST-5839) Add dynamic provider batching logic
```


koku/api/provider/models.py
```diff
@@ -84,16 +85,23 @@ def get_queryset(self):
...
         """Return a Queryset of pollable Providers that have not polled in the last 24 hours."""
         polling_delta = datetime.now(tz=settings.UTC) - timedelta(seconds=settings.POLLING_TIMER)
         if not filters:
             filters = {}
-        if limit < 1:
+        # Dynamically set batch limit (divide count by 21 hours so we always trigger a few extra + round up)
+        batch_limit = math.ceil(len(self.filter(**filters)) / 21)
+        if batch_limit <= 1:
             # Django can't do negative indexing, so just return all the Providers.
             # A limit of 0 doesn't make sense either. That would just return an empty QuerySet.
```
maskarb, line 97: If you're changing the functionality of this function, then this comment should be removed. It no longer applies here.

koku/api/provider/models.py
```diff
@@ -481,7 +489,7 @@ def _delete_from_target(self, target_info, target_values=None):
   from {qual_table_name}
  where "{target_info["column_name"]}" = any(%s)
 ;
-"""
+"""  # noqa: E231 E702 E272
```
maskarb, line 492: What causes the need for this?

koku/masu/processor/orchestrator.py
```diff
@@ -296,7 +296,7 @@ def start_manifest_processing(  # noqa: C901
                     )
                     continue
 
-                cache_key = f"{provider_uuid}:{report_file}"
+                cache_key = f"{provider_uuid}:{report_file}"  # noqa: E231
```
maskarb, line 299: Same question, why is this change needed?

koku/api/provider/models.py
lcouzens: Sure thing.

koku/masu/processor/orchestrator.py
```diff
@@ -431,6 +431,7 @@ def prepare(self):
         providers = self.get_polling_batch()
         if not providers:
             LOG.info(log_json(msg="no accounts to be polled"))
+            return
```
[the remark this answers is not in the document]
lcouzens: So I added this because the logging was a little confusing. We log `no account to be polled`. Then we log `polling accounts` even though there was nothing to poll, which means we also do `provider in providers`. I just thought returning early was more accurate here. Happy to change though.

koku/api/provider/models.py
lcouzens: just to stop pre-commit complaining instead of running `--no-verify`. I can undo these though if you prefer.

koku/masu/processor/orchestrator.py
lcouzens: same as above.

maskarb (approved):

PR #5525, opened 2025-03-03
[COST-5833] Fix incorrect delta values in previous month cost reports

bacciotti:
## Jira Ticket

[COST-5833](https://issues.redhat.com/browse/COST-5833)

## Description

This change will fix the incorrect delta value and percentage in the -2 time scope report by updating the `relativedelta` function from `time_scope_value` to 1. As a result, the delta report will correctly display the previous month values instead of an older one.

## Testing

1. Checkout Branch
2. Restart Koku
3. Hit endpoint `/openshift/costs/?filter[time_scope_value]=-2&delta=cost`
    1. Check if the `delta_value` and `delta_percent` fields are consistent

## Release Notes
- [ ] Fix incorrect delta values in previous month cost reports

```markdown
* [COST-5833](https://issues.redhat.com/browse/COST-5833) Fix incorrect delta values in previous month cost reports
```


koku/api/query_handler.py
```diff
@@ -283,7 +283,7 @@ def _create_time_interval(self):
     def _get_date_delta(self):
         """Return a time delta."""
         if self.time_scope_value in [-1, -2, -3]:
-            date_delta = relativedelta.relativedelta(months=abs(self.time_scope_value))
+            date_delta = relativedelta.relativedelta(months=1)
```
myersCody, line 286: So, the time scope value here represents the month you want to view in the endpoint. 

```
-1 == current mont (ex March)
-2 == last month (ex Feb)
-3 == two months ago  (ex Jan)
```
With this context, we would always be showing them the delta of the current month even if they pass in -2 or -3?

bacciotti: From what I understand and based on [these notes](https://github.com/project-koku/koku/compare/esebesto/previous_month_delta), when it comes to delta handlers, the report timescope is already set. So in this case, we don’t need to go two or three months back, just the previous one. We can check this by hitting the endpoints from this branch. When **time_scope** is set, with or without **delta=cost**, it always returns the correct month.

myersCody: Alright, part of my confusion here is that we utilize the `get_filter` to set the initial start & end date. Then we use it again with the delta=True parameter. However, after stepping through it with pdb this does build the delta filter correctly. Thanks for the additional clarity Lucas. 
```
self.parameters.get_filter("time_scope_value")
'-3'
delta_filter
<Q: (AND: ('usage_start__gte', datetime.date(2024, 12, 1)), ('usage_start__lte', datetime.date(2024, 12, 31)))>
self.parameters.get_filter("time_scope_value")
'-2'
delta_filter
<Q: (AND: ('usage_start__gte', datetime.date(2025, 1, 1)), ('usage_start__lte', datetime.date(2025, 1, 28)))>
self.parameters.get_filter("time_scope_value")
'-1'
delta_filter
<Q: (AND: ('usage_start__gte', datetime.date(2025, 2, 1)), ('usage_start__lte', datetime.date(2025, 2, 4)))>```

myersCody (approved):

PR #5528, opened 2025-03-03
[COST-5772] gracefully handle IntegrityError on ocppvc table

djnakabaale:
## Jira Ticket

[COST-5772](https://issues.redhat.com/browse/COST-5772)

## Description

This change gracefully handles Integrity exception on OCPPVC table

## Testing

1. Checkout Branch
2. Restart Koku
3. Load test OCP data and see everything load fine

## Release Notes
- [x] proposed release note

```markdown
* [COST-5772](https://issues.redhat.com/browse/COST-5772) gracefully handle IntegrityError on OCPPVC table
```


koku/masu/database/ocp_report_db_accessor.py
```diff
@@ -856,23 +867,51 @@ def populate_node_table(self, cluster, nodes):
...
-                if not ocppvc.csi_volume_handle:
-                    # Update the existing record's csi_volume_handle
-                    ocppvc.csi_volume_handle = pvc[2]
-                    ocppvc.save(update_fields=["csi_volume_handle"])
-            except OCPPVC.DoesNotExist:
-                # If the record does not exist, create a new one
-                OCPPVC.objects.create(
-                    persistent_volume=pvc[0], persistent_volume_claim=pvc[1], csi_volume_handle=pvc[2], cluster=cluster
-                )
+                with transaction.atomic():
```
maskarb, line 882: Why do we need a transaction here? What does using a transaction provide for us which is not provided by the `try/except IntegrityError`?

koku/masu/database/ocp_report_db_accessor.py
```diff
@@ -856,23 +867,51 @@ def populate_node_table(self, cluster, nodes):
...
+                        # If the record does not exist, create a new one
+                        OCPPVC.objects.create(
+                            persistent_volume=pvc[0],
+                            persistent_volume_claim=pvc[1],
+                            csi_volume_handle=pvc[2],
+                            cluster=cluster,
+                        )
+
+            except IntegrityError:
+                LOG.warning(f"another process inserted this pvc record: {pvc}")
```
maskarb, line 901: `another process inserted this pvc record`...how do we know this?

koku/masu/database/ocp_report_db_accessor.py
djnakabaale: Without a transaction another process can insert a record between the existence check (get/filter.first()) and the create call, leading to the `IntegrityError: duplicate key value violates unique constraint` we see. The thinking here is to ensure they happen as a single transaction.

maskarb: How does using `transaction.atomic` here prevent another process from inserting a record between the check and the create? For example, here I started a transaction, checked if an entry existed, showed it didn't, tried inserting it, then hit a duplication error:
```
postgres=# begin ;
BEGIN
postgres=*# select * from org1234567_mskarbek.reporting_ocp_pvcs where persistent_volume_claim = 'multi_pvc_1_data' and persistent_volume = 'not-multi_pvc_2' ;
(0 rows)

postgres=*# insert into org1234567_mskarbek.reporting_ocp_pvcs (uuid, persistent_volume_claim, persistent_volume, cluster_id, csi_volume_handle) values (uuid_generate_v4(), 'multi_pvc_1_data', 'not-multi_pvc_2', '76acb315-fc38-4dd4-bd59-f4507673aad2', 'vol-civil') ;
ERROR:  duplicate key value violates unique constraint "reporting_ocp_pvcs_persistent_volume_persistent_volume_clai_key"
DETAIL:  Key (persistent_volume, persistent_volume_claim, cluster_id)=(not-multi_pvc_2, multi_pvc_1_data, 76acb315-fc38-4dd4-bd59-f4507673aad2) already exists.
postgres=!# end ;
ROLLBACK
postgres=# 
```

Why did this fail?

djnakabaale: yeah, doesn't seem like a transaction will address the issue here since no rows are locked when there's no match from the select call. I guess we can just gracefully handle(LOG with proper message) the exception for now.

maskarb: Sounds good. I think the rest of the changes here are fine, I would just remove the `transaction` block.

djnakabaale: Thanks for feedback, added those updates.

koku/masu/database/ocp_report_db_accessor.py
```diff
@@ -856,23 +866,50 @@ def populate_node_table(self, cluster, nodes):
...
+                try:
+                    OCPPVC.objects.create(
+                        persistent_volume=pvc[0],
+                        persistent_volume_claim=pvc[1],
+                        csi_volume_handle=pvc[2],
+                        cluster=cluster,
+                    )
+
+                except IntegrityError:
+                    LOG.warning(f"IntegrityError raised when creating pvc: {pvc}.")
```
maskarb, line 899: It would still be good to log the exception:
```suggestion
                except IntegrityError as e:
                    LOG.warning(f"IntegrityError raised when creating pvc: {pvc}: {e}")
```

This line also needs a unit test.

maskarb (commented):
This looks fine, but we do need a unit test here.

koku/masu/database/ocp_report_db_accessor.py
djnakabaale: cleaned up the log and added a unit test.

maskarb (approved):

PR #5535, opened 2025-03-07
[COST-5039] Remove infra map cache

myersCody:
## Jira Ticket

[COST-5039](https://issues.redhat.com/browse/COST-5039)

## Description

Remove infra map cache since it appears to be barely used and the infra map SQL recently received some optimizations.

## Release Notes
- [ ] proposed release note

```markdown
```


koku/masu/processor/ocp/ocp_cloud_updater_base.py
```diff
@@ -117,10 +117,10 @@ def _generate_ocp_infra_map_from_sql_trino(self, start_date, end_date, check_cac
                     end_date,
                     gcp_provider_uuid=self._provider_uuid,
                 )
+        if infra_map is None:
```
maskarb, line 120: Could we not just make this
```suggestion
        if not infra_map:
```
and keep all the `return {}` in the ocp_report_db_accessor? Is there a point in caching a result that is `{}`?

myersCody: This is a very valid point Michael and one that I had not considered. Part of the sql changes I suggested to Luke was for us to switch us from trying to correlate with every possible OpenShift cluster to just those that are uncorrelated. I think once that optimization goes in, we probably just don't need this cache anymore. I will go ahead and update the PR to remove the cache.

lcouzens (approved):

lcouzens:
For now we have decided to leave this in place while we are transitioning from non-managed to managed tables since we are running both flows and query the cache a lot more after this [change](https://github.com/project-koku/koku/commit/3e22c22eabc8fe2d9c8f89998dc3bc989854b7ae#diff-38de8af9d67794baedf3bdc0556a78a252f31caa1b4ff3885083355eebb0cbc6R27-R39). But once we switch to managed tables and remove the old flow we should query this mapping a LOT less.

PR #5546, opened 2025-03-18
[COST-6075] Fix cpu/mem provider map

lcouzens:
## Jira Ticket

[COST-6075](https://issues.redhat.com/browse/COST-6075)

## Description

This change will fix the ocp provider map to correctly resolve the `cost_total_distributed` field when querying the compute and memory api's using `?group_by[project]=*&order_by[distributed_cost]=desc`
This also adds the distribution totals and fixes the delta queries too.

## Testing

1. Checkout Branch
2. Restart Koku
3. Load ocp data
4. Query the compute, memory api's with the following `?group_by[project]=*&order_by[distributed_cost]=desc&delta=distributed_cost`
5. Should no longer see the following error: `Cannot resolve keyword 'cost_total_distributed' into field`

## Release Notes
- [ ] proposed release note

```markdown
* [COST-6075](https://issues.redhat.com/browse/COST-6075) Fix Cannot resolve keyword 'cost_total_distributed' into field
```


bacciotti (approved):
lgtm

PR #5551, opened 2025-03-24
[COST-6089] - Fix node hour tag rates

lcouzens:
## Jira Ticket

[COST-6089](https://issues.redhat.com/browse/COST-6089)

## Description

This change will refactor a little node cost model logic in order to fix tag based rates for the node_core_cost_per_hour metric. Before this change node hour tag based rates were not applied to unallocated line items.

## Testing

1. Checkout Branch
2. Restart Koku
3. Create OCP provider
4. Create a cost model with `node_core_cost_per_hour` making sure to use tag based rates
5. Load OCP data!
6. Run the following query `select sum(cost_model_cpu_cost), node, namespace, monthly_cost_type from reporting_ocpusagelineitem_daily_summary where usage_start = '2025-03-23' and monthly_cost_type in ('Tag', 'Node_Core_Hour') group by node, namespace, monthly_cost_type order by node;`
7. See cost is applied to unallocated entries.
8. Rerun the above against main and see no cost is applied to unallocated entries.

Or run the following IQE test `test_api_ocp_source_raw_advanced_rates` seeing it fail then run the same test from Pedro's branch with the test fixes.

Example tag cost model:

`{'name': 'cost-model', 'description': 'node cost model', 'source_type': 'OCP', 'source_uuids': ['d0bf05fc-7255-4572-acf1-98e6669150e6'], 'rates': [{'metric': {'name': 'node_core_cost_per_hour'}, 'tag_rates': {'tag_key': 'node', 'tag_values': [{'tag_value': 'meliodas', 'unit': 'USD', 'usage': {'unit': 'USD', 'usage_end': None, 'usage_start': None}, 'value': '10000', 'description': '', 'default': True}, {'tag_value': 'gowther', 'unit': 'USD', 'usage': {'unit': 'USD', 'usage_end': None, 'usage_start': None}, 'value': '20000', 'description': '', 'default': False}]}}], 'currency': 'USD'}`

## Release Notes
- [ ] proposed release note

```markdown
* [COST-6089](https://issues.redhat.com/browse/COST-6089) Fix node_core_cost_per_hour tag based rate
```


koku/masu/processor/ocp/ocp_cost_model_cost_updater.py
```diff
@@ -54,8 +54,8 @@ def __init__(self, schema, provider):
             )
             self._distribution_info = cost_model_accessor.distribution_info
 
-    def _build_node_tag_cost_case_statements(
-        self, rate_dict, start_date, default_rate_dict={}, unallocated=False, node_core=False
+    def _build_node_tag_cost_case_statements(  # noqa: C901
```
lcouzens, line 57: Truthfully I hate all these tag case statement builders! It's impossible to know what happens here without printing out the sql afterwards. We should find a better solution to this, maybe structured SQL files would be better. Perhaps we should include that work into the dataclass refactor issue and just refactor cost model stuff in general.

koku/masu/processor/ocp/ocp_cost_model_cost_updater.py
```diff
@@ -106,7 +112,9 @@ def _build_node_tag_cost_case_statements(
                 )
             if default_rate_dict:
                 rate_value = default_rate_dict.get(tag_key, {}).get("default_value", 0)
-                rate = get_amortized_monthly_cost_model_rate(rate_value, start_date)
+                rate = rate_value
+                if amortized:
+                    rate = get_amortized_monthly_cost_model_rate(rate_value, start_date)
```
lcouzens, line 117: This is really what decides if a node rate metric should be monthly of not.

lcouzens:
1 test failure as expected `test_api_ocp_source_raw_advanced_rates`

koku/masu/database/ocp_report_db_accessor.py
```diff
@@ -596,6 +596,65 @@ def populate_monthly_tag_cost_sql(  # noqa: C901
         LOG.info(log_json(msg="populating monthly tag costs", context=ctx))
         self._prepare_and_execute_raw_sql_query(table_name, sql, sql_params, operation="INSERT")
 
+    def populate_node_hour_tag_cost_sql(
```
maskarb, line 599: Do we really need this function? It looks like a copy of `populate_monthly_tag_cost_sql` where the only difference is that we don't change the sql template based on the cost_type.

I would just drop `monthly` from `populate_monthly_tag_cost_sql` and just reuse this func when populating the hourly costs.

koku/masu/processor/ocp/ocp_cost_model_cost_updater.py
```diff
@@ -212,6 +220,31 @@ def _get_all_monthly_tag_based_case_statements(self, openshift_resource_type, ra
             }
         return combined_case_statements
 
+    def _get_all_node_hour_tag_based_case_statements(self, rates, default_rates, start_date):
```
maskarb, line 223: This is also just a copy of `_get_all_monthly_tag_based_case_statements`, but only for nodes and passing in `amortized`. I would factor out the `build-case-statements` for nodes, something like:
```
    def _node_statements(self, rates, start_date, default_rates, *, unallocated, node_core, amortized):
        cost_case_statements = self._build_node_tag_cost_case_statements(
            rates, start_date, default_rates, node_core=node_core, amortized=amortized
        )
        unallocated_cost_case_statements = self._build_node_tag_cost_case_statements(
            rates,
            start_date,
            default_rates,
            unallocated=unallocated,
            node_core=node_core,
            amortized=amortized,
        )
        labels_case_statement = self._build_labels_case_statement(rates, "pod_labels", default_rate=default_rates)
        return cost_case_statements, unallocated_cost_case_statements, labels_case_statement
```

Then, the node if statement in `_get_all_monthly_tag_based_case_statements` can be replaced with:
```
        if openshift_resource_type in ["Node", "Node_Core_Month"]:
            node_core = metric_constants.OCP_NODE_CORE_MONTH if openshift_resource_type == "Node_Core_Month" else ""
            cost_case_statements, unallocated_cost_case_statements, labels_case_statement = self._node_statements(
                rates,
                start_date,
                default_rates,
                unallocated=True,
                node_core=node_core,
                amortized=True,
            )
```
And then you can have:
```
    def _get_all_node_hour_tag_based_case_statements(self, rates, default_rates, start_date):
        """Call and organize cost, unallocated, and label case statements."""
        cost_case_statements, unallocated_cost_case_statements, labels_case_statement = self._node_statements(
            rates,
            start_date,
            default_rates,
            unallocated=True,
            node_core=metric_constants.OCP_NODE_CORE_HOUR,
            amortized=False,
        )
        return {
            tag_key: {
                "cost": cost_case_statements.get(tag_key, {}),
                "unallocated": unallocated_cost_case_statements.get(tag_key, {}),
                "labels": labels_case_statement.get(tag_key, {}),
            }
            for tag_key in cost_case_statements
        }
```

bacciotti:
I ran the **test_api_ocp_source_raw_advanced_rates**.

Tested on main: failed
Tested on Pedro's branch: passed

lcouzens:
Switching to hot-fix since the failing test is in an IQE update. :chicken: - :egg:

PR #5559, opened 2025-03-31
[COST-6097] Add Cluster per hour cost metric

bacciotti:
## Jira Ticket

[COST-6097](https://issues.redhat.com/browse/COST-6097)

## Description

This change will add a new Cost Model metric: `cluster_cost_per_hour`.

Basically, the value is calculated by counting how many distinct hour occurrences there were on the same day in the cluster, using `COUNT(DISTINCT interval_start) AS cluster_hours` in the Trino table `openshift_pod_usage_line_items`.

Testing and validation instructions are below.

## Testing

1. Checkout Branch
2. Restart Koku
3. Ingest OCP data
4. Create a new Cost Model with the new `cluster_cost_per_hour` metric
5. Check Trino to see how many hours the cluster ran:
```
select 
    date(interval_start) as interval_start_date, 
    count(distinct interval_start) as cluster_hours 
from 
    openshift_pod_usage_line_items 
where
    year='2025' and month='04' -- replace with your local data
group by 
    DATE(interval_start) 
limit 1;
```

6. Multiply `cluster_hours` by the `cluster_cost_per_hour` rate you created in step 4 above
7. Check if the values match the result of this Postgres query:
```
SELECT 
    cluster_id, SUM(cost_model_cpu_cost)
FROM 
    org1234567.reporting_ocpusagelineitem_daily_summary 
GROUP BY 
    cluster_id ;
```

### Example
I created a cost model with a `cluster_cost_per_hour` rate of 1.8 USD
```
"rates": [
                {
                    "metric": {
                        "name": "cluster_cost_per_hour",
                        "label_metric": "Cluster",
                        "label_measurement": "Count",
                        "label_measurement_unit": "cluster-hour"
                    },
                    "description": "",
                    "tiered_rates": [
                        {
                            "unit": "USD",
                            "usage": {
                                "usage_end": null,
                                "usage_start": null
                            },
                            "value": 1.8
                        }
                    ],
                    "cost_type": "Supplementary"
                }
```

My cluster run for 14 hours
| cluster_id                     | cluster_alias                    | interval_day | cluster_interval_hours |
|-------------------------------|----------------------------------|---------------|-------------------------|
| test_cost_x_ocp_static_cluster_3 | test_cost_x_ocp_static_cluster_3 | 2025-04-02    | 14                      |

Postgres query result (14 * 1.8 = 25.2)
| cluster_id                     | sum   |
|-------------------------------|--------|
| test_cost_x_ocp_static_cluster_3 | 25.2  |

You can also check through `/v1/reports/openshift/costs/?group_by[cluster]=*`

## Release Notes
- [ ] Add cluster per hour cost metric to the Cost Model

```markdown
* [COST-6097](https://issues.redhat.com/browse/COST-6097) Add cluster per hour cost metric to the Cost Model
```

### DIFF
--- a/koku/masu/database/trino_sql/openshift/cost_model/hourly_cost_cluster.sql
+++ b/koku/masu/database/trino_sql/openshift/cost_model/hourly_cost_cluster.sql
@@ -51,41 +51,33 @@ SELECT uuid(),
     max(report_period_id) AS report_period_id,
     lids.cluster_id,
     max(clusterhrs.cluster_alias) AS cluster_alias,
-    data_source,
+    'Pod' AS data_source,
     usage_start,
     max(usage_end) AS usage_end,
-    lids.namespace,
-    node,
-    max(resource_id) AS resource_id,
-    pod_labels,
-    all_labels,
+    NULL AS namespace,
+    NULL AS node,
+    NULL AS resource_id,
+    NULL AS pod_labels,
+    NULL AS all_labels,
     source_uuid,
     {{rate_type}} AS cost_model_rate_type,
     max(clusterhrs.cluster_interval_hours) * CAST({{cluster_cost_per_hour}} AS DECIMAL(33, 15)) AS cost_model_cpu_cost,
     0 AS cost_model_memory_cost,
     0 AS cost_model_volume_cost,
-    cost_category_id
+    NULL AS cost_category_id
 FROM postgres.{{schema | sqlsafe}}.reporting_ocpusagelineitem_daily_summary AS lids
 JOIN cte_cluster_usage_hours AS clusterhrs
     ON lids.cluster_id = clusterhrs.cluster_id
     AND clusterhrs.interval_day = lids.usage_start
-WHERE usage_start >= DATE({{start_date}})
-    AND usage_start <= DATE({{end_date}})
+WHERE usage_start BETWEEN DATE({{start_date}}) AND DATE({{end_date}})
     AND report_period_id = {{report_period_id}}
     AND data_source = 'Pod'
     AND lids.monthly_cost_type IS NULL
     AND pod_usage_cpu_core_hours IS NOT NULL
     AND pod_request_cpu_core_hours IS NOT NULL
     AND pod_request_cpu_core_hours != 0
-GROUP BY usage_start,
-         usage_end,
-         source_uuid,
-         lids.cluster_id,
-         clusterhrs.cluster_alias,
-         node,
-         lids.namespace,
-         data_source,
-         cost_category_id,
-         pod_labels,
-         all_labels
+GROUP BY
+    lids.cluster_id,
+    usage_start,
+    source_uuid
 ;
--- a/koku/masu/database/trino_sql/openshift/cost_model/hourly_cost_cluster.sql
+++ b/koku/masu/database/trino_sql/openshift/cost_model/hourly_cost_cluster.sql
@@ -23,7 +23,7 @@ WITH cte_distribution_type as (
     -- get the distribution type from the cost model associated with this source
     SELECT
         {{source_uuid}} as source_uuid,
-        coalesce(max(cm.distribution), 'cpu')  as dt -- coalesce(max()) to return 'cpu' if a cost model does not exist for source
+        coalesce(max(nullif(json_extract_scalar(distribution_info, '$.distribution_type'), 'false')), 'cpu') as dt -- coalesce(max()) to return 'cpu' if a cost model does not exist for source
     FROM postgres.{{schema | sqlsafe}}.cost_model_map as cmm
     JOIN postgres.{{schema | sqlsafe}}.cost_model as cm
         ON cmm.cost_model_id = cm.uuid
@@ -46,8 +46,7 @@ cte_cluster as (
     -- get the total number of hours a cluster ran in a day and the total number of nodes
     SELECT
         date(interval_start) as interval_day,
-        cast(count(distinct interval_start) as decimal(24, 9)) as cluster_hours,
-        cast(count(distinct node) as decimal(24, 9)) as node_count
+        cast(cast(count(distinct interval_start) as decimal(24, 9)) / count(distinct node) * cast({{cluster_cost_per_hour}} as decimal(24, 9)) as decimal(24, 9)) as cost_per_node_per_hour
     FROM hive.{{schema | sqlsafe}}.openshift_pod_usage_line_items
     WHERE source = {{source_uuid}}
         AND year = {{year}}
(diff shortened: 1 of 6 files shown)


koku/masu/database/trino_sql/openshift/cost_model/hourly_cost_cluster.sql
```diff
@@ -0,0 +1,110 @@
...
+    FROM postgres.{{schema | sqlsafe}}.reporting_ocpusagelineitem_daily_summary
+    WHERE usage_start BETWEEN DATE({{start_date}}) AND DATE({{end_date}})
+        AND report_period_id = {{report_period_id}}
+        AND data_source = 'Pod'
+        AND monthly_cost_type IS NULL
+        AND pod_usage_cpu_core_hours IS NOT NULL
+        AND pod_request_cpu_core_hours IS NOT NULL
+        AND pod_request_cpu_core_hours != 0
+    GROUP BY cluster_id, usage_start
+)
```
maskarb, line 65: What is this `usage_line_count` actually counting?

maskarb (commented):
I'm still trying to understand what this SQL is doing, but here is an initial review.

koku/masu/database/trino_sql/openshift/cost_model/hourly_cost_cluster.sql
```diff
@@ -0,0 +1,110 @@
...
+        cte_clusters.cluster_id AS cluster_id,
+        cte_clusters.cluster_alias AS cluster_alias,
+        DATE(interval_start) AS interval_day,
+        COUNT(DISTINCT EXTRACT(hour FROM pod_usage.interval_start)) AS cluster_interval_hours
+    FROM hive.{{schema | sqlsafe}}.openshift_pod_usage_line_items pod_usage
+    INNER JOIN cte_clusters
+        ON pod_usage.namespace = cte_clusters.namespace
+    WHERE source = {{source_uuid}}
+    GROUP BY cte_clusters.cluster_id, cte_clusters.cluster_alias, DATE(interval_start)
+),
```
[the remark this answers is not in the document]
bacciotti: way simpler, thanks

koku/masu/database/trino_sql/openshift/cost_model/hourly_cost_cluster.sql
bacciotti: We had to count the occurrences to split the cost, avoiding inflating the value but I found a workaround!

koku/masu/database/trino_sql/openshift/cost_model/hourly_cost_cluster.sql
```diff
@@ -0,0 +1,75 @@
...
+         source_uuid,
+         lids.cluster_id,
+         lids.cluster_alias,
+         node,
+         lids.namespace,
+         data_source,
+         cost_category_id,
+         pod_labels,
+         all_labels
+ LIMIT 1
```
maskarb, line 74: There's quite a bit to unpack with this query.
1. There is a ton of inconsistency around accessing columns. Some use `lids.` and some do not. I know we are generally pretty lax about this, but we ought to either use the `lids.` prefix for everything, or only use it when strictly required. Nothing about this query requires the use of `lids.` to access a column name, but it is more clear about which columns are used when using `lids.`
2. I presume the `limit 1` here is not intentional? Why is that here?
3. using `lids.usage_start BETWEEN DATE AND DATE` _is fine_, however we do not do this anywhere else in our codebase. Only for the sake of consistency, let's stick to using ` usage_start >= date({{start_date}}) AND usage_start <= date({{end_date}}) `
4. the join can be done like:
```
JOIN cte_cluster_usage_hours AS clusterhrs
    ON cast({{source_uuid}} as uuid) = lids.source_uuid
    AND clusterhrs.interval_day = lids.usage_start
```
5. since we pass `report_period_id` as a query param, and could replace `max(report_period_id) AS report_period_id,` with `{{report_period_id}} as report_period_id`. It's generally better to avoid using a function like `max` if we don't need to use it (we are not consistent with this in other files; some use max(), so use the passed in param).
6. We need to get clarification on how to distribute the cost.

koku/masu/database/trino_sql/openshift/cost_model/hourly_cost_cluster.sql
```diff
@@ -0,0 +1,111 @@
...
+    lids.node,
+    lids.resource_id,
+    lids.pod_labels,
+    lids.pod_labels as all_labels,
+    lids.source_uuid,
+    {{rate_type}} as cost_model_rate_type,
+    -- distribute the cost evenly amongst the namespaces across nodes
+    -- cost = (# hours running / number of nodes) * rate * (namespace usage on node / total usage of that node)
+    CASE WHEN cte_distribution_type.dt = 'cpu'
+        THEN ( max(cte_cluster.cluster_hours)/max(cte_cluster.node_count) * cast({{cluster_cost_per_hour}} as decimal(24, 9)) * sum(lids.pod_effective_usage_cpu_core_hours)/sum(cte_node_usage.cpu_usage) )
```
myersCody, line 75: Right now we are doing this math on each row even though the numbers won't change.
```
max(cte_cluster.cluster_hours)/max(cte_cluster.node_count) * cast({{cluster_cost_per_hour}} as decimal(24, 9)
```
Another alternative is to calculate these numbers one time and utilize the result here instead:
```
SELECT
        date(interval_start) as interval_day,
        cast(count(distinct interval_start) / count(distinct node) as decimal(24, 9)) * cast(1.8 as decimal(24, 9)) as cluster_cost_per_day
    FROM openshift_pod_usage_line_items
    WHERE source = 'f53aab20-a590-416a-a570-6148124ed368'
    AND year = '2025'
    AND month = '04'
    GROUP BY date(interval_start)
    
     interval_day | cluster_cost_per_day
--------------+----------------------
 2025-04-01   | 5.400000000000000000
 2025-04-03   | 5.400000000000000000
 2025-04-05   | 5.400000000000000000
 2025-04-08   | 3.600000000000000000
 2025-04-07   | 5.400000000000000000
 2025-04-04   | 5.400000000000000000
 2025-04-06   | 5.400000000000000000
 2025-04-02   | 5.400000000000000000
 ```
 You could run this as a subquery:
 ```
 JOIN (
 -- QUERY
 ) as cluster_cost
 ON lids.usage_start = cluster_cost.usage

**Review by djnakabaale — APPROVED** [review]

**Review by myersCody — APPROVED** [review]

File: koku/masu/database/trino_sql/openshift/cost_model/hourly_cost_cluster.sql (hunk 4 above)
Comment (maskarb) [reply koku/masu/database/trino_sql/openshift/cost_model/hourly_cost_cluster.sql]: The cte was updated to calculate the cost per node per day.

File: koku/masu/database/trino_sql/openshift/cost_model/hourly_cost_cluster.sql [hunk 5]
```diff
@@ -0,0 +1,111 @@
...
+    cost_model_memory_cost,
+    cost_model_volume_cost,
+    cost_category_id
+)
+
+WITH cte_distribution_type as (
+    -- get the distribution type from the cost model associated with this source
+    SELECT
+        {{source_uuid}} as source_uuid,
+        coalesce(max(cm.distribution), 'cpu')  as dt -- coalesce(max()) to return 'cpu' if a cost model does not exist for source
```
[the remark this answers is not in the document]
maskarb: This was switched to use the `distribution_type` found in the `distribution_info` field.

While we may not strictly need the `coalesce(max())` since we shouldn't be running this sql if this cluster-hour rate does not exist, it doesn't hurt to keep it anyway.

PR #5577, opened 2025-04-17
[COST-5995] - Add new GCP table source validation

lcouzens:
## Jira Ticket

[COST-5995](https://issues.redhat.com/browse/COST-5995)

## Description

- This change will add a new table check to GCP's `cost_usage_source_is_reachable` validation in order to raise table issues to customers during source creation. Both missing columns and daily partitions are captured now.
- Additionally this also fixes the original alert that was being raised during GCP table queries which was caused by an unusual project_id/table_name that requires backticks to escape invalid table names. See https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical#identifiers  

## Testing

1. Checkout Branch
2. Restart Koku
3. Create a GCP provider with our normal dev account dataset
4. See that the provider is created correctly
5. Create GCP provider with a few custom tables I created in a new project (Reach out to me for details) I created a dataset for wrong partitions `test_part` and one with missing columns `test_col`
6. Try creating a new GCP provider using these
7. You should see the provider create fail and an error message showing the missing column or missing partition

Testing the backticks, you need my new project. But essentially you will want to just run the following queries and see the result, also test with our normal dev account project to make sure it still functions correctly.

Use the following query
```

query = f"SELECT * FROM `{project_id.dataset.table}`  WHERE DATE(_PARTITIONTIME) = '2025-04-18'"
client = bigquery.Client()
query_job = client.query(query).result()

```

You should see a successful query

Run again without backticks

```
query = f"SELECT * FROM {project_id.dataset.table} WHERE DATE(_PARTITIONTIME) = '2025-04-18'"
client = bigquery.Client()
query_job = client.query(query).result()
```

The above query should fail with something like `Syntax error: Missing whitespace between literal and alias at [1:30] `

## Release Notes
- [ ] proposed release note

```markdown
* [COST-5995](https://issues.redhat.com/browse/COST-5995) Fix escaping invalid table names and return GCP table errors to integrations
```

## Summary by Sourcery

Improve GCP table query source validation by adding better error handling and table name escaping for BigQuery queries

New Features:
- Add table query check to validate GCP cost usage source during provider creation

Bug Fixes:
- Fix escaping of invalid table names in BigQuery queries
- Improve error handling for table query validation

Enhancements:
- Refactor GCP report downloader to use more modular query and table name building functions
- Improve error messaging for table query failures

## Summary by Sourcery

Improve GCP cost usage source validation by adding better error handling and table name escaping for BigQuery queries

New Features:
- Add table query check to validate GCP cost usage source during provider creation

Bug Fixes:
- Fix escaping of invalid table names in BigQuery queries
- Improve error handling for table query validation

Enhancements:
- Refactor GCP report downloader to use more modular query and table name building functions
- Improve error messaging for table query failures

Chores:
- Moved common GCP query and table name building logic to a shared utility module


koku/providers/gcp/provider.py
```diff
@@ -84,8 +85,15 @@ def _detect_billing_export_table(self, data_source, credentials):
             if bigquery_table_id:
                 data_source["table_id"] = bigquery_table_id
                 self.update_source_data_source(data_source)
+                date_str = DateHelper().now_utc.strftime("%Y-%m-%d")
+                client = bigquery.Client()
+                client.query(build_query_statement(credentials, data_source, date_str))
```
myersCody, line 90: With this implementation you are turning an entire day's worth of data just to confirm we have our required columns. That seems like a bit of overkill to me. 

 One alternative approach here is to just grab the column information out of the   `INFORMATION_SCHEMA`. 
```
SELECT column_name, REGEXP_EXTRACT_ALL(data_type, r'(\w+)\s+\w+') as struct_fields from {{project}}.{{dataset}}.INFORMATION_SCHEMA.COLUMNS where table_name='{{table_name}}';
```
This will return data like:
```
{
  "column_name": "project",
  "struct_fields": ["id", "number", "name", "labels", "key", "value", "ancestry_numbers", "ancestors", "resource_name", "display_name"]
},
{
  "column_name": "billing_account_id",
  "struct_fields": []
}
```
You can use this information to build available columns list by combining the struct_fields with the column_name for example (`project.name`) and then confirming every column in the GCP_COLUMN_LIST is in the available list that you build from the results.

myersCody:
Utilizing the python client to get the schema still accomplishes the same goal here. 

1. Have you ran this on a table set up with our minimum IAM requirements to make sure another IAM requirement isn't needed? I am 99% certain only `bigquery.tables.get` is needed for the client command, but it would be good to confirm. 

2. Can you slack me your test project names, ill add em in the morning when I get in and give you the approval. I would just like to see it in action.

PR #5592, opened 2025-04-30
[COST-6268] - Refactor Managed table tasks to run per cluster

lcouzens:
## Jira Ticket

[COST-6268](https://issues.redhat.com/browse/COST-6268)

## Description

This change was really focused around breaking up the ocp on cloud matching logic per cluster. Before this change we collected a list of possible tag matches for all clusters associated with a cloud provider, this was fine with our limited data sets but when running this on our internal PNT data we hit a trino query error because the tags list was too long.
With that in mind this change moves us to collecting tags specifically for the cluster we are running on a per cluster basis. This change also moves that `cluster in clusters for loop` up the stack so we break these into distinct tasks enabling us to run these clusters in parallel instead of in a single task.

Additionally:
1.  The original `ocp_on_cloud_trino_task` that was created for managed table processing has been removed
2. The ocp on cloud managed table processing/summary has all been tied back into the standard summary flow (Similar to the non-managed flow)
3. The cloud temp table logic has been dropped and re-factored to pull from the parquet file tables now those have hte UUID row populated. Unfortunate

## Testing

1. Checkout Branch
4. Restart Koku
5. Enabled ocp on cloud managed table summary
6. Ingest ocp on cloud data
7. Everything should process correctly.

## Release Notes
- [ ] proposed release note

```markdown
* [COST-6268](https://issues.redhat.com/browse/COST-6268) Fix managed table tag matching trino errors
```

## Summary by Sourcery

Refactor the OCP on Cloud managed table processing to run per cluster, improving the task processing workflow and simplifying the code structure

Enhancements:
- Simplified OCP on Cloud processing workflow
- Improved task chaining for managed table processing
- Removed redundant processing loops in cloud report accessors

Chores:
- Removed deprecated processing methods
- Updated task signatures and function calls
- Simplified parameter passing in SQL metadata

## Summary by Sourcery

Refactor the OCP on Cloud managed table processing to run per cluster, improving task processing workflow and simplifying code structure

New Features:
- Parallel processing of OCP on Cloud tasks per cluster

Bug Fixes:
- Fixed Trino query errors caused by long tag lists
- Resolved issues with tag matching across multiple clusters

Enhancements:
- Improved task chaining for managed table processing
- Simplified OCP on Cloud processing workflow
- Removed redundant processing loops

Chores:
- Removed deprecated processing methods
- Updated task signatures and function calls
- Simplified parameter passing in SQL metadata


koku/masu/processor/tasks.py
```diff
@@ -1262,45 +1276,152 @@ def validate_daily_data(schema, start_date, end_date, provider_uuid, ocp_on_clou
...
-            tracing_id = report.get("tracing_id", report.get("manifest_uuid", "no-tracing-id"))
+        )
+        return
+    context["start_date"] = start_date
+    context["end_date"] = end_date
+    # TODO remove trino_cloud_initilizing_signiture_list logic when row_uuid is avaiable in parquet files COST-5866
+    # This needs to be run once for the cluod provider, before splitting tasks by cluster
+    trino_cloud_initilizing_signiture_list = []
+    trino_cloud_initilizing_signiture_list.append(
+        trino_initilise_cloud_data.si(
```
myersCody, line 1314: I am really struggling to understand why the creation of the trino row_uuid tables needed to be pulled out as a separate task here? 

1. I thought it may have something to do with a delete running before creation, but:
The `determine_truncates_and_deletes` method utilizes:
- self.daily_summary_table.objects.model._meta.db_table

```
@property
    def daily_summary_table(self):
        """Return the model class for the provider's OCP on Cloud daily summary table."""
        if self.provider_type in (Provider.PROVIDER_AWS, Provider.PROVIDER_AWS_LOCAL):
            return OCPAWSCostLineItemProjectDailySummaryP
        elif self.provider_type in (Provider.PROVIDER_AZURE, Provider.PROVIDER_AZURE_LOCAL):
            return OCPAzureCostLineItemProjectDailySummaryP
        elif self.provider_type in (Provider.PROVIDER_GCP, Provider.PROVIDER_GCP_LOCAL):
            return OCPGCPCostLineItemProjectDailySummaryP
        return None
 ```
 ^ All of these are postgresql tables? 
 
 2. After that I checked order
When I examined your chain 
 ```
 c = chain(trino_cloud_initilizing, trino_processing, pg_deletes, pg_summaries)
 ```
 
 The code path goes:
 ```
trino_processing -> populate_openshift_on_cloud_managed_tables -> process_ocp_cloud_trino -> populate_ocp_on_cloud_daily_trino
```
Each `populate_ocp_on_cloud_daily_trino` calls `0_prepare_daily_summary_tables.sql` first

```
def populate_ocp_on_cloud_daily_trino(self, sql_metadata: SummarySqlMetadata) -> Any:
        """Populate the managed_azure_openshift_daily trino table for OCP on Azure.
        Args:
            sql_metadata: object of SummarySqlMetadata class
        Returns
            (None)
        """
        managed_path = "trino_sql/azure/openshift/populate_daily_summary"
        prepare_sql, prepare_params = sql_metadata.prepare_template(
            f"{managed_path}/0_prepare_daily_summary_tables.sql"
        )
 ```
 Which includes the lines you removed.
 
 From my perspective, it feels like a bunch of unnecessary restructuring unless you have a scenario I am missing here or some good justification for it.

File: koku/masu/processor/parquet/ocp_cloud_parquet_report_processor.py [hunk 2]
```diff
@@ -311,19 +311,28 @@ def find_openshift_keys_expected_values(self, start_date, end_date, ocp_provider
...
         LOG.info(
             log_json(msg=f"starting OCP on {self.provider_type} managed tables processing", context=self._context)
         )
-        if not (ocp_provider_uuids := self.get_ocp_provider_uuids_tuple()):
-            return
-        matched_tags = self.get_matched_tags(ocp_provider_uuids)
+        matched_tags = self.get_matched_tags([ocp_provider_uuid])
         matched_tag_strs = []
         if matched_tags:
             matched_tag_strs = [json.dumps(match).replace("{", "").replace("}", "") for match in matched_tags]
```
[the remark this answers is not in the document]
myersCody: @lcouzens I think this is a valid approach

koku/masu/processor/parquet/ocp_cloud_parquet_report_processor.py
```diff
@@ -311,19 +311,28 @@ def find_openshift_keys_expected_values(self, start_date, end_date, ocp_provider
...
 
         sql_metadata = SummarySqlMetadata(
-            self.db_accessor.schema, ocp_provider_uuids, self.provider_uuid, start_date, end_date, matched_tag_strs
+            self.db_accessor.schema, ocp_provider_uuid, self.provider_uuid, start_date, end_date, matched_tag_strs
         )
         self.db_accessor.populate_ocp_on_cloud_daily_trino(sql_metadata)
+
+    def initialise_managed_cloud_row_uuid_data(self, provider_type, start_date, end_date):
+        """Create table and populate cloud data with unique row_uuid. This is a workaround for COST-5866"""
```
myersCody, line 330: https://github.com/project-koku/koku/pull/5500

koku/masu/database/report_db_accessor_base.py
```diff
@@ -297,7 +297,7 @@ def find_openshift_keys_expected_values(self, ocp_provider_uuid: str, sql_metada
         matched_tag_params = sql_metadata.build_params(
             ["schema", "start_date", "end_date", "month", "year", "matched_tag_strs"]
         )
-        matched_tag_params["ocp_source_uuid"] = ocp_provider_uuid
+        matched_tag_params["ocp_source_uuid"] = sql_metadata.ocp_provider_uuid
```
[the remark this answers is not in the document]
lcouzens: Yeh we could do that :+1:

koku/masu/processor/tasks.py
lcouzens: So the problem here is that we are running these SQL files per cluster now. But I don't think we should run the populate cloud row_uuid data per cluster especially in parallel. This would mean while one cluster is doing matching another worker could be deleting and inserting the cloud data again. I could be wrong but thats the part I was attempting to prevent. Also when we can switch to using the UUID in the parquet tables (after 3 months) this separate task goes away.

myersCody: That is a fair point, how would you feel about utilizing the row_uuid we added the parquet: https://github.com/project-koku/koku/pull/5500

However, I am wondering if we missed a migration? May be good to check if we can query those tables in stage & prod.

myersCody (approved):

koku/masu/processor/tasks.py
lcouzens: Right I did look into that and had the exact same thoughts. I'm not sure if we are missing the migration or if the row uuid is just not populated in all the parquet files yet. I think its only in the last 2 months files at this point.

lcouzens:
Smoke tests passed, switching to full run smokes to catch anything else.

koku/masu/processor/parquet/ocp_cloud_parquet_report_processor.py
```diff
@@ -186,6 +186,36 @@ def get_matched_tags(self, ocp_provider_uuids):
         set_value_in_cache(cache_key, matched_tags)
         return matched_tags
 
+    def get_matched_tags_single_cluster(self, ocp_provider_uuid):
```
myersCody, line 189: Thanks for putting this together! I noticed that `get_matched_tags_single_cluster` largely duplicates the logic in `get_matched_tags`, minus the cache layer. To avoid duplication and make the code more maintainable, I’d suggest extracting the shared tag retrieval logic into a private helper (e.g. _get_matched_tags).

```
def _get_matched_tags(self, ocp_provider_uuids, matched_tags=None):
        """Retrieves the matched tags. """
        ctx = {
            "schema": self.schema_name,
            "provider_uuid": self.provider_uuid,
            "provider_type": self.provider_type,
        }
        if len(ocp_provider_uuids) == 1:
            ctx["ocp_provider_uuid"] = ocp_provider_uuids[0]

        invoice_month = self.invoice_month_date or self.start_date
        if self.has_enabled_ocp_labels:
            enabled_tags = self.db_accessor.check_for_matching_enabled_keys()
            if enabled_tags:
                LOG.info(log_json(msg="getting matching tags from Postgres", context=ctx))
                matched_tags = self.db_accessor.get_openshift_on_cloud_matched_tags(self.bill_id)
            if not matched_tags and enabled_tags:
                LOG.info(log_json(msg="matched tags not available via Postgres", context=ctx))
                # This flag is specifically for disabling trino tag matching for specific customers.
                if is_tag_processing_disabled(self.schema_name):
                    LOG.info(log_json(msg="trino tag matching disabled for customer", context=ctx))
                    return []
                LOG.info(log_json(msg="getting matching tags from Trino", context=ctx))
                matched_tags = self.db_accessor.get_openshift_on_cloud_matched_tags_trino(
                    self.provider_uuid,
                    ocp_provider_uuids,
                    self.start_date,
                    self.end_date,
                    invoice_month_date=invoice_month,
                )
        return matched_tags
```
That will turn the current `get_matched_tags` into a wrapper function that handles the cache layer:
```
def get_matched_tags(self, ocp_provider_uuids):
        """A wrapper function that caches the matched tags results"""
        # Get matching tags
        cache_key = build_matching_tags_key(self.schema_name, self.provider_type)
        matched_tags = get_value_from_cache(cache_key)

        # If the key is in the cache but the value is None, there are no matching tags
        if matched_tags or is_key_in_cache(cache_key):
            LOG.info(log_json(msg="retreived matching tags from cache", context=ctx))
            return matched_tags

        matched_tags = self._get_matched_tags(ocp_provider_uuids, matched_tags)
        set_value_in_cache(cache_key, matched_tags)
        return matched_tags
```

Then you can just call `_get_matched_tags` logic directly, without having to deal with the cache. Instead of duplicating all the logic, and removing the cache. This way, we avoid duplication and keep the logic tightly coupled. Any future changes to the matching behavior would automatically apply to both cached and uncached cases, reducing the risk of one being updated while the other is missed. Let me know what you think!
```
matched_tags = self._get_matched_tags([ocp_provider_uuid]))
```

koku/masu/database/azure_report_db_accessor.py
```diff
@@ -442,6 +456,34 @@ def check_for_matching_enabled_keys(self):
                 return False
         return True
 
+    def _get_matched_tags_strings(self, bill_id, azure_provider_uuid, ocp_provider_uuid, start_date, end_date):
```
myersCody, line 459: This functionality seems really similar across the hyperscalers and may be better suited in the base class.

lcouzens: It does but it means refactoring a lot of other stuff to which I'm not sure I want to tackle right now. Maybe I will look at it again while tests are running.

myersCody: If you feel like it would be a big lift to move it to the base class, lets just move forward with this

myersCody (approved):

PR #5609, opened 2025-05-15
[COST-5523] Update tag handler to account for tag mappings

myersCody:
## Jira Ticket

[COST-5523](https://issues.redhat.com/browse/COST-5523)

## Description

This change will apply tag mapping logic to the tags/ endpoint. 

## Testing
Instructions outlined in the issue

## Release Notes
- [ ] proposed release note

```markdown
* [COST-5523](https://issues.redhat.com/browse/COST-5523) Support tag mappings within tags API
```

## Summary by Sourcery

Apply tag mapping logic to the tags API to merge child tag values under parent tags and adjust filtering accordingly

New Features:
- Introduce support for tag mappings in the tags endpoint to aggregate child tag values under their parent tags

Enhancements:
- Add a "child_keys" filter to include mapped child tags when filtering by key
- Exclude child tags from the tag key listings to prevent duplicates

Tests:
- Add unit test to verify that child tags are merged under the parent tag and their values are aggregated correctly


lcouzens:
@myersCody I think there's a bug in here after the last changes. After I look at the tags endpoint `api/cost-management/v1/tags/aws/` after mapping child/parent tags plus hit resummary we seem to have all tag values associated with all tag keys. :confused: Might be worth double checking this.

maskarb (approved):

PR #5671, opened 2025-07-10
[COST-6477] Support fractional hours for VMs metrics

bacciotti:
## Jira Ticket

[COST-6477](https://issues.redhat.com/browse/COST-6477)

## Description

This change will add the support of fractional hours for VMs cost model metrics, using the `vm_uptime_total_seconds` field. The `is_fractional_hours_vm_cost_enabled` feature flag controls if it is using fixed hours or fractional hours.

## Testing

1. Checkout Branch
2. Restart Koku
3. Create a `vm_cost_per_hour` cost model metric 
4. Set `is_fractional_hours_vm_cost_enabled` feature flag to True
5. Ingest OCP data (with `vm_seconds` [entry](https://github.com/project-koku/nise/blob/c0de46f68816cfa18e8bdbc62cfc4af88c74c917/example_ocp_static_data.yml#L81))
6. Check if the total cost matches the cost model metric and the uptime (in seconds) 
7. You can validate the fallback behavior by setting the`is_fractional_hours_vm_cost_enabled` feature flag to False

## Release Notes
- [ ] Support fractional hours for VMs metrics

```markdown
* [COST-6477](https://issues.redhat.com/browse/COST-6477) Support fractional hours for VMs metrics
```

## Summary by Sourcery

Enable fractional-hour billing for virtual machines by adding a feature flag that switches cost calculations to use total uptime in seconds instead of whole-hour intervals and updating the SQL and data-access layer accordingly.

New Features:
- Introduce a feature flag to toggle between fixed-hour and fractional-hour VM cost calculations

Enhancements:
- Refactor VM usage SQL to pull from openshift_vm_usage_line_items and sum vm_uptime_total_seconds
- Update cost-model queries to apply vm_uptime_total_seconds/3600 when fractional hours are enabled
- Add use_fractional_hours parameter in data-access layer and implement is_fractional_hours_vm_cost_enabled flag check


koku/masu/database/trino_sql/openshift/cost_model/hourly_cost_virtual_machine.sql
```diff
@@ -30,19 +30,23 @@ SELECT uuid(),
...
-    FROM hive.{{schema | sqlsafe}}.openshift_pod_usage_line_items pod_usage
-    WHERE strpos(pod_usage.pod_labels, 'vm_kubevirt_io_name') != 0
-        AND source = {{source_uuid}}
-        AND year = {{year}}
-        AND month = {{month}}
+        vm_name,
+        DATE(interval_start) as interval_day,
+        count(interval_start) AS vm_interval_hours,
+        sum(vm_uptime_total_seconds) AS vm_uptime_total_seconds
+    FROM hive.{{schema | sqlsafe}}.openshift_vm_usage_line_items
```
maskarb: Making this change assumes that a customer has upgraded their operator (which we have not even released yet). Is that the intention? If so, why do we need an unleash flag?

bacciotti: No it's not the intention. I'm gonna fix this 🔨

koku/masu/database/trino_sql/openshift/cost_model/hourly_vm_core.sql
```diff
@@ -48,7 +48,11 @@ WITH
             vm_map.vm_cpu_request_cores AS vm_cpu_cores,
             DATE(vm_map.interval_start) AS interval_day,
             vm_map.vm_name AS vm_name,
-            count(vm_map.interval_start) AS vm_interval_hours
+            {% if use_fractional_hours %}
```
myersCody: For this specific metric, the customer is required to be on operator 4.0.0, and since this metric is not live yet, we can just go ahead and switch completely over to fractional hours here.

koku/masu/database/trino_sql/openshift/cost_model/hourly_cost_virtual_machine.sql
myersCody: If you convert the seconds into hours inside of this query and use the same alias
```
sum(vm_uptime_total_seconds) / 3600 AS vm_interval_hours
```
Then you won't need to do the if else statement above:
```
{% if use_fractional_hours %}
        max(vmhrs.vm_uptime_total_seconds) / 3600 * CAST({{hourly_rate}} AS DECIMAL(33, 15)) AS cost_model_cpu_cost,
    {% else %}
        max(vmhrs.vm_interval_hours) * CAST({{hourly_rate}} as DECIMAL(33, 15)) AS cost_model_cpu_cost,
    {% endif %}
```

myersCody: Since you are using fractional hours now, you won't need:
```
count(interval_start) AS vm_interval_hours,
```

maskarb (commented):
I think the sql here looks good. I do have a suggestion though. Given that fractional hours relies on the `openshift_vm_usage_line_items` table existing, maybe the existence of that table should control whether or not fractional-hours are used (instead of unleash).

We can use `self.table_exists_trino(name-of-vm-table)` within `OCPReportDBAccessor` to determine if the table exists or not.

If we were to enable the unleash flag for a customer that had not updated their operator, then this sql will fail.

koku/masu/processor/__init__.py
```diff
@@ -27,6 +27,14 @@ def is_cost_6356_enabled():
     )
 
 
+def is_fractional_hours_vm_cost_enabled(schema_name):
+    return UNLEASH_CLIENT.is_enabled(
+        "cost-management.backend.enable_fractional_hours_vm_cost",
```
myersCody: I am questioning whether or not an unleash flag would be the best approach for this. It may be worth exploring if we can discover what operator version a source is running here. That would allow us to dynamically decide if a customer should be using fractional hours or whole hours for this metric without developer intervention

myersCody: We could also check whether or not the vm usage table exists to control this mechanic. 

LOL I just discovered that Michael beat me to the punch on this one. I agree with him we should dynamically decide if we are using fractional hours.

bacciotti: Yeah good idea guys. Thank you! I've updated the code :)

koku/masu/database/ocp_report_db_accessor.py
```diff
@@ -731,7 +736,11 @@ def populate_usage_costs(self, rate_type, rates, distribution, start_date, end_d
                 source_uuid=provider_uuid,
                 report_period_id=report_period_id,
             )
-            hourly_params = {"rate_type": rate_type, "hourly_rate": ocp_vm_core_hour_rate}
+            hourly_params = {
+                "rate_type": rate_type,
+                "hourly_rate": ocp_vm_core_hour_rate,
+                "use_fractional_hours": use_fractional_hours,
```
maskarb: This param appears unused.

maskarb (approved):

PR #5681, opened 2025-07-21
[COST-6490] Handle special chars in tags and cost category

bacciotti:
## Jira Ticket

[COST-6490](https://issues.redhat.com/browse/COST-6490)

## Description

This change will fix the special chars handling errors in tags and cost category. 
1. Fixes the whitespace error in `&group_by[aws_category:Stage2%20OUs]=*` and
2. Fixes the tuple out of range index error in `&group_by[tag:Tirea/app-group]=*`

## Testing whitespace error fix

1. Checkout Branch
2. Restart Koku
3. Ingest AWS sample data
4. Make sure you have at least one tag key that contains a whitespace: `SELECT * FROM org1234567.reporting_awsenabledcategorykeys;`
5. Hit endpoint `/cost-management/v1/reports/aws/costs/?group_by[aws_category:<TAG-KEY-WITH-WHITESPACE-HERE>]=*`
6. It should work without any errors or warnings.

## Testing tuple out of range index error fix

1. Checkout Branch
2. Restart Koku
3. Ingest AWS sample data: you can update your `koku/dev/scripts/nise_ymls/ocp_on_aws/aws_static_data.yml` to include a tag with a backslash, for example `resourceTags/user:Tirea/app-group: testing-testing`
4. Hit endpoint `/cost-management/v1/reports/aws/costs/?group_by[tag:<TAG-WITH-BACKSLASH-HERE>]=*`
6. It should work without any errors or warnings.

## Release Notes
- [ ] Fix the special chars handling errors in tags and cost category.

```markdown
* [COST-6490](https://issues.redhat.com/browse/COST-6490) Fix the special chars handling errors in tags and cost category.
```

### DIFF
--- a/koku/api/report/queries.py
+++ b/koku/api/report/queries.py
@@ -46,6 +46,7 @@ from api.query_handler import QueryHandler
 from api.report.constants import AWS_CATEGORY_PREFIX
 from api.report.constants import TAG_PREFIX
 from api.report.constants import URL_ENCODED_SAFE
+from api.utils import safe_column_alias
 
 LOG = logging.getLogger(__name__)
 
@@ -696,7 +697,9 @@ class ReportQueryHandler(QueryHandler):
         if aws_category_column := self._mapper.provider_map.get("aws_category_column"):
             groups = self.get_aws_category_keys("group_by")
             for aws_category in groups:
-                db_name = aws_category_column + "__" + strip_prefix(aws_category, AWS_CATEGORY_PREFIX)
+                raw_key = strip_prefix(aws_category, AWS_CATEGORY_PREFIX)
+                safe_key = safe_column_alias(raw_key)
+                db_name = f"{aws_category_column}__{safe_key}"
                 aws_category = str.encode(aws_category)
                 aws_category = quote_from_bytes(aws_category, safe=URL_ENCODED_SAFE)
                 group_pos = self.parameters.url_data.index(aws_category)
--- a/koku/api/report/queries.py
+++ b/koku/api/report/queries.py
@@ -684,8 +684,8 @@ class ReportQueryHandler(QueryHandler):
         tag_groups = self.get_tag_group_by_keys()
         for tag in tag_groups:
             original_tag = strip_prefix(tag, TAG_PREFIX)
-            sanitized_tag = sanitize_tag(original_tag)
-            tag_db_name = self._mapper.tag_column + "__" + sanitized_tag
+            safe_tag = safe_column_alias(original_tag)
+            tag_db_name = self._mapper.tag_column + "__" + safe_tag
             encoded_tag_url = quote(original_tag, safe=URL_ENCODED_SAFE)
             group_pos = self.parameters.url_data.index(encoded_tag_url)
             group_by.append((tag_db_name, group_pos))
--- a/koku/api/utils.py
+++ b/koku/api/utils.py
@@ -608,4 +608,4 @@ def safe_column_alias(alias: str) -> str:
     """
     Sanitize column alias to be SQL-safe (no spaces, quotes, etc.).
     """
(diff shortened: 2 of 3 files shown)


koku/api/utils.py
```diff
@@ -602,3 +603,10 @@ def get_months_in_date_range(
 
     months = dh.list_month_tuples(dt_start, dt_end)
     return [(start, end, dt_invoice_month) for start, end in months]
+
+
+def safe_column_alias(alias: str) -> str:
+    """
+    Sanitize column alias to be SQL-safe (no spaces, quotes, etc.).
+    """
+    return re.sub(r"[^a-zA-Z0-9_]", "_", alias)
```
myersCody, line 612: So, I think you are doing this work a bit early. We can change the tag but then it won't match what is stored in the database. I think for now all we want to do is utilize the santized tag correctly.

koku/api/report/queries.py
```diff
@@ -106,6 +108,25 @@ def sanitize_tag(tag):
...
 
 
+def restore_tag_alias(value, alias_lookup, prefix="No-"):
+    """
+    Restore the original tag value from an alias.
+    Example:
+        "Tirea_app_group" → "Tirea/app-group"
+        "No-Tirea_app_group" → "No-Tirea/app-group"
+    """
+    if not isinstance(value, str):
```
myersCody, line 118: Can you give me more context about what we are trying to accomplish with this function

myersCody:
Lets jump on a call and chat about this one tomorrow

bacciotti:
@myersCody thanks for taking a look at this one... it is pretty confusing sometimes!

Regarding the `safe_column_alias()` comment, just to clarify: the sanitization here is only meant to ensure we use ORM-safe field names. Since ORM doesnt allow certain chars (like `/`, `-`, spaces...) in field lookups, we sanitize the tag to avoid breaking the query construction. The actual key used to access the JSON field in the database still corresponds to the original tag (`'Tirea/app-group'`), not the sanitized one. So the query continues to match what is stored in the DB correctly.

As for `restore_tag_alias()`, the goal is to reverse the sanitization before returning the data to the API, so that users see the original tag names (with `/`), not the sanitized version used internally.

The key piece of this change is the alias lookup dictionaries (`_tag_alias_lookup` and `_tag_plural_alias_lookup`), which preserve the mapping between sanitized tag names and their original values. This allows us to:
- Build ORM-safe field references for group by and filtering;
- Perform correct matching against db;
- Return the original tag names in the API response.

Anyway, this makes sense to me, but considering all the layers the data goes through, I might’ve missed something. Thanks again for the review! happy to dig deeper together if we think there is a better approach.

koku/api/report/queries.py
```diff
@@ -683,11 +706,15 @@ def _get_tag_group_by(self):
         tag_groups = self.get_tag_group_by_keys()
         for tag in tag_groups:
             original_tag = strip_prefix(tag, TAG_PREFIX)
-            sanitized_tag = sanitize_tag(original_tag)
-            tag_db_name = self._mapper.tag_column + "__" + sanitized_tag
             encoded_tag_url = quote(original_tag, safe=URL_ENCODED_SAFE)
             group_pos = self.parameters.url_data.index(encoded_tag_url)
-            group_by.append((tag_db_name, group_pos))
+            safe_tag = safe_column_alias(original_tag)
+            tag_db_name = self._mapper.tag_column + "__" + safe_tag
```
maskarb: This is fundamentally broken. By changing the tag value, you've changed what we are querying in the database. It is impossible for this work because we need this value as-is to do a group-by. If the original value for a key in the database was "key with space", and we try to group on that, what we end up grouping by in the db is "key_with_space". This is wrong and breaks the query.

We need a solution that passes the original value to the query, else it is wrong.

PR #5687, opened 2025-07-28
[COST-6630] Don't use cost model rows to calculate node & cluster capacity

myersCody:
## Jira Ticket

[COST-6630](https://issues.redhat.com/browse/COST-6630)

## Description

This change will not use cost model rows to calculate node & cluster capacity.

## Release Notes
- [ ] proposed release note


myersCody:
Successful OCP smokes run: https://konflux-ui.apps.stone-prd-rh01.pg1f.p1.openshiftapps.com/ns/cost-mgmt-dev-tenant/applications/koku/pipelineruns/koku-ci-6wfft

PR #5707, opened 2025-08-09
[COST-6495] - fix GCP table missing invoice month on insert

lcouzens:
## Jira Ticket

[COST-6495](https://issues.redhat.com/browse/COST-6495)

## Description

This change will include invoice month when selecting data from trino to insert into postgres table `reporting_gcpcostentrylineitem_daily_summary` and adds new logic to make sure when we run cross invoice months we insert the correct bill id per invoice month.

Before this change we would run summary for both month partitions but not consider invoice month, meaning we would insert data twice if we had already ran summary for the last day of the month for a previous manifest. 
For example we have the following manifests:

 2025-07-30|2025-07-31 04:28:58+00:00
 2025-07-31|2025-08-01 04:57:13+00:00

When we run summary for the first manifest we use start and end dates of 2025-07-30 and 2025-07-31. This means we select two days of data from the year and month partitions 2025 and 07 respectively. So far so good! 

However when we run summary for the second manifest, we collect the range of months since we cross a boundary. This then means we run the following start/end dates 2025-07-31 and 2025-08-01 with both month partitions (07 + 08). 

(datetime.date(2025, 7, 31), datetime.date(2025, 8, 1)), {'year': '2025', 'month': '08'}
(datetime.date(2025, 7, 31), datetime.date(2025, 8, 1)), {'year': '2025', 'month': '07'}

The problem here is we have already inserted data for the 2025-07-31 for the invoice month 202507 with the previous manifest, but because we are not filtering on invoice month when we select the data for the new invoice month we duplicate the data being inserted for the 31st. This woudnt be a problem if the same data was deleted prior to insert but our deletions use the bill_id (which is based of invoice month).

## Testing filter flow

1. Checkout Branch
2. Restart Koku
3. Create a GCP storage only provider (using the Dev account + the lcouzens filtered data bucket)
4. Post the report already uploaded to that bucket "gcp.csv" twice once per invoice month.
5. This is done via the ingress reports api /api/cost-management/v1/ingress/reports/
6. payload should look something like this: {'source': '1', 'bill_year': '2025', 'bill_month': '07', 'reports_list': ['gcp.csv']}
7. Once ingested check trino tables for data compared to postgress, check these tables between posting. You should see no duplicated data

    TRINO: `select sum(cost), source, invoice_month from gcp_line_items_daily group by source, invoice_month order by source, invoice_month;`
    POSTGRES: select sum(unblended_cost), invoice_month from reporting_gcpcostentrylineitem_daily_summary group by invoice_month order by invoice_month;
    

## Testing standard flow

1. Checkout Branch
2. Restart Koku
3. Run IQE test with bigquery enabled `ENV_FOR_DYNACONF=local iqe tests plugin cost_management -k "api_gcp_ingest_single or api_gcp_source_raw_calc" -vv --pdb --trace` while utilising this nise branch https://github.com/project-koku/nise/pull/564 (That creates crossover data)
4. Once ingested check trino tables for data compared to postgress, you should see no duplicated data

    TRINO: `select sum(cost), source, invoice_month from gcp_line_items_daily group by source, invoice_month order by source, invoice_month;`
    POSTGRES: select sum(unblended_cost), invoice_month from reporting_gcpcostentrylineitem_daily_summary group by invoice_month order by invoice_month;

On main this above test would have duplicated date on the 1st of Aug.

## Release Notes
- [ ] proposed release note

```markdown
* [COST-6495](https://issues.redhat.com/browse/COST-6495) Fix cross over month duplication
```

### DIFF
--- a/koku/masu/database/gcp_report_db_accessor.py
+++ b/koku/masu/database/gcp_report_db_accessor.py
@@ -60,25 +60,25 @@ class GCPReportDBAccessor(SQLScriptAtomicExecutorMixin, ReportDBAccessorBase):
     def line_item_daily_summary_table(self):
         return GCPCostEntryLineItemDailySummary
 
-    def populate_ui_summary_tables(self, start_date, end_date, source_uuid, tables=UI_SUMMARY_TABLES):
+    def populate_ui_summary_tables(
+            self, start_date, end_date, source_uuid, invoice_month, tables=UI_SUMMARY_TABLES
+            ):
         """Populate our UI summary tables (formerly materialized views)."""
-        invoice_month_list = self.date_helper.gcp_find_invoice_months_in_date_range(start_date, end_date)
-        for invoice_month in invoice_month_list:
-            for table_name in tables:
-                sql = pkgutil.get_data("masu.database", f"sql/gcp/{table_name}.sql")
-                sql = sql.decode("utf-8")
-                # Extend the end date past the end of the month & add the invoice month
-                # in order to include cross over data.
-                extended_end_date = end_date + relativedelta(days=2)
-                sql_params = {
-                    "start_date": start_date,
-                    "end_date": extended_end_date,
-                    "schema": self.schema,
-                    "source_uuid": source_uuid,
-                    "invoice_month": invoice_month,
-                }
+        for table_name in tables:
+            sql = pkgutil.get_data("masu.database", f"sql/gcp/{table_name}.sql")
+            sql = sql.decode("utf-8")
+            # Extend the end date past the end of the month & add the invoice month
+            # in order to include cross over data.
+            extended_end_date = end_date + relativedelta(days=2)
+            sql_params = {
+                "start_date": start_date,
+                "end_date": extended_end_date,
+                "schema": self.schema,
+                "source_uuid": source_uuid,
+                "invoice_month": invoice_month,
+            }
 
-                self._prepare_and_execute_raw_sql_query(table_name, sql, sql_params, operation="DELETE/INSERT")
+            self._prepare_and_execute_raw_sql_query(table_name, sql, sql_params, operation="DELETE/INSERT")
 
     def get_cost_entry_bills_query_by_provider(self, provider_uuid):
         """Return all cost entry bills for the specified provider."""
--- a/koku/masu/database/gcp_report_db_accessor.py
+++ b/koku/masu/database/gcp_report_db_accessor.py
@@ -99,7 +99,7 @@ class GCPReportDBAccessor(SQLScriptAtomicExecutorMixin, ReportDBAccessorBase):
         return GCPCostEntryBill.objects.filter(billing_period_start__lte=date)
 
     def populate_line_item_daily_summary_table_trino(
-        self, start_date, end_date, source_uuid, bill_id, markup_value, invoice_month_date
+        self, start_date, end_date, source_uuid, bill_id, markup_value, invoice_month
     ):
         """Populate the daily aggregated summary of line items table.
 
@@ -138,6 +138,7 @@ class GCPReportDBAccessor(SQLScriptAtomicExecutorMixin, ReportDBAccessorBase):
             "schema": self.schema,
             "table": TRINO_LINE_ITEM_TABLE,
             "source_uuid": source_uuid,
+            "invoice_month": invoice_month,
             "markup": markup_value or 0,
             "bill_id": bill_id,
         }
--- a/koku/masu/processor/gcp/gcp_report_parquet_summary_updater.py
+++ b/koku/masu/processor/gcp/gcp_report_parquet_summary_updater.py
@@ -110,7 +110,7 @@ class GCPReportParquetSummaryUpdater(PartitionHandlerMixin):
                     self._provider.uuid, start, end, filters
                 )
                 accessor.populate_line_item_daily_summary_table_trino(
-                    start, end, self._provider.uuid, current_bill_id, markup_value, invoice_month_date
+                    start, end, self._provider.uuid, current_bill_id, markup_value, invoice_month
                 )
                 accessor.populate_ui_summary_tables(start, end, self._provider.uuid)
             accessor.populate_tags_summary_table(bill_ids, start_date, end_date)
--- a/koku/masu/processor/gcp/gcp_report_parquet_summary_updater.py
+++ b/koku/masu/processor/gcp/gcp_report_parquet_summary_updater.py
@@ -52,7 +52,6 @@ class GCPReportParquetSummaryUpdater(PartitionHandlerMixin):
             (str, str) A start date and end date.
 
         """
-        invoice_month = kwargs.get("invoice_month")
         start_date, end_date = self._get_sql_inputs(start_date, end_date)
 
         with CostModelDBAccessor(self._schema, self._provider.uuid) as cost_model_accessor:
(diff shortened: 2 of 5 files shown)


koku/masu/database/trino_sql/gcp/get_invoice_months_and_dates.sql
```diff
@@ -0,0 +1,6 @@
+SELECT
+    invoice_month, min(usage_start_time), max(usage_start_time)
+    FROM gcp_line_items_daily
+    WHERE usage_start_time > {{start_date}}
+    AND usage_start_time <= {{end_date}}
+    group by invoice_month
```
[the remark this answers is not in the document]
lcouzens: Good catch! I need to add that.

koku/masu/database/trino_sql/gcp/get_invoice_months_and_dates.sql
```diff
@@ -0,0 +1,7 @@
+SELECT
+    invoice_month, min(usage_start_time), max(usage_start_time)
+    FROM hive.{{schema | sqlsafe}}.gcp_line_items_daily
+    WHERE usage_start_time > {{start_date}}
```
myersCody: I have a couple of notes about this query. 
1. We most likely want to include the start date in this one: 
```
usage_start_time >= DATE('2025-08-01')
```
2. This query doesn't really utilize partition pruning well. I ran explain analyze on it locally:
```
  year:string:PARTITION_KEY
                :: [[2025]]
            month:string:PARTITION_KEY
                :: [[06], [07], [08]]
            source:string:PARTITION_KEY
                :: [[4fd28550-2225-4d2c-b1a7-f2fc20b4b359]]
```
Notice it is comparing the `usage_start_time` fields for ALL of the partitions available within Trino.

This query does a better job of pruning our partitions down to the current month and the previous month:
```
SELECT
    invoice_month, min(usage_start_time), max(usage_start_time)
    FROM gcp_line_items_daily
    WHERE usage_start_time >= DATE('2025-08-01')
    AND usage_start_time <= DATE('2025-08-12')
    AND source = '4fd28550-2225-4d2c-b1a7-f2fc20b4b359'
    AND (
        month = LPAD(CAST(EXTRACT(MONTH FROM DATE('2025-08-01')) AS VARCHAR), 2, '0')
        OR month = LPAD(CAST(EXTRACT(MONTH FROM DATE('2025-08-01') - INTERVAL '1' MONTH) AS VARCHAR), 2, '0')
    )
    AND year = CAST(EXTRACT(YEAR FROM DATE('2025-08-01')) AS VARCHAR)
    group by invoice_month;
```

Example Return:
```
 invoice_month |          _col1          |          _col2
---------------+-------------------------+-------------------------
 202508        | 2025-08-01 00:00:00.000 | 2025-08-12 00:00:00.000
 202507        | 2025-08-01 00:00:00.000 | 2025-08-01 00:00:00.000
(2 rows)
```
When I ran explain analyze on it, it limited our search to the last two partitions:
```
year:string:PARTITION_KEY
                :: [[2025]]
            month:string:PARTITION_KEY
                :: [[07], [08]]
            source:string:PARTITION_KEY
                :: [[4fd28550-2225-4d2c-b1a7-f2fc20b4b359]]
```

koku/masu/database/trino_sql/gcp/openshift/populate_daily_summary/1_resource_matching_by_cluster.sql
```diff
@@ -43,7 +43,7 @@ WITH cte_usage_date_partitions as (
     select
         year,
         month
-    from gcp_line_items_daily
+    from hive.{{schema | sqlsafe}}.gcp_line_items_daily
```
myersCody, line 46: We have a similar issue with partition pruning here as well. We can either update this query too. 

Or we could potentially look into adding a for loop logic to call these files per invoice month like you did [with this logic](https://github.com/project-koku/koku/pull/5707/files#diff-670e788c13a5df6405f8bbb354667d7a69266bf8684a3538138d84e33189920bR74-R78)
```
invoice_month_dates = accessor.fetch_invoice_months_and_dates(
                    start_date, end_date, self._provider.uuid
                )
for invoice_month, invoice_start, invoice_end in invoice_month_dates:
      invoice_month_date = DateHelper().invoice_month_start(invoice_month).date()
      bills = accessor.bills_for_provider_uuid(self._provider.uuid, invoice_month_date)
```
to the `opulate_ocp_on_gcp_cost_daily_summary_trino` call. The benefit of that route is that is keeps how we grab the invoice month and usage dates consistent. So an update in one query won't get missed in the other query. However, there is likely a degree of difficulty with this route too. So totally up to you on how we improve it.

lcouzens: Right I was wondering similar and thinking about the partitions too. I like your solution above so I'll try and get that in. I also noticed I messed up the `>=` :smile: that was unintentional and broke the smoke tests.

koku/masu/database/trino_sql/gcp/openshift/populate_daily_summary/1_resource_matching_by_cluster.sql
```diff
@@ -43,10 +43,18 @@ WITH cte_usage_date_partitions as (
...
     where usage_start_time >= {{start_date}}
     AND usage_start_time <= {{end_date}}
     AND source = {{cloud_provider_uuid}}
+    AND (
+        month = LPAD(CAST(EXTRACT(MONTH FROM DATE({{start_date}})) AS VARCHAR), 2, '0')
+        OR month = LPAD(CAST(EXTRACT(MONTH FROM DATE({{end_date}})) AS VARCHAR), 2, '0')
```
myersCody: This version of the query, doesn't pick up valid cross over data:
```
trino:org1234567> SELECT
               ->     invoice_month, min(usage_start_time), max(usage_start_time)
               ->     FROM gcp_line_items_daily
               ->     WHERE usage_start_time >= DATE('2025-08-01')
               ->     AND usage_start_time <= DATE('2025-08-12')
               ->     AND source = '4fd28550-2225-4d2c-b1a7-f2fc20b4b359'
               ->     AND (
               ->         month = LPAD(CAST(EXTRACT(MONTH FROM DATE('2025-08-01')) AS VARCHAR), 2, '0')
               ->         OR month = LPAD(CAST(EXTRACT(MONTH FROM DATE('2025-08-12')) AS VARCHAR), 2, '0')
               ->     )
               ->     AND year = CAST(EXTRACT(YEAR FROM DATE('2025-08-01')) AS VARCHAR)
               ->     group by invoice_month;
 invoice_month |          _col1          |          _col2
---------------+-------------------------+-------------------------
 202508        | 2025-08-01 00:00:00.000 | 2025-08-12 00:00:00.000
(1 row)
```
For example data has a usage start of August first, but has an invoice month of `202507`

koku/masu/database/gcp_report_db_accessor.py
```diff
@@ -56,25 +56,23 @@ def __init__(self, schema):
...
-                    "schema": self.schema,
-                    "source_uuid": source_uuid,
-                    "invoice_month": invoice_month,
-                }
+        for table_name in tables:
+            sql = pkgutil.get_data("masu.database", f"sql/gcp/{table_name}.sql")
+            sql = sql.decode("utf-8")
+            # Extend the end date past the end of the month & add the invoice month
+            # in order to include cross over data.
+            extended_end_date = end_date + relativedelta(days=2)
```
myersCody: I would double check if the end_date still needs to be extended now that you are dynamically finding the date range using Trino. It depends on all the places this method is being called.

koku/masu/processor/gcp/gcp_report_parquet_summary_updater.py
```diff
@@ -64,55 +63,53 @@ def update_summary_tables(self, start_date, end_date, **kwargs):
...
+            # We need to update/insert for both invoice months when dates cross a boundry
             with schema_context(self._schema):
-                if invoice_month:
+                # Dynamically lookup invoice and date ranges from trino data
+                invoice_month_dates = accessor.fetch_invoice_months_and_dates(
+                    start_date, end_date, self._provider.uuid
+                )
+                for invoice_month, invoice_start, invoice_end in invoice_month_dates:
                     invoice_month_date = DateHelper().invoice_month_start(invoice_month).date()
                     bills = accessor.bills_for_provider_uuid(self._provider.uuid, invoice_month_date)
```
myersCody: Optional: You could add `invoice_month` as an optional param to the `bills_for_provider_uuid` method for GCP. Then move the:
```
invoice_month_date = DateHelper().invoice_month_start(invoice_month).date()
```
logic inside of the method:
```
def bills_for_provider_uuid(self, provider_uuid, start_date=None):
        """Return all cost entry bills for provider_uuid on date."""
        bills = self.get_cost_entry_bills_query_by_provider(provider_uuid)
        if start_date:
            if isinstance(start_date, str):
                start_date = parse(start_date)
            bill_date = start_date.replace(day=1)
            bills = bills.filter(billing_period_start=bill_date)
        return bills
```
This would provide us with a consistent way to get the bills for a provider_uuid based on the invoice month.

koku/masu/database/trino_sql/gcp/openshift/populate_daily_summary/1_resource_matching_by_cluster.sql
lcouzens: The crossover data is built into the dates being passed in. As in the start/end dates are based on usage dates not partition times during file processing, so we shouldnt need to deviate from those. Please test it yourself too though but it works correctly for me, so I think I'm correct.

koku/masu/database/gcp_report_db_accessor.py
lcouzens: Right I'm not sure we need this anymore either. I can test without it and see if it can be removed.

koku/masu/processor/gcp/gcp_report_parquet_summary_updater.py
lcouzens: Good idea!

lcouzens: Done

koku/masu/database/gcp_report_db_accessor.py
lcouzens: Removed it and my crossover data still showed up :+1:

koku/masu/database/trino_sql/gcp/openshift/populate_daily_summary/1_resource_matching_by_cluster.sql
myersCody: Sorry, I am really not following `the partition times during file processing` part of your reply. For additional context here, I wrote a bigquery sql statement that helped me fuel the potential cross over data scenarios.  I went ahead and added `_PARTITION_TIME`  to that SQL just in case it is helpful to you since you mentioned it. 

My previous query, was me testing it myself?  For clarity I was using the second row from the bigquery return as my example:

```
invoice_month usage_start_time usage_end_time partition_date
202204	2022-03-31	2022-04-01	2022-04-01 00:00:00 UTC
202207	2022-08-01	2022-08-01	2022-08-01 00:00:00 UTC
202210	2022-09-30	2022-09-30	2022-10-01 00:00:00 UTC
```

myersCody: Also, for more clarity we are NOT deviating or manipulating the the usage dates in any way in that SQL statement. What we are trying to determine is which trino partitions might we find the those usage dates. For the GCP flow, the data is partitioned based on the invoice month. Therefore, the `2022-08-01` data in that example lives in the `07` partition. All we are trying to do is tell Trino to search that partition as well for our date range.

lcouzens: YES! Yesterday I was totally forgetting that the invoice month drove the partitions :face_exhaling: Thanks for reminding me.

PR #5715, opened 2025-08-15
[COST-6597] - Drop bogus OCP data

lcouzens:
## Jira Ticket

[COST-6597](https://issues.redhat.com/browse/COST-6597)

## Description

This change will attempt to drop bad data when we convert CSV to Parquet for OCP payloads regarding POD and Storage csvs

## Testing

1. Checkout Branch
2. Restart Koku
3. Create OCP provider with cluster_id 'bad_data'
4. Upload the attached payload to Minio in the ocp-ingress dir
5. Hit the following masu endpoint to ingest the payload http://localhost:5042/api/cost-management/v1/ingest_ocp_payload/?payload_name=bad_data_2.tar.gz 

6. Run the following query in Trino `select * from openshift_pod_usage_line_items_daily where pod_limit_cpu_core_seconds > 1.0e+18;`
7. You should see no records returned!

Do steps 1-6 fresh on main, after you query trino you will see bogus records returned. You will also see DOUBLE trace back trino error.

Payload: 
[bad_data.tar.gz](https://github.com/user-attachments/files/21796999/bad_data.tar.gz)

## Release Notes
- [ ] proposed release note

```markdown
```

## Summary by Sourcery

Introduce anomaly filtering to OCP post-processing by dropping rows with unrealistic CPU, memory, and storage metrics during CSV-to-Parquet conversion.

New Features:
- Drop anomalous pod and storage usage rows exceeding predefined CPU, memory, and storage thresholds.

Enhancements:
- Log details of any dropped anomalous rows for diagnostics.

Tests:
- Add unit tests for anomaly removal covering no anomalies, removal of bad rows, empty DataFrame handling, and integration with process_dataframe.

### DIFF
--- a/koku/masu/util/ocp/common.py
+++ b/koku/masu/util/ocp/common.py
@@ -331,7 +331,7 @@ OCP_REPORT_TYPES = {
     },
 }
 
-# 1,000,000,000,000,000,000 is a resonable value that should remove most anomlies from bad data
+# 1,000,000,000,000,000,000 is a reasonable value that should remove most anomalies from bad data
 THRESHOLD_MAP = {
     1e18: [
         "pod_usage_cpu_core_seconds",


koku/masu/util/ocp/ocp_post_processor.py
```diff
@@ -163,7 +163,47 @@ def _generate_daily_data(self, data_frame):
...
+            # Check if the column exists in the current DataFrame
+            if col in data_frame.columns:
+                # Update the mask to True for any row that exceeds the threshold.
+                anomalous_rows_mask = anomalous_rows_mask | (data_frame[col] > threshold)
+
+        # Log the rows that are being dropped for future investigation.
+        if anomalous_rows_mask.any():
+            dropped_data = data_frame[anomalous_rows_mask]
+            for _, row in dropped_data.iterrows():
+                LOG.warning(f"Dropping anomalous row with data: {row.to_dict()}")
```
maskarb: I think let's just not log every row. I don't know what I would do with this information if it were logged. I think we'd be better off logging "anomalous rows removed" with context about what file this is so we can look at the file itself.

lcouzens: Yeh thats a good idea, I don't know that we would get that many of these but logging the filename is probably better for tracking. I just let AI do its thing with that log message for sure :smile:

koku/masu/util/ocp/ocp_post_processor.py
```diff
@@ -163,7 +163,47 @@ def _generate_daily_data(self, data_frame):
 
         return daily_data_frame
 
+    def _remove_anomalies(self, data_frame: pd.DataFrame) -> pd.DataFrame:
```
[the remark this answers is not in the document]
myersCody: @lcouzens I would consider this approach. It feels like the vectorized DataFrame would be more efficient than the python for loop. Plus I find the threshold map easier to read. 

I am guessing these thresholds match our limits in the DB?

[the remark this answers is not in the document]
lcouzens: Right its just to catch bogus values but anything below these should fit into our DB limits.

koku/masu/util/ocp/ocp_post_processor.py
```diff
@@ -163,7 +164,39 @@ def _generate_daily_data(self, data_frame):
 
         return daily_data_frame
 
-    def process_dataframe(self, data_frame):
+    def _remove_anomalies(self, data_frame: pd.DataFrame, filename: str) -> pd.DataFrame:
+        """Removes rows with anomalous values from the DataFrame."""
+
+        threshold_map = {
+            1e15: [
```
myersCody: I am still a little confused by how you are determining the threshold limits here. On the [Postgresql side](https://github.com/project-koku/koku/blob/main/koku/reporting/provider/ocp/models.py#L150-L153) of things everything appears to be a `DECIMAL(33, 15)`. From my understanding the maximum number that can fit into that decimal restriction is `1e18`.

Granted when we [build the managed table](https://github.com/project-koku/koku/blob/main/koku/masu/database/trino_sql/reporting_ocpusagelineitem_daily_summary.sql#L275-L288) in Trino we transform the seconds into hours.

So wouldn't our threshold for the maximum number of seconds we can accept here be:
```1e18 * 3600```

Comment (myersCody) [reply koku/masu/util/ocp/ocp_post_processor.py]: I should add some additional clarity here. I am totally on board for if you were aiming for what is realistic versus what is possibly accepted by Postgres. Currently our Postgres table accepts 999 trillion hours, that is obviously an obnoxiously high number, so if you were shooting for a realistic number I am totally on board, just mostly trying to understand how you came up with these thresholds.

Comment (lcouzens) [reply koku/masu/util/ocp/ocp_post_processor.py]: Right its just to catch clearly bogus numbers. If you have preferred numbers like 3.6e+21 I'd be happy to change it. Or we make them configurable so we can easily change them. The current numbers are just AI suggestions to be clear.

Comment (myersCody) [reply koku/masu/util/ocp/ocp_post_processor.py]: I think basing the limit off what is realistic is perfectly fine with me. Mostly just wanted to understand how we determined the threshold. If we want to go with the AI recommendation that is fine with me! I just feel like it might be worth making a note, so that way if we every come back to it in the future we have an understanding of how we determined the thresholds here.

Comment (lcouzens) [reply koku/masu/util/ocp/ocp_post_processor.py]: Right my last rewrite destroyed my comments :boom: That said I did have one threshold too high! I pushed up some closer max's and added some comments on what they equate too.

File: koku/masu/util/ocp/ocp_post_processor.py [hunk 4]
```diff
@@ -163,7 +164,41 @@ def _generate_daily_data(self, data_frame):
...
+                "pod_limit_memory_byte_seconds",
+                "node_capacity_memory_byte_seconds",
+                "node_capacity_memory_bytes",
+                "persistentvolumeclaim_capacity_bytes",
+                "persistentvolumeclaim_capacity_byte_seconds",
+                "volume_request_storage_byte_seconds",
+                "persistentvolumeclaim_usage_byte_seconds",
+            ],
+        }
+        thresholds = {col: thresh for thresh, cols in threshold_map.items() for col in cols}
```
maskarb: Can we make this a constant somewhere outside of this func? Otherwise we re-create this map everytime this function is called, which is quite frequently throughout the day.

maskarb (approved):

PR #5721, opened 2025-08-21
[COST-6694] - Fix OCP/GCP crossover month matching

lcouzens:
## Jira Ticket

[COST-6694](https://issues.redhat.com/browse/COST-6694)

## Description

This change removes some custom GCP invoice month logic from the downloader and summary tasks and instead relies on the newly added extended date range by invoice date lookup logic. 

This simplifies a lot of unnecessary code and moves summary tasks to a monthly model similar to that of other providers. Now we trigger summary for GCP once per invoice month when a set of manifests/files cross a month (invoice) boundary instead of triggering summary for the entire range and running that range per invoice month. Additionally the reason to do this is for the sake of OCP/GCP summary. That must maintain strict start end dates per summary task in order to correctly match the clusters report_period, without this we insert the GCP matching billing data into the incorrect OCP report period.

## Testing

1. Checkout Branch
2. Restart Koku
3. Create some amazing GCP crossover data yamls, its much easier to rely on QE automation here if you don't want to manually create and verify it all.
4. Firstly checkout the nise branch `esebesto/test_cross_edge_case` this will give you crossover data for '2025-07-31' with the invoice month 202508 (Side note, make sure nise is installed as editable in IQE plugin or it wont pick up the change)
5. Update cost-plugin yaml `gcp_static_report_basic_template.yaml` by adding the following to some generators:

```
  - ComputeEngineGenerator:
      start_date: last_month
      price: 20
      usage.amount_in_pricing_units: 1
      sku_id: CF4E-A0C7-E3BF  # ("CF4E-A0C7-E3BF", "Instance Core running in Americas", "seconds", "hour"),
      instance_type: n1-standard-4
      location.region: us-west1-a
      currency: CNY
      resource.name: projects/todays-raw-calc-id/instances/test_gcp_kikis
      resource.global_name: //compute.googleapis.com/projects/todays-raw-calc-id/zones/us-west1-a/instances/3647398860992947185
      labels: [{"OCPonGCPCommonKey": "gcp_virtual_machine_1", "openshift_node": "test_gcp_kikis", "Titan": "Jaw", "kubernetes-io-cluster-test_cost_ocp_on_gcp_cluster_kikis{{ account_suffix }}": "owned", "tm_gcp_candidate1": "tm_gcp_value1", "correlation": "resource_id_matching"}]
      credit_amount: -3  # per hour
      cross_over:                          #### This part
          month: current               #### This part
          days: [2]                           #### This part - This could also be [1] or [1, 2] MAX of [1, 2, 3] if you want multiple days
          overwrite: True               #### This part
```

6. Enable non-local sources in cost_management_default.yaml `local_sources: False`
7. Add a breakpoint to `test_api_gcp_source_raw_calc` if you want to manually check things too
8. Run the following IQE command `ENV_FOR_DYNACONF=local iqe tests plugin cost_management -k gcp_source_raw_calc -vv --pdb` (Additionally you will need vault access configured for IQE to use bigquery)
9. OCP/GCP and GCP raw calc tests should pass!!
10. There should be NO duplication or missing cost for any days!
11. Feel free to run any trino/postgres sql to compare tables and cost

For example:
```
select sum(cost), usage_start_time, invoice_month, month from gcp_line_items_daily group by invoice_month, usage_start_time, month order by invoice_month, usage_start_time, month;

select sum(unblended_cost), usage_start, invoice_month from reporting_gcp_cost_summary_p group by usage_start, invoice_month order by usage_start, invoice_month;

select sum(unblended_cost), usage_start, invoice_month, month from managed_reporting_ocpgcpcostlineitem_project_daily_summary group by invoice_month, usage_start, month order by invoice_month, usage_start, month;

select sum(unblended_cost), usage_start  from reporting_ocpgcpcostlineitem_project_daily_summary_p group by usage_start order by usage_start;
```

## Release Notes
- [ ] proposed release note

```markdown
* [COST-6694](https://issues.redhat.com/browse/COST-6694) Fix OCP/GCP crossover month matching
```

## Summary by Sourcery

Consolidate GCP invoice month handling into a unified monthly model by replacing custom cross-month logic with a new extended date range lookup and removing invoice_month propagation throughout the pipeline to ensure accurate OCP/GCP report matching.

Bug Fixes:
- Fix OCP/GCP crossover month matching by extending invoice month date ranges to correctly align billing data with OCP report periods.

Enhancements:
- Replace custom GCP invoice month logic with a single fetch_invoice_month_dates method and updated SQL.
- Refactor GCP report and cost updater tasks to execute once per invoice month boundary and drop invoice_month parameters.
- Remove GCP-specific deduplication and strip invoice_month from get_months_in_date_range, report_data API, downloader, tasks, and utilities.
- Update Trino SQL for resource matching to include adjacent month partitions for proper OCP/GCP correlation.

Tests:
- Update tests across utils, downloader, DB accessor, and tasks to remove invoice_month parameters and use the new fetch_invoice_month_dates API.


myersCody:
For the most part your changes look good Luke! I am still curious if our Nise capabilities match real world scenarios, but I want to talk more about that with you tomorrow. As for what we can produce with Nise as of today. This is working!

koku/masu/database/trino_sql/gcp/get_invoice_month_dates.sql
```diff
@@ -0,0 +1,29 @@
+SELECT
+    min(usage_start_time),
+    max(usage_start_time)
```
myersCody, line 3: This is just a potential performance enhancement. Part of me is curious if we could return the year & month here, and remove the extra filters we are doing in 

- koku/masu/database/trino_sql/gcp/openshift/populate_daily_summary/1_resource_matching_by_cluster.sql

The only reason I call it out is cause the two queries are very very similar now.

lcouzens: Hmm, I guess we could. I think it might get a little messy passing those variables along to the OCP/GCP task though. Assuming I'm following along with what your saying. Or are you saying we should generalise this sql function and use it before calling 1_resource_matching_by_cluster sql so that we pass in the year/month lists instead? That said I'm not sure how much gain we will get from that.

myersCody (approved):

PR #5722, opened 2025-08-25
[COST-6567] Improve Azure v1 reports messages

bacciotti:
## Jira Ticket

[COST-6567](https://issues.redhat.com/browse/COST-6567)

## Description

This change will change the message and the log type for Azure v1 reports.

## Release Notes
- [ ] Improve unsupported Azure reports warnings

```markdown
* [COST-6567](https://issues.redhat.com/browse/COST-6567) Improve unsupported Azure reports warnings
```

## Summary by Sourcery

Change log level and message for unsupported Azure V1 reports to provide clearer guidance and reduce log severity

Enhancements:
- Change unsupported Azure V1 report log from error to warning
- Update log message to specify the V1 schema usage and recommend switching to V2


koku/masu/external/downloader/azure/azure_report_downloader.py
```diff
@@ -179,6 +182,7 @@ def __init__(self, customer_name, credentials, data_source, report_name=None, in
         self.storage_only = data_source.get("storage_only")
         self.ingress_reports = ingress_reports
         self.compression = None
+        self.type = data_source.get("resource_group", {}).get("type", None)
```
maskarb: Why would `resource_group` contain `type`? Where is this coming from?

koku/providers/azure/provider.py
```diff
@@ -127,10 +128,23 @@ def cost_usage_source_is_reachable(self, credentials, data_source):
...
+                report = export_reports[0]
+                report_type = report.get("type")
+
+                if report_type not in SUPPORTED_REPORT_TYPES:
+                    key = "AZURE_UNSUPPORTED_REPORT_TYPE"
+                    message = (
+                        f"Unsupported report type: '{report_type}'. " f"Supported types are: {SUPPORTED_REPORT_TYPES}."
+                    )
+                    raise ValidationError(error_obj(key, message))
+
```
maskarb: I think what we should do here instead is have `describe_cost_management_exports` do the validation and raise a new error that is specific to the incorrect report type.

koku/masu/external/downloader/azure/azure_report_downloader.py
```diff
@@ -108,12 +110,13 @@ def create_daily_archives(
     process_date = get_processing_date(
         s3_csv_path, manifest_id, provider_uuid, start_date, end_date, context, tracing_id, ingress_reports
     )
+    if report_type and report_type not in SUPPORTED_REPORT_TYPES:
+        msg = f"Unsupported Azure report type: '{report_type}'. " f"Supported types are: {SUPPORTED_REPORT_TYPES}"
+        LOG.warning(log_json(msg=msg, context=context))
+        return [], {}
```
maskarb: Previously, we did validation in this method because we were actually processing the report before we knew what was in it. If we have the report type ahead of time, then we should do validation before we ever call the `create_daily_archives` method.

Validation should probably be done in `describe_cost_management_exports`.

koku/masu/external/downloader/azure/azure_report_downloader.py
```diff
@@ -209,6 +214,7 @@ def __init__(self, customer_name, credentials, data_source, report_name=None, in
             self.export_name = export_report.get("name")
             self.container_name = export_report.get("container")
             self.directory = export_report.get("directory")
+            self.type = export_report.get("type")
```
maskarb: If `describe_cost_management_exports` did the validation, then we would not need to set `self.type`.

koku/masu/external/downloader/azure/azure_service.py
```diff
@@ -232,6 +232,7 @@ def describe_cost_management_exports(self):
                     "name": report.name,
                     "container": report.delivery_info.destination.container,
                     "directory": report.delivery_info.destination.root_folder_path,
+                    "type": report.definition.type,
```
maskarb: We can write a helper func:
```
def check_report_type(report_type, context)
    if report_type not in SUPPORTED_REPORT_TYPES:
        msg = f"Unsupported Azure report type: '{report_type}'. Supported types are: {SUPPORTED_REPORT_TYPES}"
        LOG.warning(log_json(msg=msg, context=context))
        raise AzureInvalidCostReport(msg)
```
And define a new `AzureInvalidCostReport` exception type.

Then use this helper here. Wherever we call `describe_cost_management_exports`, we will need to handle this new exception type.

koku/masu/external/downloader/azure/azure_report_downloader.py
bacciotti: This "metadata" comes from here: https://github.com/project-koku/koku/blob/main/koku/masu/external/downloader/azure/azure_service.py#L229. The type should come with the other report data, such as the name.

dev/scripts/test_customer.yaml
```diff
@@ -46,6 +46,7 @@ customer:
           resource_group:
             directory: ''
             export_name: 'azure-report'
+            type: 'Usage' # invalid report type
```
maskarb: This PR looks good so far. All the changes where a `type` was added to the `resource_group` still need to be reverted tho.

koku/masu/external/downloader/azure/azure_service.py
```diff
@@ -47,13 +55,15 @@ def __init__(
...
         export_name=None,
+        context=None,
     ):
         """Establish connection information."""
         self._resource_group_name = resource_group_name
         self._storage_account_name = storage_account_name
         self._factory = AzureClientFactory(
             subscription_id, tenant_id, client_id, client_secret, cloud, scope, export_name
         )
+        self.context = context or {}
```
maskarb: Where does this context come from?

koku/masu/external/downloader/azure/azure_service.py
```diff
@@ -63,6 +73,13 @@ def __init__(
         if not self._factory.credentials:
             raise AzureServiceError("Azure Service credentials are not configured.")
 
+    def _check_report_type(self, report_type):
+        """Helper function to validate the report type."""
+        if report_type not in SUPPORTED_REPORT_TYPES:
+            msg = f"Unsupported Azure report type: '{report_type}'. " f"Supported types are: {SUPPORTED_REPORT_TYPES}"
```
maskarb:
```suggestion
            msg = f"Unsupported Azure report type: '{report_type}'. Supported types are: {SUPPORTED_REPORT_TYPES}"
```

maskarb:
This PR really needs to be tested with a real Azure source. Have you done that?

koku/masu/external/downloader/azure/azure_service.py
bacciotti: I created the context attribute. Its not directly related to the pr/task itself, but I think its an improvement. I can create a separate task to add it to the rest of the logs later. Does that make sense to you, or do you think it's better to leave it as it was originally? ty

bacciotti:
@sourcery-ai review

PR #5735, opened 2025-09-08
[COST-4903] Update SQL files to cover new Persistent Disks SKU names

bacciotti:
## Jira Ticket

[COST-4903](https://issues.redhat.com/browse/COST-4903)

## Description

This change updates the GCP SQL files to match the new Persistent Disk SKU names used in the storage queries. According to the GCP docs [1], these names include the terms "PD" and/or "Snapshot", so the queries were updated accordingly.

To make the filter more robust, the search for "PD " and "Snapshot" SKUs was restricted to occur only within the Compute Engine service. This safety measure prevents accidentally capturing costs from other services that might have similarly named SKUs in the future.

## Release Notes
- [ ] Update SQL files to cover new Persistent Disks SKU names

## Reference
[1] Google Docs
1. https://cloud.google.com/skus/sku-groups/on-demand-persistent-disk-hdd?hl=en
2. https://cloud.google.com/skus/sku-groups/persistent-disk-ssd-storage?hl=en
3. https://cloud.google.com/skus/sku-groups/on-demand-persistent-disk-ssd?hl=en
4. https://cloud.google.com/skus/sku-groups/on-demand-balanced-persistent-disk?hl=en
5. https://cloud.google.com/skus/sku-groups/on-demand-persistent-disk-hdd-capacity?hl=en
6. https://cloud.google.com/skus/sku-groups/on-demand-persistent-disk-ssd-in-india?hl=en
7. https://cloud.google.com/skus/sku-groups/regional-on-demand-persistent-disk-ssd-doha?hl=en

## Summary by Sourcery

Bug Fixes:
- Include service_alias patterns for persistent disks (pd) and snapshots in GCP storage summary queries to cover missing UI data gaps

```markdown
* [COST-4903](https://issues.redhat.com/browse/COST-4903)Update SQL files to cover new Persistent Disks SKU names
```

## Summary by Sourcery

Bug Fixes:
- Include 'Compute Engine' service_alias rows with sku_alias containing 'pd ' or 'snapshot' in GCP storage summary queries to capture persistent disk data


koku/masu/database/sql/gcp/reporting_gcp_storage_summary_by_project_p.sql
```diff
@@ -39,7 +39,18 @@ INSERT INTO {{schema | sqlsafe}}.reporting_gcp_storage_summary_by_project_p (
...
         AND source_uuid = {{source_uuid}}
-        AND service_alias IN ('Filestore', 'Storage', 'Cloud Storage', 'Data Transfer')
+        AND (
+            service_alias IN ('Filestore', 'Storage', 'Cloud Storage', 'Data Transfer')
+            OR
+            (
+                -- Gets Persistent Disk rows safely
+                service_alias = 'Compute Engine' AND
+                (
+                    LOWER(sku_alias) LIKE '%%pd %%' OR
```
myersCody: Postgresql has an `ILIKE` operator that does pattern matching string comparisons, while LIKE does a case-sensitive match. It would allow you to remove the LOWER(). Additionally, do we want to include a space before the PD as well? `'%% pd %%`? 
```
postgres=# select count(*) from reporting_gcpcostentrylineitem_daily_summary where sku_alias ILIKE '% pD%';
 count
-------
    77
(1 row)
```

bacciotti: Good catch! I updated the PG files (and left the Trino one as it was, since it doesn't support ILIKE afaik...). as for the space, that works too.

PR #5740, opened 2025-09-11
[COST-6666] Add settings dict in resource-types endpoint 

bacciotti:
## Jira Ticket

[COST-6666](https://issues.redhat.com/browse/COST-6666)

## Description

This update introduces a new field, `settings`, to the `/api/cost-management/v1/resource-types/` endpoint. It returns `1`. It also changes the `gcp.projects` key to `gcp.project`.

## Testing

1. Checkout Branch
2. Restart Koku
3. Hit endpoint `/api/cost-management/v1/resource-types/`
    1. You should see
 ```
        {
            "value": "settings",
            "available": 1,
        }
```

## Release Notes
- [ ] Add settings dict in resource-types endpoint 

```markdown
* [COST-6666](https://issues.redhat.com/browse/COST-6666) Add settings dict in resource-types endpoint 
```

## Summary by Sourcery

Add a settings entry to the /api/cost-management/v1/resource-types endpoint that reports the number of integrations (or 0 if none) and correct the naming of the GCP project resource type

New Features:
- Expose a new 'settings' resource type in the resource-types endpoint with a count of integrations

Enhancements:
- Rename the gcp.project resource-type value from 'gcp.projects' to 'gcp.project'


koku/api/resource_types/view.py
```diff
@@ -96,6 +99,11 @@ def get(self, request, **kwargs):
                 "path": "/api/cost-management/v1/resource-types/cost-models/",
                 "count": cost_model_count,
             }
+            settings_dict = {
+                "value": "settings",
+                "path": None,
+                "count": settings_count,
```
bacciotti: We could change the "count" key to something like "available" if we want. I'm using "count" just to keep it consistent with the rest of the code.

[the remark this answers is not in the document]
bacciotti: yeah probably we could drop this "path" entry

koku/api/resource_types/view.py
```diff
@@ -51,6 +52,8 @@ def get(self, request, **kwargs):
 
             cost_model_count = CostModel.objects.count()
 
+            settings_count = TenantAPIProvider.objects.count()
```
maskarb: Why are we using the number of providers as the count for `settings`?

bacciotti: The idea (suggested in the Jira issue) is to make the API endpoint return "settings" if there's at least one integration. We could use 'available' instead of 'count' if we want, I just wanted to keep the pattern.

koku/api/resource_types/view.py
```diff
@@ -96,6 +96,10 @@ def get(self, request, **kwargs):
                 "path": "/api/cost-management/v1/resource-types/cost-models/",
                 "count": cost_model_count,
             }
+            settings_dict = {
+                "value": "settings",
+                "available": 1,
```
lcouzens, line 101: Whoops I missed this. I think this needs to be count no?? Because the logic reading this api is probably doing a `.get("count")` so we need this consistent. Additionally should we have the settings path here for consistency too?

PR #5747, opened 2025-09-16
[COST-6701] Add a data check before start calculating cost model costs

bacciotti:
## Jira Ticket

[COST-6701](https://issues.redhat.com/browse/COST-6701)

## Description

This change will add a data check before start calculating cost model costs.

## Testing

1. Checkout Branch
2. Restart Koku
3. Ingest local data while watching the logs and the database. 
4. You should see the message `Skipping cost model update for source {provider_uuid}. No data has been processed yet` if there is no data yet. You can add a breakpoint [here](https://github.com/project-koku/koku/blob/b70270f00db58cc3665a84caddf00548ba008231/koku/masu/processor/tasks.py#L885) to make the analysis easier and/or clear `data_updated_timestamp` column in `api_provider` table if you want.

## Release Notes
- [ ] Add a data check before start calculating cost model costs

```markdown
* [COST-6701](https://issues.redhat.com/browse/COST-6701) Add a data check before start calculating cost model costs
```

### DIFF
--- a/koku/masu/processor/tasks.py
+++ b/koku/masu/processor/tasks.py
@@ -882,6 +882,23 @@ def update_cost_model_costs(  # noqa: C901
         None
 
     """
+    # Check if provider exists and has processed data before attempting to lock the task
+    try:
+        provider = Provider.objects.get(uuid=provider_uuid)
+    except Provider.DoesNotExist:
+        LOG.warning(f"Provider with uuid {provider_uuid} not found. Skipping cost model update.")
+        return
+
+    if not provider.data_updated_timestamp:
+        LOG.info(
+            log_json(
+                tracing_id,
+                msg=f"Skipping cost model update for source {provider_uuid}. No data has been processed yet.",
+                context={"schema": schema_name},
+            )
+        )
+        return
+
     task_name = "masu.processor.tasks.update_cost_model_costs"
     cache_args = [schema_name, provider_uuid, start_date, end_date]
     if not synchronous:


koku/masu/processor/tasks.py
```diff
@@ -882,6 +882,23 @@ def update_cost_model_costs(  # noqa: C901
...
 
     """
+    # Check if provider exists and has processed data before attempting to lock the task
+    try:
+        provider = Provider.objects.get(uuid=provider_uuid)
+    except Provider.DoesNotExist:
+        LOG.warning(f"Provider with uuid {provider_uuid} not found. Skipping cost model update.")
+        return
+
+    if not provider.data_updated_timestamp:
```
maskarb: We can avoid the try except by doing:

```suggestion
    provider = Provider.objects.filter(uuid=provider_uuid).first()
    if provider and not provider.data_updated_timestamp:
```

koku/masu/processor/tasks.py
```diff
@@ -882,6 +882,23 @@ def update_cost_model_costs(  # noqa: C901
...
+        LOG.info(
+            log_json(
+                tracing_id,
+                msg=f"Skipping cost model update for source {provider_uuid}. No data has been processed yet.",
+                context={"schema": schema_name},
```
maskarb: We are very inconsistent with logging around our code, but a best practice is to write a `msg` without an f-string, and provide whatever extra context (like, `provider_uuid`) in the `context` dict.

```suggestion
                msg="Skipping cost model update. No data has been processed yet.",
                context={"schema": schema_name, "provider_uuid": provider_uuid},
```

maskarb (dismissed):

bacciotti:
@sourcery-ui review

koku/masu/test/external/downloader/gcp/test_gcp_report_downloader.py
```diff
@@ -281,13 +281,20 @@ def test_scan_start_setup_complete(self, provider):
...
         downloader = self.downloader
-        if self.gcp_provider.setup_complete:
-            self.gcp_provider.setup_complete = False
         months_delta = Config.INITIAL_INGEST_NUM_MONTHS - 1
         expected_scan_start = self.dh.today.date() - relativedelta(months=months_delta)
         expected_scan_start = expected_scan_start.replace(day=1)
         self.assertEqual(downloader.scan_start, expected_scan_start)
+        provider.setup_complete = original_setup_complete
+        provider.data_updated_timestamp = original_timestamp
+        provider.save()
```
maskarb: I'm so confused about what is going on in this test. Why are we resetting these values here? Have you tried running the test suite without doing that?

Django tests rely on resetting the db to the previous state between every test. If it didn't do that, we would need to ensure we reset every change a test made which is basically impossible.

koku/masu/test/processor/test_tasks.py
```diff
@@ -208,7 +211,8 @@ def test_process_file_initial_ingest(self, mock_stats, mock_processor):
         mock_stats.set_completed_datetime.assert_called()
         self.aws_provider.refresh_from_db()
         self.assertTrue(self.aws_provider.setup_complete)
-        shutil.rmtree(report_dir)
+        if os.path.exists(report_dir):
+            shutil.rmtree(report_dir)
```
maskarb, line 215: Why is this necessary?

koku/masu/test/external/downloader/aws/test_aws_report_downloader.py
```diff
@@ -733,8 +733,10 @@ def test_create_daily_archives_check_leading_zeros(self, mock_copy):
 
         os.remove(temp_path)
 
-    def test_create_daily_archives_dates_out_of_range(self):
+    @patch("masu.util.aws.common.get_s3_objects_not_matching_metadata")
+    def test_create_daily_archives_dates_out_of_range(self, mock_get_s3_objects):
         """Test that we correctly create daily archive files."""
+        mock_get_s3_objects.return_value = []
```
maskarb, line 739: A simpler solution to fixing this test is to flip the `setup_complete` to False. This test is relying on this pathway:
https://github.com/project-koku/koku/blob/65fedd11b84e4e90677acc1efe2a1b0efb3b8c31/koku/masu/external/downloader/aws/aws_report_downloader.py#L84

koku/masu/test/external/downloader/aws/test_aws_report_downloader.py
```diff
@@ -744,14 +746,22 @@ def test_create_daily_archives_dates_out_of_range(self):
...
         start_date = self.dh.this_month_start.replace(year=2023, month=9).date()
+        context = {
+            "account": "account",
+            "provider_uuid": self.aws_provider_uuid,
+            "s3_filename": file_name,
+            "provider_type": Provider.PROVIDER_AWS,
+        }
         daily_file_names, date_range = create_daily_archives(
-            "trace_id", "account", self.aws_provider_uuid, temp_path, file_name, manifest_id, start_date, None
+            "trace_id", "account", self.aws_provider_uuid, temp_path, file_name, manifest_id, start_date, context
```
maskarb, line 756: You can also just pass in an empty dict for the context.

maskarb (commented):
The more I think about this issue, the more I'm convinced that the `CostModelCostUpdater` should handle this. We can add the check here instead:
https://github.com/project-koku/koku/blob/f1c06c9e71237ed80aaeb21225ae921a8e18f8b7/koku/masu/processor/cost_model_cost_updater.py#L61
```
        if not self._provider.setup_complete:
            LOG.debug("Provider setup is not complete, skipping cost model cost updater")
            return None
```
(You'll need to import and set up the logger in this file, or just not log anything)

Having the check in the task is fine, but really, the updater should be handling it since, ultimately, it is responsible for triggering the cost updates.

PR #5750, opened 2025-09-17
[COST-6759] Fix the exact and partial filters combination bug

bacciotti:
## Jira Ticket

[COST-6759](https://issues.redhat.com/browse/COST-6759)

## Description

This change will fix the bug where combining partial and exact filters for the same field resulted in an incorrect `AND` condition, introducing a handling block that identifies when both filter types are used together. It correctly combines these conditions using a logical `OR`.

## Testing

1. Checkout Branch
2. Restart Koku
3. Add `LOG.info(repr(composed_filters))` at the end of the `_get_search_filter` method in `koku/api/report/queries.py`
4. Ingest some data with `ENV_FOR_DYNACONF=local iqe tests plugin cost_management -k "test_api_ocp_ingest_source_static_nodes" -vv --trace`
5. Observe the resulting query structure for the following test cases:

#### 1. Two exact filters
URL: `?group_by[node]=*&filter[exact:node]=pag-tests-1&filter[exact:node]=pag-tests-2`

Expected Q Object:
`<Q: (AND: ('usage_start__gte', ...), (OR: ('node__exact', 'pag-tests-1'), ('node__exact', 'pag-tests-2')))>`

#### 2. partial + exact filters (Bug Fix!):
URL: `?group_by[node]=*&filter[node]=pag-tests-1&filter[exact:node]=pag-tests-2`

Expected Q Object:
`<Q: (AND: ('usage_start__gte', ...), (OR: ('node__icontains', 'pag-tests-1'), ('node__exact', 'pag-tests-2')))>`

#### 3. Overlapping partial + exact filters:
URL: `?group_by[node]=*&filter[node]=pag-tests-1&filter[exact:node]=pag-tests-11`

Expected Q Object:
`<Q: (AND: ('usage_start__gte', ...), (OR: ('node__icontains', 'pag-tests-1'), ('node__exact', 'pag-tests-11')))>`

## Release Notes
- [ ] Fix the exact and partial filters combination bug

```markdown
* [COST-6759](https://issues.redhat.com/browse/COST-6759) Fix the exact and partial filters combination bug
```

## Summary by Sourcery

Fix the combination of partial and exact filters by applying OR logic for supported fields and remove the incorrect AND of exact filters in the default filter composition.

New Features:
- Add special handling in _get_search_filter to merge partial and exact filters for specified fields using OR logic
- Introduce FILTERS_WITH_EXACT_SUPPORT constant to list fields eligible for partial+exact OR combination

Enhancements:
- Remove exact operator filters from the default AND/OR filter composition to avoid improper AND logic
- Integrate special_q_objects into the composed filters to include the new OR-based filter groups

Tests:
- Add unit test to verify that exact-only filters generate QueryFilter calls with the exact operation


koku/api/report/constants.py
```diff
@@ -33,3 +33,28 @@
 TIME_SCOPE_UNITS_DAILY = "day"
 RESOLUTION_MONTHLY = "monthly"
 RESOLUTION_DAILY = "daily"
+# Defines the set of query parameters that are eligible for the special
+# OR-logic that combines a partial match
+# (e.g., `filter[node]=...`) with an exact match (`filter[exact:node]=...`).
+FILTERS_WITH_EXACT_SUPPORT = {
```
myersCody: I am not quite sure why this list was needed, but I also have concerns about us maintaining the list in a constants file. This feels like something that should be maintain through a serializer. Why let them use the exact if it is unsupported for certain filters?

bacciotti: I added the validation in the serializer like you suggested (nice catch!). I ended up using two small lists (`unsupported_exact_filters` and `excluded_fields`) to keep it simple and non-invasive. This fixes the bug and doesnt break any existing tests. Its not super generic yet, but it does the job without touching deeper logic. Im happy to switch to something cleaner but for now I couldnt think of a better way that wouldnt break things. 🙂🫣

koku/api/report/queries.py
```diff
@@ -299,13 +298,25 @@ def _check_for_operator_specific_exclusions(self, composed_filters):
         # Tag exclusion filters are added to the self.query_filter. COST-3199
         and_composed_filters = self._set_operator_specified_filters("and", True)
         or_composed_filters = self._set_operator_specified_filters("or", True)
-        exact_composed_filters = self._set_operator_specified_filters("exact", True)
```
myersCody, line 302: This is not the only endpoint where we utilize this. Do we need to do something similar for organizations, and tags endpoints?

koku/api/report/queries.py
```diff
@@ -335,6 +346,44 @@ def _get_search_filter(self, filters):  # noqa C901
             access = self.parameters.get_access(q_param, list())
             group_by = self.parameters.get_group_by(q_param, list())
             exclude_ = self.parameters.get_exclude(q_param, list())
+            partial_list = self.parameters.get_filter(q_param, list())
```
myersCody, line 415: I would like to talk to you more about this tomorrow. I am just curious on your thoughts about if we should be handling the exact logic as its own separate method instead of placing it inside of the search filter logic.

koku/api/report/queries.py
bacciotti: afaik orgs and tags dontt need this fix because they already handle "exact:" filters correctly through the existing _set_operator_specified_filters("exact") method.

koku/api/report/queries.py
```diff
@@ -299,13 +298,25 @@ def _check_for_operator_specific_exclusions(self, composed_filters):
...
-        exact_composed_filters = self._set_operator_specified_filters("exact", True)
         if composed_filters:
-            composed_filters = composed_filters & and_composed_filters & or_composed_filters & exact_composed_filters
+            composed_filters = composed_filters & and_composed_filters & or_composed_filters
         else:
-            composed_filters = and_composed_filters & or_composed_filters & exact_composed_filters
+            composed_filters = and_composed_filters & or_composed_filters
         return composed_filters
 
+    def _is_simple_text_filter(self, filt_config):
```
myersCody: `_is_icontains_supported`

koku/api/report/queries.py
```diff
@@ -335,6 +346,44 @@ def _get_search_filter(self, filters):  # noqa C901
...
             exclude_ = self.parameters.get_exclude(q_param, list())
+            partial_list = self.parameters.get_filter(q_param, list())
+            exact_list = self.parameters.get_filter(f"exact:{q_param}", list())
+
+            # Fixes the 'partial' + 'exact' filter bug by joining them with OR instead of AND.
+            # The 'continue' prevents duplicate processing.
+            # Exclude fields that have special handling or complex business logic
+            excluded_fields = ["org_unit_id", "infrastructure"]
+            if self._is_simple_text_filter(filt) and (partial_list or exact_list) and q_param not in excluded_fields:
+                or_collection = QueryFilterCollection()
```
myersCody: I would add everything under this if as a separate method

myersCody (approved):

PR #5968, opened 2026-03-26
[FLPATH-3423]: Fix sources PATCH pause/resume (on-prem)

ELK4N4:
## Summary
Fixes [FLPATH-3423](https://redhat.atlassian.net/browse/FLPATH-3423): PATCH `/api/cost-management/v1/sources/{uuid}/` with `{"paused": true|false}` from the on-prem sources UI no longer returns 400/500.

## Changes
- **AdminSourcesSerializer**: Add `paused` to fields; run create-only logic in `validate()` only when `not self.instance` (source_type requirement, source_id/offset/source_uuid); keep `source_type_id` / `source_ref` validation when present on PATCH; sync `Provider.paused` in `update()`.
- **SourcesViewSet**: Call `super().create/update/list/retrieve(request, *args, **kwargs)` so DRF receives `partial=True` on PATCH (fixes mistaken `request=request, args=args, kwargs=kwargs` delegation).
- **Tests**: Serializer coverage for partial pause/update and `source_ref` on update; `SourcesViewPatchPauseTests` with `@override_settings(ONPREM=True)`.

## Testing
- `pipenv run tox -- sources.test`
- Manual: PATCH pause/resume + GET verify `paused` reflects on linked provider.


koku/sources/test/api/test_serializers.py
```diff
@@ -317,3 +369,30 @@ def test_validate_source_ref_ocp(self):
...
+            instance=ocp,
+            data={"source_ref": "new-cluster-id"},
+            partial=True,
+            context={"request": mock_request},
+        )
+        self.assertTrue(serializer.is_valid(raise_exception=True))
+        self.assertEqual(
+            serializer.validated_data["authentication"]["credentials"]["cluster_id"],
+            "new-cluster-id",
+        )
```
ELK4N4, line 366: @bacciotti comment:
suggestion: consider adding a test for PATCH pause when the source has no Provider (provider_id is None). The update() method has a guard for this (if paused is not None and instance.provider_id), but there's no test afaik. It would be a straightforward addition :)

ELK4N4: For on-prem, sources are guaranteed to have providers since we create both together via the POST endpoint. I'm considering removing this guard clause to improve code readability

djnakabaale (commented):
looks good overall! 

optional: i'd consider adding an OCP source to the PATCH test alongside the Azure one, since OCP pause/unpause is the primary scenario.

djnakabaale (dismissed):

djnakabaale (dismissed):

myersCody:
We skipped on prem testing due to regressions

djnakabaale (approved):

PR #5971, opened 2026-03-27
[COST-573] Configurable data retention period

jordigilh:
## Summary

Implements configurable data retention for Cost Management on-prem, replacing the static `RETAIN_NUM_MONTHS` environment variable with a database-persisted `TenantSettings` model and a new Global Settings API endpoint.

- **Model**: `TenantSettings` with `data_retention_months` (IntegerField, CHECK [3, 120], default 3)
- **API**: `GET/PUT /api/v1/account-settings/data-retention/` (on-prem only, RBAC-gated)
- **Helper**: `get_data_retention_months(schema_name)` — priority: env var > DB row > Config default (4)
- **Pipeline**: `ExpiredDataRemover` uses `relativedelta` for calendar-month precision; per-tenant retention passthrough in Celery task; skips purge on DB error (R7)
- **Kafka gate**: `extract_payload` reads per-tenant retention for OCP ingest filtering
- **Materialized views**: `materialized_view_month_start` accepts optional `schema_name` for per-tenant bounds

### Design documents

Full technical design reviewed and approved by tech lead in PR #5958:

| Document | Description |
|----------|-------------|
| [README.md](docs/architecture/data-retention/README.md) | Architecture overview, decisions D1-D6, resolved questions IQ-0 through IQ-4 |
| [data-model.md](docs/architecture/data-retention/data-model.md) | TenantSettings schema, helper priority chain, migration strategy (Approach A) |
| [api.md](docs/architecture/data-retention/api.md) | Endpoint spec, serializer, permissions, validation rules |
| [retention-pipeline.md](docs/architecture/data-retention/retention-pipeline.md) | Purge flow, calendar-month fix, Kafka gate, blast radius analysis |
| [phased-delivery.md](docs/architecture/data-retention/phased-delivery.md) | Phase plan, risk register R1-R10, rollback strategy |
| [test-plan.md](docs/architecture/data-retention/test-plan.md) | IEEE 829 test plan — 3 tiers, 73 tests, traceability matrix |

### Implementation (TDD — 5 cycles)

| Cycle | Scope | Tests |
|-------|-------|-------|
| 1 | `TenantSettings` model + migration + validators | 10 |
| 2 | `GlobalSettingsView` + serializer + URL routing + RBAC | 30 |
| 3 | `get_data_retention_months` helper (priority chain, R7 error handling) | 11 |
| 4 | Calendar-month expiration (`relativedelta`), per-tenant task passthrough | 8 |
| 5 | Kafka gate + `materialized_view_month_start` + behavioral Kafka tests | 6 |
| Review | RBAC test gap fix + Kafka behavioral tests | 3 |
| **Total** | | **73** (all green) |

### Test tiers

| Tier | Scope | Tests | Status |
|------|-------|-------|--------|
| Unit | Model validation, serializer logic | 20 | Pass |
| Integration | API views + DB, helper, pipeline consumers | 45 | Pass |
| Contract | RBAC enforcement, env-var overrides | 8 | Pass |

### Key design decisions

- **D3**: `RETAIN_NUM_MONTHS` env var overrides DB value and blocks PUT (403)
- **D4**: `relativedelta(months=N)` replaces `months * timedelta(days=30)` approximation
- **D6**: Backend logic is environment-agnostic — `ONPREM` only gates URL registration
- **R7**: On DB read failure, helper returns `None`; purge skips tenant (no data loss)

### Known limitations / follow-ups

- ~~**G4**: `materialized_view_month_start` callers in `api/report/serializers.py` don't pass `schema_name` yet~~ — **Resolved**: `_schema_name` property added to `ParamSerializer`; `validate_start_date`, `validate_end_date`, and `validate_order_by` now thread `schema_name` into `materialized_view_month_start`
- **IQ-1**: `MASU_RETAIN_NUM_MONTHS_LINE_ITEM_ONLY` is dead code — removal deferred (separate cleanup)
- ~~`int(RETAIN_NUM_MONTHS)` edge case: non-numeric env var value raises unhandled `ValueError`~~ — **Resolved**: `get_data_retention_months` now catches `ValueError`/`TypeError` and returns `None` (consistent with DB error handling)

## Test plan

- [x] Tier 1 (Unit): 20/20 passed
- [x] Tier 2 (Integration): 45/45 passed
- [x] Tier 3 (Contract): 8/8 passed
- [x] Full regression suite: 73/73 passed, zero regressions
- [x] Pre-PR review: design-to-implementation cross-reference complete

**Jira**: [COST-573](https://redhat.atlassian.net/browse/COST-573)
**Design PR**: #5958


koku/api/urls.py
```diff
@@ -129,6 +130,11 @@
     path("exchange-rates/", get_exchange_rates, name="exchange-rates"),
     path("cost-type/", UserCostTypeSettings.as_view(), name="cost-type"),
     path("account-settings/", AccountSettings.as_view(), name="account-settings"),
+    *(
```
myersCody: *Minor NitPick:*
At the bottom of this file we already have logic set up to optionally include a url:
```
if settings.ONPREM:
    urlpatterns += [
        path("source_types", SourceTypesView.as_view(), name="source-types"),
        path("application_types", ApplicationTypesView.as_view(), name="application-types"),
        path("applications", ApplicationsView.as_view(), name="applications"),
    ]
```
I prefer this pattern because it makes it easier for us to determine which urls are only support for ONPREM.

koku/api/utils.py
```diff
@@ -510,9 +510,21 @@ def get_year_month_list_from_start_end(self, start, end):
         return [{"year": date.strftime("%Y"), "month": date.strftime("%m")} for date in dates]
 
 
-def materialized_view_month_start(dh=DateHelper()):
-    """Datetime of midnight on the first of the month where materialized summary starts."""
-    return dh.this_month_start - relativedelta(months=settings.RETAIN_NUM_MONTHS - 1)
+def materialized_view_month_start(dh=DateHelper(), schema_name=None):
```
myersCody, line 513: I can tell this function was extended to support per-tenant retention, but the serializers were not updated to pass in the schema. All API validation still uses the global `settings.RETAIN_NUM_MONTHS` for all tenants. A tenant within 12-month retention configured via the API will have their data retained correctly by the purge functionality, but the API serializers will still reject queries for dates outside of the global 4-month window. 

I can't tell if this is by design, where you want to limit the amount of data that can selected at once or if a piece of the puzzle was missed. Please provide some clarity on your end.

koku/api/urls.py
jordigilh: Good catch — agreed. I'll move the `data-retention` path into the existing `if settings.ONPREM` block at the bottom of the file to keep all on-prem-only URLs grouped together.

koku/api/utils.py
jordigilh: You're right — this is a gap, not by design. The serializers should honor per-tenant retention so that a tenant with 12-month retention can actually query all 12 months via the API. I'll thread `schema_name` from the request context through `validate_start_date`, `validate_end_date`, and `validate_order` into `materialized_view_month_start`. Will fix in the next push.

koku/reporting/tenant_settings/models.py
```diff
@@ -0,0 +1,27 @@
...
+    class Meta:
+        db_table = "tenant_settings"
+
+    data_retention_months = models.IntegerField(
+        default=DEFAULT_RETENTION_MONTHS,
+        validators=[
+            MinValueValidator(MIN_RETENTION_MONTHS),
+            MaxValueValidator(MAX_RETENTION_MONTHS),
+        ],
+    )
```
jordigilh, line 28: **Question for review**: The design calls for a single row per tenant, but there's no DB-level unique constraint enforcing it. `get_or_create` with `select_for_update` in the view mitigates concurrent writes, but:

- The read path (`TenantSettings.objects.first()`) returns an arbitrary row if duplicates exist (no ordering guarantee).
- Direct DB manipulation (migration scripts, Django shell) can create duplicates silently.

Should we add a `UniqueConstraint` (e.g. a sentinel boolean column with `unique=True`) or a `max_num` check? Alternatively, at minimum we could use `.order_by("-id").first()` on the read path to deterministically pick the latest row.

koku/api/settings/views.py
```diff
@@ -104,6 +110,49 @@ def put(self, request, **kwargs):
...
+        new_value = serializer.validated_data["data_retention_months"]
+        schema = request.user.customer.schema_name
+        with schema_context(schema), transaction.atomic():
+            settings_row, created = TenantSettings.objects.select_for_update().get_or_create(
+                defaults={"data_retention_months": new_value},
+            )
+            if not created:
+                settings_row.data_retention_months = new_value
+                settings_row.full_clean()
+                settings_row.save()
```
jordigilh, line 152: **Observation**: When `created=True`, Django's `get_or_create` bypasses `full_clean()` — it goes straight to `INSERT`. Model validators (MinValueValidator/MaxValueValidator) are only enforced via `full_clean()`, which is only called on the `not created` (update) path here.

Currently safe because `TenantSettingsSerializer` validates the same bounds (3–120) before this code runs, so the serializer acts as the gate. But if the serializer and model bounds ever diverge, the create path would skip model validation entirely.

Two options:
1. Add `settings_row.full_clean()` after the `get_or_create` on the create path too (belt-and-suspenders).
2. Accept the serializer as the sole validator and add a comment noting the intentional asymmetry.

Either way, wanted to flag it for awareness.

koku/api/settings/utils.py
```diff
@@ -241,3 +247,31 @@ def set_cost_type(schema, cost_type_code=KOKU_DEFAULT_COST_TYPE):
...
+
+    Priority: env var > DB > Config default.
+    Returns None on DB read failure (caller should skip purge).
+    """
+    env_val = os.environ.get("RETAIN_NUM_MONTHS")
+    if env_val is not None:
+        try:
+            return int(env_val)
+        except (ValueError, TypeError):
+            LOG.error("RETAIN_NUM_MONTHS env var is not a valid integer: %r", env_val)
```
jordigilh, line 263: **Self-review note — ValueError guard on env var path**

Added `try/except (ValueError, TypeError)` around `int(env_val)` so that a non-numeric `RETAIN_NUM_MONTHS` (e.g. `"abc"`) returns `None` instead of crashing every caller (view, purge task, Kafka gate, materialized views). This makes the env var path consistent with the DB error path which already returns `None` + logs.

Previously this was listed as a "Known limitation" in the PR description — now resolved.

koku/api/settings/utils.py
```diff
@@ -241,3 +247,31 @@ def set_cost_type(schema, cost_type_code=KOKU_DEFAULT_COST_TYPE):
...
+            row = TenantSettings.objects.first()
+            if row:
+                return row.data_retention_months
+    except Exception:
+        LOG.error(
+            "Failed to read tenant_settings for %s; skipping purge for this tenant",
+            schema_name,
+            exc_info=True,
+        )
+        return None
```
jordigilh, line 276: **Self-review note — traceback on DB error**

Added `exc_info=True` to the `LOG.error(...)` call. Previously the exception was swallowed with just a message string — operators would see "Failed to read tenant_settings for org123" but no traceback, connection error, or SQL detail. This one-word change gives full diagnostic context when investigating production DB issues.

koku/api/urls.py
```diff
@@ -562,6 +563,17 @@
...
+    for _i, _p in enumerate(urlpatterns):
+        if getattr(_p, "name", None) == "get-account-setting":
+            urlpatterns.insert(
+                _i, path("account-settings/data-retention/", GlobalSettingsView.as_view(), name="data-retention")
+            )
+            break
+    else:
+        import logging as _logging
+
+        _logging.getLogger(__name__).warning("get-account-setting URL not found; data-retention route not registered")
```
jordigilh, line 576: **Self-review note — defensive guard on URL registration**

Added a `for/else` clause so that if `get-account-setting` is ever renamed or removed, we get an explicit warning in the logs instead of the `data-retention` route silently never being registered. Without this, the `break` never fires and the loop finishes with no effect — a subtle failure mode that would only surface as 404s in production.

koku/api/urls.py
```diff
@@ -562,6 +563,17 @@
     path("settings/", SunsetView, name="settings"),
 ]
 if settings.ONPREM:
+    # data-retention must precede the <str:setting> catch-all to avoid shadowing
+    for _i, _p in enumerate(urlpatterns):
+        if getattr(_p, "name", None) == "get-account-setting":
+            urlpatterns.insert(
+                _i, path("account-settings/data-retention/", GlobalSettingsView.as_view(), name="data-retention")
+            )
```
jordigilh, line 571: **Self-review note — why insert, not append**

Per your review, we moved the `data-retention` route into the `if settings.ONPREM` block at the bottom. However, a straight `urlpatterns += [...]` caused the `account-settings/<str:setting>/` catch-all (line 133) to shadow our route — Django matches top-down, so `data-retention` was captured as `setting="data-retention"` → 404 from `AccountSettings`.

The fix uses `enumerate` + `insert` to place `data-retention` immediately before the catch-all, keeping both the grouping you requested and correct URL resolution order. The previous push had the `append` bug and its fix as two separate commits — squashed those into this single commit to avoid a broken intermediate state in git history.

koku/api/report/serializers.py
```diff
@@ -336,6 +336,13 @@ class ParamSerializer(BaseSerializer):
 
     _tagkey_support = True
 
+    @property
+    def _schema_name(self):
+        request = self.context.get("request")
+        if request and hasattr(request, "user") and hasattr(request.user, "customer"):
+            return request.user.customer.schema_name
+        return None
```
jordigilh, line 348: **Self-review note — closing the G4 gap**

Per your review comment, added `_schema_name` property to `ParamSerializer` that extracts the tenant schema from the request context. This is now threaded through `validate_start_date`, `validate_end_date`, and `validate_order_by` into `materialized_view_month_start(dh, self._schema_name)`.

Without this, a tenant with 12-month retention could not query months 5–12 via the API because date validation was using the global `settings.RETAIN_NUM_MONTHS` (4 months). The property returns `None` when no request context is available, which safely falls back to the global default.

koku/api/settings/views.py
myersCody: Agreed — let's go with Option 1 (belt-and-suspenders).

koku/reporting/tenant_settings/models.py
myersCody: So, the design doc you put together has a call out about this in the `phased-delivery.md` R2. When I initially read over it, I felt that the surface is narrow (just the API) so .first() is manageable today. I did overlook the need for the order_by, so that is a good call out. But I agree that a DB-level constraint removes the cognitive overhead entirely and prevents future surprises.

jordigilh: Implemented in ba34a88 — added `singleton = BooleanField(default=True, unique=True)` to `TenantSettings` with migration 0346. This enforces exactly one row per tenant schema at the DB level, so the `get_or_create` + `.first()` pattern is now backed by a hard constraint.

koku/api/settings/views.py
jordigilh: Done in ba34a88 — moved `full_clean()` outside the `if not created` branch so it runs on both create and update paths. Belt-and-suspenders with the serializer validation.

myersCody (dismissed):

myersCody (dismissed):

myersCody (approved):

PR #5980, opened 2026-04-02
COST-7249: Align cost breakdown design with COST-575 PriceList infrastructure

jordigilh:
## Summary

- **Align Phase 1 with COST-575**: Phase 1 now builds on top of the existing `price_list` table and `PriceListCostModelMap` from COST-575 ([#5963](https://github.com/project-koku/koku/pull/5963), [#5957](https://github.com/project-koku/koku/pull/5957)). Migration numbers renumbered (M1-M5). References tech lead's [migration coordination gist](https://gist.github.com/myersCody/91195a531ada9c09d39a5c5234993654).
- **R20: Unleash write-freeze flag**: Add `cost-management.backend.disable-cost-model-writes` to block API writes during M2 and M5 migration windows. Helper function in `masu/processor/__init__.py`, gating in `CostModelSerializer`.
- **Critical review fixes (v4.2)**: Move write-freeze gating from `CostModelManager` to `CostModelSerializer` (manager lacks schema context), add `@transaction.atomic` to `create()` pseudocode, fix risk range, add `Decimal()` conversion to M2 pseudocode.
- **Open concerns document**: [`open-concerns.md`](docs/architecture/cost-breakdown/open-concerns.md) with 5 code-grounded concerns for tech lead review (see below).

## Open Concerns for Review

The following concerns were identified during a critical review. Each is grounded in code evidence and includes a proposed action. See [open-concerns.md](docs/architecture/cost-breakdown/open-concerns.md) for full details.

1. **Distribution SQL tests verify orchestration, not correctness** — Tests mock `_execute_raw_sql_query` and assert it was called; no test runs actual distribution SQL against data or checks math. R18 mitigation may be weaker than stated.
2. **No observability instrumentation for the cost model pipeline** — `ocp_cost_model_cost_updater.py` has zero timing. Only metric is `charge_update_attempts_count` (attempts, not duration). Phase 2 replaces the core cost path with no production metrics to monitor.
3. **`_sync_rate_table()` delete-all and `RatesToUsage.rate_id` FK** — Every PUT triggers delete-all + recreate of Rate rows, NULLing `rate_id` on all historical `RatesToUsage` rows via `SET_NULL`. Not a bug, but undocumented tradeoff.
4. **On-prem M2 migration runs without write protection** — `MockUnleashClient.is_enabled()` returns `False`; M2 pseudocode lacks `select_for_update()` (unlike COST-575's 0011 migration). On-prem has no defense against concurrent writes during migration.
5. **No defined fallback trigger from IQ-9 Option 1 to Option 2** — Option 2 is documented as "viable fallback" but no decision gate defines when to switch. Team could spend sprints debugging Option 1 without a pre-agreed trigger.

## Test plan

- [ ] All design documents render correctly in GitHub markdown
- [ ] Cross-document references and links are valid
- [ ] Open concerns are grounded in code evidence (file paths, method names, code snippets verified against `main`)
- [ ] Tech lead reviews and comments on open-concerns.md

### DIFF
--- a/docs/architecture/cost-breakdown/open-concerns.md
+++ b/docs/architecture/cost-breakdown/open-concerns.md
@@ -1,0 +1,359 @@
+# Concerns Register — Cost Breakdown Design Review
+
+This document captures concerns identified during a critical review of the
+cost breakdown technical design ([COST-7249](https://redhat.atlassian.net/browse/COST-7249)).
+Each concern is grounded in code evidence, not supposition.
+
+All five concerns are now resolved. Concerns 1, 2, 3, and 4 were
+self-resolved by folding fixes directly into the design specs.
+Concern 5 was resolved by the tech lead (Option 1 is non-negotiable;
+remove Option 2 fallback language). Concern 3's six implementation
+gaps (A-F) have been closed with conservative defaults documented in
+[api-and-frontend.md](./api-and-frontend.md).
+
+---
+
+## Design Strengths
+
+Before diving into concerns, the design has several notable strengths
+worth acknowledging:
+
+- **Grounded in source code triage.** All 4 open questions (OQ-1 through
+  OQ-4) and 9 implementation questions (IQ-1 through IQ-9) were resolved
+  by reading koku source, not by assumption.
+- **Reuses COST-575 infrastructure.** Phase 1 builds on the existing
+  `price_list` table and `PriceListCostModelMap` junction table rather
+  than creating parallel tables.
+- **Every phase has a realistic rollback.** Phases 1-4 can be reverted
+  with code changes + data cleanup. Phase 5's rollback honestly requires
+  a backup.
+- **Risk register with concrete decision rationales.** 20 risks (R1-R20)
+  with alternatives-evaluated tables, benchmarks with acceptance criteria,
+  and a phase matrix.
+
+---
+
+## Self-Resolved Concerns (Concern 3 and Gaps A-F)
+
+### Concern 3: `_sync_rate_table()` Delete-All Pattern and `RatesToUsage.rate_id` FK — RESOLVED
+
+**Status**: Self-resolved. All 6 implementation gaps (A-F) closed with
+conservative defaults documented in
+[api-and-frontend.md](./api-and-frontend.md#diff-based-sync-design-decisions).
+No tech lead escalation required — decisions use safe defaults that can
+be refined later without rework.
+
+#### Tech Lead Response Summary
+
+The tech lead proposed:
+1. Replace delete-all + recreate with **diff-based sync** (update
+   existing rates by `rate_id`, create new rates, delete only removed
+   rates) — preserves `Rate.uuid` for unchanged rates.
+2. Include `rate_id` in **API payloads** so the frontend can round-trip
+   stable Rate UUIDs.
+3. Change `RatesToUsage.rate` FK from **`SET_NULL` to `CASCADE`**.
+4. **Scoped recalculation** — only recalculate the changed rate's costs.
+
+#### Blast Radius Re-Triage
+
+##### Critical finding: `rate_id` is never populated in the SQL pipeline
+
+The PoC SQL `insert_usage_rates_to_usage.sql` INSERT column list omits
+`rate_id` entirely. The `custom_name` column serves as the denormalized
+per-rate identifier. This means:
+
+- Under the current design, `RatesToUsage.rate_id = NULL` for **all** rows
+- `CASCADE` vs `SET_NULL` on the `rate` FK is **moot** — NULL FKs don't cascade
+- The FK constraint is decorative until the SQL pipeline populates it
+
+**Why this gap exists**: Under the original delete-all + recreate
+pattern, `Rate.uuid` values were **ephemeral** — every cost model API
+update deleted all Rate rows and created new ones with new UUIDs.
+Populating `rate_id` in `RatesToUsage` would have been pointless because
+the referenced UUIDs would go stale on the next update. The `custom_name`
+denormalization was the intentional per-rate identifier. The `rate` FK
+on the model was future-proofing, not a functional dependency.
+
+The tech lead's diff-based sync changes this: `Rate.uuid` values are now
+**stable** across updates (unchanged rates keep their UUID). This makes
+`rate_id` in `RatesToUsage` a durable, meaningful link — but the SQL
+pipeline hasn't been updated to reflect this architectural pivot.
+
+**Adopted mitigation**: Populate `rate_id` starting in **Phase 2**,
+when the first SQL INSERT into `rates_to_usage` ships. The Python
+orchestration already knows which rate it's processing — it passes
+`custom_name` as a SQL parameter and can pass `Rate.uuid` alongside
+it. Each SQL file adds one column to its INSERT list. Phase 3's 25
+files follow the same pattern. This unlocks CASCADE, scoped
+recalculation, and any future JOIN from `RatesToUsage` back to `Rate`.
+
+For `CASCADE` to have real effect, **every SQL INSERT into
+`rates_to_usage` must populate `rate_id`** — this is now scoped for
+Phase 2 onward. `SET_NULL` stays on the FK until `rate_id` is
+populated; CASCADE switch deferred until Phase 2 benchmarks confirm
+performance on large partitions.
+
+##### Layer 1: Diff-based `_sync_rate_table()` (Phase 1)
+
+| What changes | Where | Impact |
+|---|---|---|
+| `_sync_rate_table()` rewrite | `cost_model_manager.py` | Match by `rate_id` (if present) or `custom_name`, update/create/delete individually |
+| `RateSerializer` adds `rate_id` | `serializers.py` | Optional UUID, read-write. GET includes it; PUT may include it. |
+| Frontend `RateRequest` adds `rate_id` | `koku-ui` `api/rates.ts` | Coordinated frontend change. Backward-compatible: omitting degrades to match-by-`custom_name`. |
+| API contract documented | `api-and-frontend.md` | New field in rate objects |
+| Phase 1 validation expanded | `phased-delivery.md` | Test: update cost model → unchanged rates keep their UUID |
+
+**Backward compatibility**: If the frontend doesn't send `rate_id`, the
+backend diffs by `custom_name` (using the existing `unique_together`
+constraint). Rename = delete + create rather than update. Full
+diff-based sync activates once the frontend ships `rate_id` support.
+
+##### Layer 2: `rate_id` in SQL pipeline (currently unscoped)
+
+For the `rate` FK to be meaningful, every SQL INSERT into `rates_to_usage`
+must populate `rate_id`. This requires either:
+- Passing `rate_id` as a SQL parameter (Python looks up `Rate.uuid` by
+  `(price_list_id, custom_name)` before running SQL), or
+- Adding a JOIN to `cost_model_rate` in the SQL itself
+
+**Scope impact**: 1 PoC SQL + 1 Phase 2 SQL + 25 Phase 3 SQL files +
+5 Phase 4 distribution SQL files + 6 accessor methods = **38 file
+modifications** not currently scoped in any phase.
+
+Markup rows: `rate_id = NULL` (correct — markup has no Rate row).
+
+##### Layer 3: `SET_NULL` → `CASCADE` (only meaningful after Layer 2)
+
+| Change | Impact |
+|---|---|
+| Django model FK and M3 DDL | `on_delete=CASCADE` |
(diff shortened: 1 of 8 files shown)


docs/architecture/cost-breakdown/open-concerns.md
```diff
@@ -0,0 +1,203 @@
...
+```python
+def _sync_rate_table(self, price_list, rates_data):
+    price_list.rate_rows.all().delete()
+    Rate.objects.bulk_create([...])
+```
+
+The `RatesToUsage` model defines the FK as:
+
+```python
+rate = models.ForeignKey("cost_models.Rate", on_delete=models.SET_NULL, null=True)
```
myersCody: # Alternative Approach: Per-Rate Granularity
The delete-and-recreate pattern feels like a relic from our old flow. When all cost model costs were summed into a single row, it made sense to clear everything and reprocess. With this feature, we're adding rate-level granularity, and I think our architecture should reflect that.

## Current Behavior
Deleting all Rate rows and recreating them on every cost model update which sets rate_id to NULL on all RatesToUsage rows. This approach requires a full recalculation of all rates. This would be expensive for minor changes. For example, updating a single rate from .3 to .4 recalculates everything. This feels like a lot of unnecessary churn, especially if a customer is making multiple small edits in a row.

## Proposed Behavior
Diff incoming rates against existing and only act on what changed:

1. Rate Unchanged: No action, no recalculation
2. Rate Row Changed: Update rate row in place (preserves UUID), recalculate only that rate
3. Rate deleted: Delete Rate row, CASCADE delete cleans up for us
4. Rate added: Create Rate row, recalculate only that rate

### How It Works
API change: Include rate_id in the rate payload for the API. Then handle the different states

- GET response: Each rate includes its rate_id (UUID)
- PUT request: Client sends rate_id for existing rates, omits it for new rates

Sync logic: The rate_id tells us intent:
- Has rate_id that exists in DB → Update this rate
- Has rate_id that doesn't exist → Error (invalid ID)
- No rate_id → Create new rate
- Existing rate not in payload → Delete this rate

Recalculation: Instead of recalculating all rates, pass a list of affected rate_ids to the Celery task. The task filters its SQL to only process those rates.

#### Pros
- CASCADE becomes safe — Deleting one rate only affects that rate's RatesToUsage rows
- Minimal database churn — Only affected rates are touched
- Efficient for minor edits — Changing one rate doesn't trigger full recalculation
- Explicit intent — rate_id makes it clear whether we're updating or creating

#### Cons
- More complex sync logic — Diffing instead of simple delete-all + recreate
- API update required — Need to return and accept rate_id in rate payloads
- SQL/Celery changes — Task needs to support filtering by rate ID(s)
- Change detection — Need to define which fields constitute a "change" requiring recalculation

### Backward Compatibility
If a client doesn't send rate_id (legacy behavior):
- All incoming rates are treated as new
- All existing rates are treated as deleted
- Behaves like delete-all + recreate (no optimization, but no breakage)

----- 
@jordigilh Let me know your thoughts; however, I think this approach allows us to reduce unnecessary processing and turns the rate_id & cost_model_id into CASCADE DELETES which removes the problem.

docs/architecture/cost-breakdown/open-concerns.md
```diff
@@ -0,0 +1,203 @@
...
+   rows?
+2. If yes: should we switch from delete-all + recreate to an UPSERT
+   pattern (match on `price_list` + `custom_name`) that preserves
+   `Rate` UUIDs for unchanged rates?
+3. If no: we will document this as an intentional design choice in
+   `data-model.md` and proceed with the current pattern.
+
+---
+
+### Concern 5: No Defined Fallback Trigger from IQ-9 Option 1 to Option 2
```
myersCody: # Resolution: Back-allocation is not needed as a fallback.

With the per-rate granularity approach, rate_id is tracked throughout the pipeline. Distribution doesn't need to reverse-engineer anything — it just maintains the rate_id when moving costs:

RatesToUsage already has per-rate rows for "Platform" and "Worker unallocated" namespaces (each row has a rate_id)
Distribution reads per-rate costs from the source namespace
Distribution writes proportional costs to target namespaces with the same rate_id
The rate_id flows through naturally. Back-allocation (Option 2) was designed for an architecture where distribution operated on aggregated costs and we'd need to reconstruct per-rate detail afterward. That's not the architecture we're building.

Decision: Remove Option 2 as a documented fallback. If distribution hits issues, we debug and fix — the rate_id tracking is fundamental to the design, not an optional optimization.

This eliminates the need for a fallback trigger.

docs/architecture/cost-breakdown/open-concerns.md
jordigilh: Adopted. The updated `api-and-frontend.md` now has the full diff-based `_sync_rate_table()` pseudocode with the refinements we identified during due diligence:

- **Backward compat**: matches by `custom_name` when `rate_id` is absent (preserves UUIDs, avoids CASCADE deleting historical `RatesToUsage` data if the frontend hasn't shipped `rate_id` support yet).
- **Ownership validation**: `rate_id` is checked against the current `price_list`'s rates, not just existence.
- **Operation ordering**: delete → update → create to avoid transient `UniqueConstraint` violations on `(price_list, custom_name)`.
- **Change detection**: `COST_AFFECTING_FIELDS = {default_rate, tag_values, metric, cost_type}` — metadata-only changes (`description`, `custom_name`) update the Rate row but skip Celery recalculation.
- **Scoped recalculation**: documented as a deferred optimization. Phase 1 uses full-window recalc (correct, just less efficient). The monolithic pipeline in `ocp_cost_model_cost_updater.py` needs a non-trivial refactor for per-rate filtering.
- **`rate_id` in SQL pipeline**: scoped for Phase 2 onward. `SET_NULL` stays on the FK until `rate_id` is populated; CASCADE switch deferred until benchmarks confirm performance on large partitions.

`rate_id` is declared as an explicit `UUIDField` on `RateSerializer` with `to_representation()` updated to emit it. The `_sync_rate_table()` injects `Rate.uuid` back into the JSON blob during dual-write so GET responses include it without changing the read path.

All 6 gaps (A-F) from the due diligence are now closed with conservative defaults. No open items remain on Concern 3.

docs/architecture/cost-breakdown/open-concerns.md
jordigilh: Agreed — Option 2 fallback language has been removed from active design guidance in `README.md`.

To compensate, we've strengthened the R18 integration tests from 4 to 10 assertions. Assertion #5 specifically encodes Option 2's back-allocation formula (`rate_cost / total_cost × aggregate_distributed_cost`) as an independent cross-check against the actual per-rate `RatesToUsage` distributed rows. This gives us the mathematical safety net that Option 2 would have provided at runtime, but as a test-time verification.

The full list is in `phased-delivery.md` § Phase 4 distribution integration tests and `risk-register.md` § R18 acceptance criteria.

myersCody (dismissed):

docs/architecture/cost-breakdown/phased-delivery.md
```diff
@@ -32,35 +32,50 @@ graph LR
...
-| Migration M3 | `cost_models/migrations/XXXX_migrate_json_to_rate.py` | Data migration: JSON → Rate rows |
-| `RateSerializer` update | `cost_models/serializers.py` | Add `custom_name` field (required) |
-| `CostModelSerializer` update | `cost_models/serializers.py` | Dual-write to JSON + Rate table |
-| `CostModelDBAccessor` update | `masu/database/cost_model_db_accessor.py` | Read from Rate table (dual-write preserves JSON as fallback) |
+| `Rate` model | `cost_models/models.py` | New Django model with `custom_name`, FK to existing `PriceList` |
+| Migration M1 | `cost_models/migrations/0012_create_rate.py` | DDL for `cost_model_rate` (depends on `0011`) |
+| Migration M2 | `cost_models/migrations/0013_migrate_json_to_rate.py` | Data migration: PriceList.rates JSON → Rate rows |
+| `RateSerializer` update | `cost_models/serializers.py` | Add `custom_name` field (`required=False`) |
+| `CostModelSerializer` update | `cost_models/serializers.py` | Extend dual-write: JSON + PriceList.rates + Rate table |
+| `CostModelManager` update | `cost_models/cost_model_manager.py` | Add `_sync_rate_table()` — delete + recreate Rate rows on every write |
```
bacciotti: This row still describes delete-all + recreate on every write, but `api-and-frontend.md` documents diff-based sync (delete → update → create, match by `rate_id` or `custom_name`). Is this intended or we want to update it?

docs/architecture/cost-breakdown/risk-register.md
```diff
@@ -18,19 +18,20 @@ design documents link here for details.
 | R4 | Monthly cost rates produce more rows than expected | **MITIGATED** | 3 | GROUP BY quantified (OQ-2). |
 | R5 | Cost category reclassification invalidates breakdown tree | **MITIGATED** | 4 | Existing `update_summary_tables` chain handles it (OQ-4). |
 | R6 | 25 SQL file modifications introduce regressions | Active | 3 | Per-file-per-PR + 8-point checklist. |
-| R7 | Dual-write divergence (JSON ↔ Rate table) | Active | 1-4 | Delete-all + recreate on every write. |
+| R7 | Dual-write divergence (JSON ↔ PriceList.rates ↔ Rate table) | Active | 1-4 | COST-575 already dual-writes JSON ↔ PriceList.rates. Phase 1 extends with delete-all + recreate Rate rows on every write. Three-way sync wrapped in `@transaction.atomic`. |
```
bacciotti: R7’s mitigation text still references “delete-all + recreate on every write”. The design has moved to diff-based `_sync_rate_table()` with transactional three-way sync. Could we refresh this one-liner so the risk register summary stays accurate?

bacciotti (changes requested):
Couple of general comments @jordigilh 

1. One clarity tweak I’d find helpful: the docs already acknowledge that `rate_id` won’t be meaningfully populated until the Phase 2+ SQL work lands. Could we make that even more visible in the phase overview (separate “target architecture” from “what the database actually enforces in each phase”)? That would reduce the risk of someone assuming joins/CASCADE are fully active before the INSERT paths populate `rate_id`.
2. R20 / on-prem ops: I like the split (SaaS: Unleash gate vs on-prem: maintenance window + DB-side locking), but it’s easy for that to get lost for someone running migrations outside the core dev path. Would it make sense to add a tiny “checklist” (migrate sequencing + validation + when it’s safe to resume writes)?
3. On `_sync_rate_table()` matching: when `rate_id` is absent, matching by `custom_name` preserves compatibility but makes renames ambiguous (effectively delete+create vs in-place update). As `RatesToUsage`/`Rate` relationships harden, that ambiguity becomes an operational cost. Could we document an explicit recommendation (client/UI): always round-trip `rate_id` after it appears in GET responses, and treat missing `rate_id` as legacy-only?

bacciotti:
Also: the PR Labeler is failing due to an old change. We should rebase this to pass.

docs/architecture/cost-breakdown/phased-delivery.md
jordigilh: Updated in commit 1c31fe09: CostModelManager artifact now describes diff-based sync (match by `rate_id`/`custom_name`; delete → update → create) preserving stable Rate UUIDs. Also rebased onto main per your other comment.

docs/architecture/cost-breakdown/risk-register.md
jordigilh: Updated in commit 1c31fe09: R7 one-liner now reads "diff-based `_sync_rate_table()` (match by `rate_id`/`custom_name`; delete → update → create) preserving stable Rate UUIDs. Three-way sync wrapped in `@transaction.atomic`."

jordigilh:
@bacciotti — Addressed all three points in commit aa9554e8:

**1. `rate_id` phasing visibility** — Added a "rate_id Enforcement by Phase" table to `phased-delivery.md` that maps each phase to what the DB actually enforces: `Rate` table state, `rate_id` in API, `rate_id` in SQL INSERT, FK mode, and CASCADE safety. Includes a callout box: "Until Phase 2 ships, `RatesToUsage.rate_id = NULL` for all rows. JOINs and CASCADE have no effect."

**2. R20 on-prem migration checklist** — Added an 8-step checklist to `risk-register.md` § R20: announce maintenance window → scale down API replicas → run migrations → validate Rate row counts and `custom_name` population → validate M5 column drop → scale up → resume operations.

**3. `rate_id` round-trip recommendation** — Added to `api-and-frontend.md` § Diff-based sync design decisions: clients should always round-trip `rate_id` once it appears in GET responses. Missing `rate_id` is explicitly documented as legacy-only behavior with ambiguity tradeoffs (renames become delete+create, backend cannot distinguish "rename" from "replace").

myersCody (approved):

bacciotti (approved):

PR #5990, opened 2026-04-09
[COST-7372] Add mig instance uuid & fix count

myersCody:
## Jira Ticket

[COST-7372](https://issues.redhat.com/browse/COST-7372)

## Description

This change will ...

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Release Notes
- [ ] proposed release note

```markdown
```


bacciotti (dismissed):

koku/masu/database/trino_sql/openshift/ui_summary/reporting_ocp_gpu_summary_p_usage_only.sql
```diff
@@ -29,7 +31,9 @@ SELECT uuid(),
     gpu.gpu_vendor_name,
     gpu.gpu_model_name,
     max(gpu.gpu_memory_capacity_mib) * 0.001048576 as memory_capacity_gb,
-    count(DISTINCT COALESCE(gpu.mig_instance_id, gpu.gpu_uuid)) as gpu_count,
+    -- count(DISTINCT COALESCE(gpu.mig_instance_id, gpu.gpu_uuid)) as gpu_count,
+    max(gpu.mig_instance_id) as mig_instance_id,
+    max(gpu.gpu_uuid) as physical_gpu_uuid,
```
djnakabaale: Do we need to rename `gpu_uuid` to `physical_gpu_uuid`? `gpu_uuid` already represents the parent GPU throughout the pipeline.  To me the distinction from `mig_instance_id` is already clear.

koku/masu/database/self_hosted_sql/openshift/ui_summary/reporting_ocp_gpu_summary_p_usage_only.sql
```diff
@@ -36,6 +38,8 @@ SELECT uuid_generate_v4(),
     gpu.gpu_model_name,
     max(gpu.gpu_memory_capacity_mib) * 0.001048576 as memory_capacity_gb,
     count(DISTINCT COALESCE(gpu.mig_instance_id, gpu.gpu_uuid)) as gpu_count,
+    max(gpu.mig_instance_id) as mig_instance_id,
+    max(gpu.gpu_uuid) as physical_gpu_uuid,
```
djnakabaale: Should the GROUP BY here include `gpu_uuid` and `mig_instance_id`? Currently it's `(namespace, node, vendor, model, mig_profile, interval_start)`, so if a node has multiple GPUs of the same model used by the same namespace, they'd collapse into one row and only one UUID is kept. At the API level, `Count("physical_gpu_uuid", distinct=True)` would then undercount.

The cost model path, `reporting_ocp_gpu_summary_p.sql` includes `resource_id` and `all_labels->>'mig-instance-id'` in the GROUP BY. 
Do the usage_only templates need the same treatment?

koku/api/report/ocp/provider_map.py
```diff
@@ -1099,6 +1100,7 @@ def __init__(self, provider, report_type, schema_name):
                         "annotations": {
                             "gpu_model": F("model_name"),
                             "gpu_vendor": F("vendor_name"),
+                            "mig_uuid": F("mig_instance_id"),
```
djnakabaale, line 1103: Can we use `mig_id` instead of mig_uuid? MIG instance ids are small integers, not UUIDs, and not maintained across MIG reconfiguration. I wonder if calling it `mig_uuid` would mislead API consumers into treating it as a stable unique identifier.

koku/reporting/provider/ocp/models.py
```diff
@@ -1007,7 +1007,9 @@ class Meta:
     vendor_name = models.CharField(max_length=128, null=True)
     model_name = models.CharField(max_length=128, null=True)
     memory_capacity_gb = models.DecimalField(max_digits=33, decimal_places=15, null=True)
-    gpu_count = models.IntegerField(null=True)
+    gpu_count = models.IntegerField(null=True)  # TODO: Remove this field
+    mig_instance_id = models.CharField(max_length=128, null=True)
```
djnakabaale: small niptpick: `ming_instance_id`s from the operator are small integers. `max_length=128` seems oversized, maybe something like `max_length=16` would better signal the expected value range.

djnakabaale (commented):
A few comments, mostly around naming and one question on the GROUP BY in the usage_only SQL.

koku/api/report/ocp/provider_map.py
myersCody: Will need to do this update in a follow up, Pedro already released the IQE changes to check for `mig_uuid`

koku/reporting/provider/ocp/models.py
```diff
@@ -1007,7 +1007,9 @@ class Meta:
     vendor_name = models.CharField(max_length=128, null=True)
     model_name = models.CharField(max_length=128, null=True)
     memory_capacity_gb = models.DecimalField(max_digits=33, decimal_places=15, null=True)
-    gpu_count = models.IntegerField(null=True)
+    gpu_count = models.IntegerField(null=True)  # TODO: Remove this field
```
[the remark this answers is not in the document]
myersCody: We can't because we have to update the SQL to stop using it first otherwise we could run into prod issues.

djnakabaale (approved):

PR #5992, opened 2026-04-10
[COST-7379] Mig profiles API fixes

myersCody:
## Jira Ticket

[COST-7379](https://issues.redhat.com/browse/COST-7379)

## Description

MIG profiles API & GPU summary behavior

### filter[exact:project] (and related project filters)

- The MIG profiles filter serializer did not declare project, so keys like exact:project were treated as unknown and returned `“Unsupported parameter or invalid value.”`
-  project is now on `OCPMigProfilesFilterSerializer` (with and: / or: / exact: via _opfields), matching other OCP reports and the existing provider map mapping project → namespace.

### Tag query params on MIG profiles

- The GPU report strips `tag: / and:tag: / or:tag: / exact:tag:` from filter / exclude / group_by before nested validation so the UI can send the same params without `400s`. 
- That logic was factored into `_strip_tag_keys_from_param_dict` and `OCPMigProfilesQueryParamSerializer` now uses the same `__init__ `behavior as `OCPGpuQueryParamSerializer`. 
- Tag filters are still not applied to the MIG query (same as GPU); they are only ignored for validation.

### mig_profile data quality in queries

The MIG profiles report type filters out non–MIG rows by requiring a non-null, non-empty mig_profile (via `isnull + gt ''` on text) so empty strings are not included.

### MIG compute / memory in API responses

mig_compute / mig_memory (and unit fields) are derived from mig_profile segments: numeric value from digits-only parsing, mig_compute_units / mig_memory_units from the suffix after stripping leading digits (e.g. g, gb), with PACK_DEFINITIONS in the OCP query handler shaping { value, units } for the UI.

## Tests

MIG profiles: filter[exact:project]=* with required GPU filters returns 200.
MIG profiles: required filters plus filter[tag:application]=… returns 200 (parity with GPU tag stripping).
Use this as the narrative for reviewers; adjust if your branch also includes doc-only edits under docs/architecture/mig-gpu-support.md.

### DIFF
--- a/docs/architecture/mig-gpu-support.md
+++ b/docs/architecture/mig-gpu-support.md
@@ -2,7 +2,7 @@
 
 **Purpose:** This document provides a comprehensive overview of NVIDIA Multi-Instance GPU (MIG) cost tracking and allocation in Koku, enabling accurate chargeback for partitioned GPU resources in OpenShift environments.
 
-**Last Updated:** March 4, 2026
+**Last Updated:** April 10, 2026
 
 ---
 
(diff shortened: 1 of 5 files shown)


bacciotti (approved):

myersCody:
Pass with ocp smokes: https://github.com/project-koku/koku/runs/71100727872

PR #5997, opened 2026-04-14
COST-7391: Move smokes-labeler to pull_request_target for fork support

bacciotti:
## Jira Ticket
[COST-7391](https://issues.redhat.com/browse/COST-7391)

## Description

Follow-up to [koku-test-container#84](https://github.com/project-koku/koku-test-container/pull/84).

The `smokes-labeler` job in `ci.yml` fails silently for fork PRs because the `GITHUB_TOKEN` is read-only in `pull_request` events from forks. This means fork PRs never receive `smokes-required` or `ok-to-skip-smokes` labels automatically.

This PR moves the labeler to a dedicated workflow (`.github/workflows/pr-labeler.yml`) using `pull_request_target`, which runs in the context of the base repository and has write permissions even for fork PRs.

### Key changes

- **New file:** `.github/workflows/pr-labeler.yml` — labeler using `pull_request_target`
- **Modified:** `.github/workflows/ci.yml` — removed `smokes-labeler` job and `issues: write` permission

### Security

`pull_request_target` runs the workflow code from the **base branch** (main), not from the fork. The default checkout is also the base branch. No fork code is executed — only the GitHub API is used to detect changed files and apply labels.

Also removed `continue-on-error: true` so label failures are now visible instead of silently swallowed.

### How to test

1. Merge this PR
2. Open a PR from a fork (or use an existing one from the Flight Path team)
3. The `PR Labeler` workflow should run and add `smokes-required` or `ok-to-skip-smokes`
4. Verify internal branch PRs still get labeled correctly


.github/workflows/pr-labeler.yml
```diff
@@ -0,0 +1,111 @@
...
+  contents: read
+  issues: write
+  pull-requests: write
+
+jobs:
+  smokes-labeler:
+    name: Smoke Test Label
+    runs-on: ubuntu-latest
+
+    steps:
```
djnakabaale, line 17: should the action versions here match what's already in ci.yml? `harden-runner` is v2.14.2 vs v2.17.0, `changed-files` is v47.0.1 vs v47.0.5, and `github-script` is v8.0.0 vs v9.0.0.

bacciotti: yes! good catch, thanks.

djnakabaale (approved):

PR #5998, opened 2026-04-14
Add architecture docs for Price List feature

myersCody:
Adds architecture docs for the Lifecycles of price list feature


docs/architecture/price-lists/README.md
```diff
@@ -0,0 +1,137 @@
...
+no map yet (under row lock).
+
+---
+
+## Design notes (implemented behavior)
+
+| Topic | Behavior |
+|-------|----------|
+| **Priority** | Lowest `priority` value on `PriceListCostModelMap` wins among lists whose date range contains the billing day. |
+| **`enabled`** | New attachments require `enabled=True`. Disabled lists **still** participate in `get_effective_price_list` for dates inside their window (see docstring in [`price_list_manager.py`](../../../koku/cost_models/price_list_manager.py)). |
```
bacciotti, line 132: the query in `get_effective_price_list` has no `enabled=True` filter, is this intentional? without a brief rationale here, someone might "fix" it later and silently break cost calculation. A sentence on the design intent would help..

docs/architecture/price-lists/calculation-and-resolution.md
```diff
@@ -0,0 +1,107 @@
...
+[`OCPReportDBAccessor._populate_gpu_ui_summary_table_with_usage_only`](../../../koku/masu/database/ocp_report_db_accessor.py)
+builds `CostModelDBAccessor(..., price_list_effective_on=parse_to_date(start_date))`
+to decide whether GPU monthly metrics apply — without using the `with`
+context manager (comment in file explains schema reset interaction).
+
+---
+
+## Cloud provider cost models
+
+AWS / Azure / GCP cost model updaters use `CostModelDBAccessor` for markup
```
bacciotti, line 90: I might be wrong, but I think this section implies `Price Lists` are OCP-only without ever stating it directly. Could we add a sentence clarifying whether cloud provider support is planned or intentionally out of scope?

docs/architecture/price-lists/README.md
```diff
@@ -0,0 +1,137 @@
...
+resolution (or equivalent), drop dual-write in `CostModelManager`, and
+remove or hollow out the JSON column / API surface according to migration and
+compatibility policy. Exact sequencing is a delivery decision; this doc only
+records the intent.
+
+See [api-and-lifecycle.md § Cost model rates and dual-write](./api-and-lifecycle.md#cost-model-rates-and-dual-write).
+
+---
+
+## What a price list is
```
bacciotti, line 44: Nit: `currency` is missing from the field list here, so it might be worth mentioning alongside the other fields.

docs/architecture/price-lists/README.md
```diff
@@ -0,0 +1,137 @@
...
+   create/update replaces mappings via
+   [`PriceListManager.attach_price_lists_to_cost_model`](../../../koku/cost_models/price_list_manager.py).
+3. **Dual-write sync** — cost model `rates` updates copy into the first linked
+   price list (or auto-create list + map) ([`CostModelManager`](../../../koku/cost_models/cost_model_manager.py)); see [Dual-write transition](#dual-write-transition-product-plan).
+4. **Calculate** — for each month in range, OCP cost application loads rates
+   with that month’s start date as `price_list_effective_on` ([`_load_rates`](../../../koku/masu/processor/ocp/ocp_cost_model_cost_updater.py)).
+
+---
+
+## Migrations and backfill
```
bacciotti, line 117: nit: the doc doesn't mention that `_get_or_create_price_list` uses a hardcoded `effective_start_date`. If someone reprocesses data before that date, rates silently resolve to `{}`. might be worth documenting this? so future engs know its intentional.

docs/architecture/price-lists/calculation-and-resolution.md
myersCody: I feel like this is covered in statement below: `they do not implement the same per-day price list loop as
OCP.` 

Essentially I am calling out that the cloud providers don't use the price list here.

docs/architecture/price-lists/README.md
myersCody: Not intentional just missed, sounds like a bug that needs to be fixed.

bacciotti: Opened a PR to fix this [https://github.com/project-koku/koku/pull/6020](https://github.com/project-koku/koku/pull/6020)

bacciotti (approved):

djnakabaale (approved):

PR #6005, opened 2026-04-15
[COST-7252] Implement constant currency Phase 1

ELK4N4:
## Summary

Implements Phase 1 of the constant currency feature, enabling per-month exchange rate resolution in report queries and forecasts instead of using a single daily rate.

- **New models** (`cost_models` app, tenant-scoped): `StaticExchangeRate` (user-defined rates with monthly validity), `MonthlyExchangeRate` (single source of truth per month), `EnabledCurrency` (UI dropdown control)
- **New API endpoints**: `exchange-rate-pairs/` (static rate CRUD), `settings/currency/enabled-currencies/` (list/update enablement), `settings/currency/available-currencies/` (dropdown source)
- **Pipeline changes**: `get_daily_currency_rates` Celery task now discovers currencies, creates `EnabledCurrency` records, and upserts dynamic rates into `MonthlyExchangeRate` (respecting static overrides)
- **Query handler changes**: Base, OCP, and forecast query handlers use `Subquery` annotation from `MonthlyExchangeRate` for per-month rate resolution with earliest-rate fallback
- **Report metadata**: Responses include `exchange_rates_applied` in `meta` for transparency
- **Cache invalidation**: Both static rate writes and Celery task updates invalidate tenant view caches

## Test plan

- [ ] Unit tests for `StaticExchangeRate` serializer (validation, versioning, `MonthlyExchangeRate` side effects)
- [ ] Unit tests for `StaticExchangeRateViewSet` (CRUD operations, filtering, error cases)
- [ ] Unit tests for `EnabledCurrencyView` and `AvailableCurrencyView`
- [ ] Unit tests for `MonthlyExchangeRate` and `EnabledCurrency` model constraints
- [ ] Integration test: report queries return per-month exchange rates from `MonthlyExchangeRate`
- [ ] Integration test: static rate overrides dynamic rate for overlapping months
- [ ] Migration test: seed data populates `MonthlyExchangeRate` from `ExchangeRateDictionary`
- [ ] Manual test: full on-prem dev env with cost model rates + currency conversion


koku/api/currency/models.py
```diff
@@ -6,15 +6,12 @@
...
 
-from api.currency.currencies import CURRENCIES
 from koku.type_json_transcode import TypedJSONDecoder
 from koku.type_json_transcode import TypedJSONEncoder
 
 
 class ExchangeRates(models.Model):
-    SUPPORTED_CURRENCIES = tuple((curr.get("code", "").lower(), curr.get("code")) for curr in CURRENCIES)
-
-    currency_type = models.CharField(max_length=5, choices=SUPPORTED_CURRENCIES, unique=False, blank=True)
```
ELK4N4, line 17: Why `blank` was `True` in first place?

koku/cost_models/static_exchange_rate_serializer.py
```diff
@@ -0,0 +1,168 @@
...
+        old_end = instance.end_date
+        old_base = instance.base_currency
+        old_target = instance.target_currency
+
+        for attr, value in validated_data.items():
+            setattr(instance, attr, value)
+        instance.save()
+
+        if old_base != instance.base_currency or old_target != instance.target_currency:
+            remove_static_and_backfill_dynamic(old_base, old_target, old_start, old_end)
```
bacciotti, line 139: qq: if only the validity dates change (same currency pair, different start/end),
`remove_static_and_backfill_dynamic` isnt called for the old date range. Is this the expected behavior?

Example: rate defined for 2026-01 to 2026-03, updated to 2026-01 to
2026-01. After the update, `MonthlyExchangeRate` still has STATIC rows for
2026-02 and 2026-03, so the query handler keeps applying the static rate for
those months instead of falling back to dynamic IIUC

koku/api/currency/view.py
```diff
@@ -12,26 +12,24 @@
...
-    Returns:
-        (Response): The report in a Response object
+    """Get available currencies.
 
+    Returns currencies that have been enabled by an administrator via
+    the EnabledCurrency table.  Name, symbol, and description are
+    computed at response time via babel.
     """
-    return ListPaginator(CURRENCIES, request).paginated_response
+    enabled_codes = EnabledCurrency.objects.values_list("currency_code", flat=True)
```
bacciotti, line 31: get_currency uses AllowAny, which means it can be called without an
x-rh-identity header. Before this PR the view returned a static list; now it queries `EnabledCurrency`, which is a tenant-scoped table (cost_models app). If KokuTenantMiddleware does not set a tenant schema for the request (no identity header - public schema), this query will raise `ProgrammingError: relation "enabled_currency" does not exist`. 

maybe we could consider one of: require authentication, wrap with an explicit schema_context fallback, or catch `ProgrammingError` and return an empty list with a clear log message.

bacciotti (commented):
couple of comments

koku/cost_models/static_exchange_rate_serializer.py
ELK4N4: Good catch! I added dates to the check: 
```python
        if (
            old_base != instance.base_currency
            or old_target != instance.target_currency
            or old_start != instance.start_date
            or old_end != instance.end_date
        ):
            remove_static_and_backfill_dynamic(old_base, old_target, old_start, old_end)
```
With this change it will remove the entire instance from `MonthlyExchangeRate` every time there any change in the instance.
Then it will use `upsert_static_monthly_rates` to update the rates with new records

koku/api/currency/view.py
ELK4N4: Good point. I will go with the auth option

bacciotti (dismissed):

bacciotti (dismissed):

myersCody (changes requested):
You should just combine the migrations.

koku/api/currency/currencies.py
```diff
@@ -2,143 +2,70 @@
...
+"""
+from babel.core import get_global
+from babel.numbers import get_currency_name
+from babel.numbers import get_currency_symbol
+from babel.numbers import UnknownCurrencyError
+from rest_framework import serializers
+
+from cost_models.models import EnabledCurrency
+
+_ISO_4217_CURRENCIES = get_global("all_currencies")
```
myersCody, line 22: It appears you are allowing the customer to select any currency that is in the `get_global("all_currencies")` library. However, for our dynamic rates logic pulls pulls the exchange rates from: https://open.er-api.com/v6/latest/USD

@ELK4N4 @bacciotti Has anyone done any checking to ensure the list provided by `get_global` does not include currencies that can not be found in URL we use to grab dynamic exchange rates?

ELK4N4: A user can defines a currency that doesn't exist in https://open.er-api.com/v6/latest/USD using static rates

koku/api/currency/currencies.py
```diff
@@ -2,143 +2,70 @@
...
+def get_currency_info(code):
+    """Return a dict with code, name, symbol, and description for a currency.
+
+    All metadata is resolved via babel at call time.  Falls back to the
+    code itself for currencies babel does not recognise.
+    """
+    code = code.upper()
+    try:
+        name = get_currency_name(code, locale="en_US")
+        symbol = get_currency_symbol(code, locale="en_US")
```
myersCody, line 73: The symbol and description are directly utilize by our UI. The previous approach utilize Unicode escape sequence for the more complex currency symbols. The purpose around this is because the UX supports multiple locals. 

For example: 
```
{
        "code": "CNY",
        "name": "Chinese Yuan",
        "symbol": "CN\u00a5",
        "description": "CNY (CN\u00a5) - Chinese Yuan",
},
```

Would return `CN¥` for `en_US` local; however, if your local is Chinese it just returns `¥`. 


This approach forces all customers into the `en_US` local for currency. A minor change, but also different functionality from our previous approach. Let me ping the team to see if there is any impact to this change that I don't recognize  real quick.

koku/api/currency/currencies.py
myersCody: The current process from what I can see allows user to enable a currency that can not be found in the CURRENCY_URL. Then if they try to utilize it they git a:

```
{"currency": ["No exchange rate available for <CODE>. Ask your administrator to configure static exchange rates or enable dynamic exchange rates."]}
```
The problem is that they did enable the dynamic rate, it just can't be supported. 

I think we may need some type of validation for the dynamic flow that ensure they currency code is found in whatever url they decide to use? Thoughts? 

That way they can "enable a dynamic rate" without being able to get a rate if that makes sense.

ELK4N4: I can change the error message to
```
No exchange rate available for <CODE>. Ask your administrator to configure static exchange rates
```
if `CURRENCY_URL` is defined.

In general, I think the enabled currencies capability should be decoupled from the actual available rates as long the list is ISO 4217

ELK4N4: I have checked the differences between Babel and the API.
Babel support all ISO 4217 currencies(~300) and the API support only 162.
The exchange rate API (`https://open.er-api.com/v6/latest/USD`) returns 6 currency
codes that are **not** in babel's ISO 4217 registry. These are silently skipped
during the exchange rate fetch in `masu/celery/tasks.py`.

| Code | Currency | Territory | Official Currency | Pegged To |
|------|----------|-----------|-------------------|-----------|
| FOK | Faroese Króna | Faroe Islands | DKK (Danish Krone) | 1:1 DKK |
| GGP | Guernsey Pound | Guernsey | GBP (British Pound) | 1:1 GBP |
| IMP | Isle of Man Pound | Isle of Man | GBP (British Pound) | 1:1 GBP |
| JEP | Jersey Pound | Jersey | GBP (British Pound) | 1:1 GBP |
| KID | Kiribati Dollar | Kiribati | AUD (Australian Dollar) | 1:1 AUD |
| TVD | Tuvaluan Dollar | Tuvalu | AUD (Australian Dollar) | 1:1 AUD |

koku/api/currency/currencies.py
ELK4N4: Today the UI converts from code to symbol? If that the case I can just returns the code and the UI will do the rest

koku/api/currency/currencies.py
bacciotti: I think the design is correct. decoupling "which currencies are enabled" from "which currencies have a dynamic rate" is the right call. however id say we have two things that worth addressing:
1. fix the error message: if saas customer enables a currency thats not in the er-api and tries to use it, they get told to "enable dynamic exchange rates" -> but they already did

2. would be good to, at least, warn the admin at "enable-time". Foe example: if `CURRENCY_URL` is configured and the currency being enabled isnt in `ExchangeRateDictionary`, return a warnings field in the response saying, maybe, "this currency has no dynamic rate available, you'll need to configure a static rate before users can use it."

myersCody: From the SaaS perspective they don't get to control the CURRENCY url. Therefore, there is no way for them to "fix" it to where they could provide a dynamic rate to the currencies not in the URL. It is probably worth a discussion a team level on how to handle it.

bacciotti: after the team discussion, i think we landed on two things that need to be addressed before merge:

1) `error message`: if a currency has no rate available, the current message suggests enabling dynamic exchange rates, but that won't help if the currency simply isn't in the er-api. the message needs to be clear and actionable, something like "this currency requires a static exchange rate to be configured by your administrator."

2) `static-only signal at enable-time`: if the currency being enabled isn't in the er-api, the admin should know upfront that it's static only. without a static rate configured, users will hit an error in reports. 

the dynamic fallback only makes sense for currencies that exist in the er-api. for everything else, static is the only path, and both the admin and the end user need to know that clearly.

ELK4N4: Done

koku/api/currency/currencies.py
```diff
@@ -2,143 +2,75 @@
...
+    except UnknownCurrencyError:
+        name = code
+        symbol = code
+
+    has_dynamic_rate = ExchangeRates.objects.filter(currency_type=code.lower()).exists()
```
bacciotti: suggestion: this fires one DB query per currency. since this is called in a loop, consider fetching all dynamic codes in a single query before the loop and passing them in, instead of hitting the DB per currency.

bacciotti: `has_dynamic_rate` checks `ExchangeRates`, but `get_missing_rate_warning` checks `MonthlyExchangeRate`. for a new tenant where the celery task hasn't run yet, `ExchangeRates` will have data but `MonthlyExchangeRate` will be empty so the ui could has `has_dynamic_rate: true` but the posgt returns a warning. is this the expected behavior?

ELK4N4: You are right! I thought about 3 options to fix it:

1. Change the warning text to `Dynamic rate available but not yet applied. It will be applied automatically within 24 hours.`
2. Suppress the warning when `has_dynamic_rate` is `true`
3. When POST enables a currency, immediately upsert `MonthlyExchangeRate` for that tenant from `ExchangeRateDictionary`.

What do you prefer?

bacciotti: option 3 makes the most sense to me. it fixes the inconsistency at the root ! :)

ELK4N4: I think maybe we can do it on tenant creation - when a tenant will be created it will fulfill it's `MonthlyExchangeRate` with all dynamic currencies instead of do it for every time a currency is enabled.
What do you think?

bacciotti: that would help for new tenants but i think it doesnt cover the case where an admin enables a currency after the tenant is already created... id say option 3 (post) handles both.

ELK4N4: I thought you are talking about an edge case when a tenant is created after the celery run.
Can you detail?

bacciotti: the main case is simpler than that.. imagine a tenant that already exists and has MonthlyExchangeRate populated. an admin goes to settings and enables a new currency (say NOK). at that moment, ExchangeRates has NOK (so has_dynamic_rate: true), but MonthlyExchangeRate doesn't have NOK yet for that tenant because the celery task only upserts currencies it finds in ExchangeRateDictionary on its own schedule. so the admin enables NOK, sees has_dynamic_rate: true in the response, but also gets a warning saying there's no rate available. the warning and the flag contradict each other. with option 3, the POST itself would immediately upsert NOK into MonthlyExchangeRate from ExchangeRateDictionary, making both consistent right away.

ELK4N4: The celery task also upserts the `MonthlyExchangeRate` immediately after updating the `ExchangeRateDictionary`
https://github.com/ELK4N4/koku/blob/COST-7252/constant-currency-phase1/koku/masu/celery/tasks.py#L353

bacciotti: youre right! my bad. still worth fixing but a narrower edge case.... option 3 or 1 both work, up to yuou!

ELK4N4: > I think maybe we can do it on tenant creation - when a tenant will be created it will fulfill it's MonthlyExchangeRate with all dynamic currencies instead of do it for every time a currency is enabled.

So what about this approach?

bacciotti: yes, makes sense

ELK4N4: During manual test I found the `MonthlyExchangeRate` table is already populated for new tenants through the template schema mechanism.
When a tenant is created, [Koku clones `template0` with `copy_data => true`](https://github.com/project-koku/koku/blob/17f4089e37de374fe802b969d961098d7d571905/koku/api/iam/models.py#L137-L152), which copies all table data including `MonthlyExchangeRate` rows.
Since the celery task [updates all tenants](https://github.com/ELK4N4/koku/blob/c21e1f3ac77d7bbf5177910e2fbc1efce637fcc3/koku/masu/celery/tasks.py#L312-L332)(including `template0`) the template always has the latest dynamic rates. Any new tenant cloned after the task runs inherits them automatically.

koku/cost_models/static_exchange_rate_utils.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+def upsert_static_monthly_rates(static_rate):
+    """Upsert MonthlyExchangeRate rows for each month, including the inverse direction.
+
+    The forward direction is always written unconditionally. The inverse (1/rate)
+    is written only when no explicit StaticExchangeRate defines the reverse pair
+    for that month, ensuring explicit user-defined rates always take precedence.
+    """
+    inverse_rate = Decimal(1) / static_rate.exchange_rate
```
bacciotti, line 44: `exchange_rate should` be > 0 here?

ELK4N4: Yes

ELK4N4: I added validations in the last commit

myersCody:
```
postgres=# SELECT count(DISTINCT base_currency)
FROM monthly_exchange_rate
WHERE base_currency NOT IN (
    SELECT currency_code
    FROM enabled_currency
);
 count
-------
   138
(1 row)
```
There are 138 currencies in the `monthly_exchange_rate` table that are not in the enabled currency table. Each currency must have a rate with each other currency in the table. Considering that I have 21 currencies enabled at the moment, that is 159 rows per currency. This means 138 * 159 = `21,942 rows` in that table will go unused. This seems like unnecessary bloat in my opinion. Especially considering there is no vacuuming being done on the `monthly_exchange_rate` table at the moment. 
```
SELECT count(*)
FROM monthly_exchange_rate
WHERE base_currency NOT IN (
    SELECT currency_code
    FROM enabled_currency
);
 count
-------
 21942
(1 row)
```

koku/api/currency/currencies.py
ELK4N4: Any progress about it?
I understand today the UI do the conversion from currency codes, so we can returns only the code and UI will go from there.
The thing is that the UI today only [supports](https://github.com/project-koku/koku-ui/blob/dbb08c695eae10db1c7905a46594cd66cee620d1/apps/koku-ui-hccm/src/locales/messages.ts#L1235-L1263) 22 currency conversion. So there will be a task to add the rest to the UI

ELK4N4:
@myersCody 

> 
> ```
> postgres=# SELECT count(DISTINCT base_currency)
> FROM monthly_exchange_rate
> WHERE base_currency NOT IN (
>     SELECT currency_code
>     FROM enabled_currency
> );
>  count
> -------
>    138
> (1 row)
> ```
> There are 138 currencies in the `monthly_exchange_rate` table that are not in the enabled currency table. Each currency must have a rate with each other currency in the table. Considering that I have 21 currencies enabled at the moment, that is 159 rows per currency. This means 138 * 159 = `21,942 rows` in that table will go unused. This seems like unnecessary bloat in my opinion. Especially considering there is no vacuuming being done on the `monthly_exchange_rate` table at the moment. 
> ```
> SELECT count(*)
> FROM monthly_exchange_rate
> WHERE base_currency NOT IN (
>     SELECT currency_code
>     FROM enabled_currency
> );
>  count
> -------
>  21942
> (1 row)
> ```

Done

myersCody:
@ELK4N4 Awesome work reducing the bloat! We just need to add the new tables to our vacuum process so that way we are not keeping data past the data retention period for the monthlyExchangeRate table.

ELK4N4:
@myersCody What about [autovacuum_tune_schemas](https://github.com/project-koku/koku/blob/be5cbfc6555b9e1aa6c60bc0a00197a8fa6e6d64/koku/masu/celery/tasks.py#L251)? It's not enough for vacuuming `MonthlyExchangeRate`?

myersCody:
> @myersCody What about [autovacuum_tune_schemas](https://github.com/project-koku/koku/blob/be5cbfc6555b9e1aa6c60bc0a00197a8fa6e6d64/koku/masu/celery/tasks.py#L251)? It's not enough for vacuuming `MonthlyExchangeRate`?

@ELK4N4 No it is not, That method adjusts PostgreSQL's internal autovacuum configuration parameters (like how aggressively it cleans up dead row versions left behind by updates/deletes). It doesn't delete any application data. What I am specifically talking about is the cleaning schedule we have for data that is outside of our retention period. 

You could look at the `remove_expired_data` celery task. As of right now I feel like the rows int he MonthlyExchangeRate table will persist forever even if we set our retention period to keep only 3 months.

docs/architecture/constant-currency/api-and-frontend.md
```diff
@@ -136,143 +144,104 @@ This endpoint is always available. No Unleash feature flag gating.
...
+
+Currency enablement is managed via a dedicated endpoint under
+`settings/currency/enabled-currencies/`. The enabled status for each currency
+is also visible in the `GET settings/currency/exchange_rate/` list response and
+can be toggled individually.
 
 ### URL
 
 ```
-GET/PUT   /api/cost-management/v1/settings/currency/enabled-currencies/
```
pedrolp85, line 144: We still have a GET method in the class view bellow

koku/api/settings/currency_views.py
```diff
@@ -0,0 +1,80 @@
...
+    permission_classes = [SettingsAccessPermission]
+
+    def _validate_code(self, code):
+        code = code.upper()
+        if not is_valid_iso_currency(code):
+            raise ValidationError({"code": f"Invalid ISO 4217 currency code: {code}"})
+        return code
+
+    @method_decorator(never_cache)
+    def get(self, request, *args, **kwargs):
```
pedrolp85, line 40: we have a get() method but GET is removed form allowed HTTP Methods in the docs

koku/api/settings/currency_views.py
```diff
@@ -0,0 +1,80 @@
...
+        if warning:
+            LOG.warning(log_json(msg="Currency enabled with warning", currency=code, warning=warning))
+        return Response({"warning": warning}, status=status.HTTP_200_OK)
+
+    @method_decorator(never_cache)
+    def delete(self, request, *args, **kwargs):
+        code = self._validate_code(kwargs["code"])
+        EnabledCurrency.objects.filter(currency_code=code).delete()
+        current_month = DateHelper().this_month_start.date()
+        deleted_count, _ = MonthlyExchangeRate.objects.filter(
```
pedrolp85, line 87: this seems incomptatible with

> "Enabling or disabling a currency only affects its visibility in the target currency dropdown. It does not affect the MonthlyExchangeRate."

docs/architecture/constant-currency/api-and-frontend.md
```diff
@@ -136,143 +144,104 @@ This endpoint is always available. No Unleash feature flag gating.
...
-  "currencies": [
-    { "currency_code": "GBP", "enabled": true },
-    { "currency_code": "CNY", "enabled": true }
-  ]
-}
-```
+**Permission**: `SettingsAccessPermission` — requires the **Cost Management
+Administrator** role.
 
 **Side effects**: Enabling or disabling a currency only affects its visibility
```
pedrolp85: This comment seems to incompatible with the delete method in enabled currencies view

docs/architecture/constant-currency/api-and-frontend.md
```diff
@@ -418,3 +409,7 @@ The frontend will:
 | v1.5 | 2026-04-09 | Replaced stale `MonthlyExchangeRateSnapshot` → `MonthlyExchangeRate`, removed `StaticExchangeRateDictionary` references (removed in pipeline-changes v1.6). |
 | v1.6 | 2026-04-12 | Updated `exchange_rates_applied` implementation to reflect `Subquery`-based rate resolution (removed `effective_exchange_rates` reference). |
 | v1.7 | 2026-04-13 | Removed stale "snapshotted" terminology (remnant from `MonthlyExchangeRateSnapshot` rename). |
+| v1.8 | 2026-04-28 | Consolidated endpoints under `settings/currency/exchange_rate/`. List returns grouped response with enabled status. Currency enablement via POST/DELETE at `settings/currency/enabled-currencies/{code}/` (no body). Removed separate `AllCurrencyView` and `available-currencies` endpoints. |
```
pedrolp85: I we don't have an ALLCurrencyView, how does a customer enable a currency that is not in cloud reports or enabled by Default? UI needs a source of truth for disabled currencies

ELK4N4: `settings/currency/enabled-currencies/` returns all currencies(disabled/enabled).

ELK4N4: I'm refactoring those endpoints

ELK4N4:
Restructured currency settings endpoints:

**Before:**
```
GET/POST  settings/currency/exchange_rate/
PUT/DEL   settings/currency/exchange_rate/<uuid>/
GET       settings/currency/enabled-currencies/
POST/DEL  settings/currency/enabled-currencies/<code>/
```

**After:**
```
GET       settings/currency/
POST      settings/currency/static-rates/
PUT/DEL   settings/currency/static-rates/<uuid>/
POST/DEL  settings/currency/enabled/<code>/
```

`GET settings/currency/` now returns all ISO 4217 currencies (with `?search=` and `?enabled=` filters), removing the need for a separate `enabled-currencies` list endpoint.

docs/architecture/constant-currency/api-and-frontend.md
ELK4N4: Fixed — the docs were restructured in `f964535` and `fa96925`. The `GET` is now documented on the separate `settings/currency/` endpoint (`CurrencyListView`), which correctly has a `get()` method. The `EnabledCurrencyView` at `settings/currency/enabled/{code}/` only exposes `POST` and `DELETE` — no `GET`. The URL table in the docs now reads:

```
GET             settings/currency/
POST/DELETE     settings/currency/enabled/{code}/
```

koku/api/settings/currency_views.py
ELK4N4: Fixed — same restructure as above. The `get()` method lives on `CurrencyListView` which is now documented at `GET settings/currency/`. The `EnabledCurrencyView` (POST/DELETE only) has no `get()` method and the docs reflect that.

koku/api/settings/currency_views.py
ELK4N4: Fixed in `fa96925`. The old blanket statement was removed and replaced with accurate side-effects:

> **Side effects**:
> - **POST (enable)**: Creates an `EnabledCurrency` row. Returns a warning if no exchange rate is available for the currency.
> - **DELETE (disable)**: Removes the `EnabledCurrency` row and deletes dynamic `MonthlyExchangeRate` rows for the current month where the currency appears as base or target. Does not affect `ExchangeRateDictionary`, `ExchangeRates`, or `StaticExchangeRate` tables.

docs/architecture/constant-currency/api-and-frontend.md
ELK4N4: Fixed — same doc update as the previous comment. The contradictory text was replaced with the accurate side-effects description in `fa96925`.

docs/architecture/constant-currency/api-and-frontend.md
ELK4N4: Fixed in `f964535`. The new `CurrencyListView` at `GET settings/currency/` returns **all ~170 ISO 4217 currencies** with an `enabled` flag, `has_dynamic_rate` flag, and `description`. It supports `?search=` and `?enabled=` query params. This serves as the source of truth for the UI to discover any currency (enabled or disabled) and toggle enablement via `POST/DELETE settings/currency/enabled/{code}/`.

PR #6012, opened 2026-04-20
COST-7284: Add guard to prevent redundant GPU finalization

bacciotti:
## Jira Ticket
[COST-7284](https://issues.redhat.com/browse/COST-7284)

## Description

This change adds an `EXISTS` check to `populate_distributed_cost_sql` that prevents redundant GPU finalization on the 2nd of each month. Before running the artificial GPU distribution for the previous month, the code now queries `OCPUsageLineItemDailySummary` for existing `gpu_distributed` rows. If they already exist, the distribution is skipped with an informational log message.

Previously, high-frequency customers sending hourly payloads would trigger the full GPU finalization up to 24 times on the 2nd of the month (or 72 times across a 3-day window), causing significant redundant processing.

**Architecture doc**: [GPU Finalization section in mig-gpu-support.md](https://github.com/project-koku/koku/blob/main/docs/architecture/mig-gpu-support.md#gpu-finalization)

### Changes
- **`ocp_report_db_accessor.py`**: Added an inline `OCPUsageLineItemDailySummary.objects.filter(...).exists()` check before running GPU distribution for the previous month. The check runs before `sql_params` are modified, and uses `config.cost_model_rate_type` to support any future `requires_full_month` distribution.
- **`test_ocp_report_db_accessor.py`**: Added 3 unit tests covering skip behavior, run behavior, and natural finalization bypass.

## Testing

1. Checkout branch
2. Restart Koku: `docker compose restart koku-server koku-worker`
3. Ingest OCP data with GPU for previous + current month
4. Verify `gpu_distributed` rows exist for previous month:
   ```sql
   SET search_path TO org1234567;
   SELECT COUNT(*) FROM reporting_ocpusagelineitem_daily_summary
   WHERE cost_model_rate_type = 'gpu_distributed'
     AND usage_start >= date_trunc('month', now() - interval '1 month')
     AND usage_start < date_trunc('month', now());
   ```
5. Temporarily change `dh.now_utc.day == 2` to today's day in `ocp_report_db_accessor.py`
6. Trigger cost model update:
   ```bash
   curl -s "http://localhost:5042/api/cost-management/v1/update_cost_model_costs/?provider_uuid=<UUID>&schema=org1234567"
   ```
7. Check worker logs — you should see: `Skipping gpu_unallocated distribution - previous month already finalized`
8. Delete the `gpu_distributed` rows and trigger again — this time the distribution should run normally (no "Skipping" message)
9. Revert the temporary day change

## Release Notes
- [ ] proposed release note

```markdown
* [COST-7284](https://issues.redhat.com/browse/COST-7284) Prevent redundant GPU finalization on the 2nd of each month
```


koku/masu/database/ocp_report_db_accessor.py
```diff
@@ -651,6 +660,20 @@ def populate_distributed_cost_sql(  # noqa: C901
                     if dh.now_utc.day == 2:
                         sql_params["start_date"] = summary_range.start_of_previous_month
                         sql_params["end_date"] = summary_range.end_of_previous_month
```
esebesto, line 662: do we want` sql_params["start_date"] = summary_range.start_of_previous_month ` and `sql_params["end_date"] = summary_range.end_of_previous_month` in case that the distribution is not supposed to run? i.e, will this affect anything? I feel like we had issues with something like this before, but may be it is ok now

esebesto (commented):
:)

koku/masu/database/ocp_report_db_accessor.py
```diff
@@ -190,6 +190,15 @@ def _populate_gpu_ui_summary_table_with_usage_only(self, params):
             # SaaS: execute via Trino
             self._execute_trino_multipart_sql_query(populate_gpu_usage_info, bind_params=sql_params)
 
+    def _distribution_exists_for_month(self, provider_uuid, start_date, end_date, rate_type) -> bool:
```
myersCody, line 193: I don't really see the need/benefit of creating a separate method for this.

koku/masu/database/ocp_report_db_accessor.py
bacciotti: Good catch! Moved the exists check before the `sql_params` assignments so they're only set when the distribution actually runs.

koku/masu/database/ocp_report_db_accessor.py
bacciotti: Agreed, inlined the query directly. Also removed the standalone test for the method — the behavior is covered by the skip/run tests.

myersCody (approved):

PR #6027, opened 2026-04-24
COST-7405: apply filter[limit] to MIG instances instead of profiles

bacciotti:
## Summary

`filter[limit]` on the `/mig_profiles/` endpoint was ranking by distinct MIG profiles, not by individual MIG instances. Since a node can have many instances per profile, the limit was effectively ignored — e.g. with 2 profiles and `filter[limit]=3`, both profiles (and all their instances) were returned.

This fix introduces a `rank_group_by` key in the provider map, allowing reports to rank at a finer granularity than their response `group_by`. For `mig_profiles`, ranking is now done at the `(mig_profile, mig_id)` level, so `filter[limit]=N` returns exactly N individual MIG instances.

**Changes:**
- `provider_map.py`: added `rank_group_by: ["mig_profile", "mig_id"]` to the `mig_profiles` report type
- `queries.py` (`_group_by_ranks`): reads `rank_group_by` from the provider map; explicitly annotates alias fields (e.g. `mig_id = F("mig_instance_id")`) that are not in `self.annotations`, skipping direct model fields to avoid Django's "conflicts with a field on the model" error
- `queries.py` (`_ranked_list`): accepts `rank_group_by` parameter and uses it for `groupby` and `merge` operations instead of the response-level `group_by`

## Testing

Manual verification with local data (2 profiles: `1g.10gb` with 3 instances, `4g.40gb` with 1 instance):

| `filter[limit]` | Before (main) | After (this PR) |
|---|---|---|
| 3 | 4 instances — both profiles returned | 3 instances — only `1g.10gb` |
| 1 | 4 instances — both profiles returned | 1 instance — only `1g.10gb` |
| 4 | 4 instances | 4 instances (all) |

Redis cache flushed between branch switches to ensure fresh computation.

Fixes: [COST-7405](https://redhat.atlassian.net/browse/COST-7405)

### DIFF
--- a/koku/api/report/ocp/provider_map.py
+++ b/koku/api/report/ocp/provider_map.py
@@ -1167,6 +1167,9 @@ class OCPProviderMap(ProviderMap):
                             {"field": "mig_profile", "operation": "gt", "parameter": ""},
                         ],
                         "group_by": ["mig_profile"],
+                        # filter[limit] ranks individual MIG instances (mig_id level), not profiles.
+                        # A node can have many instances per profile, so limiting by profile is not useful.
+                        "rank_group_by": ["mig_profile", "mig_id"],
                         # Do not synthesize an "Other(s)" bucket when filter[limit] is used; return top-N only.
                         "rank_limit_include_others": False,
                         "cost_units_key": "raw_currency",
--- a/koku/api/report/queries.py
+++ b/koku/api/report/queries.py
@@ -1342,14 +1342,33 @@ class ReportQueryHandler(QueryHandler):
         if self.order_field == "subscription_name":
             group_by_value.append("subscription_name")
 
-        # Do not re-annotate names that are already GROUP BY columns (often the rank key).
+        # Some reports rank at a finer granularity than their response group_by.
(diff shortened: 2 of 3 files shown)


esebesto:
Results of [ocp-smokes](https://konflux-ui.apps.stone-prd-rh01.pg1f.p1.openshiftapps.com/ns/cost-mgmt-dev-tenant/applications/koku/pipelineruns/koku-ci-6hgdq?integrationTestName=koku-ci) can be considered as a clean pass: only couple of unimportant test failures caused by test regression -> unless we change the code, we can switch this pr to hot-fix-smokes

esebesto:
Returning the ocp-smokes label now to make it obvious that it was originally tested and passed with this label

djnakabaale (commented):
changes look good to me overall. one small suggestion is adding a deterministic unittest asserting `filter[limit]=N` returns exactly N instance rows. current tests cover 200/no-Other behavior, but this would lock in the precise bug fix.

djnakabaale (approved):

PR #6028, opened 2026-04-24
COST-7422: Allow re-enabling disabled price list via PUT

bacciotti:
## Summary

- The serializer always auto-injects `currency` into the validated data on every update. When a price list is disabled, the manager blocked any field outside `{name, description, enabled}`, causing `currency` to trigger a rejection even on a minimal `PUT {"name": "...", "enabled": true}`.
- The fix skips the field restriction when the payload is re-enabling the price list (`enabled=True`). A disabled list that *stays* disabled still enforces the restriction as before.
- Made `name` optional on updates (`required=False`) since the UUID already identifies the resource — no need to repeat the name on every PUT.

## Test plan

- [ ] Existing tests for disabled price list restrictions still pass (rates/validity still blocked when staying disabled)
- [ ] New test `test_reenable_disabled_price_list_with_full_payload` covers Dan's exact scenario (full payload + `enabled=true`)
- [ ] `PUT {"enabled": true}` on a disabled price list returns 200

Fixes: COST-7422

### DIFF
--- a/koku/cost_models/price_list_manager.py
+++ b/koku/cost_models/price_list_manager.py
@@ -66,7 +66,8 @@ class PriceListManager:
         if not self._model:
             raise PriceListException("No price list to update.")
 
-        if not self._model.enabled:
+        re_enabling = data.get("enabled") is True
+        if not self._model.enabled and not re_enabling:
             allowed_fields = {"name", "description", "enabled"}
             disallowed = set(data.keys()) - allowed_fields
             if disallowed:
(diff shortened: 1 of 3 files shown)


esebesto:
The latest [pr-check](https://konflux-ui.apps.stone-prd-rh01.pg1f.p1.openshiftapps.com/ns/cost-mgmt-dev-tenant/applications/koku/pipelineruns/koku-ci-mtj57) can be considered as a clean pass (only couple of unimportant test failures due to a test regression). I just wonder if it was intional to run only ocp-on-prem-smoke-tests against SaaS deployement?

myersCody (approved):

PR #6043, opened 2026-05-05
[COST-7249] Phase 3: Migrate monthly/tag/VM cost SQL to RTU pipeline

jordigilh:
## Summary

Completes the RTU (RatesToUsage) pipeline migration for **all remaining cost types** — monthly, tag-based, and VM costs — building on the Phase 2 foundation merged in #6017.

After the initial SQL migration, a comprehensive readiness audit (FedRAMP, QE, SRE, Security, Architecture, DBA) identified several issues. This PR includes the original migration **plus all audit fixes**.

---

#### 1. Original Phase 3 SQL migration (`2fec011`)
- **25 SQL files converted** to INSERT into `rates_to_usage` instead of direct-writing to `reporting_ocpusagelineitem_daily_summary`:
  - 4 PostgreSQL tag rate files (infra/supp specific + default)
  - 5 PostgreSQL monthly cost files (node/cluster, PVC, VM, node_cost_by_tag, PVC_by_tag)
  - 8 Trino SQL files (VM hourly/monthly, VM core, project, GPU)
  - 8 self-hosted SQL files (mirrors of Trino)
- **6 accessor methods updated** in `ocp_report_db_accessor.py` to pass `cost_model_id`, `rate_uuid`, and `custom_name` to SQL templates via new `rate_info_map` property on `CostModelDBAccessor`

#### 2. Orchestration: RTU pipeline unconditional (`da8808b`)
- **All cost types now flow through RTU unconditionally.** Phase 3 converts monthly/tag/VM SQL to write to `rates_to_usage`, making the legacy direct-write path non-functional for these cost types.
- The `COST_BREAKDOWN_RTU_UNLEASH_FLAG` is still read; when OFF, a **warning log** is emitted but the RTU path runs anyway. This prevents silent data loss — if usage costs went through legacy direct-write while aggregation runs (triggered by Phase 3 costs in RTU), the aggregation DELETE would wipe the legacy rows.
- The legacy `_update_usage_costs()` method is preserved (not called) for git-revert rollback.
- **⚠️ This changes rollback semantics**: reverting Phase 3 requires a git revert, not a flag toggle. See risk register R20 update.

#### 3. SQL audit fixes (`97767b7`)
- **`metric_type` centralized**: Added `_get_routing_metric_type()` static helper to `OCPReportDBAccessor`. Replaces hardcoded `'cpu'`, `'storage'`, `'gpu'` and `CASE` expressions in 25 SQL files with `{{metric_type}}` template variable. The routing logic mirrors Phase 2's SQL-level decisions (R13 IQ-2):
  - Node/Cluster → `'cpu'` if distribution is cpu, else `'memory'`
  - PVC → `'storage'`
  - GPU → `'gpu'`
  - Tag usage → passthrough from `usage_type`
- **SHA-256 upgrade**: Replaced `md5()` with `sha256()` across all PostgreSQL, Trino, and self-hosted SQL files. `sha256(bytea)` is built-in PostgreSQL 11+ (proven by `koku/db_functions/jsonb_sha256_text.sql`, in production since 2021). No `pgcrypto` needed.
  - Migration `0350` widens `label_hash` from `VARCHAR(32)` to `VARCHAR(64)`
  - No data migration needed — delete-and-rebuild on next RTU run
  - Aligns with NIST 800-53 Rev 5 (FedRAMP) guidance discouraging md5 for new implementations
  - Risk register R13 already documented this as an available option
- **DELETE scope restored**: Added `AND rtu.pod_labels ? {{tag_key}}` / `AND rtu.volume_labels ? {{tag_key}}` to the DELETE statements in `node_cost_by_tag.sql` and `monthly_cost_persistentvolumeclaim_by_tag.sql` to prevent over-deletion across tag keys

#### 4. Risk register update (`bc59356`)
- **R20**: Documented Phase 3 realization — RTU unconditional, flag warning-only, rollback via git revert
- **R13**: Recorded md5 → sha256 upgrade with rationale and computation patterns
- **Changelog v1.3** added

#### 5. Tests (`891753e`)
- **TC-R20-01**: Updated expected orchestration order to `rtu → monthly → vm → agg → markup → dist`
- **TC-R20-03**: Asserts monthly/vm/agg/markup now called even when `cost_model_id` is None
- **TC-55**: Asserts RTU is called when flag is OFF (override behavior)
- **TestRoutingMetricType**: 12 new unit tests covering all aggregation buckets (cpu, memory, storage, gpu) for `_get_routing_metric_type()`

#### 6. Cleanup (`b726ce0`)
- `rate_info_map` on `CostModelDBAccessor` cached with `@cached_property`

---

### Key design decisions

| Decision | Rationale |
|----------|-----------|
| RTU unconditional (flag → warning) | Phase 3 SQL writes to RTU; legacy path can't coexist without maintaining dual SQL versions (rejected by TL in PR #5948) |
| SHA-256 over md5 | FedRAMP alignment; built-in PG function; risk register R13 documented this option; zero-cost migration via delete-and-rebuild |
| Centralized `metric_type` helper | Prevents drift between SQL files; mirrors Phase 2 routing logic; 12 unit tests |
| Rollback via git revert | Consistent with R18 (distribution SQL); flag toggle no longer viable |

### Alignment with Phase 2

| Area | Phase 2 | Phase 3 | Status |
|------|---------|---------|--------|
| metric_type routing | In SQL (rate table + overrides) | Python helper, same logic | ✅ Aligned |
| label_hash | md5, VARCHAR(32) | sha256, VARCHAR(64) | ✅ Upgraded (documented option) |
| DELETE scoping | By cost_model_id + period | Tag-key filter restored | ✅ Aligned |
| Test patterns | `_make_orchestration_patches` | Same decorator, extended | ✅ Aligned |
| Feature flag | Gates usage between RTU/legacy | Warning-only (RTU unconditional) | ⚠️ Evolved |

## Test plan

- [ ] CI passes (`koku-ci` pipeline)
- [ ] RTU path processes all cost types (usage, monthly, tag, VM)
- [ ] Flag OFF: warning logged, RTU path still executes
- [ ] Aggregation correctly routes `metric_type` into `cost_model_cpu_cost`, `cost_model_memory_cost`, `cost_model_volume_cost`, `cost_model_gpu_cost`
- [ ] Tag cost DELETE only removes rows for the current tag_key
- [ ] `label_hash` is 64-char hex (sha256) in `rates_to_usage`
- [ ] Migration 0350 applies cleanly
- [ ] GPU costs (allocated + unallocated) with MIG slice support preserved


jordigilh:
## FedRAMP Readiness Review — Phase 3 RTU Migration

Reviewed the full diff (29 files, 1,256+/1,116−) against NIST 800-53 Rev 5 controls. The migration pattern is solid overall — tenant isolation is preserved, SQL injection surface is reduced by removing `{{labels_field | sqlsafe}}`, and the RTU staging model is consistent with Phase 2. There are a few items that need attention before merge.

---

### Critical: Flag-OFF path silently drops monthly/tag/VM costs

**Problem.** All 25 SQL templates now INSERT into `rates_to_usage` unconditionally. The flag-OFF branch in `update_summary_cost_model_costs` calls `_update_monthly_cost`, tag cost methods, and `_update_vm_usage_costs` — all of which now write to RTU — but never calls `_aggregate_rates_to_daily_summary`. Those costs never reach the daily summary table.

**Root cause.** Phase 2 kept monthly/tag processing _outside_ the flag check (they still targeted the daily summary directly). This PR moved them _inside_ both branches and changed the SQL target, but only added aggregation to the flag-ON branch.

**Recommended fix.** Preserve the Phase 2 structural pattern: the flag gates only usage-cost routing. Monthly/tag/VM are now fully migrated and should sit outside the flag check, with aggregation always running after them:

```python
# Flag gates ONLY usage costs (Phase 2 scope):
if rtu_enabled:
    if self._cost_model_id:
        self._update_usage_rates_to_usage(...)
    else:
        self._cleanup_stale_rtu_costs(...)
else:
    self._update_usage_costs(...)

# Phase 3: always through RTU (migration complete):
self._update_monthly_cost(...)
# ... tag cost methods ...
self._update_vm_usage_costs(...)

# Aggregation always runs:
self._aggregate_rates_to_daily_summary(...)

self._update_markup_cost(...)
```

This matches the Phase 2 test `TC-R20-01` contract, avoids duplicate SQL files, and keeps the flag as a safety valve for usage costs only.

---

### High: DELETE scope broadened in tag-based cost files

`node_cost_by_tag.sql` and `monthly_cost_persistentvolumeclaim_by_tag.sql` removed the `tag_key` containment filter (`pod_labels ? {{tag_key}}` / `volume_labels ? {{tag_key}}`) from their DELETE statements. When multiple tag keys have rates, processing tag_key_A will delete rows created for tag_key_B.

**Fix:** Restore an equivalent filter on `rates_to_usage` — either the JSONB containment check or a filter on `rate_id` / `custom_name`.

---

### Medium: `metric_type` hardcoded — regression from Phase 2 pattern

Phase 2's `insert_usage_rates_to_usage.sql` reads `metric_type` from the Rate table via a `rate_names` CTE (`rn.metric_type`), populated by `derive_metric_type()` in `rate_sync.py`. Phase 3's SQL files hardcode `'cpu' AS metric_type` everywhere (including all VM cost files), bypassing the Rate table.

**Fix:** Add `metric_type` to `rate_info_map` (it's already on the Rate model), and use `{{metric_type}}` in the SQL templates instead of hardcoded literals.

---

### Medium: MD5 for `label_hash` — use SHA-256

MD5 is used across all 25 SQL files for `label_hash`. The codebase already has `public.jsonb_sha256_text` (installed via migration `0046`), and the risk register doc mentions widening to `sha256 / VARCHAR(64)`.

**Fix:** Switch to `encode(sha256(...), 'hex')` in PostgreSQL and `to_hex(sha256(...))` in Trino. Update `label_hash` column to `VARCHAR(64)`.

---

### Low / Informational

- **No new tests.** 29 files changed, 0 test files. At minimum: `rate_info_map` property, flag-ON vs flag-OFF orchestration paths, aggregation metric_type routing.
- **No audit logging** distinguishing RTU vs legacy path — add structured log entries with `cost_model_id` and pipeline path.
- **Duplicated orchestration logic** across both flag branches — extracting shared logic into a helper method would reduce maintenance drift (this becomes moot if the fix above is adopted).
- **Trino `CAST(NULL AS json)` chain** in label_hash — simplify to `CAST(NULL AS varchar)` or `''` directly.

---

### Positive

- Removing `{{labels_field | sqlsafe}}` in favor of hardcoded column names (`pod_labels`, `volume_labels`, `all_labels`) reduces SQL injection surface. Good change.
- Schema-based tenant isolation fully preserved in all RTU paths.
- Feature flag + aggregation staging model is sound once the flag-OFF path is fixed.

jordigilh:
@myersCody — heads-up on one architectural decision in this PR that needs your sign-off before merge.

### Feature flag is now warning-only (RTU unconditional)

Phase 3 converts all 25 monthly/tag/VM SQL files to INSERT into `rates_to_usage`. This makes the `COST_BREAKDOWN_RTU_UNLEASH_FLAG` unable to gate between RTU and legacy paths:

- **Why**: If the flag is OFF and usage costs go through legacy direct-write (`_update_usage_costs`), the aggregation DELETE — which now runs unconditionally because monthly/tag/VM costs are in RTU — would wipe those legacy usage rows. Maintaining two versions of every SQL file (RTU + legacy) would reintroduce the dual-path architecture you rejected in PR #5948.

- **What we did**: The RTU pipeline runs for all cost types regardless of the flag. When the flag is OFF, a `LOG.warning` is emitted for operational visibility. The legacy `_update_usage_costs()` method is preserved but not called, keeping git-revert clean.

- **Impact on rollback**: Phase 3 rollback requires a git revert of the SQL migration commits, not a flag toggle. This is consistent with R18 (distribution SQL rewrite) which also uses git revert as its rollback mechanism.

**Question**: Are you comfortable with the flag becoming warning-only, with git revert as the rollback mechanism? If you'd prefer to preserve the flag as a kill switch, the alternative is maintaining dual SQL versions for all 25 files — which we assessed as higher risk.

The full rationale is documented in the updated risk register (R20, Phase 3 section) in this PR.

---

Everything else in the audit fixes (SHA-256, metric_type centralization, DELETE scope, tests) aligns with Phase 2 patterns and documented options. See the updated PR description for a commit-by-commit breakdown.

koku/masu/processor/ocp/ocp_cost_model_cost_updater.py
```diff
@@ -765,33 +797,56 @@ def update_summary_cost_model_costs(self, summary_range: SummaryRangeConfig) ->
...
+                        report_accessor.populate_tag_based_costs(
+                            start_date,
+                            end_date,
+                            self._provider.uuid,
+                            self.metric_to_tag_params_map,
+                            cluster_params,
+                        )
+                if not (self._tag_infra_rates or self._tag_supplementary_rates):
+                    self._delete_tag_usage_costs(start_date, end_date, self._provider.uuid)
             self._update_markup_cost(start_date, end_date)
```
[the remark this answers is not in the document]
jordigilh: Resolved in commit `da8808b`. The duplicated flag branches and the broken legacy path were both addressed by making the RTU pipeline unconditional — all cost types now flow through a single code path (RTU INSERT → aggregation → daily summary). The feature flag is still read but emits a warning when OFF; the legacy `_update_usage_costs()` method is preserved for git-revert rollback. See updated PR description § "Orchestration: RTU pipeline unconditional" for full rationale.

koku/masu/database/ocp_report_db_accessor.py
```diff
@@ -1176,6 +1201,7 @@ def populate_tag_usage_default_costs(  # noqa: C901
                     for value_to_skip in value_names:
                         key_value_pair.append(json.dumps({tag_key: value_to_skip}))
                     json.dumps(key_value_pair)
+                    tag_rate_info = rate_info_map.get((metric, cost_type_label, tag_key), {})
```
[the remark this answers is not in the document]
jordigilh: Good catch. Fixed in commit `c39760b` — `populate_tag_usage_default_costs` now falls back to the metric-level `rate_info` (matching the pattern in `populate_tag_usage_costs` at line 1150):

```python
rate_info = rate_info_map.get((metric, cost_type_label, ""), {})
# ...
tag_rate_info = rate_info_map.get((metric, cost_type_label, tag_key), rate_info)
```

koku/masu/database/sql/openshift/cost_model/usage_rates/aggregate_rates_to_daily_summary.sql
```diff
@@ -148,3 +134,58 @@ GROUP BY
...
+)
+SELECT
+    uuid_generate_v4(),
+    rtu.report_period_id,
+    rtu.cluster_id,
+    rtu.cluster_alias,
+    rtu.namespace,
+    rtu.node,
+    rtu.usage_start,
+    rtu.usage_start + interval '1 day' AS usage_end,
```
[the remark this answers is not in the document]
jordigilh: This is a false positive. Both Step 2 (line 87) and Step 3 (line 158) use the same expression: `rtu.usage_start + interval '1 day' AS usage_end`. Neither uses `MIN(rtu.usage_end)`. The pattern is consistent across both INSERT blocks and matches the daily granularity convention where `usage_end = usage_start + 1 day`.

myersCody:
> @myersCody — heads-up on one architectural decision in this PR that needs your sign-off before merge.
> 
> ### Feature flag is now warning-only (RTU unconditional)
> Phase 3 converts all 25 monthly/tag/VM SQL files to INSERT into `rates_to_usage`. This makes the `COST_BREAKDOWN_RTU_UNLEASH_FLAG` unable to gate between RTU and legacy paths:
> 
> * **Why**: If the flag is OFF and usage costs go through legacy direct-write (`_update_usage_costs`), the aggregation DELETE — which now runs unconditionally because monthly/tag/VM costs are in RTU — would wipe those legacy usage rows. Maintaining two versions of every SQL file (RTU + legacy) would reintroduce the dual-path architecture you rejected in PR [[COST-7249] Technical design: Cost breakdown for OpenShift price list costs #5948](https://github.com/project-koku/koku/pull/5948).
> * **What we did**: The RTU pipeline runs for all cost types regardless of the flag. When the flag is OFF, a `LOG.warning` is emitted for operational visibility. The legacy `_update_usage_costs()` method is preserved but not called, keeping git-revert clean.
> * **Impact on rollback**: Phase 3 rollback requires a git revert of the SQL migration commits, not a flag toggle. This is consistent with R18 (distribution SQL rewrite) which also uses git revert as its rollback mechanism.
> 
> **Question**: Are you comfortable with the flag becoming warning-only, with git revert as the rollback mechanism? If you'd prefer to preserve the flag as a kill switch, the alternative is maintaining dual SQL versions for all 25 files — which we assessed as higher risk.
> 
> The full rationale is documented in the updated risk register (R20, Phase 3 section) in this PR.
> 
> Everything else in the audit fixes (SHA-256, metric_type centralization, DELETE scope, tests) aligns with Phase 2 patterns and documented options. See the updated PR description for a commit-by-commit breakdown.

Yes! So, this is part of the reason I pushed for these to be done in a separate phases is because I knew these files will not be controllable through an unleash flag. 

The hope is to get phase 2 into production before we merge phase 3. Once phase 2 is in prod with no issues, we can remove the unleash flag. I always planned for the rollback strategy for phase 3 would be to revert the commit. 

However, I did not foresee you needing another migration, which complicates the revert strategy. I think what we need to do is pull the migration out of this PR and merge it beforehand. We can extend the capacity of the label hash prior to these changes with no impact on the system. 

That allows us to push in these changes and have a clean revert if we experience problems.

koku/masu/database/self_hosted_sql/openshift/cost_model/hourly_cost_virtual_machine.sql
```diff
@@ -1,37 +1,52 @@
...
     usage_start,
     usage_end,
-    namespace,
     node,
-    max(resource_id) AS resource_id,
+    namespace,
+    cluster_id,
+    cluster_alias,
+    data_source,
+    NULL AS persistentvolumeclaim,
```
myersCody: Why add the column to the insert if it is always null?

myersCody (changes requested):

myersCody:
Github is down right now and I can't leave comments on the PR lines directly. Sorry for the annoying comments that are going to start rolling in, but I want to get you some feedback ASAP. 

*Reference:*
https://github.com/project-koku/koku/pull/6043/changes#diff-f0f05a0d5883c67612ca839a0a04e95f28cb6044612c5140be4e19bfb7563004R141-R148

*Feedback:*
This structure doesn't match the preexisting pattern for the repo on how we write our SQL files, [Example](https://github.com/project-koku/koku/pull/6043/changes#diff-e8ae434c3a8db06ab7b4b05f87fbe6ff0896fb5c3870191ef6b8a99c89953a25R10-R14)

myersCody:
```
@cached_property
    def rate_info_map(self):
        """Return (metric, cost_type, tag_key) -> {rate_uuid, custom_name} for all Rate rows."""
        if not self.cost_model:
            return {}
        rate_rows = Rate.objects.filter(price_list__cost_model_maps__cost_model=self.cost_model).only(
            "uuid", "metric", "cost_type", "custom_name", "tag_key"
        )
        return {
            (rate.metric, rate.cost_type, rate.tag_key or ""): {
                "rate_uuid": str(rate.uuid),
                "custom_name": rate.custom_name,
            }
            for rate in rate_rows
        }
```

What is the benefit of basically recreating the json structure that we removed here? I feel like we have the ability to query the Rate table directly now to see if the rate exists?

Additionally how are we handling scenarios where we have more than one price list. Are we taking validity period into consideration here?

koku/masu/database/self_hosted_sql/openshift/cost_model/hourly_cost_virtual_machine.sql
jordigilh: Valid point — `persistentvolumeclaim` is always NULL for VM costs since these rows come from `data_source = 'Pod'`. The column is nullable with a NULL default, and the aggregation JOIN handles NULLs via `IS NOT DISTINCT FROM`, so omitting it from the INSERT is safe.

Will remove it. The same pattern applies to `volume_labels` (also always NULL in this file) — will trim that too for consistency. Applies to `hourly_vm_core.sql`, `hourly_cost_vm_tag_based.sql`, and `hourly_vm_core_tag_based.sql` as well.

jordigilh:
Agreed — will split the migration out into a standalone PR. The migration (`0350_widen_ratestousage_label_hash`) just widens `label_hash` from `VARCHAR(32)` to `VARCHAR(64)` for SHA-256. It's backwards-compatible and has zero runtime impact on Phase 2's MD5 hashes.

One change from the current migration: removing the reverse-only `TRUNCATE TABLE rates_to_usage`. That was added to handle test-session edge cases where 64-char hashes existed. For the standalone migration:

- **Forward**: Widens the column — no-op for existing 32-char data
- **Reverse**: Django's `AlterField` narrows back to `VARCHAR(32)`. If 64-char data exists (from Phase 3), PostgreSQL will reject it — which is correct because Phase 3 must be reverted first. No silent data loss.

Will open the migration PR shortly and rebase Phase 3 to exclude it.

jordigilh:
> This structure doesn't match the preexisting pattern for the repo on how we write our SQL files

Agreed — the `sql/` directory Phase 3 files (tag rates, monthly costs, aggregate) use a compacted multi-column-per-line INSERT column list, while the repo's established pattern is one column per line. The `self_hosted_sql/` and `trino_sql/` files already follow the correct pattern. Will reformat all `sql/openshift/cost_model/` files to match.

jordigilh:
> What is the benefit of basically recreating the json structure that we removed here? I feel like we have the ability to query the Rate table directly now to see if the rate exists?

The intent was a cached lookup table (`@cached_property`) to avoid N+1 queries — `rate_info_map` is accessed ~15 times per update across `populate_monthly_cost`, `populate_tag_usage_costs`, `populate_tag_usage_default_costs`, and `_update_vm_usage_costs`. Each lookup resolves `rate_uuid` and `custom_name` for the SQL template parameters.

That said, you're right that it partially recreates the structure we removed. An alternative is to query Rate directly per-invocation with the indexes already on the table (`rate_metric_idx`, `rate_price_list_idx`). The tradeoff is query count vs. dict overhead. Happy to refactor to direct queries if you prefer that pattern — the per-query cost should be negligible with the indexes.

> Additionally how are we handling scenarios where we have more than one price list. Are we taking validity period into consideration here?

Valid catch — **this is a bug**. The current `rate_info_map` query fetches from all price lists:
```python
Rate.objects.filter(price_list__cost_model_maps__cost_model=self.cost_model)
```
This does not scope to the effective price list for the billing period. All other rate properties (`infrastructure_rates`, `supplementary_rates`, etc.) correctly resolve via `self.effective_rates` → `PriceListManager.get_effective_price_list(cost_model_uuid, price_list_effective_on)`, but `rate_info_map` bypasses that resolution.

Will fix by scoping the Rate query to the effective price list when `price_list_effective_on` is set, matching the pattern used by `effective_rates`.

myersCody:
> > What is the benefit of basically recreating the json structure that we removed here? I feel like we have the ability to query the Rate table directly now to see if the rate exists?
> 
> The intent was a cached lookup table (`@cached_property`) to avoid N+1 queries — `rate_info_map` is accessed ~15 times per update across `populate_monthly_cost`, `populate_tag_usage_costs`, `populate_tag_usage_default_costs`, and `_update_vm_usage_costs`. Each lookup resolves `rate_uuid` and `custom_name` for the SQL template parameters.
> 
> That said, you're right that it partially recreates the structure we removed. An alternative is to query Rate directly per-invocation with the indexes already on the table (`rate_metric_idx`, `rate_price_list_idx`). The tradeoff is query count vs. dict overhead. Happy to refactor to direct queries if you prefer that pattern — the per-query cost should be negligible with the indexes.
> 
> > Additionally how are we handling scenarios where we have more than one price list. Are we taking validity period into consideration here?
> 
> Valid catch — **this is a bug**. The current `rate_info_map` query fetches from all price lists:
> 
> ```python
> Rate.objects.filter(price_list__cost_model_maps__cost_model=self.cost_model)
> ```
> 
> This does not scope to the effective price list for the billing period. All other rate properties (`infrastructure_rates`, `supplementary_rates`, etc.) correctly resolve via `self.effective_rates` → `PriceListManager.get_effective_price_list(cost_model_uuid, price_list_effective_on)`, but `rate_info_map` bypasses that resolution.
> 
> Will fix by scoping the Rate query to the effective price list when `price_list_effective_on` is set, matching the pattern used by `effective_rates`.



> > What is the benefit of basically recreating the json structure that we removed here? I feel like we have the ability to query the Rate table directly now to see if the rate exists?
> 
> The intent was a cached lookup table (`@cached_property`) to avoid N+1 queries — `rate_info_map` is accessed ~15 times per update across `populate_monthly_cost`, `populate_tag_usage_costs`, `populate_tag_usage_default_costs`, and `_update_vm_usage_costs`. Each lookup resolves `rate_uuid` and `custom_name` for the SQL template parameters.
> 
> That said, you're right that it partially recreates the structure we removed. An alternative is to query Rate directly per-invocation with the indexes already on the table (`rate_metric_idx`, `rate_price_list_idx`). The tradeoff is query count vs. dict overhead. Happy to refactor to direct queries if you prefer that pattern — the per-query cost should be negligible with the indexes.
> 
> > Additionally how are we handling scenarios where we have more than one price list. Are we taking validity period into consideration here?
> 
> Valid catch — **this is a bug**. The current `rate_info_map` query fetches from all price lists:
> 
> ```python
> Rate.objects.filter(price_list__cost_model_maps__cost_model=self.cost_model)
> ```
> 
> This does not scope to the effective price list for the billing period. All other rate properties (`infrastructure_rates`, `supplementary_rates`, etc.) correctly resolve via `self.effective_rates` → `PriceListManager.get_effective_price_list(cost_model_uuid, price_list_effective_on)`, but `rate_info_map` bypasses that resolution.
> 
> Will fix by scoping the Rate query to the effective price list when `price_list_effective_on` is set, matching the pattern used by `effective_rates`.

@jordigilh You bring up a good point about the N+1 query. Another additional benefit of this approach is that it allows you to do the work of figuring out the validity periods of price lists in one place instead of 15 different places. I am sold on keeping it a cache property.

koku/masu/database/self_hosted_sql/openshift/cost_model/hourly_cost_virtual_machine.sql
jordigilh: Done — removed \`persistentvolumeclaim\` and \`volume_labels\` from all four VM hourly SQL files in commit \`7794ad9\`. The INSERT columns now exactly match the non-NULL values the SELECT produces.

myersCody: @jordigilh The nulls seem to be back 😓

koku/masu/database/sql/openshift/cost_model/supplementary_tag_rates.sql
```diff
@@ -69,6 +54,9 @@ FROM (
         lids.persistentvolume,
         lids.storageclass,
         lids.source_uuid,
+        lids.pod_labels,
```
myersCody: So, the previous logic here was to only keep the label that matched. Keeping the entire label dictionary is a behavior change in the code that I am uncertain we need or want. 
```
{{k_v_pair}}::jsonb as {{labels_field | sqlsafe}},
{{k_v_pair}}::jsonb as all_labels,
```

I am not sure if there would be any unintentional consequences of keeping the whole label. If we want to go this route, we should loop in the QE team to ensure there is decent test coverage for supplementary tag rates.

koku/reporting/migrations/0350_widen_ratestousage_label_hash.py
```diff
@@ -12,4 +12,8 @@ class Migration(migrations.Migration):
             name="label_hash",
             field=models.CharField(max_length=64, null=True),
         ),
+        migrations.RunSQL(
```
myersCody: What is the purpose behind this?

koku/masu/database/self_hosted_sql/openshift/cost_model/hourly_cost_virtual_machine.sql
jordigilh: Fixed in commit `aacf76b`. Removed `persistentvolumeclaim` and `volume_labels` from all 8 VM hourly SQL files (4 self-hosted + 4 Trino), and simplified the `label_hash` to use `''` instead of `COALESCE((NULL::jsonb)::text, '')`.

Root cause: commit `7794ad9` fixed this on May 6, but the fix was on a parallel commit line that didn't get integrated during the Phase 3 rebase — the rebased branch reintroduced the NULLs from the original Phase 3 migration commit.

koku/masu/database/sql/openshift/cost_model/supplementary_tag_rates.sql
jordigilh: Agreed — storing the full label dictionaries was an unintended behavior change. Fixed in commit `aacf76b`.

**Specific and default tag rate files** now restore the pre-Phase-3 label semantics:

- `supplementary_tag_rates.sql` and `infrastructure_tag_rates.sql`: store `{{k_v_pair}}::jsonb` for `pod_labels`, `volume_labels`, and `all_labels` — only the matched key-value pair, matching the original `{{k_v_pair}}::jsonb as {{labels_field}}, {{k_v_pair}}::jsonb as all_labels` pattern. Also removed full labels from the subquery SELECT and GROUP BY.

- `default_supplementary_tag_rates.sql` and `default_infrastructure_tag_rates.sql`: store `{{labels_field}}` for all three label columns — matching the original behavior where the relevant label field was kept whole (since there's no single k_v_pair to extract for default rates).

This ensures `label_hash` is computed from the same label scope as pre-Phase-3, so aggregation in `aggregate_rates_to_daily_summary.sql` produces identical grouping behavior.

koku/reporting/migrations/0350_widen_ratestousage_label_hash.py
jordigilh: This migration widens `rates_to_usage.label_hash` from `VARCHAR(32)` to `VARCHAR(64)`.

**Why:** Migration 0348 created the column at `max_length=32` (sized for MD5 hex), but the SQL files compute `label_hash` using SHA-256 which produces a 64-character hex string. Without this migration, the hash would be silently truncated to 32 chars on INSERT, causing collisions during the `aggregate_rates_to_daily_summary` GROUP BY (different label combinations mapping to the same truncated hash → incorrect cost aggregation).

**Reverse safety:** The `RunSQL(noop, reverse_sql=TRUNCATE)` ensures that if we ever roll back past this migration, the table is emptied before the column is narrowed back to 32 chars — otherwise existing 64-char values would cause `value too long for type character varying(32)` errors. The RTU table is ephemeral (rebuilt every cost model run), so truncation on rollback is safe.

jordigilh:
@myersCody — Three proactive fixes pushed from a GA readiness audit. Requesting your feedback.

### 1. Monthly VM NULL columns (commit `3ef1663`)

Same cleanup you approved for the hourly files — `persistentvolumeclaim` and `volume_labels` (always NULL for VM costs) were still present in the 4 monthly VM SQL files (`monthly_vm_core.sql`, `monthly_vm_core_tag_based.sql` in both self-hosted and Trino). Removed from INSERT/SELECT and simplified `label_hash` middle segment to `''` in the 2 self-hosted files (Trino already had this correct).

### 2. Normalize tag-rate `label_hash` (commit `947a9c7`)

`supplementary_tag_rates.sql` and `infrastructure_tag_rates.sql` were the **only** SQL files computing `label_hash` from the raw bind parameter string (`COALESCE({{k_v_pair}}, '')`). Every other file — usage rates, VM rates, default tag rates — hashes from the jsonb-normalized text representation (e.g., `lids.pod_labels::text` or `labels.combined_labels::text`).

Changed to `COALESCE(({{k_v_pair}}::jsonb)::text, '')` so the hash is computed from the same representation PostgreSQL stores. In practice, for simple key-value pairs like `{"app": "banking"}`, the hash values are identical — this makes the guarantee structural rather than coincidental.

This is also a prerequisite for Fix 3 below: the CI validation query recomputes `label_hash` from the daily summary's `pod_labels::text`, which goes through jsonb normalization. Without this fix, the validation JOIN could produce false-positive diff rows if Python's `json.dumps` formatting ever diverges from PostgreSQL's `jsonb::text` output.

### 3. Extend RTU validation to cover tag rates (commit `68dd099`)

Per your [comment on supplementary_tag_rates.sql](https://github.com/project-koku/koku/pull/6043#discussion_r2161746785) about ensuring "decent test coverage for supplementary tag rates" — extended the CI-only `validate_rates_against_daily_summary.sql` to also check tag-rate rows (`monthly_cost_type = 'Tag'`), not just usage-rate rows (`monthly_cost_type IS NULL`).

The change is a single WHERE clause expansion. The RTU subquery already aggregates all `monthly_cost_type` values and groups by it, and the JOIN matches via `COALESCE(lids.monthly_cost_type, '')` — so no structural changes were needed.

---

Would appreciate your review on these, especially the hash normalization rationale. Happy to discuss.

koku/masu/database/self_hosted_sql/openshift/cost_model/hourly_cost_virtual_machine.sql
jordigilh: Follow-up: extended the same cleanup to the **4 monthly VM files** (`monthly_vm_core.sql`, `monthly_vm_core_tag_based.sql` in both self-hosted and Trino) in commit `3ef1663`. These had the same `NULL AS persistentvolumeclaim` and `NULL::jsonb AS volume_labels` pattern that the hourly files had. All VM SQL files — hourly and monthly — are now clean.

Verified: `grep` for `NULL AS persistentvolumeclaim`, `NULL::jsonb AS volume_labels`, `CAST(NULL AS varchar) AS persistentvolumeclaim`, and `CAST(NULL AS json) AS volume_labels` returns 0 matches across all VM SQL files.

koku/masu/database/sql/openshift/cost_model/supplementary_tag_rates.sql
jordigilh: Follow-up on test coverage: two additional commits address this.

**Hash normalization** (commit `947a9c7`): The `label_hash` in these files was the only place computing the hash from a raw bind parameter string (`COALESCE({{k_v_pair}}, '')`) instead of from the jsonb-normalized text (`COALESCE(({{k_v_pair}}::jsonb)::text, '')`). Every other SQL file — usage rates, VM rates, default tag rates — already hashes from `::text` on a jsonb value. Changed to match, ensuring the hash is deterministic regardless of Python serialization choices.

**Validation coverage** (commit `68dd099`): Extended the CI-only `validate_rates_against_daily_summary.sql` to also check tag-rate rows (`monthly_cost_type = 'Tag'`), not just usage-rate rows. The hash normalization above is a prerequisite — without it, the validation JOIN could produce false-positive diff rows since it recomputes `label_hash` from the daily summary's `pod_labels::text` (jsonb-roundtripped).

This gives tag rates the same CI regression coverage that usage rates already had.

koku/reporting/migrations/0350_widen_ratestousage_label_hash.py
jordigilh: Removed, it should've not been added.

/Jordi

jordigilh:
@myersCody — Follow-up on your [migration 0350 question](https://github.com/project-koku/koku/pull/6043#discussion_r3313093424):

The `TRUNCATE` reverse-SQL we had added to `0350_widen_ratestousage_label_hash.py` has been removed. The file is now identical to what's already on `main` (merged via PR #6048). This PR introduces **zero migration changes**.

myersCody (approved):

PR #6062, opened 2026-05-14
[COST-7476] Fall back to credential-based auth when storage account key access is disabled

bacciotti:
## Jira Ticket
[COST-7476](https://redhat.atlassian.net/browse/COST-7476)

## Description

This change fixes a silent failure in `AzureClientFactory.blob_service_client` when a storage account has **"Allow storage account key access" disabled** at the storage account level.

### Root cause

The existing fallback to credential-based auth only triggers when `listKeys` **fails** (i.e. the service principal lacks `Microsoft.Storage/storageAccounts/listKeys/action`). However, if the service principal has `Storage Account Contributor`, `listKeys` **succeeds** and returns keys — but any attempt to use those keys is rejected by the storage account with `KeyBasedAuthenticationNotPermitted`. The fallback never kicks in, and the integration silently fails with a 403 during blob listing.

This means that having *more* permission (`Storage Account Contributor`) paradoxically breaks the integration when key access is disabled, because the code never reaches the token-based fallback.

### Fix

After creating the key-based `BlobServiceClient`, immediately call `get_account_information()` to validate that key-based authentication is actually permitted. If it raises `HttpResponseError` with `KeyBasedAuthenticationNotPermitted`, the code falls back to the credential-based `BlobServiceClient` — the same path already used when `listKeys` is denied.

Also replaces `httpError.message` (which can be `None`) with `str(httpError)` for safer string matching.

## Testing

1. Checkout branch
2. Run the Azure client tests:
   ```
   pipenv run tox -- providers.test.azure.test_client
   ```
   - `test_blob_service_client_key_auth_not_permitted` — new test covering the fixed scenario
   - `test_blob_service_client_none_message` — new test covering the `None` message edge case
   - All existing tests should continue to pass

3. To verify end-to-end: configure an Azure integration where the storage account has "Allow storage account key access" disabled and the service principal has `Storage Account Contributor` + `Storage Blob Data Reader`. Data should now ingest successfully without needing to remove `Storage Account Contributor`.

## Release Notes
- [ ] proposed release note

```markdown
* [COST-7476](https://redhat.atlassian.net/browse/COST-7476) Fix Azure integration failure when storage account key access is disabled but service principal has Storage Account Contributor role
```


koku/providers/azure/client.py
```diff
@@ -78,16 +79,20 @@ def blob_service_client(self, resource_group_name, storage_account_name):
...
                 f"EndpointSuffix=core.windows.net"
             )
-            return BlobServiceClient.from_connection_string(connect_str)
+            client = BlobServiceClient.from_connection_string(connect_str)
+            client.get_account_information()
+            return client
         except HttpResponseError as httpError:
-            err = "does not have authorization to perform action 'Microsoft.Storage/storageAccounts/listKeys/action'"
-            if err not in str(httpError):
+            error_msg = str(httpError)
```
lcouzens, line 86: I feel like we obfuscated these errors before because they would occasionally leak sensitive information into our logs. I could be miss remembering here but we should also check or add in some logic that can recognise sensitive content and obfuscate it.

koku/providers/azure/client.py
```diff
@@ -78,16 +79,20 @@ def blob_service_client(self, resource_group_name, storage_account_name):
                 f"AccountKey={key.value};"
                 f"EndpointSuffix=core.windows.net"
             )
-            return BlobServiceClient.from_connection_string(connect_str)
+            client = BlobServiceClient.from_connection_string(connect_str)
+            client.get_account_information()
```
lcouzens, line 83: Just for my understanding, this is the real line that fixes thingss correct? Essentially this check would cause a error to be raised for the customer and enable the flow to fall back to account access keys?

lcouzens (commented):
a couple of clarifying questions

koku/providers/azure/client.py
bacciotti: i didnt find any obfuscation pattern... Also the original code already logged `httpError.message` in the same warning. For these errors we check here the messages don't contain secrets, just ids and metadata.

koku/providers/azure/client.py
bacciotti: correct, without it, the error only surfaces later during blob listing, outside where the fallback can trigger

lcouzens (approved):

PR #6097, opened 2026-06-04
[COST-7252] Refactor currency list to ISO 4217 and add EnabledCurrency

ELK4N4:
## Summary

Replace the hardcoded 22-currency list with a dynamic, database-driven currency management system backed by babel's ISO 4217 registry.

Previously, supported currencies were a static tuple in `currencies.py` -- adding or removing currencies required a code change. Now, the new `EnabledCurrency` tenant-scoped model controls which currencies are available, while babel provides name/symbol/description at runtime.

**Key changes:**
- **`CurrencyField` serializer field** -- normalizes to uppercase, validates against ISO 4217 via babel, optionally checks the enabled list
- **Removed all static currency constants** (`CURRENCIES`, `VALID_CURRENCIES`, `CURRENCY_CHOICES`) -- serializers, views, filters, and Celery tasks now resolve dynamically
- **`User.is_authenticated` property** -- added to Koku's custom User model so DRF's `IsAuthenticated` permission works on the new endpoints
- **Disable safety net** -- disabling a currency resets any `UserSettings` referencing it to `KOKU_DEFAULT_CURRENCY`

**PR 1 of 7** for the constant currency feature (split from #6005).

**Depends on** #6126 (migrations PR -- must be merged first).

## Endpoints

| Method | Path | Status | Description |
|--------|------|--------|-------------|
| `GET` | `currency/` | **Modified** | Now queries `EnabledCurrency` table instead of a static list. Permission changed to `IsAuthenticated`. |
| `GET` | `settings/currency/` | **New** | Lists all ISO 4217 currencies with `enabled` and `has_dynamic_rate` flags. Supports `?search=` and `?enabled=` filters. |
| `POST` | `settings/currency/enabled/<code>/` | **New** | Enables a currency for the tenant (idempotent). |
| `DELETE` | `settings/currency/enabled/<code>/` | **New** | Disables a currency; resets affected user settings to default. |


koku/api/iam/models.py
```diff
@@ -71,6 +71,10 @@ def __init__(self, *args, **kwargs):
         self.identity_header = None
         self.beta = False
 
+    @property
+    def is_authenticated(self):
```
bacciotti, line 75: not sure if I follow why we automatically return True here

bacciotti (commented):
Some comments:
1) disabling a currency that is in use by active cost models succeeds silently. no warning, no guard. The cost model remains in the database with the disabled currency, but becomes uneditable (400: "is not an enabled currency" on any PUT/PATCH). No data is corrupted, but the admin has no way to know they just locked their own cost models. Suggest adding a check on DELETE that returns a warning (or 409 Conflict) if the currency is referenced by any CostModel. Not a data integrity blocker, but a UX issue worth tracking as a follow-up.

2) It is possible to disable all currencies (tested by clearing `enabled_currency` directly). With zero enabled currencies, `GET /currency/` returns an empty list, cost model creation returns 400, and the system is effectively unusable for currency-related operations. The state is fully reversible (re-enabling any currency restores normal behavior), so this is not a data-loss scenario. A minimum-one guard on DELETE would prevent this.

koku/api/iam/models.py
ELK4N4: This `User` model doesn't inherit from `AbstractBaseUser` (which provides `is_authenticated` by default), so we implement it manually. DRF's `IsAuthenticated` permission class checks `request.user.is_authenticated` — without this property, all requests would be rejected as unauthenticated. It's always `True` because by the time a `User` instance is on the request, auth has already been verified at the middleware layer.

ELK4N4:
> It is possible to disable all currencies (tested by clearing enabled_currency directly). With zero enabled currencies, GET /currency/ returns an empty list, cost model creation returns 400, and the system is effectively unusable for currency-related operations. The state is fully reversible (re-enabling any currency restores normal behavior), so this is not a data-loss scenario. A minimum-one guard on DELETE would prevent this.

@bacciotti  What is the behavior you want in this case? to block the user from disabling the last enabled currency?

ELK4N4:
> disabling a currency that is in use by active cost models succeeds silently. no warning, no guard. The cost model remains in the database with the disabled currency, but becomes uneditable (400: "is not an enabled currency" on any PUT/PATCH). No data is corrupted, but the admin has no way to know they just locked their own cost models. Suggest adding a check on DELETE that returns a warning (or 409 Conflict) if the currency is referenced by any CostModel. Not a data integrity blocker, but a UX issue worth tracking as a follow-up.

@bacciotti  Added a warning for this use case

koku/api/iam/models.py
bacciotti: let's add a short comment on the property documenting this contract? Something like:
`    # Always True: by the time a User instance is on the request,
    # IdentityHeaderMiddleware has already validated the x-rh-identity header.`

bacciotti:
> > It is possible to disable all currencies (tested by clearing enabled_currency directly). With zero enabled currencies, GET /currency/ returns an empty list, cost model creation returns 400, and the system is effectively unusable for currency-related operations. The state is fully reversible (re-enabling any currency restores normal behavior), so this is not a data-loss scenario. A minimum-one guard on DELETE would prevent this.
> 
> @bacciotti What is the behavior you want in this case? to block the user from disabling the last enabled currency?

let;s just check if it's the last row before deleting and return 409 if so.

ELK4N4:
> > > It is possible to disable all currencies (tested by clearing enabled_currency directly). With zero enabled currencies, GET /currency/ returns an empty list, cost model creation returns 400, and the system is effectively unusable for currency-related operations. The state is fully reversible (re-enabling any currency restores normal behavior), so this is not a data-loss scenario. A minimum-one guard on DELETE would prevent this.
> > 
> > 
> > @bacciotti What is the behavior you want in this case? to block the user from disabling the last enabled currency?
> 
> let;s just check if it's the last row before deleting and return 409 if so.

Done. I return 400 and not 409. I don't think it's a resource conflict but a validation rule.

bacciotti (dismissed):
can you add a short inline comment to is_authenticated documenting why it always returns True?

ELK4N4:
> can you add a short inline comment to is_authenticated documenting why it always returns True?

Done

bacciotti (dismissed):

myersCody (dismissed):
We need to have the migrations be separate than the code changes in case we need to roll back.

ELK4N4:
> We need to have the migrations be separate than the code changes in case we need to roll back.

@myersCody Done #6126

bacciotti (dismissed):

martinpovolny:
# Expanded version of the explanation:

currencies.py — We took the PR's version entirely. Main had symbol fixes (#6143) and added GHS (#6089) to the static CURRENCIES list, but this PR deletes that list and replaces it with babel-backed dynamic lookups, which produce the correct symbols.

test_rate_serializer.py — We took main's version. The PR's tests asserted that rate_id is included in RateSerializer output, but COST-7703 (#6123) landed on main and changed the serializer to strip rate_id from API responses (it's an internal DB identifier). The PR's tests would fail against the current serializer. Main's replacement tests cover the same scenarios but assert the new exclusion behavior.

bacciotti (approved):

PR #6116, opened 2026-06-11
Auto-heal org_id mismatch after Stage refresh (COST-7342)

martinpovolny:
## Jira Ticket

[COST-7342](https://redhat.atlassian.net/browse/COST-7342)  
**Runbook**: https://gitlab.cee.redhat.com/cost-management/service-docs/-/blob/main/docs/operations/sync-org-id-after-stage-refresh.md

## Description

Implements automatic self-healing for username collisions that occur after Stage environment refreshes, when users are recreated in Ethel with new `org_id` values but keep the same `username`.

**The Problem:**  
After a Stage refresh, when a user with a new `org_id` makes their first request, the middleware creates a new Customer successfully, but the old `api_user` record still exists with the same username pointing to the old Customer. This blocks any future attempt to persist a User with that username due to the UNIQUE constraint on `api_user.username`, requiring manual SQL intervention by SRE.

**The Solution:**  
When `Customer.DoesNotExist` is raised and a username/org_id mismatch is detected, the middleware checks the Unleash feature flag `cost-management.backend.auto-heal-org-mismatch`. If enabled (Stage only), it automatically renames the stale user to `old-{username}` (with collision handling via counter suffix), freeing the username for the new identity.

**See full documentation:** [docs/COST-7342.md](https://github.com/project-koku/koku/blob/mpovolny/COST-7342/docs/COST-7342.md)

### Key Features

- ✅ **Production-safe**: Feature flag disabled by default
- ✅ **Stage-only enablement**: Auto-heal only when explicitly enabled via Unleash UI
- ✅ **Collision handling**: Falls back to `old-{username}-1`, `-2`, etc. if needed
- ✅ **Read-only safety**: Skips rename on GET/HEAD requests
- ✅ **Comprehensive logging**: All detection and healing actions logged with `log_json()`
- ✅ **Emergency disable**: Can be toggled off instantly via Unleash without redeployment

### Code
- **`koku/koku/middleware.py`**: Added `_handle_org_id_mismatch()` method with detection and auto-heal logic
- **`koku/koku/feature_flags.py`**: Added flag to `MockUnleashClient.ONPREM_FLAG_DEFAULTS`
- **`koku/koku/test_middleware.py`**: Added 6 comprehensive unit tests in `HandleOrgIdMismatchTest`

### Documentation & Reproduction
- **`docs/COST-7342.md`**: Complete analysis, implementation details, and testing guide
- **`Makefile.COST-7342`**: Automated reproduction with `reproduce` and `reproduce-with-fix` targets
- **`scripts/COST-7342/*.py`**: Python scripts for each reproduction step (clean, create, trigger, fix, verify)
- **`docs/COST-7342-erd.svg`**: Entity relationship diagram showing the UNIQUE constraint issue

### Unit Tests
```bash
python koku/manage.py test koku.test_middleware.HandleOrgIdMismatchTest
```

Six test cases covering:
1. Auto-heal enabled → old user renamed
2. Auto-heal disabled → alert only (no rename)
3. GET request → rename skipped
4. HEAD request → rename skipped
5. Username collision → counter suffix used
6. Logging → mismatch always logged

### Automated Reproduction
```bash
# Start dev environment
make docker-up-min

# Reproduce the bug (shows IntegrityError)
make -f Makefile.COST-7342 reproduce

# Reproduce with manual fix applied
make -f Makefile.COST-7342 reproduce-with-fix
```

**See detailed testing instructions:** [docs/COST-7342.md#reproducing-the-issue](https://github.com/project-koku/koku/blob/mpovolny/COST-7342/docs/COST-7342.md#reproducing-the-issue)

### Manual Verification

1. **Checkout branch**
   ```bash
   git checkout mpovolny/COST-7342
   ```

2. **Run reproduction**
   ```bash
   make docker-up-min
   make -f Makefile.COST-7342 reproduce
   ```
   - You should see the `IntegrityError` demonstrating the bug

3. **Test auto-heal** (with mocked Unleash flag)
   ```bash
   python koku/manage.py test koku.test_middleware.HandleOrgIdMismatchTest::test_auto_heal_renames_old_user -v 2
   ```
   - Test passes → old user successfully renamed

4. **Verify logging**
   - Check test output for structured log messages with `log_json()`
   - Mismatch detection always logs WARNING
   - Auto-heal logs INFO when renaming
   - Alert-only logs ERROR when flag disabled

### Unleash Flag Configuration

**Before merging**, create the feature flag in Unleash:

```
Flag name: cost-management.backend.auto-heal-org-mismatch
Type: Boolean kill switch
Default: OFF (disabled globally)
```

**After merging**, enable in Stage only:
```
Environment: stage
Strategy: ON (enabled for all Stage users)

Environment: production  
Strategy: OFF (explicitly disabled)
```

### Monitoring

After deployment to Stage, monitor logs for:
```
"msg": "Org ID mismatch detected for user after environment refresh"
"msg": "Org ID mismatch: renamed stale user to free username (COST-7342)"
```

These indicate the auto-heal is working correctly.

## Release Notes

```markdown
* [COST-7342](https://redhat.atlassian.net/browse/COST-7342) Auto-heal org_id mismatch after Stage refresh - automatically renames stale users when username collisions are detected (Stage only, controlled by Unleash feature flag)
```


martinpovolny:
@myersCody , @esebesto : after review, I would remove all the `./**/*COST-7342*` files. Those are to make the review easier.

koku/koku/middleware.py
```diff
@@ -272,6 +273,97 @@ def create_customer(account, org_id, request_method):
...
+                    msg="Org ID mismatch: skipping rename on read-only request",
+                    username=username,
+                    request_method=request_method,
+                )
+            )
+            return
+
+        new_username = f"old-{username}"
+        counter = 1
+        while User.objects.filter(username=new_username).exists():
```
myersCody: From my perspective this while loop seems to be fixing a problem with our current renaming strategy. Instead of choosing usernames that have the potential to collide with each other we could utilize a timestamp suffix when we are renaming the old username.

```
old-test-user-20260615T110000
```

This would ensure that our new username would never collide with a previous username. 

Additionally do you know if we do anything to clean up the stale user objects caused by the stage refresh? 

If we don't then the counter approach may only last 10 stage refreshes before we hit our safeguard and the problem persist.

martinpovolny: That's a great point. I do not think we do anything. Let me look into that.

martinpovolny: @myersCody : no cleanup in the playbook or anywhere else IMO.

Implementing the change you suggested.

myersCody (approved):

PR #6120, opened 2026-06-11
COST-7243: Include unmatched GPU models in reports with cost=0

martinpovolny:
## Summary

- When a cost model has per-model GPU rates (`value_rates`) and a cluster has a GPU model not listed in those rates, `monthly_cost_gpu.sql` was filtering those rows out via a `WHERE` clause that only allowed GPU models with a defined rate.
- Unmatched GPU models were silently dropped from the API response instead of appearing with `cost=0.00`.
- Fix: remove the restrictive `AND (gpu.gpu_model_name = ...)` WHERE filter from both the pod-usage INSERT and the CTE subquery in the unallocated INSERT; add `ELSE 0` to both CASE expressions so unmatched models get `calculated_cost=0`.

**Details:** https://docs.google.com/document/d/1fwXOn6s6RzotHPATDwg-zVWomZMi3a_SHN9RGf6gH8Q/edit?usp=sharing

**jira:** https://redhat.atlassian.net/browse/COST-7243


martinpovolny:
@esebesto : the test I have provided is very shallow, but my understanding is that a proper integration test would go to iqe-cost-management-plugin and quality engineering would do that. Or shall I? What is the expectation and flow. Be my guide, please?

djnakabaale (commented):
suggestion: can we mirror this fix in the on-prem SQL (`self_hosted_sql`) template as well to keep behavior consistent in `koku/masu/database/self_hosted_sql/openshift/cost_model/monthly_cost_gpu.sql`?

djnakabaale (approved):

PR #6123, opened 2026-06-15
[COST-7703] Remove rate_id from CostModel API response

jordigilh:
## Summary

Fixes [COST-7703](https://redhat.atlassian.net/browse/COST-7703) — the CostModel API was returning the internal database identifier `rate_id` in rate objects. When the UI round-tripped this field back on update, the backend rejected it with a 400 Bad Request.

- **Strip `rate_id` from output**: `RateSerializer.to_representation()` no longer emits `rate_id` — this is an internal DB PK for the dual-write sync mechanism, not consumer-facing data
- **Keep `custom_name` in output**: `custom_name` remains in the API response as requested by the UI team for rate identification and future edit/delete functionality
- **Strip `rate_id` from input**: `RateSerializer.to_internal_value()` silently pops `rate_id` instead of validating it, so old clients sending it back get silent acceptance
- **Lenient fallback**: `_classify_incoming_rates()` now logs a warning and falls back to `custom_name` matching when a `rate_id` doesn't match any existing Rate row, instead of raising `ValueError`

### What stays the same

- Internal JSON storage (`CostModel.rates`, `PriceList.rates`) still contains `rate_id`/`custom_name` — needed for sync diff logic
- `Rate` model and migration 0012 — unchanged
- PriceList API (`_build_rate_response`) — separate endpoint with its own `rate_id` semantics
- `test_cost_model_manager.py` and `test_price_list_view.py` — unaffected

### FedRAMP control alignment

| Control | Objective | Verified by |
|---------|-----------|-------------|
| SC-8 / SC-28 | Internal DB PKs must not leak through API | `test_to_representation_excludes_rate_id`, `test_api_response_does_not_expose_rate_id` |
| SI-11 | Round-tripping API response must not cause 400 | `test_to_internal_value_tolerates_malformed_rate_id`, `test_round_trip_update_succeeds_without_rate_id` |
| CM-7 | API surface exposes only consumer-needed fields | `test_to_representation_preserves_only_business_fields` |

## Test plan

- [x] Unit tests: 6 new tests in `test_rate_serializer.py` (rate_id output stripping, custom_name output inclusion, input tolerance, business-field allowlist)
- [x] Unit tests: 2 new + 1 updated in `test_sync_rate_table.py` (lenient fallback on unknown rate_id)
- [x] Integration tests: 2 new + 2 updated in `test_rate_integration.py` (rate_id exclusion, custom_name inclusion, full round-trip)
- [x] Removed 10 obsolete tests that asserted the old broken behavior
- [x] flake8 passes on all modified files
- [ ] CI pipeline passes full test suite


koku/cost_models/rate_sync.py
```diff
@@ -151,21 +151,22 @@ def _classify_incoming_rates(rates_data, existing_by_uuid, existing_by_name, all
...
-                raise ValueError(f"Invalid rate_id: {rate_id}")
-            rate_obj = existing_by_uuid[rate_uuid]
-            incoming_ids.add(rate_uuid)
+                raise ValueError(f"Invalid rate_id format: {rate_id}")
+            if rate_uuid in existing_by_uuid:
+                rate_obj = existing_by_uuid[rate_uuid]
+                incoming_ids.add(rate_uuid)
+                to_update.append((rate_obj, rate_data))
+                continue
+            LOG.warning("rate_id %s not found; falling back to custom_name matching", rate_id)
```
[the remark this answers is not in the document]
jordigilh: Good catch — this was a real bug. Stripping `custom_name` from input meant `_resolve_custom_name` had to regenerate the name, but `used_names` was seeded with `all_existing_names`, forcing a `-2` suffix that broke `existing_by_name` matching and caused UUID churn.

Fixed in 95c119a:
- `used_names` now initialized with only explicit `custom_name` values from the incoming request (not `all_existing_names`)
- `rate_id` match path now adds the matched rate's `custom_name` to `used_names` to prevent intra-batch collisions
- Added `test_update_without_custom_name_preserves_uuid` to verify UUID preservation when neither `rate_id` nor `custom_name` is in the input

koku/cost_models/serializers.py
```diff
@@ -405,22 +398,13 @@ def to_representation(self, rate_obj):
...
-            if value is not None:
-                try:
-                    self.fields[field_name].run_validation(value)
-                except serializers.ValidationError as e:
-                    errors[field_name] = e.detail
-        if errors:
-            raise serializers.ValidationError(errors)
+            data.pop(field_name, None)
         metric = data.get("metric", {})
         new_metric = {"name": metric.get("name")}
```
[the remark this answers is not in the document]
jordigilh: Good defensive catch. This was technically pre-existing (the `data.get("metric", {})` pattern was in the original code), but since we're touching this method anyway, fixed in the same commit (95c119a):

```python
metric = data.get("metric") or {}
new_metric = {"name": metric.get("name") if isinstance(metric, dict) else None}
```

The `or {}` handles explicit `None`, and the `isinstance` guard prevents `AttributeError` if someone passes a non-dict type.

koku/cost_models/serializers.py
```diff
@@ -213,7 +213,6 @@ class RateSerializer(serializers.Serializer):
     DECIMALS = ("value", "usage_start", "usage_end")
     RATE_TYPES = ("tiered_rates", "tag_rates")
 
-    rate_id = serializers.UUIDField(required=False, allow_null=True)
     custom_name = serializers.CharField(max_length=50, required=False, allow_blank=True)
```
djnakabaale, line 216: Do we need field validation for `custom_name` in this method?
Since `to_internal_value()` bypasses `super()`, the declared `custom_name = serializers.CharField(max_length=50, ...)` validation no longer runs, so oversized/invalid values can flow to the DB write path instead of failing at serializer validation time.

koku/cost_models/serializers.py
```diff
@@ -405,24 +401,14 @@ def to_representation(self, rate_obj):
     def to_internal_value(self, data):
         """Convert the JSON representation of rate to DB representation.
 
-        Validates rate_id and custom_name via their declared fields, then
-        returns the original data dict with the metric normalized. We cannot
-        call super().to_internal_value() because validate_cost_type has a
-        non-standard signature (2 args) that DRF's field-level validation
-        would call incorrectly.
+        Silently strips internal enrichment fields (rate_id, custom_name) that
```
djnakabaale: nitpick suggestion: Update this docstring to match behavior.
It currently says we strip `rate_id` and `custom_name`, but the implementation only strips rate_id.

djnakabaale (commented):
Changes look good, just a couple small comments:

koku/cost_models/serializers.py
jordigilh: Good catch — since `to_internal_value()` bypasses `super()`, the `CharField(max_length=50)` declared on the field never runs. Added an explicit length check inside `to_internal_value()` that raises a 400 with the same message DRF would produce. See c18b40a.

koku/cost_models/serializers.py
jordigilh: Fixed — updated the docstring to only mention `rate_id` since `custom_name` is not stripped. See c18b40a.

djnakabaale (approved):
lgtm

PR #6134, opened 2026-06-19
[COST-7701] fix expired trino partition for month-partitioned managed tables

djnakabaale:
## Jira Ticket

[COST-7701](https://issues.redhat.com/browse/COST-7701)

## Description

This change updates expired Trino partition discovery SQL to support both day-partitioned and month-partitioned managed OCP tables, preventing `TrinoUserError: Column 'day' cannot be resolved` during cleanup.

## Testing

1. Checkout branch.

2. Start shell in tenant schema and run runtime validation against affected tables.
   ```bash
   TRINO_HOST=localhost TRINO_PORT=8080 make shell-schema schema=org1234567
   ```
   ```python
   from masu.database.ocp_report_db_accessor import OCPReportDBAccessor

   schema = "org1234567"
   tables = [
       ("managed_gcp_openshift_disk_capacities_temp", "ocp_source"),
       ("managed_aws_openshift_disk_capacities_temp", "ocp_source"),
       ("managed_azure_openshift_disk_capacities_temp", "ocp_source"),
       ("managed_reporting_ocpgcpcostlineitem_project_daily_summary_temp", "ocp_source"),
       ("reporting_ocpusagelineitem_daily_summary", "source"),
   ]

   with OCPReportDBAccessor(schema) as accessor:
       for table, source_col in tables:
           result = accessor.find_expired_trino_partitions(table, source_col, "2026-07-01")
           print(table, "OK", type(result).__name__, result[:2] if isinstance(result, list) else result)
   ```
   - You should see `OK` for all tables.
   - You should **not** see `Column 'day' cannot be resolved`.

## Release Notes
- [ ] proposed release note

```markdown
* Fix expired Trino partition cleanup for month-partitioned managed OCP tables.
```

### DIFF
--- a/koku/koku/reportdb_accessor_trino.py
+++ b/koku/koku/reportdb_accessor_trino.py
@@ -13,6 +13,13 @@ from koku.reportdb_accessor import ReportDBAccessor
 
 LOG = logging.getLogger(__name__)
 
+MONTH_PARTITIONED_MANAGED_TABLES = {
+    "managed_aws_openshift_disk_capacities_temp",
+    "managed_azure_openshift_disk_capacities_temp",
+    "managed_gcp_openshift_disk_capacities_temp",
+    "managed_reporting_ocpgcpcostlineitem_project_daily_summary_temp",
+}
+
 
 class TrinoReportDBAccessor(ReportDBAccessor):
     """Trino implementation of report database accessor."""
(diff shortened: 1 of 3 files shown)


koku/koku/reportdb_accessor_trino.py
```diff
@@ -13,6 +13,13 @@
 
 LOG = logging.getLogger(__name__)
 
+MONTH_PARTITIONED_MANAGED_TABLES = {
```
myersCody:
```
def purge_expired_trino_partitions(self, expired_date, simulate=False):
        """Removes expired trino partitions."""
        LOG.debug(f"purge_expired_trino_partitions: {expired_date}, {simulate}")

        with OCPReportDBAccessor(self._schema) as accessor:
            for table, source_column in TRINO_MANAGED_TABLES.items():
                LOG.debug(f"{table}, {source_column}")
                results = accessor.find_expired_trino_partitions(table, source_column, str(expired_date.date()))
                if results:
                    LOG.info(f"Discovered {len(results)} expired partitions")
                else:
                    LOG.info("No expired partitions")
                    return
                for partition in results:
                    LOG.info(f"partition_info: {partition}")
                    if not simulate:
                        year, month, source_value = partition
                        accessor.delete_hive_partition_by_month(table, source_value, year, month, source_column)
```

From what I can see it looks like we drop the partitions be they month. Do we even need the day information? Also, we have `EXPIRE_MANAGED_TABLES` but it looks unused.  May be worth cleaning up that code.

djnakabaale: Thanks for review. Removed `EXPIRE_MANAGED_TABLES` because it was unused, and updated expired-partition filtering to use month-level date anchoring (`-01`) so it does not depend on a day partition column.

lcouzens (approved):

PR #6145, opened 2026-06-25
[COST-7469] add post-write deduplication for OCP parquet files from concurrent payloads

djnakabaale:
## Jira Ticket

[COST-7469](https://issues.redhat.com/browse/COST-7469)

## Description

This change adds a post-write deduplication step for OCP parquet files in S3. When concurrent OCP payloads with overlapping daily data are processed by separate workers, the existing pre-write dedup can miss duplicates due to a race condition (both workers check S3 before either writes). The post-write dedup re-scans S3 after writing, compares `reportnumhours` metadata across manifests, and deterministically removes the superseded files, eliminating inflated usage/cost data without manual intervention.

The feature is gated behind the Unleash flag `cost-management.backend.ocp_post_write_parquet_dedup`. 

## Testing

1. Checkout branch and start the stack with Trino:

    ```bash
    make docker-up-min-trino
    ```

2. Confirm `DEVELOPMENT=True` is set in `.env` which auto-enables the feature flag via `dev_fallback=True`,  no Unleash configuration needed.

3. Create an OCP source:

    ```bash
    curl -s http://localhost:8000/api/cost-management/v1/sources/ \
      -H "x-rh-identity: '<X_RH_IDENTITY>'" \
      -H "Content-Type: application/json" \
      -d '{"name":"test-dedup-source","source_type":"OCP","authentication":{"credentials":{"cluster_id":"my-test-cluster"}}}'
    ```

4. Upload two overlapping OCP payloads to MinIO (same cluster, same day, different assembly IDs, different hour counts):

    ```bash
    # Upload payload A (e.g. 10 hours) and payload B (e.g. 11 hours) to ocp-ingress bucket
    # Trigger both ingestions back-to-back:
    curl -s "http://localhost:5042/api/cost-management/v1/ingest_ocp_payload/?org_id=1234567&payload_name=payloadA.2026_06" &
    curl -s "http://localhost:5042/api/cost-management/v1/ingest_ocp_payload/?org_id=1234567&payload_name=payloadB.2026_06" &
    wait
    ```

5. Check masu-server logs for dedup activity:
    1. You should see `OCP parquet post-write dedup complete` with `files_deleted` > 0 for the manifest with fewer hours.
    2. You should see `OCP parquet dedup post-write winner selection complete` or `OCP parquet dedup: current manifest superseded` indicating deterministic winner selection.

6. Verify only one manifest's parquet files remain in S3 for the overlapping day.

## Release Notes
- [ ] proposed release note

```markdown
* Add post-write deduplication for OCP parquet files from concurrent payloads

### CODE REVIEW

File: koku/masu/processor/parquet/parquet_report_processor.py [hunk 1]
```diff
@@ -422,6 +425,11 @@ def convert_to_parquet(self):  # noqa: C901
                     )
                 )
                 raise ParquetReportProcessorError(msg)
+
+            if self.provider_type == Provider.PROVIDER_OCP and is_feature_flag_enabled_by_schema(
```
bacciotti, line 429: I think this is running in onprem mode too? Maybe we can add a safeguard `if not isOnPrem`?

koku/masu/util/ocp/common.py
```diff
@@ -745,3 +749,179 @@ def table_exists(self, schema: str) -> bool:
...
+
+    Returns:
+        S3 keys of the superseded files to delete. Empty list if no duplicates exist.
+    """
+    if len(manifest_groups) <= 1 or current_manifest_id not in manifest_groups:
+        return []
+
+    # Two-pass approach: first check if we lose to any manifest,
+    # then collect superseded keys only if we're the winner.
+    winner_manifest_id = None
```
bacciotti: not a blocker: maybe we can drop this line?

koku/masu/processor/parquet/parquet_report_processor.py
djnakabaale: Onprem never writes OCP parquet files to S3, goes through `handle_daily_frames_postgres / write_to_self_hosted_table` instead...the dedup scan would find no objects and return an empty list (no-op).

koku/masu/util/ocp/common.py
djnakabaale: thanks for the review, i've cleaned this up.

bacciotti (approved):

PR #6149, opened 2026-06-29
[COST-7729] Fix env_override logic and expose min/max bounds in data-retention endpoint

bacciotti:
## Jira
[COST-7729](https://redhat.atlassian.net/browse/COST-7729) — Missing type-ahead APIs for UI integration

## Summary

Enriches the existing `GET /api/cost-management/v1/account-settings/data-retention/` endpoint with the fields the UI team needs to fully integrate the data retention settings feature.

### Changes to `GlobalSettingsView`

**1. Fix `env_override` logic (aligns with COST-7728)**

**Before:** `env_override` was `true` whenever `RETAIN_NUM_MONTHS` was set in the environment, regardless of its value.

**After:** `env_override` is `true` only when `RETAIN_NUM_MONTHS` is set **and** its value differs from the system default (`DEFAULT_RETAIN_NUM_MONTHS = 4`). If the env var equals the default, the UI treats the setting as unconfigured and still shows the dropdown.

The same logic applies to `PUT`: blocked (`403`) only when the env var is set to a non-default value.

**2. Expose `min_retention_months` and `max_retention_months`**

The GET response now includes the input bounds so the UI does not need to hardcode them.

### Final GET response

```json
{
  "data_retention_months": 12,
  "env_override": false,
  "min_retention_months": 3,
  "max_retention_months": 120
}
```

### What the UI does with this

```
env_override: false  →  show the settings dropdown (user can configure)
env_override: true   →  hide the dropdown / show value as read-only

data_retention_months >= 6   →  show "Last 6 months" preset in Cost Explorer
data_retention_months >= 12  →  show "Last 12 months" preset in Cost Explorer
```

> The date picker presets are derived by the UI from `data_retention_months` — no separate endpoint is needed (confirmed in sync with @danLabrecque and @myersCody).

## Scenarios

| `RETAIN_NUM_MONTHS` | `env_override` | PUT allowed | Effective retention |
|---|---|---|---|
| not set | `false` | ✓ | DB value or 4 |
| `4` (== default) | `false` | ✓ | 4 (env wins) |
| `12` | `true` | ✗ | 12 |
| `"abc"` (invalid) | `true` | ✗ | 503 |

## Testing

```bash
# Unit tests (25 total)
DJANGO_READ_DOT_ENV_FILE=True pipenv run python koku/manage.py test api.settings.test.test_global_settings_views -v 2

# Live testing — requires ONPREM=True in .env and containers running

# env_override: false (RETAIN_NUM_MONTHS=4 in .env, then force-recreate)
curl http://localhost:8000/api/cost-management/v1/account-settings/data-retention/ \
  -H "x-rh-identity: $IDENTITY"
# → env_override: false, data_retention_months: 4

# env_override: true (set RETAIN_NUM_MONTHS=12 in .env, then force-recreate)
curl http://localhost:8000/api/cost-management/v1/account-settings/data-retention/ \
  -H "x-rh-identity: $IDENTITY"
# → env_override: true, data_retention_months: 12
```

Force-recreate command:
```bash
docker compose -f docker-compose.yml -f docker-compose.debug.yml up -d --force-recreate koku-server
```


koku/api/settings/views.py
```diff
@@ -115,19 +116,37 @@ class GlobalSettingsView(APIView):
...
             )
-        return Response({"data_retention_months": effective, "env_override": env_override})
+        return Response(
+            {
+                "data_retention_months": effective,
+                "env_override": env_override,
+                "min_retention_months": TenantSettings.MIN_RETENTION_MONTHS,
+                "max_retention_months": TenantSettings.MAX_RETENTION_MONTHS,
+            }
+        )
```
djnakabaale, line 146: question: The response here returns a flat JSON object, but all other account-settings/ endpoints (e.g. currency/, cost-type/). Should we consider to wrap the response in the standard {meta, links, data} envelope:

```
{
    "meta": {"count": 1, "limit": 10, "offset": 0},
    "links": {"first": "...", "next": null, "previous": null, "last": "..."},
    "data": {"currency": "USD"}
}
```

koku/api/settings/views.py
```diff
@@ -154,6 +173,40 @@ def put(self, request):
...
+            return Response(
+                {"error": "Unable to read retention settings."},
+                status=status.HTTP_503_SERVICE_UNAVAILABLE,
+            )
+        options = list(self._BASE_OPTIONS)
+        for threshold, option in self._CONDITIONAL_OPTIONS:
+            if retention >= threshold:
+                options.append(option)
+        options.extend(self._TAIL_OPTIONS)
+        return Response({"options": options})
```
djnakabaale: suggestion (same as above): The date-range-options response returns `{"options": [...]}` as flat JSON. Consider wrapping in the standard {meta, links, data} envelope for consistency with sibling endpoints under account-settings/.

koku/api/settings/views.py
```diff
@@ -154,6 +173,40 @@ def put(self, request):
...
+        "last_2_months",
+        "last_3_months",
+    ]
+    _CONDITIONAL_OPTIONS = [
+        (6, "last_6_months"),
+        (12, "last_12_months"),
+    ]
+    _TAIL_OPTIONS = ["maximum", "custom"]
+
+    def get(self, request):
```
djnakabaale: Also, this endpoint is `GET` only and accepts no query parameters, but unknown params (e.g. ?foo=bar) are silently ignored and still return 200. Do we need to validate/reject unknown params? Worth deciding if we want consistent behavior here or if silent-ignore is acceptable.

djnakabaale (commented):
From first pass, the endpoint logic looks clean. A few suggestions below, mostly around aligning with existing API conventions.


Also, the new `date-range-options` endpoint (and `data-retention`, which was added in PR #5971) are not documented in the `docs/specs/openapi.json`. The spec currently only has /account-settings/ and /account-settings/{setting}.
Since the UI team will be integrating against these, could we add the endpoint definitions here or in a follow-up pr?

koku/api/settings/views.py
bacciotti: following a sync we agreed to keep the flat JSON response

koku/api/settings/utils.py
```diff
@@ -296,13 +296,17 @@ def set_cost_type(schema, cost_type_code=KOKU_DEFAULT_COST_TYPE):
 def get_data_retention_months(schema_name: str) -> "int | None":
     """Return the effective data-retention period for the given tenant.
 
-    Priority: env var > DB > Config default.
+    Priority: env var (only when set to a non-default value) > DB > Config default.
     Returns None on DB read failure (caller should skip purge).
     """
+    from koku.settings import DEFAULT_RETAIN_NUM_MONTHS
```
djnakabaale: nitpick: this import can be moved to the top of the file alongside the other imports for consistency.

koku/api/settings/views.py
```diff
@@ -115,19 +116,37 @@ class GlobalSettingsView(APIView):
...
+    @staticmethod
+    def _is_env_retention_locked():
+        """Return True when RETAIN_NUM_MONTHS is set to a non-default value."""
+        env_val = os.environ.get("RETAIN_NUM_MONTHS")
+        if env_val is None:
+            return False
+        try:
+            return int(env_val) != DEFAULT_RETAIN_NUM_MONTHS
+        except (ValueError, TypeError):
+            return True
```
djnakabaale, line 128: for `GET` with `RETAIN_NUM_MONTHS=3months`, this returns `True`, but then `get_data_retention_months` returns `None` and the view responds `503` with (`Unable to read retention settings`). should the error message distinguish between an invalid env var, or is a generic 503 acceptable here?

bacciotti: the generic message is intentional, the actual cause is logged at ERROR level, so the admin could check. happy to improve this if you'd prefer a more descriptive message, though!

djnakabaale (approved):

PR #6152, opened 2026-06-29
[COST-7249] Fix RTU deadlock caused by overlapping deletes

myersCody:
## Jira Ticket

### Root Cause
The OCP cost model updater writes to RTU in multiple separate transactions across two phases. The DELETE in insert_usage_rates_to_usage.sql uses a broad predicate — source_uuid + report_period_id — that covers rows owned by the distribution and markup DELETEs as well.

When two pipeline runs overlap, each tries to DELETE rows the other has just inserted but not yet committed, producing a ShareLock wait cycle that PostgreSQL resolves by killing one transaction as the deadlock victim.

A secondary contributing factor: the distribution DELETE filtered on report_period_id without source_uuid, causing PostgreSQL to fall back to a heap scan rather than using the composite index (usage_start, source_uuid, report_period_id).

The different lock-acquisition order between the heap scan and the indexed DELETE is a lock-order inversion that can trigger the same deadlock even without concurrent runs.

### Changes
insert_usage_rates_to_usage.sql — narrow the DELETE to only the rows it owns by adding AND monthly_cost_type IS NULL AND metric_type != 'markup'. Each DELETE in the pipeline now targets a disjoint set of rows.

## Testing

1. Checkout Branch
2. Restart Koku
    1. You should see ...
4. Do more things...

## Release Notes
- [ ] proposed release note

```markdown
```


djnakabaale (dismissed):
SQL changes lgtm! 👍🏾 

one small note: the deadlock ([#4511209](https://glitchtip.devshift.net/insights/issues/4511209)) seems to originate in [`delete_archived_data`](https://github.com/project-koku/koku/blob/ad7c8c93956982bc31068f778e275659a6586c4d/koku/masu/celery/tasks.py#L253), not the cost model SQL path. PR helps reduce lock surface, but that ORM delete is still unguarded, maybe worth a follow-up.

PR #6162, opened 2026-07-01
[COST-7249] Fix deadlock on rates_to_usage during concurrent cost model updates

jordigilh:
## Summary

`#6100` was reverted in `#6158` after a deadlock surfaced in stage:

```
django.db.utils.OperationalError: deadlock detected
DETAIL: Process 922 waits for ShareLock on transaction 131824610; blocked by process 920.
        Process 920 waits for ShareLock on transaction 131824621; blocked by process 922.
CONTEXT: while updating tuple (204591,13) in relation "rates_to_usage_2026_05"
```

This PR fixes it with two independent changes. A companion PR reapplies #6100's content plus separate SQL performance fixes ([gist](https://gist.github.com/jordigilh/806be98b9ea71a14b530c8980e0720a2)) — the two PRs touch disjoint files and can be reviewed/merged independently.

Full writeup with root-cause analysis, the reasoning behind what we did and didn't do, and precise claims about what is/isn't prevented: **[gist](https://gist.github.com/jordigilh/3f6cca5b0d4c815f5aa1436443541f1b)**.

## What each fix targets

**Fix 1 — narrow the `WorkerCache` dedup key** (`masu/processor/tasks.py`)

`update_cost_model_costs` keyed its dedup cache on `[schema, provider_uuid, start_date, end_date]`. Two invocations for the *same provider* with different (possibly overlapping) date ranges — e.g. a normal monthly trigger racing a backfill — were never recognized as duplicates, so both ran concurrently and did DELETE+INSERT against the same `rates_to_usage` partition. Narrowing the key to `[schema, provider_uuid]` makes the second invocation correctly detected and requeued instead of running concurrently.

This **prevents** the deadlock for this mechanism — if the two transactions never run at the same time, there's no lock-order inversion to occur. It reuses the exact `WorkerCache` dedup pattern already used identically elsewhere in this same file (`update_summary_tables`, the OCP-on-cloud summary task) — no new locking mechanism, just a one-line change to which arguments go into the existing key.

**Fix 2 — retry raw SQL execution on deadlock detection** (`masu/database/report_db_accessor_base.py`)

`RatesToUsage` is tenant-scoped and its physical partition is shared by *every provider* under one customer schema, not per-provider. Fix 1 intentionally still lets different providers under the same schema run concurrently (for throughput), so two different providers could in principle still collide on the same partition — a residual case Fix 1 doesn't address. Rather than leave that as a silent gap, we made any deadlock self-healing: `ReportDBAccessorBase._execute_raw_sql_query` (the single choke point for all raw SQL execution, including every `rates_to_usage` write) is now decorated with `@retry(retry_on=(ExtendedDeadlockDetected,))`.

This does **not** prevent that residual deadlock from occurring — it changes the consequence from a fatal task failure to a transparent retry. Both pieces already existed in this codebase and were already proven in production, just never wired together: `ExtendedDeadlockDetected` (`koku/database_exc.py`) already precisely classifies a raw Postgres deadlock (it even parses out both colliding process/transaction IDs), and `retry()` (`koku/trino_database.py`) is the same backoff/jitter decorator already used for transient Trino errors elsewhere in this file. Safe to retry because every call here is its own autocommitted statement — nothing spans a `transaction.atomic()` block, so a deadlock only aborts the one statement, not a larger transaction.

## How they relate

| | Prevents the deadlock event? | Prevents it from becoming a task failure? |
|---|---|---|
| Fix 1 | Yes, for the proven same-provider/overlapping-date-range mechanism | N/A — nothing to fail if it's prevented |
| Fix 2 | No — occurrence for the cross-provider case is unchanged | Yes — any deadlock that still occurs now retries instead of failing |

They are deliberately complementary, not redundant: Fix 1 removes the one root cause we could reproduce and prove from the incident logs; Fix 2 is a safety net for the case Fix 1 can't structurally close without a throughput trade-off we don't think is justified (see the gist for the options we considered and rejected, including a schema-wide serialization and a `(schema, month)`-scoped advisory lock, and why).

## What to expect

- The specific deadlock mechanism visible in the incident logs (same-provider, overlapping date ranges) should no longer occur.
- If a different provider under the same customer schema happens to collide with another provider's cost-model update on the same monthly partition (a case we have no direct evidence fired in the original incident, but can't rule out architecturally), it will now surface as a brief, logged retry rather than a failed Celery task.
- Full structural prevention of that residual cross-provider case would require restructuring transaction boundaries around the whole per-month cost-model pipeline (to safely hold a schema/month-scoped advisory lock) — we scoped that out of this PR as a separate follow-up rather than bundling an under-tested architectural change here; the reasoning is in the gist.

## Validation

- New unit tests (`masu.test.database.test_report_db_accessor_base`) added via TDD, covering: a transient deadlock recovers after one retry, a persistent deadlock is retried a bounded number of times then raised, and a non-deadlock `OperationalError` fails immediately with no retry.
- Full `masu.test.processor.test_tasks` and `masu.test.database.test_report_db_accessor_base` suites pass on this branch in isolation (71 tests).
- Full IQE integration suite run on a helios08 environment (concurrent Celery workers, multiple providers) completed with no deadlock errors observed on `rates_to_usage`.
- We did not construct a synthetic test that deliberately races two overlapping-date-range invocations for the same provider end-to-end through Celery — Fix 1 is validated by code inspection of the dedup key plus absence of the failure mode under real concurrent load in the helios08 run, not by a targeted repro test.


koku/masu/test/database/test_report_db_accessor_base.py
```diff
@@ -3,14 +3,48 @@
...
+    set to the original driver exception.
+    """
+    deadlock = DeadlockDetected(
+        "deadlock detected"
+        + os.linesep
+        + "DETAIL: Process 12  transaction 34  blocked by process 56"
+        + os.linesep
+        + "Process 56  transaction 78  blocked by process 12"
+        + os.linesep
+    )
```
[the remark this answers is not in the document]
jordigilh: Confirmed and fixed in d8f304bd1.

You're right that real Postgres `DETAIL` continuation lines are indented (verified by re-checking the original incident log), and the regex had no `\s*` allowance between the first line's trailing `\n` and the second line's `Process` token, so it silently failed to match on that shape.

Two things worth calling out:
- This doesn't affect the retry behavior itself — `get_extended_exception_by_base_type` classifies purely on `isinstance(_db_exception, DeadlockDetected)`, so `retry_on=(ExtendedDeadlockDetected,)` still fires correctly even when the regex fails to parse. The blast radius was limited to the `process1`/`process2`/`txaction1`/`txaction2` fields staying `None`, which degrades the `DEADLOCKED DATABASE PIDS: [...]` log line used for diagnosing *which* two backends collided.
- Rather than only tweaking this PR's mock, I added a regression test (`test_deadlock_exception_with_indented_continuation_line` in `test_database_exc.py`) using the exact indented shape from the original incident, confirmed it failed against the old regex (`TypeError` on `sorted([None, None])`), then fixed `ExtendedDeadlockDetected.REGEXP` to tolerate the leading whitespace, and updated this file's mock to match the real format too.

All 16 tests across `test_database_exc.py` and `test_report_db_accessor_base.py` pass with the fix.

jordigilh:
Addressed the review feedback:

- **gemini-code-assist**: fixed the deadlock-message regex to tolerate the indented continuation line that real Postgres `DETAIL` blocks use — see [inline reply](https://github.com/project-koku/koku/pull/6162#discussion_r3508333419) for details. Added a regression test using the real indented message shape.
- **CI triager**: added the `ocp-smoke-tests` label (the retry logic is exercised by `update_cost_model_costs`, which is OCP-specific).

Pushed as d8f304bd1.

koku/masu/processor/tasks.py
```diff
@@ -932,7 +932,8 @@ def update_cost_model_costs(  # noqa: C901
 
     """
     task_name = "masu.processor.tasks.update_cost_model_costs"
-    cache_args = [schema_name, provider_uuid, start_date, end_date]
+    # Only key on schema + provider so overlapping date ranges serialize instead of deadlocking on rates_to_usage
+    cache_args = [schema_name, provider_uuid]
```
myersCody: These cache args do act as guard rails to prevent duplicate runs on the same date range. However, these args have been in place in production for the past 5 years with no issues. 

For this reason, I would argue that the the recent DeadLock issues likely has to do with write or delete structure of the RatesToUsage table. I am not convinced we are solving the root cause of the deadlock.

jordigilh:
## RCA: `koku-ci` failure is a known, pre-existing flake unrelated to this PR

The bot's diagnosis attributes the failure to this PR's `_flush_db_operation` retry/re-raise path in `report_db_accessor_base.py`. That method **does not exist** anywhere in this codebase (checked the current tree and the full git history) — the bot fabricated the mechanism. Grounding the RCA in what's actually true:

**1. Same failure occurred on an unrelated PR.** [#6116](https://github.com/project-koku/koku/pull/6116) (touches only `middleware.py`, `feature_flags.py`, and docs — nothing DB/SQL/Celery related) hit the identical assertion pattern in the same test:
```
tests/rest_api/v1/test_ros.py::test_api_ocp_ros_kafka_content
AssertionError: assert 1 == 2  (expected 2 kafka messages, got 1)
```
vs. this run's:
```
tests/rest_api/v1/test_ros.py::test_api_ocp_ros_kafka_content
AssertionError: assert 3 == 4  (expected 4 kafka messages, got 3)
```
Two different PRs, two different code areas, same test, same "N-1 messages received" shape — that's the signature of a timing-dependent flake in the test, not a functional regression tied to either PR's diff.

**2. No code path connects the two.** ROS Kafka event shipping happens synchronously in `masu/external/kafka_msg_handler.py` during initial manifest processing (`ROSReportShipper.process_manifest_reports`, called from `process_report`/`handle_message` before any cost-model task is even scheduled). This PR's changes are confined to `_execute_raw_sql_query`'s deadlock-retry decorator (used by `OCPReportDBAccessor` for cost-model SQL) and the `WorkerCache` key for `update_cost_model_costs`. Those run later in the pipeline and never touch the Kafka producer or `ROSReportShipper`.

**3. Failure shape matches a timing race, not a logic bug.** The log shows `Topic scan time limit elapsed` after receiving 3 of 4 expected messages — a bounded-time consumer poll racing message delivery, not a dropped/failed upload (all 4 files were confirmed uploaded, and 4174/4175 other tests passed in the same run).

**Conclusion:** unrelated, pre-existing IQE flakiness in `test_api_ocp_ros_kafka_content`, not a regression from this PR's deadlock-retry logic. Re-triggering to confirm.

/retest

koku/masu/database/report_db_accessor_base.py
```diff
@@ -104,8 +105,21 @@ def _prepare_and_execute_raw_sql_query(self, table, tmp_sql, tmp_sql_params=None
         sql, sql_params = self.prepare_query(tmp_sql, tmp_sql_params)
         return self._execute_raw_sql_query(table, sql, bind_params=sql_params, operation=operation)
 
+    @retry(
```
myersCody, line 108: My concern about a blanket retry here is idempotency. 

I still don't think we have investigated WHY the deadlock is happening for the RatesToUsage table specifically.

jordigilh:
Swapped the smoke label from \`ocp-smoke-tests\` to \`full-run-smoke-tests\` so this run validates against all providers, not just OCP. Re-triggering to make sure the in-flight pipeline picks up the new label from the start.

/retest

myersCody (changes requested):
My concern here is that I don't see a root cause analysis of what is specifically different about the rates_to_usage table that is causing this deadlock. We never experienced this pattern with the old flow, which tells me something about the new write pipeline itself is creating the conditions for it.

The two proposed fixes are:

1. Prevent more than one cost model task from running per provider at a time
2. Catch the deadlock and retry

Both feel like treating symptoms rather than the root cause. If a behavior in the pipeline is what creates the deadlock, we should identify and fix that behavior — not just serialize execution or retry around the failure.

Before this merges, I'd want to see an explanation of why RTU specifically is vulnerable when other tables in the same pipeline are not. The hypothesis I shared with you is that it has to do with the how the DELETEs are scoped.

jordigilh:
## RCA follow-up: reproduced the deadlock end-to-end, confirmed both fixes empirically

Thanks for pushing on this — you were right that the original writeup leaned too much on static analysis. I spiked a full empirical reproduction in `helios08` to settle it. Summary below; happy to share the harness/scripts if useful.

### 1. Why `rates_to_usage` and not the legacy tables?

Confirmed via `EXPLAIN (ANALYZE, BUFFERS)` and live `pg_stat_activity` monitoring, not just reading the SQL:

- `insert_usage_rates_to_usage.sql`'s initial DELETE is scoped only by `usage_start` + `source_uuid` + `report_period_id` — it removes rows for **all 6 cost types in one statement**. Legacy `daily_summary` deletes always added a further discriminator (`cost_model_rate_type`, `monthly_cost_type`), spreading the same total footprint across several smaller statements.
- Row amplification means RTU's DELETE touches an order of magnitude more physical rows per invocation than the legacy path did for the same source data.
- I directly observed the exact production wait via `pg_stat_activity` during two concurrent overlapping-range deletes: `wait_event_type=Lock, wait_event=transactionid` — this is Postgres's `XactLockTableWait`, the identical mechanism behind the `"waiting for ShareLock on transaction NNN"` line in the incident log. It reproduces reliably any time two overlapping-range invocations for the same provider run concurrently against RTU.

### 2. Does a single overlapping DELETE actually deadlock (not just contend)?

Tested exhaustively: 2-way nested overlap, 2-way crossing overlap, 3-way circular overlap, and DELETE+INSERT round-trip — **160 trials, 0 deadlock cycles**, despite confirmed contention in every case. Root cause: Postgres's Bitmap Heap Scan (confirmed via `EXPLAIN`) visits matching rows in deterministic ascending physical-block order for any parameterization of this query, so two (or three) concurrent invocations always encounter shared rows in the same relative order — they serialize (wait), they don't cycle.

### 3. Full-pipeline reproduction (the actual smoking gun)

Single-statement testing wasn't sufficient, so I stood up the real code path: seeded a production-representative dataset (~26K `daily_summary` rows → 51K+ RTU rows per phase), created a real cost model + price list, and ran `OCPCostModelCostUpdater.update_summary_cost_model_costs()` **3 times concurrently for the same provider with overlapping date ranges**, exercising the full 9-phase pipeline (usage rates, monthly costs, tag costs, VM costs, distribution).

**Before this PR's retry fix** (WorkerCache narrowing only): 15 trials → **2 real, unhandled deadlocks (13%)**, byte-for-byte the same signature as the incident:
```
deadlock detected
DETAIL:  Process 1359695 waits for ShareLock on transaction 691784; blocked by process 1359697.
Process 1359697 waits for ShareLock on transaction 691783; blocked by process 1359695.
CONTEXT:  while deleting tuple (304,2) in relation "rates_to_usage_2026_06"
```

**After cherry-picking this PR's retry-on-deadlock commit onto the same checkout** and re-running the identical 15-trial batch: 3 deadlocks occurred (same signature) — **all 3 caught, logged as `"Deadlock detected, retrying statement (attempt 1)"`, retried, and completed successfully. Zero unhandled failures.**

### Conclusion

- The deadlock is real, reproduces reliably at realistic scale, and requires 3+ concurrent overlapping-range invocations against the full pipeline (not a single statement) — consistent with why it took a while to surface in staging.
- **Fix 1** (`WorkerCache` key narrowed to `[schema, provider_uuid]`) sits in front of this in production: it serializes *all* same-provider tasks at dispatch time regardless of date range, so 3 same-provider Celery tasks can never reach this SQL concurrently through the normal path. (I bypassed Celery/WorkerCache intentionally to stress-test the underlying DB mechanism — this validates the SQL-level race and Fix 2, not whether Fix 1's guard itself can be bypassed.)
- **Fix 2** (retry-on-deadlock) is now empirically verified, not just theoretically argued, to make the residual failure mode (anything Fix 1 doesn't cover — cross-provider concurrency under the same schema, manual/API-triggered re-runs, worker restarts mid-flight) self-healing.

Let me know if you'd like the reproduction scripts attached to the PR or kept as a reference gist.

jordigilh:
## RCA: raw_calc IQE failures (pipeline `koku-ci-bj9c5`)

I traced this at the code level rather than taking the bot's causal chain at face value (it's previously hallucinated a nonexistent method name on this same PR).

### The bot's specific mechanism is not supported by the diff

The bot claims COST-7252 changed `/exchange-rates/` in a way that breaks the raw_calc tests. That's not what the code shows:

- `/exchange-rates/` → `get_exchange_rates()` in `koku/api/currency/view.py`, backed by the `ExchangeRateDictionary` model. **COST-7252 touches neither this file nor this model.**
- COST-7252's actual footprint: `koku/api/settings/currency_views.py` (`CurrencySettingsView`, which serves the unrelated `/settings/currency/` endpoint) gains a `StaticExchangeRate.objects.all()` lookup, plus new additive CRUD endpoints/serializers for the new `StaticExchangeRate` model.
- The new `MonthlyExchangeRate` model (added in the same merge, #6161) has **zero references anywhere outside its own model file and migration** — it isn't read by any query handler or report path yet. It cannot be influencing computed raw_calc cost values.
- COST-7729 (`#6149`, also pulled in by the same `merge main` commit) only touches `api/settings/views.py` / `api/settings/utils.py` (data-retention bounds) — unrelated to sources/stats/cost endpoints.

So on the evidence, the currency-model changes that rode in via the merge are not a plausible mechanism for these specific failures.

### What I can't rule out, and why I can't confirm it either

This PR's `koku-ci` run on `d8f304bd` (before the `main` merge) failed for a **confirmed, unrelated Konflux infra reason** (`SnapshotCreationFailed`, all visible tests passing at 12% before truncation) — so we have no signal either way from that run on whether `raw_calc` specifically passed or failed on the pre-merge code.

More importantly: **this is the first time this PR has run under `full-run-smoke-tests`.** It previously ran under `ocp-smoke-tests`, which does not exercise `test__i_source_g_cloud_provider_raw_calc.py` (AWS/Azure/GCP) at all. There is no historical baseline for these exact tests on this branch, and no other recent PR has run the full suite against a post-COST-7252 `main` to serve as a clean control (`#6099`/`#6101` are the only other recent `full-run-smoke-tests` runs, and both predate the COST-7252/COST-7729 merge).

That leaves this PR's own behavioral change — the `WorkerCache` key narrowing (`[schema, provider_uuid]` instead of `[schema, provider_uuid, start_date, end_date]`) — as the remaining candidate: it serializes all `update_cost_model_costs` calls per-provider, which could plausibly increase recompute latency in an ephemeral IQE environment where many test files trigger overlapping updates against a small shared pool of fixture providers. I don't have access to the actual pytest tracebacks (Konflux log storage requires internal RH auth I don't have from here), so I can't confirm whether the failures are actually a stale-cost-data assertion (which would support this theory) or something unrelated.

### Recommendation

1. Retrigger (`/retest`) — infra flakiness has already occurred once on this exact PR, so a clean rerun is the cheapest next signal.
2. If it reproduces the same `raw_calc` failures again, I'd want the actual pytest traceback for one of them before concluding root cause — that will show directly whether it's a cost-value staleness assertion. Happy to bisect by temporarily dropping just the `WorkerCache` narrowing commit and rerunning full smoke to isolate it from the retry-on-deadlock commit, if that's useful.

_Posted after static verification of the diffs referenced above — not automated._

myersCody:
> ## RCA follow-up: reproduced the deadlock end-to-end, confirmed both fixes empirically
> Thanks for pushing on this — you were right that the original writeup leaned too much on static analysis. I spiked a full empirical reproduction in `helios08` to settle it. Summary below; happy to share the harness/scripts if useful.
> 
> ### 1. Why `rates_to_usage` and not the legacy tables?
> Confirmed via `EXPLAIN (ANALYZE, BUFFERS)` and live `pg_stat_activity` monitoring, not just reading the SQL:
> 
> * `insert_usage_rates_to_usage.sql`'s initial DELETE is scoped only by `usage_start` + `source_uuid` + `report_period_id` — it removes rows for **all 6 cost types in one statement**. Legacy `daily_summary` deletes always added a further discriminator (`cost_model_rate_type`, `monthly_cost_type`), spreading the same total footprint across several smaller statements.
> * Row amplification means RTU's DELETE touches an order of magnitude more physical rows per invocation than the legacy path did for the same source data.
> * I directly observed the exact production wait via `pg_stat_activity` during two concurrent overlapping-range deletes: `wait_event_type=Lock, wait_event=transactionid` — this is Postgres's `XactLockTableWait`, the identical mechanism behind the `"waiting for ShareLock on transaction NNN"` line in the incident log. It reproduces reliably any time two overlapping-range invocations for the same provider run concurrently against RTU.
> 
> ### 2. Does a single overlapping DELETE actually deadlock (not just contend)?
> Tested exhaustively: 2-way nested overlap, 2-way crossing overlap, 3-way circular overlap, and DELETE+INSERT round-trip — **160 trials, 0 deadlock cycles**, despite confirmed contention in every case. Root cause: Postgres's Bitmap Heap Scan (confirmed via `EXPLAIN`) visits matching rows in deterministic ascending physical-block order for any parameterization of this query, so two (or three) concurrent invocations always encounter shared rows in the same relative order — they serialize (wait), they don't cycle.
> 
> ### 3. Full-pipeline reproduction (the actual smoking gun)
> Single-statement testing wasn't sufficient, so I stood up the real code path: seeded a production-representative dataset (~26K `daily_summary` rows → 51K+ RTU rows per phase), created a real cost model + price list, and ran `OCPCostModelCostUpdater.update_summary_cost_model_costs()` **3 times concurrently for the same provider with overlapping date ranges**, exercising the full 9-phase pipeline (usage rates, monthly costs, tag costs, VM costs, distribution).
> 
> **Before this PR's retry fix** (WorkerCache narrowing only): 15 trials → **2 real, unhandled deadlocks (13%)**, byte-for-byte the same signature as the incident:
> 
> ```
> deadlock detected
> DETAIL:  Process 1359695 waits for ShareLock on transaction 691784; blocked by process 1359697.
> Process 1359697 waits for ShareLock on transaction 691783; blocked by process 1359695.
> CONTEXT:  while deleting tuple (304,2) in relation "rates_to_usage_2026_06"
> ```
> 
> **After cherry-picking this PR's retry-on-deadlock commit onto the same checkout** and re-running the identical 15-trial batch: 3 deadlocks occurred (same signature) — **all 3 caught, logged as `"Deadlock detected, retrying statement (attempt 1)"`, retried, and completed successfully. Zero unhandled failures.**
> 
> ### Conclusion
> * The deadlock is real, reproduces reliably at realistic scale, and requires 3+ concurrent overlapping-range invocations against the full pipeline (not a single statement) — consistent with why it took a while to surface in staging.
> * **Fix 1** (`WorkerCache` key narrowed to `[schema, provider_uuid]`) sits in front of this in production: it serializes _all_ same-provider tasks at dispatch time regardless of date range, so 3 same-provider Celery tasks can never reach this SQL concurrently through the normal path. (I bypassed Celery/WorkerCache intentionally to stress-test the underlying DB mechanism — this validates the SQL-level race and Fix 2, not whether Fix 1's guard itself can be bypassed.)
> * **Fix 2** (retry-on-deadlock) is now empirically verified, not just theoretically argued, to make the residual failure mode (anything Fix 1 doesn't cover — cross-provider concurrency under the same schema, manual/API-triggered re-runs, worker restarts mid-flight) self-healing.
> 
> Let me know if you'd like the reproduction scripts attached to the PR or kept as a reference gist.

## The Problem:
- `insert_usage_rates_to_usage.sql`'s initial DELETE is scoped only by `usage_start` + `source_uuid` + `report_period_id` — it removes rows for **all 6 cost types in one statement**. Legacy `daily_summary` deletes always added a further discriminator (`cost_model_rate_type`, `monthly_cost_type`), spreading the same total footprint across several smaller statements.
- Row amplification means RTU's DELETE touches an order of magnitude more physical rows per invocation than the legacy path did for the same source data.

I agree that the issue is that the `insert_usage_rates_to_usage.sql's` delete is too broad:
```
DELETE FROM {{schema | sqlsafe}}.rates_to_usage
WHERE usage_start >= {{start_date}}
  AND usage_start <= {{end_date}}
  AND source_uuid = {{source_uuid}}
  AND report_period_id = {{report_period_id}};
```

Now we did have a narrower scoped delete in the initial proof of concept:
```
DELETE FROM {{schema | sqlsafe}}.rates_to_usage AS rtu
WHERE rtu.usage_start >= {{start_date}}
    AND rtu.usage_start <= {{end_date}}
    AND rtu.source_uuid = {{source_uuid}}
    AND rtu.report_period_id = {{report_period_id}}
    AND rtu.cost_model_rate_type = {{rate_type}}
    AND rtu.monthly_cost_type IS NULL
;
```

Could we adjust the delete to match what you had in the POC if we turned `rtu.cost_model_rate_type = {{rate_type}}` to ` AND cost_model_rate_type in ('Infrastructure', 'Supplementary')` it would help narrow the scope of the `DELETE` footprint and it means this step no longer clobbers work done by other phases in the pipeline. 

## Fix One

```
**Fix 1** (`WorkerCache` key narrowed to `[schema, provider_uuid]`) sits in front of this in production: it serializes *all* same-provider tasks at dispatch time regardless of date range, so 3 same-provider Celery tasks can never reach this SQL concurrently through the normal path. (I bypassed Celery/WorkerCache intentionally to stress-test the underlying DB mechanism — this validates the SQL-level race and Fix 2, not whether Fix 1's guard itself can be bypassed.)
```

The date range was included in the cache key intentionally -- it allows cost model summaries for different months to run concurrently for the same provider. If a task for both Jan & Feb are queued, Feb starts immediately rather than waiting in the requeue loop until Jan finishes.

Narrowing to `[schema, provider_uuid]` sacrifices that concurrency to paper over the SQL problem. That's a real cost.

## Fix Two (keeping it, but as a safety net — not the primary fix)

The retry-on-deadlock is good defense-in-depth, I still have some concerns regarding dempotency. `_execute_raw_sql_query` is what we use for every database query, so we will blindly be retrying on any deadlock not just cost model scenario you outlined in the RCA.

jordigilh:
Thanks for the detailed pushback — both points land. Pushed changes addressing each:

## Fix 1: reverted the WorkerCache key narrowing, fixed the root cause in SQL instead

You're right that `[schema_name, provider_uuid]` was too blunt — it serialized Jan/Feb (or any non-overlapping range) for the same provider even though they never touch the same rows. Reverted to `[schema_name, provider_uuid, start_date, end_date]` in `2f113e4f3`, with a regression test (`test_update_cost_model_costs_cache_key_includes_date_range`) pinning down the expected cache key so it can't silently regress again.

In its place, narrowed `insert_usage_rates_to_usage.sql`'s DELETE (`f533c2f29` / `66d4eb678` on #6163) close to what you suggested, with one adjustment: instead of `cost_model_rate_type = {{rate_type}} AND monthly_cost_type IS NULL`, it's now

```sql
AND cost_model_rate_type IN ('Infrastructure', 'Supplementary')
AND (monthly_cost_type IS NULL OR monthly_cost_type = 'Tag')
```

The `OR monthly_cost_type = 'Tag'` is necessary, not cosmetic: tag-based usage costs (`infrastructure_tag_rates.sql` / `supplementary_tag_rates.sql` / the two `default_*_tag_rates.sql` files) are co-located in this same phase and have no other dedicated DELETE anywhere in the pipeline. A strict `IS NULL`-only filter would silently stop cleaning those rows up, duplicating them on every re-run — the exact bug class #6168 just fixed for monthly costs. Monthly costs (Node/Cluster/PVC/OCP_VM) stay excluded since `_delete_monthly_cost_model_data` already owns their delete+reinsert, scoped per `cost_type`.

This also exposed two pre-existing test defects that were relying on the old, over-broad DELETE to accidentally wipe monthly-cost rows as test isolation: `TestRTUCapacityColumns` was asserting capacity on rows that never carry it by design, and `TestDistributionIntegration._source_qs` was excluding monthly-cost rows from valid distribution sources when the SQL itself treats them as valid sources. Both fixed in `66d4eb678`.

Validated on helios08: 3 consecutive full runs of the targeted OCP/RTU/distribution suite (283 tests, 0 regressions) plus the broader `masu.test.processor` + `masu.test.database` suite (668 tests, only 2 pre-existing unrelated AWS Parquet failures, confirmed via zero source diff against upstream/main).

## Fix 2: idempotency concern — added a self-verifying guard instead of a one-time audit

Fair concern — `_execute_raw_sql_query`'s retry is only safe because every current call site runs as an independent, autocommitted statement (no surrounding `transaction.atomic()`). Rather than just asserting that in the PR description, added `RawSqlRetryIdempotencyGuardTest` (`2f113e4f3`): an AST-based static test across every `ReportDBAccessorBase` subclass that fails loudly if `transaction.atomic()` is ever introduced as a real `with` block or decorator in those files. If someone later wraps a raw-SQL call in a real transaction alongside other statements, this trips before the retry-safety assumption silently breaks, rather than relying on it staying true forever by convention.

Let me know if you want the reproduction scripts/harness attached here or kept as a reference gist — happy to share either way.

koku/masu/processor/tasks.py
jordigilh: You called this exactly right, and the empirical follow-up (see my comment below) bore it out: the recent deadlock wasn't caused by this cache key's shape — it's the RTU table's own write footprint. Root cause is `insert_usage_rates_to_usage.sql`'s DELETE being scoped too broadly (wiping all 6 RTU cost types per invocation instead of just the ones this phase repopulates), amplifying per-invocation row-lock contention far beyond the legacy tables' equivalent. Fixed at the SQL layer in #6163 (`f533c2f29`/`66d4eb678`), and reverted this cache-key narrowing back to `[schema, provider_uuid, start_date, end_date]` in `2f113e4f3` so Jan/Feb-style non-overlapping ranges keep running concurrently like they always have. Full RCA + reproduction numbers in the PR-level comment: https://github.com/project-koku/koku/pull/6162#issuecomment-4907128146

koku/masu/database/report_db_accessor_base.py
jordigilh: Both fair — addressed in `2f113e4f3`:

**Idempotency**: added `RawSqlRetryIdempotencyGuardTest`, an AST-based static test across every `ReportDBAccessorBase` subclass. It fails loudly if `transaction.atomic()` is ever introduced as a real `with` block or decorator around one of these files, since that's the exact condition that would break the "each raw-SQL call is an independent, autocommitted statement" assumption the retry currently relies on. Makes the safety argument self-verifying instead of a one-time manual audit.

**Root cause on rates_to_usage specifically**: did the full empirical dig — confirmed via `EXPLAIN (ANALYZE, BUFFERS)` and live `pg_stat_activity` monitoring that `insert_usage_rates_to_usage.sql`'s DELETE was scoped only by usage_start/source_uuid/report_period_id, wiping all 6 RTU cost types per invocation (vs. the legacy tables' deletes, which always added a further discriminator). That row amplification is the proximate cause. Full writeup with reproduction numbers (160 deadlock-contention trials, 15-trial full-pipeline reproduction before/after) here: https://github.com/project-koku/koku/pull/6162#issuecomment-4907128146. Fixed the scope itself in #6163.

jordigilh:
## Status: on hold — awaiting team decision on DB migration impact

Per team discussion, this PR is intentionally paused: the deadlock/race-condition fix here may be partially or fully superseded by the RTU schema migrations that just landed (`ON DELETE CASCADE` on `rate_id`/`cost_model_id`/`source_uuid`/`report_period`, indexes on `rate_id`/`cost_model_id` — see #6172/#6187/#6207/#6218/#6219). We want to observe how the migration affects deadlocks/race conditions in production before deciding whether the retry/locking logic in this PR is still needed, needs to be scaled back, or should be dropped entirely.

**Current state (for whoever picks this back up):**
- CI (`Red Hat Konflux / koku-ci / koku`) is currently failing at `init-pipeline-context` — this PR has `smokes-required` but no specific smoke-suite label (e.g. `ocp-smoke-tests`). Not investigated further since we're not actively pushing this forward.
- Branch is ~38 commits behind `main` and will need a rebase before CI/merge is attempted again.
- No code changes needed based on review feedback so far (`myersCody`'s WorkerCache-key and idempotency concerns were already addressed).

**Next step:** revisit after we have enough production signal from the migration rollout to know whether deadlocks/races still occur, then rebase + add the smoke label + re-open for review.

Leaving `on-hold` label in place until then.

koku/masu/database/report_db_accessor_base.py
```diff
@@ -104,8 +105,21 @@ def _prepare_and_execute_raw_sql_query(self, table, tmp_sql, tmp_sql_params=None
...
+        retries=settings.DB_DEADLOCK_RETRIES,
+        retry_on=(ExtendedDeadlockDetected,),
+        max_wait=4,
+        log_message="Deadlock detected, retrying statement",
```
myersCody, line 112: @jordigilh Right now when we experience a deadlock this is the only information we will get. Could you look into providing more detail here?

jordigilh: Good catch, thanks — you're right that this line was next to useless in practice. Fixed in 720ee6c00.

Two compounding bugs:
1. The `@retry` decorator's `context` is derived by inspecting a `sql_params` kwarg, but this method was actually called with `bind_params=`, so `context` was silently always empty — no schema/date-range/provider info ever made it into the log line.
2. The exception object itself was stuffed under an `exc_info` key rather than serialized, so nothing from `ExtendedDeadlockDetected` (query, tables, deadlocked PIDs, pgcode, backend PID, traceback) ever surfaced on the retry line — you'd have had to cross-reference the separate `_execute_raw_sql_query` warning logged one line earlier by timestamp.

The fix:
- Renamed `bind_params` -> `sql_params` on `_execute_raw_sql_query` so the decorator's existing context extraction actually fires.
- The decorator now merges `ex.as_dict()` (when the caught exception supports it, e.g. any `ExtendedDBException` subclass) into the logged context under an `exception_detail` key, so the retry/failure log line is self-sufficient -- no more correlating two separate WARNING lines.

Added `test_execute_raw_sql_query_retry_log_includes_deadlock_detail` to pin this down, and verified on helios08 that a simulated deadlock now logs (attempt 1) with `exception_detail` containing `process1_pid`/`process2_pid`, the deadlock message, etc., instead of an empty context.

myersCody (dismissed):

koku/masu/database/report_db_accessor_base.py
jordigilh: Reverted `0b5003a93` — it chased schema/date context you didn't ask for, and didn't fully work as described anyway. Back to `720ee6c00`, which already covers this: `ex.as_dict()` is merged into the retry-warning log, so a deadlock now logs query type, tables, both PIDs, pgcode, and the Postgres message right on the `(attempt N)` line. Re-requesting review since the push reset your approval — nothing new to look at.

myersCody (approved):

jordigilh:
@esebesto @myersCody  switched to hot-fix since the PR has been approved and passed the full-smoke-tests in the past

PR #6163, opened 2026-07-01
[COST-7249] Phase 4 PR 3: Reapply orchestration rewire + per-rate distribution with performance fixes

jordigilh:
## Summary

This reapplies `#6100` ("Phase 4 PR 3: Orchestration rewire + per-rate distribution"), reverted in `#6158`, plus targeted fixes for the SQL performance regressions that contributed to the revert (a companion PR separately fixes the deadlock that also contributed — see below). Full writeup with root-cause analysis and complete benchmark methodology/results: **[gist](https://gist.github.com/jordigilh/806be98b9ea71a14b530c8980e0720a2)**.

**Companion PR:** #6162 fixes the deadlock on `rates_to_usage`. The two PRs touch disjoint files (this PR never modifies `masu/processor/tasks.py`'s `WorkerCache` dedup key) and are independent — either can merge first.

## What changed

`#6100`'s original content (orchestration rewire, per-rate distribution SQL, breakdown population SQL) is reapplied unchanged as the base of this PR. On top of that, we found and fixed three distinct root causes of the performance regression in the distribution SQL (`masu/database/sql/openshift/cost_model/distribute_cost/*.sql`):

- **Row amplification** — the original per-rate distribution ran the full expensive proportional-distribution join once *per rate*, multiplying both compute cost and output row count by the number of configured rates. Fixed with a two-phase rewrite: compute the total distributed amount once per namespace/day, then split it across rates with cheap arithmetic. A dedicated test (`test_11_two_phase_rewrite_equivalence_all_distribution_types`) proves this is algebraically equivalent to the original single-phase computation for all four distribution types, not just faster.
- **Redundant scans** — several shared CTEs (`denominator`, `namespace_usage`) were being recomputed from scratch independently inside all four distribution SQL files. Fixed by materializing them into temp tables once per month (`create_distribution_temp_tables.sql`) and having all four files read from there.
- **Expensive joins** — `IS NOT DISTINCT FROM` on nullable join keys prevented Postgres from using index/hash join plans in some cases; replaced with `COALESCE`-based equality. Several of these joins existed only to pull capacity/metadata columns (`node_capacity_*`, `resource_id`, `persistentvolume`, `storageclass`, etc.) from `daily_summary`; we denormalized those directly onto `RatesToUsage` (migration `0352_rtu_capacity_columns_and_distribution_indexes.py`) so distribution no longer needs to join back for them.

## What to expect (see gist for full numbers)

Benchmarked end-to-end and at the SQL-component level against **the exact commit that was reverted** (`a9751bd47`), not a reconstructed baseline, on a synthetic dataset scaled to 6.3M distributed output rows:

- **No performance regression** at any scale we tested.
- A modest (~1.6%) mean wall-clock improvement at this large scale, where bulk `INSERT` cost dominates over the query-planning inefficiencies we fixed (the old code was insert-bound, not join-bound, at 6.3M rows) — we're not overselling this number.
- A **>10x reduction in run-to-run variance** (5.8s spread vs 65.7s spread across 3 runs) — the more operationally meaningful result for avoiding another multi-hour-timeout-style incident, since a predictable ~274s ± 3s runtime is far easier to reason about for timeout budgets than one swinging between 254s and 320s.
- Component-level `EXPLAIN ANALYZE` isolates ~6.5–7.5s of eliminated redundant computation per run on this dataset, independent of INSERT overhead, which is smaller than the full wall-clock delta but explains the tighter variance.

## Validation

- Full unit test suite passes on this branch in isolation (205 tests: `test_phase4_distribution`, `test_phase2_rates_to_usage`, `test_ocp_report_db_accessor`).
- New dedicated equivalence test for the two-phase distribution rewrite, since that's the change with the highest correctness risk (rewriting the actual distribution math).
- Full IQE integration suite run on a helios08 environment completed with no regressions.
- Migration `0352_rtu_capacity_columns_and_distribution_indexes.py` adds columns/indexes only — no data migration required.

## Summary by Sourcery

Reapply the Phase 4 RTU distribution pipeline with performance, concurrency, currency, and cost-breakdown fixes while retaining a safe legacy path.

New Features:
- Reapply the Phase 4 rates-to-usage orchestration, per-rate cost distribution, and cost-breakdown summary pipeline behind the feature flag.
- Add cost-breakdown UI summary population from per-rate usage and distributed-cost records.
- Support currency conversion for unattributed infrastructure distribution and propagate cost-model currency through aggregation.

Bug Fixes:
- Prevent rates-to-usage DELETE operations from removing unrelated cost categories and reduce concurrent update conflicts.
- Prevent distributed cost double-counting during concurrent overlapping runs with provider-scoped serialization.
- Avoid markup double-counting and correctly handle infrastructure-only and zero-rate distribution pools.
- Correct multi-GPU-model distribution so each model's cost is allocated once and conserved.
- Preserve RTU capacity and metadata during aggregation and remove expensive daily-summary joins.

Enhancements:
- Improve distribution SQL performance by reducing repeated scans, eliminating per-rate row amplification, and using planner-friendly joins.
- Support separate legacy and RTU distribution paths, including appropriate cost-breakdown handling when RTU is disabled.
- Make distribution and aggregation processing month-aware and maintain derived-cost timestamps.

Tests:
- Add integration and regression coverage for distribution correctness, cost conservation, idempotency, orchestration order, markup handling, GPU allocation, breakdown population, RTU capacity propagation, cleanup, and concurrency.


koku/masu/processor/ocp/ocp_cost_model_cost_updater.py
```diff
@@ -772,21 +777,166 @@ def _delete_tag_usage_costs(self, start_date, end_date, source_uuid):
...
+        """Return the markup percentage as a Decimal, or None if no markup configured."""
+        with CostModelDBAccessor(
+            self._schema, self._provider_uuid, price_list_effective_on=None
+        ) as cost_model_accessor:
+            markup = cost_model_accessor.markup
+            if not markup:
+                return None
+            value = Decimal(markup.get("value", 0))
+            if not value:
+                return None
```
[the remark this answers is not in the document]
jordigilh: Good catch, fixed in ad70bc997.

`markup.get("value", 0)` only falls back to the default when the key is *absent*; a JSON `null` value (`{"value": null, "unit": "percent"}`) passes straight through as `None`, and `Decimal(None)` raises `TypeError`. Applied your suggested approach: read the raw value first, short-circuit on falsy/`None` before constructing the `Decimal`, and convert via `str()` (matching the existing convention a few lines up in `test_returns_decimal_percentage`) rather than passing the raw JSON value directly.

Added a regression test (`test_returns_none_when_markup_value_is_none` in `TestGetMarkupPercentage`), confirmed it raised `TypeError: conversion from NoneType to Decimal is not supported` against the old code, then applied the fix. All 4 tests in that class pass now.

koku/reporting/migrations/0352_rtu_capacity_columns_and_distribution_indexes.py
```diff
@@ -0,0 +1,78 @@
+# Generated by Django 5.2.15 on 2026-07-01 13:04
+
+from django.db import migrations, models
```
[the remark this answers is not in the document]
jordigilh: Applied, fixed in ad70bc997 (ran `reorder-python-imports` directly to match the exact pre-commit hook output).

koku/masu/test/processor/ocp/test_phase4_distribution.py
```diff
@@ -0,0 +1,945 @@
...
+    def test_11_two_phase_rewrite_equivalence_all_distribution_types(self):
+        """Cross-check formula + cost conservation hold for every distribution type."""
+        for dist_type in ("platform_distributed", "worker_distributed", "unattributed_storage"):
+            with self.subTest(dist_type=dist_type):
+                if not self._skip_if_no_distributed_type(dist_type):
+                    continue
+
+                with schema_context(self.schema):
+                    total_distributed = (
+                        self._recipient_qs(dist_type).aggregate(t=Sum("distributed_cost"))["t"] or Decimal(0)
```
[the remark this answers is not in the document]
jordigilh: Applied, fixed in ad70bc997 (ran `black` directly to match the exact pre-commit hook output).

jordigilh:
Addressed the review feedback:

- **gemini-code-assist**: fixed `_get_markup_percentage` to handle an explicit `null` markup value without raising `TypeError` — see [inline reply](https://github.com/project-koku/koku/pull/6163#discussion_r3508334418) for details. Added a regression test that reproduced the crash before the fix.
- **CI triager (sanity/pre-commit)**: applied the `reorder-python-imports` and `black` auto-fixes to the migration and test file.
- **CI triager (labels)**: added the `ocp-smoke-tests` label (changes touch OCP cost model distribution logic).

Pushed as ad70bc997.

koku/reporting/migrations/0354_rtu_capacity_columns_and_distribution_indexes.py
```diff
@@ -0,0 +1,78 @@
+# Generated by Django 5.2.15 on 2026-07-01 13:04
+from django.db import migrations
+from django.db import models
+
+
+class Migration(migrations.Migration):
```
djnakabaale, line 6: @jordigilh can we split this `0352` migration into its own PR. It will need to land first before the code that depends on it ships, especially here since it adds columns that the new SQL reads/writes.

jordigilh: Good call. Split it out into #6173 (schema-only: the 9 nullable columns + 2 indexes, verified via `makemigrations --check` = no changes detected against the model state). It's purely additive so it can merge/deploy independently. Once it lands I'll rebase this PR to drop the duplicate `0352` migration and depend on it instead.

jordigilh:
Dropped `summary_ocp_rp_start_ds_idx` from this branch's copy of migration `0352` (commit b12dfb06e) to mirror #6173's fix. Per EXPLAIN ANALYZE validation on #6173 (https://github.com/project-koku/koku/pull/6173#discussion_r3539677910), the planner never used this index for the queries it targeted — dropped as pure write overhead with no measured read benefit. This is a draft pending rebase onto #6173 once that merges, so this is just keeping the duplicate copy consistent in the meantime.

koku/reporting/migrations/0354_rtu_capacity_columns_and_distribution_indexes.py
jordigilh: Done — split out as #6173 (schema-only: adds the RTU capacity/metadata columns + index). That PR needs to land first; #6163 will rebase onto it once merged.

esebesto:
> :

this comment is not correct - most likely, there is no perf. issue. The pipeline was killed after 7h 30 min, because the midnight hold was applied and prolonged the overall execution time (expected), but for some reason, the total timeouts we tried to bump yesterday didn't work in expected way. We will fix the total timeouts shortly.

jordigilh:
@esebesto @myersCody  full-smoke-test suite passed, switching to hot-fix-smoke-test labels and rebasing. 
https://github.com/project-koku/koku/pull/6163/checks?check_run_id=93702993453

koku/masu/processor/ocp/ocp_cost_model_cost_updater.py
```diff
@@ -907,4 +1071,4 @@ def update_summary_cost_model_costs(self, summary_range: SummaryRangeConfig) ->
 
             self._update_markup_cost(start_date, end_date, use_rtu=rtu_enabled)
 
-        self.distribute_costs_and_update_ui_summary(summary_range)
+        self.distribute_costs_and_update_ui_summary(summary_range, use_rtu=rtu_enabled)
```
djnakabaale, line 1091: With the flag on, `_update_markup_cost(..., use_rtu=rtu_enabled)` still runs before this. Won’t that double-count markup via `platform_rtu_cost` and `platform_infra.infrastructure_markup_cost`? Should RTU markup wait until after distribute/aggregate?

jordigilh: good catch, it looks like there is a test gap that is missing this scenario. I'm going to extend the scope to see if I find other missing use cases. I'll push a new commit with the test + fix for this bug. Thanks for catching this up!. 

I'll run again with full-smoke-test to ensure no regressions again.

jordigilh: done in https://github.com/project-koku/koku/pull/6163/commits/94358b3e10071d7963a2c48228524f683460c3be

jordigilh:
@djnakabaale koku-ci passed https://github.com/project-koku/koku/pull/6163/checks?check_run_id=94255774220
Is it ok to review again? Thanks!

koku/masu/test/processor/ocp/test_phase4_distribution.py
```diff
@@ -0,0 +1,1860 @@
...
+                )
+                rate_share = rate_cost / total_source
+                for entry in rate_rows:
+                    ns_total = ns_totals.get(entry["namespace"], Decimal(0))
+                    expected = rate_share * ns_total
+                    self.assertAlmostEqual(
+                        float(entry["actual"]),
+                        float(expected),
+                        places=6,
+                        msg=f"Option 2 cross-check failed for rate={rate_name}, ns={entry['namespace']}",
```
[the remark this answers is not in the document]
jordigilh: Good catch, applied in e5b49018a. This is the same flake class already fixed for the ratio check a few tests down — confirmed with 5x reruns plus the full suite on a clean DB, no regressions.

koku/masu/processor/ocp/ocp_cost_model_cost_updater.py
```diff
@@ -777,21 +781,181 @@ def _delete_tag_usage_costs(self, start_date, end_date, source_uuid):
...
+                    t0 = time.monotonic()
+                    accessor.populate_markup_rates_to_usage(
+                        month_range.start_date,
+                        month_range.end_date,
+                        self._provider.uuid,
+                        self._cluster_id,
+                        self._cost_model_id,
+                    )
+                    RTU_MARKUP_DURATION.labels(provider_type=self._provider.type).observe(time.monotonic() - t0)
+                accessor.populate_ui_summary_tables(month_range, self._provider_uuid)
```
djnakabaale: The `populate_ui_summary_tables` always rebuilds UI summary tables, including `reporting_ocp_cost_breakdown_p` (via `UI_SUMMARY_TABLES`). That table is populated from `rates_to_usage`, so with the RTU flag off we can still rewrite breakdown from stale RTU data. Should this path gate breakdown population on `use_rtu`?

djnakabaale (commented):
@jordigilh thanks for addressing the earlier feedback. One last note on my side, otherwise this looks ready to move along.

koku/masu/processor/ocp/ocp_cost_model_cost_updater.py
jordigilh: Good catch — you're right, this was a real gap. `reporting_ocp_cost_breakdown_p` is built entirely from `rates_to_usage`, and the legacy path never writes there. Since the flag gets re-evaluated per schema per billing month, a period that was processed once with RTU on (and has real RTU/breakdown data) could later get reprocessed with the flag off, and we'd happily rebuild breakdown from that stale data without anyone noticing it no longer matches the fresh legacy costs.

Went back to the Phase 4 design doc to make sure I wasn't inventing a new policy here — turns out the rollback plan already says an empty breakdown table is the expected state when the RTU source isn't populated, so I went with that: when `use_rtu` is off, we skip rebuilding breakdown and just clear whatever's there for that period instead. Nothing changes on the RTU-on path.

Fixed in e84d50a37, with new tests covering both branches (`TestLegacyDistributePath` in `test_phase2_rates_to_usage.py`). Also did a quick red/green check — reverted the fix locally, confirmed the new test fails, put it back, confirmed it passes — plus ran the full distribution/orchestration suite against real Postgres to make sure nothing else broke.

jordigilh:
@djnakabaale @myersCody all checks passed
https://github.com/project-koku/koku/pull/6163/checks?check_run_id=95746856983
PTAL.

djnakabaale (approved):

PR #6150, opened 2026-06-29
[COST-7252] Static exchange rate CRUD

ELK4N4:
## Summary

**PR 2** of the currency refactor series, split from #6005.

- Add CRUD endpoints for `StaticExchangeRate` (tenant-scoped, admin-defined rates with monthly validity periods)
- Serializer with validation: ISO 4217 codes, no same-pair, `end_date >= start_date`, no past-month edits, no overlapping date ranges, `exchange_rate > 0`
- `GET settings/currency/` now includes `static_rates` nested under each currency
- Computed `name` property (`"{base}-{target}"`)
- No impact on rate calculations — dynamic rates are still used everywhere

**Depends on:**
- Migration PR: #6148

**Split from:**
- #6005

---

### `POST /api/cost-management/v1/settings/currency/static-rates/`

Create a new static exchange rate.

**Request:**
```json
{
    "base_currency": "USD",
    "target_currency": "EUR",
    "exchange_rate": 0.92,
    "start_date": "2026-07-01",
    "end_date": "2026-09-30"
}
```

**Response (201):**
```json
{
    "uuid": "a1b2c3d4-e5f6-7890-abcd-ef1234567890",
    "name": "USD-EUR",
    "base_currency": "USD",
    "target_currency": "EUR",
    "exchange_rate": "0.920000",
    "start_date": "2026-07-01",
    "end_date": "2026-09-30",
    "created_timestamp": "2026-06-29T12:00:00Z",
    "updated_timestamp": "2026-06-29T12:00:00Z"
}
```

### `PUT /api/cost-management/v1/settings/currency/static-rates/<uuid>/`

Update an existing static exchange rate (cannot edit past-month rates).

**Request:**
```json
{
    "base_currency": "USD",
    "target_currency": "EUR",
    "exchange_rate": 0.95,
    "start_date": "2026-07-01",
    "end_date": "2026-09-30"
}
```

**Response (200):**
```json
{
    "uuid": "a1b2c3d4-e5f6-7890-abcd-ef1234567890",
    "name": "USD-EUR",
    "base_currency": "USD",
    "target_currency": "EUR",
    "exchange_rate": "0.950000",
    "start_date": "2026-07-01",
    "end_date": "2026-09-30",
    "created_timestamp": "2026-06-29T12:00:00Z",
    "updated_timestamp": "2026-06-29T12:05:00Z"
}
```

### `DELETE /api/cost-management/v1/settings/currency/static-rates/<uuid>/`

[...]

PR #6167, opened 2026-07-06
[COST-7252] Sync MonthlyExchangeRate on rate and currency changes

ELK4N4:
## Summary

**PR 3** of the currency refactor series, split from #6005.

Keeps `MonthlyExchangeRate` rows in sync whenever exchange rate data changes:

- **Celery daily task** — populates dynamic rates for all tenants after fetching new rates
- **Static rate CRUD** — upserts/removes monthly rows on create, update, and delete
- **Currency enable/disable** — adds or removes monthly rows for the toggled currency
- **Past-month immutability** — only the current month is writable; finalized months are read-only

No impact on report queries — reports still consume rates the old way.

**Depends on:**

- Static exchange rate CRUD: #6150
- MonthlyExchangeRate model + migration: #6161

**Split from:**

- [COST-7252] Implement constant currency Phase 1 #6005


koku/cost_models/static_exchange_rate_view.py
```diff
@@ -75,23 +78,29 @@ def put(self, request, *args, **kwargs):
...
 
         today = timezone.now().date()
         current_month_start = today.replace(day=1)
-        if instance.end_date < current_month_start:
-            raise ValidationError(
-                f"This rate ended on {instance.end_date} and all its months have been finalized. "
-                "Finalized rates cannot be deleted."
-            )
-
+        if instance.start_date < current_month_start:
```
ELK4N4, line 87: @bacciotti Out of scope, but I slipped in a quick fix here.
Resolved a bug where users could delete finalized rates. Now, finalized rates cannot be deleted

bacciotti (approved):

PR #6172, opened 2026-07-07
Improve unleash feature control for RTU

myersCody:
# Summary

This PR implements a **smart revert** of the Phase 3 RTU migration so that the existing SQL filenames return to legacy behavior by default, while preserving the RTU logic in parallel `*_rtu.sql` templates behind a feature flag.

- Restores original monthly/tag/VM SQL files to write directly to `reporting_ocpusagelineitem_daily_summary`
- Adds `*_rtu.sql` variants that retain Phase 3 behavior writing to `rates_to_usage`
- Reintroduces a real runtime flag gate in the cost model updater:
  - **Flag OFF (default/safe):** legacy write path to daily summary
  - **Flag ON:** RTU write path + aggregate back to daily summary

# Why

Phase 3 rewired core cost-model SQL to RTU-only behavior, which increased rollout risk.  
This change preserves the RTU work but makes the default path safe and reversible while remaining ready for staged re-enable after known RTU gaps are addressed.

# What Changed

- Added `use_rtu: bool = False` selector in OCP DB accessor methods to choose between legacy vs `*_rtu.sql` templates
- Threaded `use_rtu` through cost updater private methods and restored true flag-based orchestration in `update_summary_cost_model_costs`
- Restored legacy SQL and added paired RTU variants:
  - PostgreSQL cost-model SQL (`sql/openshift/cost_model/`) — 9 legacy + 9 `_rtu`
  - Trino cost-model SQL (`trino_sql/openshift/cost_model/`) — 8 legacy + 8 `_rtu`
  - Self-hosted SQL (`self_hosted_sql/openshift/cost_model/`) — mirrored 8 legacy + 8 `_rtu`
- Updated orchestration tests in `test_phase2_rates_to_usage.py`:
  - TC-55 now validates legacy behavior when flag is OFF
  - TC-56 added for full legacy execution order
  - RTU-path mocks updated for new keyword args (`use_rtu=True`)

# Behavior by Flag

- **Flag OFF** (`cost-management.backend.cost_breakdown_rates_to_usage` disabled):
  - usage/monthly/tag/vm SQL writes directly to daily summary
  - no RTU aggregate step
- **Flag ON**:
  - writes to `rates_to_usage` via `*_rtu.sql`
  - runs aggregate to rebuild daily summary from RTU rows

# Validation / Test Plan

- [x] Unit tests updated for both flag-off and flag-on orchestration expectations
- [ ] Manual verify flag OFF:
  - trigger `update_cost_model_costs`
[...]

PR #6173, opened 2026-07-07
[COST-7249] Add RTU capacity/metadata columns + indexes (schema only, split from #6163)

jordigilh:
## Summary

Splits the schema change out of #6163 per [review feedback](https://github.com/project-koku/koku/pull/6163#discussion_r3537972228): the RTU capacity/metadata columns and index added here are read/written by #6163's SQL changes (`insert_usage_rates_to_usage.sql`, `aggregate_rates_to_daily_summary.sql`), so this schema-only change needs to land and deploy first.

- Adds `node_capacity_cpu_cores`, `node_capacity_cpu_core_hours`, `node_capacity_memory_gigabytes`, `node_capacity_memory_gigabyte_hours`, `cluster_capacity_cpu_core_hours`, `cluster_capacity_memory_gigabyte_hours`, `resource_id`, `persistentvolume`, `storageclass` to `RatesToUsage` — all nullable, no backfill required.
- Adds `ratestousage_src_rp_mct_idx` on `RatesToUsage` (`source_uuid`, `report_period`, `usage_start`, `monthly_cost_type`).

Purely additive: no code in this PR reads or writes the new columns/index yet, so it's safe to merge and deploy independently of #6163. Once this lands, #6163 will rebase to depend on it.

**Rebased on top of the merged Unleash/migration rework** (#6172, #6187, #6207, #6218, #6219): the migration was renumbered from `0352_rtu_capacity_columns_and_distribution_indexes` to `0354_rtu_capacity_columns_and_distribution_indexes` to avoid colliding with the now-merged `0352_rtu_schema_improvements`/`0353_rtu_schema_improvements`, and its dependency updated accordingly. The `ratestousage_src_rp_mct_idx` index and model field now reference `report_period` (renamed from `report_period_id` upstream) instead of the stale name.

Dropped `summary_ocp_rp_start_ds_idx` on `OCPUsageLineItemDailySummary` from the original scope: `EXPLAIN (ANALYZE, BUFFERS)` testing against production-sized data showed no query plan benefit (planner never chose it) and ~7.3% write overhead, so it wasn't worth carrying forward (see #6163 discussion).

Also fixed `reporting/provider/ocp/test/test_rtu_migrations.py` (added upstream in #6219) to stay correct once a migration exists beyond `0353_rtu_schema_improvements`: it previously used a hardcoded `MIGRATE_TO = 0353` target and the live `RatesToUsage` model for ORM writes/cascades while manipulating the DB schema directly via `MigrationExecutor`, which broke once `0354` added new columns. Now:
- `_run_migration()` captures the historical model matching the state it just migrated to, and row creation writes through that model.
- `tearDown()` restores to the latest migration graph leaf *before* running fixture cleanup (which cascades through the live model).
- The CASCADE tests and restore helper target the dynamic leaf node of the `reporting` app's migration graph instead of a hardcoded `0353`.
- FK fields are now assigned by raw id (`rate_id=`, `cost_model_id=`, `source_uuid_id=`) instead of model instance, since historical and live model classes aren't interchangeable for FK assignment even when the underlying table is identical.

## Test plan

- [x] `makemigrations --check --dry-run` — no changes detected (migration matches model state)
- [x] `reporting.provider.ocp.test.test_rtu_migrations` — 5/5 passing (verified locally against PostgreSQL 16, fresh test DB)
- [x] `reporting.provider.ocp.test` (full package) — 5/5 passing
- [x] `masu.test.processor.ocp.test_phase2_rates_to_usage` — 88/88 passing (5 skipped)
- [ ] CI: migration applies cleanly, no model/migration drift

## Summary by Sourcery

Add RTU schema support for capacity and metadata fields and make RTU migration tests resilient to future migrations.

New Features:
- Add nullable RTU capacity and metadata columns to the RatesToUsage model for node and cluster resources.
- Add a composite index on RatesToUsage covering source, report period, usage start, and monthly cost type to support downstream queries.

Enhancements:
- Introduce a new RTU schema migration to add capacity fields and the composite index without backfilling data.
- Update RTU migration tests to use historical models, dynamic migration targets, and model_bakery to remain compatible with new migrations.


koku/reporting/provider/ocp/models.py
```diff
@@ -129,6 +129,10 @@ class Meta:
             models.Index(fields=["data_source"], name="summary_data_source_idx"),
             models.Index(fields=["monthly_cost_type"], name="monthly_cost_type_idx"),
             models.Index(fields=["cost_model_rate_type"], name="cost_model_rate_type_idx"),
+            models.Index(
+                fields=["report_period", "usage_start", "data_source"],
+                name="summary_ocp_rp_start_ds_idx",
+            ),
```
djnakabaale: Can we verify the performance impact of `summary_ocp_rp_start_ds_idx` using `EXPLAIN ANALYZE` before merging?

My main concern is index bloat and write amplification on a table that already handles heavy summarization INSERTs. The temp table materialization (pr #6163) seems to halve the number of scans on this table. I'd like to understand whether the remaining scans are bottlenecked on scan cost (where an index helps) vs. aggregation/join cost (where it doesn't).

jordigilh: Did the EXPLAIN ANALYZE on helios08 (seeded a synthetic 150K-row dataset — 5,000 namespaces × 200 nodes × 30 days — layered onto the existing OCP fixture data, ~250K rows total in the partition) against the two live query shapes this index targets, from `create_distribution_temp_tables.sql`'s `tmp_dist_denominator_pod`/`tmp_dist_namespace_usage_pod` CTEs (`report_period_id = X AND usage_start BETWEEN ... AND data_source = 'Pod'`), plus the "all data source" variant.

**Your hypothesis is correct — the planner never uses this index, at any row count tested:**

```
Parallel Seq Scan on reporting_ocpusagelineitem_daily_summary_default lids
  Filter: (usage_start >= ... AND usage_start <= ... AND report_period_id = 1 AND data_source = 'Pod' AND ...)
```

This held with the index both absent *and* freshly created + `ANALYZE`d — same plan, same scan node, every time. Reason: the filter matches ~82-99% of the rows in the partition (a single-provider/single-report-period query against a table already scoped to one tenant schema and, in production, one monthly partition, isn't selective — it's most of the partition). At that selectivity, a sequential scan is cheaper than an index scan's per-row heap fetches, so Postgres correctly ignores the index regardless of whether it exists.

**Where the time actually goes**, confirming the second half of your question — scan/join cost is small and constant; the aggregation is what dominates and scales with output cardinality:

| Query | Scan+Join | Total execution |
|---|---|---|
| `denominator_pod` shape (GROUP BY usage_start, cluster_id — 60 output rows) | ~40ms | 77ms |
| `namespace_usage_pod` shape (GROUP BY + namespace, node — 174K output rows) | ~50ms | 766ms (Sort + GroupAggregate, external merge to disk) |

The namespace/node-grained query is bottlenecked entirely on `Sort`/`GroupAggregate` (spilling to disk at this row count), not the scan — an index on `(report_period, usage_start, data_source)` can't help that; it would need a different structure entirely (e.g., an index matching the GROUP BY columns, or reducing output cardinality upstream) if that aggregation cost needs addressing later.

**Write-side cost, since the index provides no measured read benefit:** timed bulk `INSERT` of 50K rows into `daily_summary`, 3 trials each way:

| | Trial 1 | Trial 2 | Trial 3 | Mean |
|---|---|---|---|---|
| With index | 1755.6ms | 1691.5ms | 1679.3ms | 1708.8ms |
| Without | 1630.4ms | 1556.5ms | 1590.0ms | 1592.3ms |

~7.3% slower INSERT with this one additional index, on a table that already carries 13 other indexes and sees heavy summarization write volume — a real, non-trivial cost for an index the planner never picks up.

**Conclusion: dropping `summary_ocp_rp_start_ds_idx` from #6173.** It doesn't help the query it was added for, and it's pure write overhead. `ratestousage_src_rp_mct_idx` (the other new index, on `RatesToUsage`) is separate — happy to run the same EXPLAIN ANALYZE treatment on it if you'd like before merging, but wanted to get this one's answer back to you first since it's the more clear-cut case.

djnakabaale (approved):
PR lgtm!

One note on the unit tests: I think the CI's migrate_schemas + makemigrations check may be sufficient for these. Otherwise once they run in production they're immediately obsolete, and the maintenance cost is visible here. I'd question the long-term ROI of unit-testing these additive schema migrations.

jordigilh:
@djnakabaale @esebesto @myersCody I'll be switching to full-smoke-tests and validate no regressions, once it passes I'll switch to hot fix smokes and rebase as needed.

> I'd question the long-term ROI of unit-testing these additive schema migrations.

I can remove them if that is a general opinion. I don't mind :)

jordigilh:
retesting with the full smoke tests suite, once it passes I'll switch to hot fix smokes

jordigilh:
Closing/reopening to re-trigger the GH Actions CI workflow, which never ran for the latest commit (f0d0b832) — likely dropped during yesterday's GitHub Actions outage (now resolved).

jordigilh:
## Post-mortem: IQE smoke-test failures on `koku-ci-hm6wk` are a known midnight date-boundary flake, not a code regression

**Run:** `koku-ci-hm6wk` (`s3://konflux-artifacts/cost-mgmt-dev-tenant/koku-ci-hm6wk/`)
**IQE run window:** `2026-08-06 23:49:16 UTC` (pipeline start) → IQE suite itself ran `2026-08-07 00:17:53` → `04:52:02 UTC`
**Result:** 10,412 tests, **27 failed**, 0 errors, 39 skipped — all 27 reduce to one root cause.

### Root cause

`test_all_data_ingest_check.py:427` computes the expected day-count like this:

```python
hours = today.hour
expected_number_days = get_end_day_previous_month() + today.day
# we have no data for today until 1 a.m.
if hours == 0:
    expected_number_days -= 1
```

The suite kicked off at `00:17:53 UTC` — squarely inside the `hour == 0` window the code is trying to special-case. The heuristic assumes "no data for today until 1 a.m.," but in this run the current day's data *had* already landed by 00:17, so `expected_number_days` was decremented when it shouldn't have been. Every failure is this off-by-one, just surfacing through different assertions:

- **12 failures** — `test_api_ocp_for_{aws,azure,gcp}_ingest_single_source_check[cost|memory|compute|volume]`: literal `assert 38 == 37`.
- **12 failures** — the `*_multi_sources_check` variants of the same tests: same `38 == 37`.
- **3 failures** — GCP raw-calc reconciliation (`test_api_ocp_on_gcp_source_raw_calc`, `test_api_gcp_source_raw_calc`, `test_api_ocp_on_gcp_storage_raw_calc_clusters_sharing_csi_disk`): totals mismatched by ~$81–86, consistent with exactly one extra day of GCP daily cost being summed into the "supplied" total vs. an "expected" total computed for 37 days.

### Why this isn't a code issue

This PR is schema-only (migration + model fields/indexes on `rates_to_usage`, no processing/API behavior change). I grepped the full `iqe-cost_management-sequential.log` for `500`/`Internal Server Error` (0 hits) and for any migration/DB error referencing `rates_to_usage` or the new columns — none found. The failures are 100% explained by the `hour == 0` day-count heuristic above.

### Suggested fix for IQE

The `hours == 0` special-case is racing real ingestion timing instead of checking it. A more robust fix would query the actual max `date` present in the report response and compare day-count against that, rather than guessing based on wall-clock hour. Happy to pair on this if useful — flagging so it's tracked rather than re-discovered next time a run happens to start near UTC midnight.

jordigilh:
## Second post-mortem: remaining `koku-ci` failure is unrelated API-contract drift, not a code regression

**Run:** `koku-ci-wlvch` — big improvement over the last run: **1 failure** (down from 27), 10,412 tests, 0 errors. The midnight date-boundary flake from the previous run appears to be already fixed on the QE side, since this run didn't reproduce any of those 27 failures.

**Remaining failure:** `test_api_settings_currency_discovery_endpoint` asserts `actual_fields == required_fields` (strict set equality) against `GET /settings/currency/`, and fails because the response now includes a field the test doesn't know about:
```
AssertionError: Currency AED field mismatch - missing: set(), extra: {'is_disableable'}
```

**Root cause:** `is_disableable` was added by [PR #6245](https://github.com/project-koku/koku/pull/6245) ([COST-7993](https://redhat.atlassian.net/browse/COST-7993)), merged into `main` after this branch's last sync. The QE coverage ticket for this endpoint, [COST-7947](https://redhat.atlassian.net/browse/COST-7947), was closed **2026-07-29**, before `is_disableable` existed, and its test's required-field list hasn't been updated since.

This gap has been latent on `main` since #6245 merged — it just hadn't surfaced yet. #6245 ran under `ocp-smoke-tests` (~4,175 tests), which doesn't include this general settings test at all. This PR is labeled `full-run-smoke-tests` (the full ~10,412-test suite), which is what actually exercises it. So this isn't specific to our schema-only RTU changes — it's the first `full-run-smoke-tests` run to hit `main` since #6245 landed.

**Not retesting yet** — verified the test's expected field list still doesn't account for `is_disableable`, so a retest would just reproduce this. @pedrolp85  (tagged as assignee on the underlying test) — could you confirm if a fix is in flight, or should we file a QE follow-up ticket against COST-7993/COST-7947?

esebesto:
[full-run-smoke-tests](https://konflux-ui.apps.stone-prd-rh01.pg1f.p1.openshiftapps.com/ns/cost-mgmt-dev-tenant/pipelinerun/koku-ci-wlvch):
`= 1 failed, 10372 passed, 39 skipped, 1013 deselected, 316 warnings in 17331.90s (4:48:51) =`
```
FAILED tests/rest_api/v1/test_user_settings.py::test_api_settings_currency_discovery_endpoint - AssertionError: Currency AED field mismatch - missing: set(), extra: {'is_disableable'}
assert {'code', 'des..., 'name', ...} == {'code', 'des...c_rates', ...}
```
Since this failure was caused by a regression in main, unrelated to this PR, we can consider this full smoke run as a clean pass -> switching to `hot-fix-smoke-tests label`

PR #6175, opened 2026-07-08
[COST-7252] Wire per-month exchange rate resolution into query handlers and forecasts

ELK4N4:
## Summary

- Wire per-month exchange rate resolution from `MonthlyExchangeRate` into report query handlers and forecast handlers, gated behind a feature flag
- Add exchange rate validation that returns 400 when rates are missing or have monthly coverage gaps
- Add `base_currency` and `exchange_rate` columns to CSV exports for per-row currency transparency
- Add admin masu endpoint (`monthly_exchange_rates/`) to inspect stored exchange rate rows

## Test plan

- [ ] Verify report endpoints return correct costs when feature flag is enabled
- [ ] Verify report endpoints behave unchanged when feature flag is disabled
- [ ] Verify forecast endpoints validate exchange rate coverage before running predictions
- [ ] Verify CSV exports include `base_currency` and `exchange_rate` columns
- [ ] Verify 400 is returned when exchange rates are missing for a required currency pair
- [ ] Verify OCP reports validate both infrastructure and cost model currencies
- [ ] Verify masu `monthly_exchange_rates/` endpoint returns stored rates with optional filters

### DIFF
--- a/koku/masu/api/monthly_exchange_rates.py
+++ b/koku/masu/api/monthly_exchange_rates.py
@@ -6,6 +6,7 @@
 import datetime
 
 from django.views.decorators.cache import never_cache
+from django_tenants.utils import schema_context
 from rest_framework.decorators import api_view
 from rest_framework.decorators import permission_classes
 from rest_framework.decorators import renderer_classes
@@ -25,11 +26,16 @@ def monthly_exchange_rates(request):
     """Return stored MonthlyExchangeRate rows, optionally filtered by date range and currency pair.
 
     Query parameters:
-        start_date  - filter effective_date >= (YYYY-MM-DD, default: no lower bound)
-        end_date    - filter effective_date <= (YYYY-MM-DD, default: no upper bound)
+        schema          - tenant schema name (required)
+        start_date      - filter effective_date >= (YYYY-MM-DD, default: no lower bound)
+        end_date        - filter effective_date <= (YYYY-MM-DD, default: no upper bound)
         base_currency   - filter by base currency code
         target_currency - filter by target currency code
     """
+    schema = request.query_params.get("schema")
+    if not schema:
+        raise ValidationError({"schema": "This parameter is required."})
+
     filters = {}
     if start := request.query_params.get("start_date"):
         try:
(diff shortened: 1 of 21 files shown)


bacciotti:
I noticed we added CONSTANT_CURRENCY_FLAG as a gate for the query handler behavior. just confirming: is this intentionally SaaS only for now? On-prem uses MockUnleashClient, which returns False for any flag not listed in ONPREM_FLAG_DEFAULTS... so the feature would be permanently off onprem as written. Just want to make sure this is the intended scope for this PR.

koku/masu/api/monthly_exchange_rates.py
```diff
@@ -0,0 +1,57 @@
...
+    """Return stored MonthlyExchangeRate rows, optionally filtered by date range and currency pair.
+
+    Query parameters:
+        start_date  - filter effective_date >= (YYYY-MM-DD, default: no lower bound)
+        end_date    - filter effective_date <= (YYYY-MM-DD, default: no upper bound)
+        base_currency   - filter by base currency code
+        target_currency - filter by target currency code
+    """
+    filters = {}
+    if start := request.query_params.get("start_date"):
```
bacciotti, line 51: `masu/api/monthly_exchange_rates.py` queries `MonthlyExchangeRate` without a
`schema_context`, causing a 500 in any environment:

    ProgrammingError: relation "monthly_exchange_rate" does not exist

`MonthlyExchangeRate` lives in the tenant schema, not `public`. The endpoint
needs a `schema` query param (same pattern as `purge_trino_files.py`) and a
`with schema_context(schema):` block around the ORM query.

Reproduced locally: HTTP 500 regardless of identity header.

ELK4N4:
> I noticed we added CONSTANT_CURRENCY_FLAG as a gate for the query handler behavior. just confirming: is this intentionally SaaS only for now? On-prem uses MockUnleashClient, which returns False for any flag not listed in ONPREM_FLAG_DEFAULTS... so the feature would be permanently off onprem as written. Just want to make sure this is the intended scope for this PR.

@bacciotti Good point. I didn't know about the ONPREM flags. Fixed

koku/masu/api/monthly_exchange_rates.py
ELK4N4: Fixed and added tests for this endpoint as well

bacciotti (approved):

PR #6177, opened 2026-07-10
Rtu improvements

jvcsizilio:
## Description
Jira: https://redhat.atlassian.net/browse/COST-7736

PR for truncate, indexes and schema improvements to `rates_to_usage`.

**Do not merge until #6172 is merged and the Unleash flag `cost-management.backend.cost_breakdown_rates_to_usage` is OFF in prod.**

### What this PR does

| Commit | Change |
|--------|--------|
| 1 | **`0352`** — `TRUNCATE TABLE rates_to_usage` across all tenant schemas + indexes on `rate_id` and `cost_model_id` (GAP-5) |
| 2 | **`0353`** — Change `rate` and `cost_model` FK `on_delete` from `SET_NULL` to `CASCADE` (team policy; Django ORM metadata only) |
| 3 | Migration tests for truncate, indexes, and CASCADE behavior |

### What this PR does **not** do (follow-up commits after team answers)

- DB-level `ON DELETE CASCADE` via `RunSQL`
- GAP-1 (`sync_rate_table` / `schema_context`)
- GAP-3 (PriceList → Rate → RTU path)
- GAP-4 (`report_period_id` orphan cleanup)
- GAP-6 / GAP-7 (pipeline SQL scoped DELETEs)
- Architecture doc updates (deferred until changes are final)

### Context

- RTU is derived pipeline data; `rate_id` and `cost_model_id` are write-only traceability fields.
- Prod may still have ~15.6M RTU rows when this runs — `0352` truncates as part of migration (not relying on a manual ops clear).
- Django 4.2 keeps PG FK constraints as `NO ACTION DEFERRABLE`; `AlterField` only changes Django Collector behavior, not the database constraint.

---

### Open questions for team review

Please comment on each — follow-up commits will depend on answers.

---

#### Q1: DB-level `ON DELETE CASCADE` on `rate_id` and `cost_model_id`?

**Context:** `0353` sets Django `on_delete=CASCADE` only. The PG constraint remains `NO ACTION DEFERRABLE INITIALLY DEFERRED`. Django Collector handles CASCADE in Python; the DB does not.

**Why it matters:** GAP-3 (PriceList delete → DB-level CASCADE to `Rate` → RTU) bypasses Django Collector. ORM-only CASCADE may not fix that path.

| Option | Pros | Cons |
|--------|------|------|
| **A. Django-only CASCADE (this PR)** | Minimal migration; matches Martin review short-term recommendation | GAP-3 may still 500; deferred FK trigger bugs on partitioned tables (PG #16908) |
| **B. Add `RunSQL` to set `ON DELETE CASCADE` on both FKs** | DB enforces cleanup on all delete paths including PriceList → Rate cascade | Constraint names per schema; partitioned FK quirks; need to verify on stage |
| **C. Hybrid: DB CASCADE on `rate_id` only; Django CASCADE on `cost_model_id`** | Fixes GAP-3/PriceList path for the hot `rate_id` FK | Asymmetric; still may hit PG partitioned-FK bugs |

**Recommendation:** Start with A (this PR), add B in a follow-up if GAP-3 is confirmed on stage/prod — unless team wants B before merge.

---

#### Q2: Production PostgreSQL version — is prod on 16.5+?

**Context:** PG Bug [#16908](https://www.postgresql.org/message-id/16908-5b53ef9b31fb94c9@postgresql.org) (deferred FK trigger corruption on partitioned tables after ATTACH/DETACH) was fixed in **PG 16.5** (Nov 2024). Koku targets PG 16; deploy config does not pin minor version.

**Why it matters:** May explain why `on_delete=SET_NULL` failed in prod despite Django Collector doing the right thing (COST-7736 / PROD-5ER). Affects whether we need `cascade_delete()`, DB-level CASCADE, GAP-1 app fixes, or an ops upgrade.

| Option | Pros | Cons |
|--------|------|------|
| **A. Verify prod minor version** (gabi-cli / platform team) | Informs all FK strategy decisions | Ops task, not code |
| **B. Upgrade to PG 16.5+** if below | Fixes root trigger corruption class | Ops coordination, not this PR |
| **C. Assume bug still present; add app-level workarounds** | Defense in depth regardless of PG version | More code paths to maintain |

**Recommendation:** Run Martin's diagnostic FK query on a sample tenant schema and confirm prod PG minor version before deciding on Q1 and GAP-1.

---

#### Q3: GAP-4 — how should we handle `report_period_id` orphans?

**Context:** `report_period_id` is intentionally a bare `IntegerField` (no FK), likely because partitioned tables + PG FKs are painful. Orphans occur on single-period re-ingestion and provider-UUID purge (`ocp_report_db_cleaner` uses `cascade_delete` on report periods but RTU is not linked).

| Option | Pros | Cons |
|--------|------|------|
| **A. Explicit `filter().delete()` at call sites** (GAP-4 option A in gap analysis) | Simple; no model change | Provider-wide delete by `source_uuid` can seq-scan millions of rows → timeout risk on whale tenants |
| **B. `ForeignKey(OCPUsageReportPeriod, on_delete=CASCADE, null=True, db_constraint=False)`** (GAP-4 option B) | Django Collector deletes RTU when report period deleted via ORM; scoped by `report_period_id` | No DB enforcement; still needs correct `schema_context`; migration + pipeline SQL may need updates |
| **C. B + index on `(source_uuid, report_period_id)`** if provider-wide deletes remain | Helps any remaining bulk delete paths | Extra index; composite `(usage_start, source_uuid, report_period_id)` already exists but is usage_start-leading |

**Recommendation:** Prefer **B** over provider-wide manual delete (A) for performance semantics. Confirm `cascade_delete()` / purge paths run in tenant `schema_context` so Django CASCADE fires. No DB-level CASCADE needed for B (`db_constraint=False`).

---

#### Q4: GAP-1 — explicit RTU cleanup in `sync_rate_table()`?

**Context:** `sync_rate_table()` deletes `Rate` rows without guaranteed `schema_context`; production hit COST-7736. #6166 (null-before-delete) was abandoned in favor of CASCADE.

| Option | Pros | Cons |
|--------|------|------|
| **A. Defer** (current plan) | Avoids redundant code if Q1=B or Q2=B fixes root cause | GAP-1 may recur when RTU flag is re-enabled |
| **B. Add `RatesToUsage.objects.filter(rate_id__in=...).delete()` before Rate delete** (+ `schema` kwarg) | Defense in depth; works regardless of PG bugs | Extra code; may be redundant after DB CASCADE |
| **C. Use `cascade_delete()` from `koku/database.py`** on cost model / rate delete paths | Project pattern; `SET CONSTRAINTS ALL IMMEDIATE` | Broader refactor; overlaps with CASCADE migration |

**Recommendation:** **Defer** until Q1 and Q2 are answered. If we stay Django-only CASCADE on PG < 16.5, add **B** before re-enabling the flag.

---

#### Q5: GAP-3 — PriceList delete → Rate CASCADE → RTU?

**Context:** `Rate.price_list` is DB-level CASCADE. PriceList delete can delete Rates without Django Collector seeing RTU.

| Option | Pros | Cons |
|--------|------|------|
| **A. Depends on Q1=B** (DB CASCADE on `rate_id`) | Single migration fix | Same partitioned-FK risks as Q1 |
| **B. Pre-delete RTU in `PriceListManager.delete()`** within `schema_context` | ORM-only; no constraint migration | Duplicates logic; must pass `schema` kwarg (same as GAP-1) |
| **C. Both** | Belt and suspenders | Most code |

**Recommendation:** Tie to **Q1** — if no DB CASCADE, add **B** before RTU re-enable.

---

#### Q6: GAP-6 / GAP-7 — pipeline duplicate RTU rows?

**Context:** Monthly and tag cost SQL INSERT into RTU without scoped DELETE (duplicate costs when flag is ON). Mitigated while flag is OFF.

| Option | Pros | Cons |
|--------|------|------|
| **A. Defer until #6172 merges** (`*_rtu.sql` variants land) | Edit the right files once | Duplicates accumulate if flag turned on early |
| **B. Add scoped DELETE now** to current SQL + `ocp_cost_model_cost_updater._delete_tag_usage_costs` | Ready before flag re-enable | May need rework when smart revert lands |

**Recommendation:** **A** — wait for #6172, then one commit per gap (GAP-6 SQL, GAP-7 updater).

---

#### Q7: Anything else before re-enabling the RTU flag?

Checklist for team:

- [x] #6172 merged and deployed
- [ ] Resummarize daily summary via legacy path if needed (`update_cost_model_costs`)
- [ ] This PR merged and migrated in prod
- [ ] Q1–Q6 resolved
- [ ] Stage integration test with flag ON
- [ ] Incremental prod rollout per tenant

---

## Testing

1. Checkout branch `rtu_improvements`
2. Ensure test DB is running (`localhost:15432`, `KEEPDB=True`)
3. Run migration tests:

```bash
pipenv run python koku/manage.py test reporting.provider.ocp.test.test_rtu_migrations --no-input -v 2
```

4. Manual verification on test schema `org1234567` (optional):

```bash
pipenv run python koku/manage.py migrate reporting 0353 --schema=org1234567
```

```sql
SET search_path TO org1234567;
SELECT count(*) FROM rates_to_usage;  -- expect 0 after 0352

SELECT indexname FROM pg_indexes
WHERE schemaname = 'org1234567' AND tablename = 'rates_to_usage'
  AND indexname IN ('ratestousage_rate_id_idx', 'ratestousage_cost_model_id_idx');

-- FK still NO ACTION at DB level until Q1 is decided:
SELECT c.conname, c.confdeltype  -- 'a' = NO ACTION
FROM pg_constraint c
WHERE c.conrelid = 'org1234567.rates_to_usage'::regclass AND c.contype = 'f';
```

5. Expected test results: 4 tests pass (truncate, indexes, CASCADE on rate delete, CASCADE on cost model delete)

---

## References

- [rates-to-usage-gap-analysis.md](docs/architecture/cost-breakdown/rates-to-usage-gap-analysis.md)
- [rtu-issues-and-path-forward.md](docs/architecture/cost-breakdown/rtu-issues-and-path-forward.md)
- Smart revert prerequisite: [#6172](https://github.com/project-koku/koku/pull/6172)
- Martin adversarial review: [pr-6166-review.md](https://github.com/project-koku/koku/blob/pr-6166-adversarial-review/pr-6166-review.md)


koku/reporting/provider/ocp/models.py
```diff
@@ -1059,11 +1059,13 @@ class Meta:
...
+            models.Index(fields=["cost_model_id"], name="ratestousage_cost_model_id_idx"),
         ]
 
     uuid = models.UUIDField(primary_key=True, default=uuid4)
-    rate = models.ForeignKey("cost_models.Rate", on_delete=models.SET_NULL, null=True)
-    cost_model = models.ForeignKey("cost_models.CostModel", on_delete=models.SET_NULL, null=True)
+    rate = models.ForeignKey("cost_models.Rate", on_delete=models.CASCADE, null=True)
+    cost_model = models.ForeignKey("cost_models.CostModel", on_delete=models.CASCADE, null=True)
     report_period_id = models.IntegerField(null=True)
     source_uuid = models.UUIDField()
```
myersCody: @jvcsizilio You are doing an awesome job here! 

I think the `source_uuid` & `report_period_id` are awkwardly configured for the RTU table as well. Since we will have a clean slate for the RTU table, now might a good time to fix them. 

Both of these should be FK to other tables. 

Having the source_uuid as a uuid field is forcing us to do an [anti-pattern](https://github.com/project-koku/koku/pull/6163/changes#diff-1677f6b8109045120a44a4396ec04c2b66ac344510a1851fef6e47651761d891R249-R255) for our architecture when it comes to doing data retention later on. 

Would you mind doing a little investigation into if the report_period_id should be a FK as well. My gut is telling me yes, but could be worth seeing if there is a valid reason not to.

jvcsizilio: Thanks @myersCody! 
I Investigated both fields. Confirmed the FK targets from pipeline code, tests, and sibling models (not just the current RTU model definition).

**report_period_id** >> **OCPUsageReportPeriod**

- `ocp_cost_model_cost_updater._update_usage_rates_to_usage()` calls `report_periods_for_provider_uuid()`, takes `report_period.id`, and passes it to `populate_usage_rates_to_usage()`.
- Tests assert `sql_params["report_period_id"] == OCPUsageReportPeriod.id` (test_phase2_rates_to_usage.py TC-30).
- RTU INSERT SQL filters/joins daily summary on `lids.report_period_id`, and daily summary already has `report_period = ForeignKey(OCPUsageReportPeriod)` (same integer PK).
- data-model.md line 217 documents this: "FK to reporting_ocpusagereportperiod". It was IntegerField because the FK was never wired. This is GAP-4.


**source_uuid** >> **TenantAPIProvider** (not `api.Provider`)

- Pipeline passes `self._provider.uuid` (api.Provider in public schema).
- On provider create, `TenantAPIProvider.uuid = provider.uuid` (api/provider/serializers.py).
- `OCPUsageReportPeriod.provider` already FKs to `TenantAPIProvider` and is queried via `provider_id=self._provider.uuid`.
- Sibling tables in the same feature already use the FK: `OCPCostUIBreakDownP` and all OCP `*P` summary tables use `ForeignKey(TenantAPIProvider, on_delete=CASCADE, db_column="source_uuid")`.
- RTU is the outlier with a bare UUIDField; that blocks CASCADE cleanup on provider delete (the retention anti-pattern you flagged).


Plan: Add both FKs in the migration PR with `db_column` unchanged (`source_uuid`, `report_period_id`) so SQL templates need no changes. Open question: should `source_uuid` be `null=True` like other `*P` tables?

jvcsizilio:
> 🤖 **CI Triager — Diagnosis**
> 
> **Check:** `Units - 3.11` **Commit:** `99a2482`
> 
> **Root cause:** Migration `0354_rtu_provider_report_period_fk` adds a real PostgreSQL FK constraint from `rates_to_usage.report_period_id` → `reporting_ocpusagereportperiod.id`. Existing provider/source cleanup tests fail because Django's ORM cascade does not cross the tenant schema boundary — when those tests delete an OCP provider (which cascades to delete `OcpUsageReportPeriod` rows), the `rates_to_usage` rows referencing those report periods are NOT cleaned up by the ORM. At test teardown, Django calls `SET CONSTRAINTS ALL IMMEDIATE` which then triggers the DB-level FK violation.
> 
> **Evidence:**
> 
> ```
> ERROR: test_delete_all_providers_from_queryset (api.provider.test.test_models.ProviderModelTest)
> ERROR: test_remove_all_ocp_providers (api.provider.test.test_provider_manager.ProviderManagerTest)
> ERROR: test_delete_missing_sources (masu.test.api.test_source_cleanup.SourceCleanupTests)
> ERROR: test_delete_providers_without_sources (masu.test.api.test_source_cleanup.SourceCleanupTests)
> 
> IntegrityError: update or delete on table "reporting_ocpusagereportperiod" violates foreign key
> constraint "rates_to_usage_report_period_id_1d1c7756_fk_reporting" on table "rates_to_usage"
> DETAIL:  Key (id)=(7) is still referenced from table "rates_to_usage".
>   (triggered by: cursor.execute("SET CONSTRAINTS ALL IMMEDIATE"))
> ```
> 
> **Action:** The `rates_to_usage` rows with `report_period_id` pointing to the standard test fixture's `OcpUsageReportPeriod` (id=7) must be cleaned up before provider/report-period deletion. Two approaches:
> 
> 1. **Preferred — add cascade at the DB level in the migration**: Change migration 0354's FK definition for `report_period` to use a raw `RunSQL` that adds the constraint with `ON DELETE CASCADE` instead of Django's default `NO ACTION DEFERRABLE INITIALLY DEFERRED`. This way the DB itself cascades the delete and the test teardown constraint check will not fire.
> 2. **Alternative — clean up in tests**: In the failing tests (or their base class tearDown), delete `rates_to_usage` rows explicitly before provider deletion using `with schema_context(self.schema): RatesToUsage.objects.filter(report_period__isnull=False).delete()`. But this is a symptom fix.
> 
> The deeper question to investigate: which test creates a `rates_to_usage` row with `report_period_id=7`? If it is from the new `test_rtu_migrations.py` test (run as `TransactionTestCase`, which commits data), the `_cleanup_rtu_migration_fixtures` method needs to also clean up rows that have a `report_period` set.
> 
> _Generated automatically. Review before applying._

Addressed in `squashed 0352_rtu_schema_improvements.py`. The deleted 0354 only added Django AlterField FKs (NO ACTION at PG level). The migration now ends with RunSQL that applies ON DELETE CASCADE for all RTU FKs including `report_period_id`. This fixes the four provider-delete CI failures without modifying those tests. `test_rtu_migrations` rows don't set `report_period`, so they weren't the source of the id=7 references.

myersCody:
After running migration
```
postgres=# \d rates_to_usage;
                  Partitioned table "org1234567.rates_to_usage"
        Column         |          Type          | Collation | Nullable | Default
-----------------------+------------------------+-----------+----------+---------
 uuid                  | uuid                   |           | not null |
 report_period_id      | integer                |           |          |
 source_uuid           | uuid                   |           | not null |
 usage_start           | date                   |           | not null |
 usage_end             | date                   |           | not null |
 node                  | character varying(253) |           |          |
 namespace             | character varying(253) |           |          |
 cluster_id            | text                   |           | not null |
 cluster_alias         | text                   |           |          |
 data_source           | character varying(63)  |           |          |
 persistentvolumeclaim | character varying(253) |           |          |
 pod_labels            | jsonb                  |           |          |
 volume_labels         | jsonb                  |           |          |
 all_labels            | jsonb                  |           |          |
 custom_name           | character varying(50)  |           | not null |
 metric_type           | character varying(20)  |           | not null |
 cost_model_rate_type  | text                   |           |          |
 monthly_cost_type     | text                   |           |          |
 calculated_cost       | numeric(33,15)         |           |          |
 distributed_cost      | numeric(33,15)         |           |          |
 labels                | jsonb                  |           |          |
 label_hash            | character varying(64)  |           |          |
 cost_category_id      | integer                |           |          |
 cost_model_id         | uuid                   |           |          |
 rate_id               | uuid                   |           |          |
Partition key: RANGE (usage_start)
Indexes:
    "rates_to_usage_pkey" PRIMARY KEY, btree (usage_start, uuid)
    "rates_to_usage_cluster_id_idx" btree (cluster_id)
    "rates_to_usage_cost_category_id_idx" btree (cost_category_id)
    "rates_to_usage_cost_model_id_idx" btree (cost_model_id)
    "rates_to_usage_custom_name_idx" btree (custom_name)
    "rates_to_usage_label_hash_idx" btree (label_hash)
    "rates_to_usage_monthly_cost_type_idx" btree (monthly_cost_type)
    "rates_to_usage_namespace_idx" btree (namespace)
    "rates_to_usage_rate_id_idx" btree (rate_id)
    "rates_to_usage_usage_start_source_uuid_report_period_id_idx" btree (usage_start, source_uuid, report_period_id)
    "ratestousage_cost_model_id_idx" btree (cost_model_id)
    "ratestousage_rate_id_idx" btree (rate_id)
    "ratestousage_start_src_rp_idx" btree (usage_start, source_uuid, report_period_id)
Foreign-key constraints:
    "rates_to_usage_cost_category_id_85b5b064_fk_reporting" FOREIGN KEY (cost_category_id) REFERENCES reporting_ocp_cost_category(id) DEFERRABLE INITIALLY DEFERRED
    "rates_to_usage_cost_model_id_656f6c76_fk_cost_model_uuid" FOREIGN KEY (cost_model_id) REFERENCES cost_model(uuid) ON DELETE CASCADE DEFERRABLE INITIALLY DEFERRED
    "rates_to_usage_rate_id_ee9b159e_fk_cost_model_rate_uuid" FOREIGN KEY (rate_id) REFERENCES cost_model_rate(uuid) ON DELETE CASCADE DEFERRABLE INITIALLY DEFERRED
    "rates_to_usage_report_period_id_fk" FOREIGN KEY (report_period_id) REFERENCES reporting_ocpusagereportperiod(id) ON DELETE CASCADE DEFERRABLE INITIALLY DEFERRED
    "rates_to_usage_source_uuid_fk" FOREIGN KEY (source_uuid) REFERENCES reporting_tenant_api_provider(uuid) ON DELETE CASCADE DEFERRABLE INITIALLY DEFERRED
Number of partitions: 3 (Use \d+ to list them.)
postgres=# select * from rates_to_usage;
(0 rows)
```

myersCody (dismissed):

jvcsizilio:
> After running migration
> 
> ```
> postgres=# \d rates_to_usage;
>                   Partitioned table "org1234567.rates_to_usage"
>         Column         |          Type          | Collation | Nullable | Default
> -----------------------+------------------------+-----------+----------+---------
>  uuid                  | uuid                   |           | not null |
>  report_period_id      | integer                |           |          |
>  source_uuid           | uuid                   |           | not null |
>  usage_start           | date                   |           | not null |
>  usage_end             | date                   |           | not null |
>  node                  | character varying(253) |           |          |
>  namespace             | character varying(253) |           |          |
>  cluster_id            | text                   |           | not null |
>  cluster_alias         | text                   |           |          |
>  data_source           | character varying(63)  |           |          |
>  persistentvolumeclaim | character varying(253) |           |          |
>  pod_labels            | jsonb                  |           |          |
>  volume_labels         | jsonb                  |           |          |
>  all_labels            | jsonb                  |           |          |
>  custom_name           | character varying(50)  |           | not null |
>  metric_type           | character varying(20)  |           | not null |
>  cost_model_rate_type  | text                   |           |          |
>  monthly_cost_type     | text                   |           |          |
>  calculated_cost       | numeric(33,15)         |           |          |
>  distributed_cost      | numeric(33,15)         |           |          |
>  labels                | jsonb                  |           |          |
>  label_hash            | character varying(64)  |           |          |
>  cost_category_id      | integer                |           |          |
>  cost_model_id         | uuid                   |           |          |
>  rate_id               | uuid                   |           |          |
> Partition key: RANGE (usage_start)
> Indexes:
>     "rates_to_usage_pkey" PRIMARY KEY, btree (usage_start, uuid)
>     "rates_to_usage_cluster_id_idx" btree (cluster_id)
>     "rates_to_usage_cost_category_id_idx" btree (cost_category_id)
>     "rates_to_usage_cost_model_id_idx" btree (cost_model_id)
>     "rates_to_usage_custom_name_idx" btree (custom_name)
>     "rates_to_usage_label_hash_idx" btree (label_hash)
>     "rates_to_usage_monthly_cost_type_idx" btree (monthly_cost_type)
>     "rates_to_usage_namespace_idx" btree (namespace)
>     "rates_to_usage_rate_id_idx" btree (rate_id)
>     "rates_to_usage_usage_start_source_uuid_report_period_id_idx" btree (usage_start, source_uuid, report_period_id)
>     "ratestousage_cost_model_id_idx" btree (cost_model_id)
>     "ratestousage_rate_id_idx" btree (rate_id)
>     "ratestousage_start_src_rp_idx" btree (usage_start, source_uuid, report_period_id)
> Foreign-key constraints:
>     "rates_to_usage_cost_category_id_85b5b064_fk_reporting" FOREIGN KEY (cost_category_id) REFERENCES reporting_ocp_cost_category(id) DEFERRABLE INITIALLY DEFERRED
>     "rates_to_usage_cost_model_id_656f6c76_fk_cost_model_uuid" FOREIGN KEY (cost_model_id) REFERENCES cost_model(uuid) ON DELETE CASCADE DEFERRABLE INITIALLY DEFERRED
>     "rates_to_usage_rate_id_ee9b159e_fk_cost_model_rate_uuid" FOREIGN KEY (rate_id) REFERENCES cost_model_rate(uuid) ON DELETE CASCADE DEFERRABLE INITIALLY DEFERRED
>     "rates_to_usage_report_period_id_fk" FOREIGN KEY (report_period_id) REFERENCES reporting_ocpusagereportperiod(id) ON DELETE CASCADE DEFERRABLE INITIALLY DEFERRED
>     "rates_to_usage_source_uuid_fk" FOREIGN KEY (source_uuid) REFERENCES reporting_tenant_api_provider(uuid) ON DELETE CASCADE DEFERRABLE INITIALLY DEFERRED
> Number of partitions: 3 (Use \d+ to list them.)
> postgres=# select * from rates_to_usage;
> (0 rows)
> ```

Your output showed duplicate indexes on rates_to_usage. I verified this in Postgres: three pairs index the same columns twice.

Duplicates:
- rates_to_usage_rate_id_idx and ratestousage_rate_id_idx on rate_id
- rates_to_usage_cost_model_id_idx and ratestousage_cost_model_id_idx on cost_model_id
- rates_to_usage_usage_start_source_uuid_report_period_id_idx and ratestousage_start_src_rp_idx on usage_start, source_uuid, report_period_id

Root cause: migration 0348 created FK auto-indexes on rate_id and cost_model_id (Django db_index defaults to True). Migration 0352 then added explicit GAP-5 indexes with ratestousage_* names without dropping the existing ones. The composite index had a similar overlap depending on migration path.

There is no query performance difference between the two copies. Postgres uses one index per query. The problem is extra write work and disk on every pipeline INSERT and DELETE, which matters at scale.

Why drop the auto-generated indexes instead of skipping the new AddIndex in 0352?

We considered not adding the ratestousage_* indexes and keeping only the Django auto names. We rejected that because:
- The model Meta.indexes already declare ratestousage_* names; that is what migrations and tests expect
- Auto names are long and inconsistent (rates_to_usage_rate_id_idx vs ratestousage_rate_id_idx vs hash-suffixed variants on other FKs)
- If we only kept auto indexes, Django migration state and the model would drift, and a future makemigrations could try to recreate the explicit indexes

So the fix is: keep ratestousage_* (explicit, stable, matches model), drop the three auto-named duplicates, and set db_index=False on the rate and cost_model FK fields so Django does not recreate FK auto-indexes on top of Meta.indexes going forward.

Added test_0352_drops_duplicate_auto_named_indexes to guard against this coming back.

If you already ran `0352` before this commit, you will need to roll back to `0351` and re-apply `0352`, or run the three DROP INDEX statements manually while the table is still empty.

koku/reporting/migrations/0352_rtu_schema_improvements.py
```diff
@@ -0,0 +1,205 @@
...
+    END IF;
+END $$;
+"""
+
+# Django FK db_index creates rates_to_usage_* indexes; Meta.indexes uses ratestousage_* names.
+RTU_DROP_DUPLICATE_INDEXES_SQL = """
+DROP INDEX IF EXISTS rates_to_usage_rate_id_idx;
+DROP INDEX IF EXISTS rates_to_usage_cost_model_id_idx;
+DROP INDEX IF EXISTS rates_to_usage_usage_start_source_uuid_report_period_id_idx;
+"""
```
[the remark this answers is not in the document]
jvcsizilio: The static DROP INDEX list ran after AddIndex and missed the _idx1 variants PostgreSQL created when the base _idx names were already taken.

We are not accepting the one-click suggestion to swap _idx for _idx1 because that is still fragile (suffix can vary).

Instead, the follow-up commit:
1. Drops legacy FK auto-indexes on rate_id and cost_model_id BEFORE AddIndex (RTU_DROP_LEGACY_FK_INDEXES_SQL)
2. Runs a DO block AFTER FK CASCADE that drops any duplicate index on those columns by column set, keeping only ratestousage_rate_id_idx, ratestousage_cost_model_id_idx, and ratestousage_start_src_rp_idx

This covers _idx, _idx1, and any other auto-generated name. test_0352_drops_duplicate_auto_named_indexes passes locally.

djnakabaale (approved):
Pr lgtm!  One side note on `clone_schema` + custom index names: tested this locally after ingesting real OCP data, your fix works, zero duplicate indexes and zero orphaned fks on `rates_to_usage`. 

There's an existing quirk worth knowing: tenant schemas provisioned via `clone_schema()` (`db_functions/clone_schema.sql`) uses `CREATE TABLE (LIKE ... INCLUDING ALL)`, and Postgres silently renames indexes to its own `<table>_<column>_idx` convention on clone. So `ratestousage_rate_id_idx` etc. won't necessarily exist by that literal name on a cloned tenant. Doesn't affect your fix, there's still only one index per column (no duplicates).

Worth keeping in mind for the future though: if a later migration does something like `RemoveIndex("ratestousage_rate_id_idx")`, it could silently no-op on cloned tenants.

PR #6179, opened 2026-07-12
[COST-7252] Block disabling currencies that are in use

ELK4N4:
## Summary
- Return **400** instead of allowing a disable with a warning when the currency is referenced by cost models, price lists, or is the system/user default currency.
- Per PRD requirement: *"Customers are prevented from disabling base currencies currently in use (Cloud providers, cost models, or defined as the default currency). The system MUST block the user from disabling these."*
- Adds a new test for blocking the system default currency (USD).

## Test plan
- [x] `test_disable_currency_in_use_by_cost_model_blocked` — returns 400, currency remains enabled
- [x] `test_disable_currency_in_use_by_price_list_blocked` — returns 400, currency remains enabled
- [x] `test_disable_default_currency_blocked` — returns 400, USD remains enabled
- [x] `test_disable_currency` — currencies not in use can still be disabled (204)
- [x] All existing currency view tests pass

### DIFF
--- a/koku/api/currency/exceptions.py
+++ b/koku/api/currency/exceptions.py
@@ -13,7 +13,7 @@ class ExchangeRateNotFound(ValidationError):
     def __init__(self, base_currencies, target_currency, start_date, end_date, missing_months=None):
         missing_pairs = ", ".join(f"{base} -> {target_currency}" for base in base_currencies)
         if missing_months:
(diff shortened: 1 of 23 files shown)


ELK4N4:
Closing — reopening against the correct base branch on the fork.

PR #6182, opened 2026-07-12
[COST-7252] Block disabling currencies that are in use

ELK4N4:
## Summary
- Return **400** instead of allowing a disable with a warning when the currency is referenced by cost models, price lists, or is the system/user default currency.
- Per PRD requirement: *"Customers are prevented from disabling base currencies currently in use (Cloud providers, cost models, or defined as the default currency). The system MUST block the user from disabling these."*
- Adds a new test for blocking the system default currency (USD).

## Test plan
- [x] `test_disable_currency_in_use_by_cost_model_blocked` — returns 400, currency remains enabled
- [x] `test_disable_currency_in_use_by_price_list_blocked` — returns 400, currency remains enabled
- [x] `test_disable_default_currency_blocked` — returns 400, USD remains enabled
- [x] `test_disable_currency` — currencies not in use can still be disabled (204)
- [x] All existing currency view tests pass


koku/api/settings/currency_views.py
```diff
@@ -109,44 +109,59 @@ def post(self, request, *args, **kwargs):
...
+
         if EnabledCurrency.objects.count() == 1:
             raise ValidationError({"error": "At least one currency must be enabled."})
 
+        reasons = []
+
+        if code == KOKU_DEFAULT_CURRENCY:
+            reasons.append("it is the system default currency")
+
+        account_settings = UserSettings.objects.first()
```
bacciotti, line 123: we are using `.first()` here to check if the currency is an account default. Since `UserSettings` is per-user (the previous code used `.filter()` to iterate over all of them), `.first()` may miss users who have this currency set as their default, so the block would only trigger if the first object returned happens to match.

would it make sense to use `.filter().exists()`instead, similar to how it was done before?

ELK4N4: That makes sense, but the issue is that the database currently allows multiple rows for `UserSettings` due to a bug or lack of constraints.

Because of this, using `.filter().exists()` risks matching against stale or orphaned rows that shouldn't even be there. Elsewhere in the project, we use `.first()` to fetch settings, meaning `.first()` is what the user actually sees, and the app ignores those duplicate, bugged rows anyway (see [api/utils.py](https://github.com/project-koku/koku/blob/main/koku/api/utils.py#L27-L57) and [settings/utils.py](https://github.com/project-koku/koku/blob/main/koku/api/settings/utils.py#L252-L264)).

bacciotti (approved):

PR #6194, opened 2026-07-20
[COST-7930] Block disabling currencies used by cloud provider sources

ELK4N4:
## Summary

- Prevents disabling a currency that is actively used as the billing currency by AWS, Azure, or GCP provider sources
- Queries `AWSCostSummaryP`, `AzureCostSummaryP`, and `GCPCostSummaryP` to find providers billing in the target currency, and blocks the disable with a 400 response listing the affected providers
- Adds the `affected_cloud_providers` list to the error response alongside the existing `affected_cost_models` and `affected_price_lists`

## Test plan

- [x] Unit test: `test_disable_currency_in_use_by_cloud_provider_blocked` — verifies all three cloud provider types (AWS/Azure/GCP) block currency disable
- [x] Verify existing currency disable tests still pass
- [x] Manual validation against a dev environment with cloud provider data

### DIFF
--- a/koku/api/settings/currency_views.py
+++ b/koku/api/settings/currency_views.py
@@ -6,6 +6,7 @@
 import logging
 from collections import defaultdict
 
+from django.db.models import Q
 from django.utils.decorators import method_decorator
 from django.views.decorators.cache import never_cache
 from rest_framework import status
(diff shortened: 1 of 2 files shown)


koku/api/settings/currency_views.py
```diff
@@ -30,11 +32,28 @@
...
 from reporting.user_settings.models import UserSettings
 
 LOG = logging.getLogger(__name__)
 
 
+def _get_cloud_providers_using_currency(code, customer):
+    """Return cloud providers whose billing data uses ``code`` as a base currency."""
+    aws_uuids = AWSCostSummaryP.objects.filter(currency_code=code).values_list("source_uuid", flat=True)
+    azure_uuids = AzureCostSummaryP.objects.filter(currency=code).values_list("source_uuid", flat=True)
+    gcp_uuids = GCPCostSummaryP.objects.filter(currency=code).values_list("source_uuid", flat=True)
```
bacciotti: I noticed the codebase consistently uses .distinct() when querying source_uuid from summary tables. Maybe it's worth adding   .distinct() to the three values_list calls?

ELK4N4: Good call. Updated

bacciotti (approved):

PR #6206, opened 2026-07-22
[COST-7953] Return exchange_rate as JSON number

ELK4N4:
## Summary
- Serialize `exchange_rate` as a JSON number (`float`) instead of a decimal string in static exchange rate Settings responses
- Normalize monthly exchange rate Masu responses the same way
- Update unit tests to expect numeric `exchange_rate` values

## Notes on precision
JSON itself can write numbers with many digits, but typical clients parse JSON numbers as IEEE 754 doubles (~15–17 significant digits). Our models still store `DecimalField(max_digits=33, decimal_places=15)`, so full 33-digit fidelity is preserved in the database; the API response may round beyond float precision.

For normal exchange rates (e.g. `0.92`, `1.17`) this is fine and matches what most clients expect. Exact round-trip of the full stored decimal would require a string (or another exact decimal encoding).

## Test plan
- [ ] `pipenv run tox -- cost_models.test.test_static_exchange_rate_view`
- [ ] `pipenv run tox -- masu.test.api.test_monthly_exchange_rates`
- [ ] Create/list/update a static exchange rate and confirm `exchange_rate` is a number in the JSON body
- [ ] Call Masu `monthly_exchange_rates/` and confirm `exchange_rate` values are numbers

### DIFF
--- a/koku/cost_models/static_exchange_rate_serializer.py
+++ b/koku/cost_models/static_exchange_rate_serializer.py
@@ -6,7 +6,6 @@
 import calendar
 import logging
 from datetime import timedelta
-from decimal import Decimal
 
 from django.db import transaction
 from django.utils import timezone
@@ -36,13 +35,7 @@ class TrailingZeroStrippingDecimalField(serializers.DecimalField):
         if value is None:
             return None
         # DecimalField quantizes; normalize strips trailing zeros before float().
-        val = super().to_representation(value)
-        if isinstance(val, Decimal):
-            return float(val.normalize())
-        val_str = format(val, "f") if not isinstance(val, str) else val
-        if "." in val_str:
-            val_str = val_str.rstrip("0").rstrip(".")
-        return float(val_str)
+        return float(super().to_representation(value).normalize())
 
 
 class StaticExchangeRateSerializer(serializers.ModelSerializer):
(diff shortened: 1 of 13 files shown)


bacciotti (approved):

bacciotti (changes requested):
accidentally approved before noticing @djnakabaale was already assigned as reviewer. Leaving the review to him.

koku/cost_models/static_exchange_rate_serializer.py
```diff
@@ -22,22 +22,15 @@
 
 
 class TrailingZeroStrippingDecimalField(serializers.DecimalField):
-    """Serialize Decimals as strings without unnecessary trailing zeros.
+    """Serialize Decimals as JSON numbers.
```
djnakabaale: nit suggestion: Should we rename `TrailingZeroStrippingDecimalField` to something that reflects what it actually does? It no longer strips trailing zeros which was the old implementation

ELK4N4: I thought the same but I wanted to reduce the changes in this PR.
I will rename it to `NumericDecimalField`

djnakabaale: thanks for addressing the nit feedback.

djnakabaale (approved):

bacciotti (approved):

PR #6239, opened 2026-08-01
[COST-8005] Add RequestTimeoutMiddleware to replace gunicorn SIGABRT noise

martinpovolny:
## Summary
- Adds `RequestTimeoutMiddleware` using `signal.SIGALRM` to enforce a soft timeout (90s, configurable via `REQUEST_SOFT_TIMEOUT`) before gunicorn's hard kill, replacing ~1,400+ useless `SystemExit:1` GlitchTip events with actionable `RequestTimeoutError` events containing request path, method, and duration.
- Bumps gunicorn hard timeout from 90s → 100s so the middleware fires first while preserving a 10s window for the error to propagate to GlitchTip.
- Adds `sentry_before_send` hook in `middleware.py` to tag timeout events with `timeout=soft` for GlitchTip filtering.
- Thread-safe: only arms SIGALRM in the main thread. With threaded workers (`GUNICORN_THREADS=True`, active in stage), the middleware is a no-op and gunicorn's hard timeout remains the fallback.

## Changes
| File | Change |
|------|--------|
| `koku/koku/middleware.py` | Add `RequestTimeoutError`, `sentry_before_send`, `RequestTimeoutMiddleware` (SIGALRM + main-thread guard) |
| `koku/koku/settings.py` | Insert middleware after `RequestTimingMiddleware` |
| `koku/koku/sentry.py` | Import and wire `sentry_before_send` from middleware |
| `koku/gunicorn_conf.py` | Bump default `TIMEOUT` from 90 → 100 |
| `koku/koku/test_middleware.py` | 9 new tests (6 middleware, 3 sentry hook) |
| `CLAUDE.md` | Fix smoke test label docs |

## Configuration
| Env var | Default | Description |
|---------|---------|-------------|
| `REQUEST_SOFT_TIMEOUT` | 90 | Seconds before middleware aborts the request |
| `TIMEOUT` | 100 (was 90) | Gunicorn hard timeout (SIGABRT fallback) |

## Test plan
- [x] All 9 new tests pass (100% coverage on new middleware code)
- [x] No regressions in existing middleware tests
- [x] Pre-commit lint passes
- [x] Thread-safety: SIGALRM skipped in non-main threads (tested)
- [x] Safe env parsing: bad `REQUEST_SOFT_TIMEOUT` values fall back to default (tested)
- [ ] Deploy to stage, verify GlitchTip events contain request context
- [ ] Monitor for 1 week — confirm noise reduction and actionable data

JIRA: https://redhat.atlassian.net/browse/COST-8005
Design doc: https://docs.google.com/document/d/19f3DbCfW6NoV9vXABDPjeScnY4WA_aAThTKkxCVpmVY/edit

### DIFF
--- a/koku/gunicorn_conf.py
+++ b/koku/gunicorn_conf.py
@@ -44,7 +44,7 @@ workers = ENVIRONMENT.int("GUNICORN_WORKERS", default=(cpu_resources * 2 + 1))
 gunicorn_threads = ENVIRONMENT.bool("GUNICORN_THREADS", default=False)
 if gunicorn_threads:
     threads = cpu_resources * 2 + 1
-timeout = ENVIRONMENT.int("TIMEOUT", default=90)
+timeout = ENVIRONMENT.int("TIMEOUT", default=92)
 graceful_timeout = ENVIRONMENT.int("GRACEFUL_TIMEOUT", default=180)
 
 
--- a/koku/gunicorn_conf.py
+++ b/koku/gunicorn_conf.py
@@ -44,7 +44,7 @@ workers = ENVIRONMENT.int("GUNICORN_WORKERS", default=(cpu_resources * 2 + 1))
 gunicorn_threads = ENVIRONMENT.bool("GUNICORN_THREADS", default=False)
 if gunicorn_threads:
     threads = cpu_resources * 2 + 1
-timeout = ENVIRONMENT.int("TIMEOUT", default=92)
+timeout = ENVIRONMENT.int("TIMEOUT", default=100)
 graceful_timeout = ENVIRONMENT.int("GRACEFUL_TIMEOUT", default=180)
 
 
--- a/koku/koku/middleware.py
+++ b/koku/koku/middleware.py
@@ -463,16 +463,31 @@ class RequestTimeoutError(Exception):
     """Raised when a request exceeds the soft timeout."""
 
 
+def _parse_soft_timeout(default=90):
+    raw = os.environ.get("REQUEST_SOFT_TIMEOUT")
+    if raw is None:
+        return default
+    try:
+        return int(raw)
+    except (TypeError, ValueError):
+        return default
+
+
 class RequestTimeoutMiddleware(MiddlewareMixin):
     """Abort requests that exceed a soft timeout, before gunicorn kills the worker.
 
     Uses SIGALRM to raise RequestTimeoutError with full request context,
     replacing the uninformative SystemExit:1 that gunicorn's SIGABRT produces.
+    Only active in the main thread (sync workers); with threaded workers,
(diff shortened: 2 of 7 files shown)


koku/koku/middleware.py
```diff
@@ -457,6 +459,50 @@ def process_response(self, request, response):
...
+    Uses SIGALRM to raise RequestTimeoutError with full request context,
+    replacing the uninformative SystemExit:1 that gunicorn's SIGABRT produces.
+    Only active in the main thread (sync workers); with threaded workers,
+    gunicorn's hard timeout remains the fallback.
+    """
+
+    SOFT_TIMEOUT = _parse_soft_timeout()
+
+    def process_request(self, request):
+        if threading.current_thread() is not threading.main_thread():
```
bacciotti, line 495: Prod is fine today (GUNICORN_THREADS=false in app-interface).

But this soft timeout only works on sync/main thread. Stage still defaults to
threads=True, so the COST-8005 stage verification likely will not hit this code.

Maybe it's worth documenting the `GUNICORN_THREADS=false` dependency.

koku/gunicorn_conf.py
```diff
@@ -44,7 +44,7 @@
 gunicorn_threads = ENVIRONMENT.bool("GUNICORN_THREADS", default=False)
 if gunicorn_threads:
     threads = cpu_resources * 2 + 1
-timeout = ENVIRONMENT.int("TIMEOUT", default=90)
+timeout = ENVIRONMENT.int("TIMEOUT", default=92)
```
bacciotti: I's day soft/hard gap looks too small.

Ticket suggests ~80s soft before ~90s hard. This PR uses 90s soft / 92s hard.
After `RequestTimeoutError` we still need unwind + Sentry; with only 2s, gunicorn
can still SIGABRT and we keep SystemExit:1 noise.

Can we use a bigger buffer (e.g. 80/90), aligned with the ticket?

bacciotti (commented):
Direction looks good for the GlitchTip noise.

Before approve, please:
1) clarify GUNICORN_THREADS (stage validation vs prod-only override)
2) widen soft vs hard timeout to match COST-8005

Happy to re-review after that :)

koku/gunicorn_conf.py
martinpovolny: I went for 92/90 to avoid changing the behavior that we currently have (and thus potentially changing what customers experience). The currect timeout is after 90s.

I can go for 90/100, that will be consistent with what we have and give a log of time for the processing.

martinpovolny:
created https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/200559 to unify the prod/stage settings

bacciotti (dismissed):

bacciotti (approved):

PR #6245, opened 2026-08-04
[COST-7993] Add is_disableable to GET /settings/currency/ response

bacciotti:
## Summary

- Adds `is_disableable` boolean field to every currency entry returned by `GET /settings/currency/`
- Allows the UI to render the enable/disable toggle correctly **before** a user attempts a `DELETE`, without needing to rely on a `400` error response

## Changes

- `currency_views.py`: new `_get_non_disableable_codes()` helper using **batched queries** (one per blocking criterion, not N per currency) to determine which codes cannot be disabled
- Also includes the `if not customer or not customer.pk: return []` guard on `_get_cloud_providers_using_currency` (defensive fix for unsaved customer instances)

## `is_disableable` logic

Returns `true` only when **all** of the following hold:
- The currency is currently enabled
- It is not the only enabled currency
- It is not the system default currency (`KOKU_DEFAULT_CURRENCY`)
- It is not the account default currency (`UserSettings`)
- It is not used by any Cost Model
- It is not used by any Price List
- It is not a base currency in AWS/Azure/GCP billing data

## Test plan

- [ ] `GET /settings/currency/?search=USD` → `is_disableable: false` (system default)
- [ ] Enable a free currency (e.g. CHF) with no cost models/price lists → `is_disableable: true`
- [ ] Assign CHF to a cost model → `is_disableable: false`
- [ ] Disabled currency → `is_disableable: false`

Closes #COST-7993

## Summary by Sourcery

Expose disableability metadata for currencies in settings API and guard currency usage checks for unsaved customers.

New Features:
- Add is_disableable flag to each currency returned by GET /settings/currency/ based on enablement and usage constraints.

Bug Fixes:
- Avoid querying cloud provider usage when the customer object is missing or unsaved in _get_cloud_providers_using_currency.

Tests:
- Add comprehensive tests covering is_disableable behavior for enabled, disabled, default, cost model, price list, account default, and cloud provider base currencies.

### DIFF
--- a/koku/api/settings/currency_views.py
+++ b/koku/api/settings/currency_views.py
@@ -44,6 +44,8 @@ LOG = logging.getLogger(__name__)
 
 def _get_cloud_providers_using_currency(code, customer):
     """Return cloud providers whose billing data uses ``code`` as a base currency."""
+    if not customer or not customer.pk:
+        return []
     aws_uuids = AWSCostSummaryP.objects.filter(currency_code=code).values_list("source_uuid", flat=True).distinct()
     azure_uuids = AzureCostSummaryP.objects.filter(currency=code).values_list("source_uuid", flat=True).distinct()
     gcp_uuids = GCPCostSummaryP.objects.filter(currency=code).values_list("source_uuid", flat=True).distinct()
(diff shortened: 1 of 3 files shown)


koku/api/settings/currency_views.py
```diff
@@ -56,6 +58,29 @@ def _get_cloud_providers_using_currency(code, customer):
...
+    if account_settings:
+        account_currency = account_settings.settings.get("currency")
+        if account_currency:
+            blocked.add(account_currency)
+
+    blocked |= set(CostModel.objects.values_list("currency", flat=True).distinct())
+    blocked |= set(PriceList.objects.values_list("currency", flat=True).distinct())
+    blocked |= set(AWSCostSummaryP.objects.values_list("currency_code", flat=True).distinct())
+    blocked |= set(AzureCostSummaryP.objects.values_list("currency", flat=True).distinct())
+    blocked |= set(GCPCostSummaryP.objects.values_list("currency", flat=True).distinct())
```
djnakabaale: should these be to `enabled_codes` only or is it intentional not to? Otherwise it looks like we make unconstrained scans over the AWS/Azure/GCP summary partitions on every `GET`.

koku/api/settings/currency_views.py
```diff
@@ -56,6 +58,29 @@ def _get_cloud_providers_using_currency(code, customer):
     )
 
 
+def _get_non_disableable_codes():
```
djnakabaale: Should we be filtering currencies by the request customer?

djnakabaale (commented):
here are a couple questions from first pass:

koku/api/settings/currency_views.py
bacciotti: good catch! updated

koku/api/settings/currency_views.py
bacciotti: id say taht no customer filter needed here, similar to how EnabledCurrency, CostModel, and PriceList are queried in the same function. AWSCostSummaryP, AzureCostSummaryP etc are tenant-scoped models, so all queries on these tables are already customer-isolated without an explicit filter.

djnakabaale (approved):

PR #6249, opened 2026-08-05
[COST-8031] - Add retry logic to sync_hive_partitions for transient errors

esebesto:
## Jira Ticket

[COST-8031](https://issues.redhat.com/browse/COST-8031)

## Description
Co-Authored-By: Claude <noreply@anthropic.com>

Cost data is intermittently missing for some sources and billing months during testing. Investigations revealed that while data is successfully written to S3 as Parquet files, it never becomes visible in Trino because Hive partition metadata registration silently fails.

### Root Cause

sync_hive_partitions() relies on _execute_trino_sql(), which silently swallows all Trino exceptions.

The primary failure mode seems to be a race condition: multiple sources within the same schema call sync_partition_metadata('FULL') concurrently. Because 'FULL' drops and re-adds all partition metadata at the table level (rather than per source), concurrent calls collide with an ALREADY_EXISTS error. This race condition became more frequent after the MinIO-to-S4 proxy switch, which increased latency and widened the collision window.

### Proposed Fix

Extracts a new _execute_trino_sql_with_retries() method and switches sync_hive_partitions() to use it. Key behaviours:

- Retries up to 3 times with exponential backoff (~1s, ~2s, ~4s + jitter, capped at 30s), but only for known transient errors: ALREADY_EXISTS, HIVE_METASTORE_ERROR, HIVE_FILESYSTEM_ERROR, JDBC_ERROR. This is consistent with how the rest of the codebase handles Trino transient errors.

- Unknown TrinoQueryErrors (e.g. syntax errors, permission failures) are logged immediately at ERROR level with error_name in the structured log so they surface clearly and can be added to the allowlist if found to be transient.

- All log calls pass exc_info to the LOG call (not inside log_json) so exception tracebacks are correctly captured.

- On final failure, logs an error without raising so the ingest process is not completely blocked.

## Testing

1. Use main branch
2. run full local smokes -> most likely you will get **multiple test failures**
3. check koku logs for "One or more partitions already exist for table" errors
4. checkout to this branch
5. run full local smokes 
6. you should get a **clean pass**
9. check koku logs for "One or more partitions already exist for table" errors -> they should all be part of "sync_hive_partitions retrying" log

## Release Notes
- [COST-8031](https://issues.redhat.com/browse/COST-8031) Add retry logic to sync_hive_partitions to prevent transient errors

```markdown
* [COST-8031](https://issues.redhat.com/browse/COST-8031) Add retry logic to sync_hive_partitions to prevent transient errors
```

## Summary by Sourcery

Enhancements:
- Inline Trino execution in sync_hive_partitions with bounded retries and exponential backoff to handle transient errors without blocking ingest.

### DIFF
--- a/koku/masu/processor/report_parquet_processor_base.py
+++ b/koku/masu/processor/report_parquet_processor_base.py
@@ -29,6 +29,8 @@ from reporting.models import TenantAPIProvider
 
 LOG = logging.getLogger(__name__)
 
+_RETRYABLE_TRINO_ERRORS = frozenset({"ALREADY_EXISTS", "HIVE_METASTORE_ERROR", "HIVE_FILESYSTEM_ERROR", "JDBC_ERROR"})
+
 
 class PostgresSummaryTableError(Exception):
     """Postgres summary table is not defined."""
@@ -209,10 +211,11 @@ class ReportParquetProcessorBase:
         return created
 
     def _execute_trino_sql_with_retries(self, sql, schema_name: str, caller="", max_retries=3):
-        """Execute Trino SQL with retries on transient TrinoQueryError exceptions.
+        """Execute Trino SQL with retries on known transient TrinoQueryError exceptions.
 
-        Returns the result rows on success, or an empty list on failure.
-        Non-retryable errors (ProgrammingError, Error) are logged and return immediately.
+        Only errors in _RETRYABLE_TRINO_ERRORS are retried with exponential backoff.
+        Unknown TrinoQueryErrors are logged at ERROR (with error_name) and return [] immediately.
+        ProgrammingError and Django Error are also non-retryable and return [] immediately.
         """
         for attempt in range(max_retries + 1):
             try:
@@ -229,6 +232,17 @@ class ReportParquetProcessorBase:
                         LOG.debug(f"{caller} returned {len(rows)} rows")
                 return rows
             except TrinoQueryError as err:
+                if err.error_name not in _RETRYABLE_TRINO_ERRORS:
+                    LOG.error(
+                        log_json(
+                            msg=f"{caller} failed with non-retryable TrinoQueryError",
+                            schema=schema_name,
+                            table=self._table_name,
+                            error_name=err.error_name,
+                        ),
+                        exc_info=err,
+                    )
+                    return []
                 if attempt < max_retries:
                     backoff = min(2**attempt, 30) + random.uniform(0, 1)
                     LOG.warning(
@@ -236,6 +250,7 @@ class ReportParquetProcessorBase:
                             msg=f"{caller} retrying (attempt {attempt + 1})",
                             schema=schema_name,
                             table=self._table_name,
+                            error_name=err.error_name,
                         ),
                         exc_info=err,
                     )
@@ -246,6 +261,7 @@ class ReportParquetProcessorBase:
                             msg=f"{caller} failed after {attempt + 1} attempts",
                             schema=schema_name,
                             table=self._table_name,
+                            error_name=err.error_name,
                         ),
                         exc_info=err,
                     )
--- a/koku/masu/processor/report_parquet_processor_base.py
+++ b/koku/masu/processor/report_parquet_processor_base.py
@@ -4,6 +4,7 @@
 #
 """Processor for Parquet files."""
 import logging
+import random
 import time
 
 from dateutil.relativedelta import relativedelta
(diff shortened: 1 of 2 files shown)


esebesto:
[ocp-smoke-tests](https://konflux-ui.apps.stone-prd-rh01.pg1f.p1.openshiftapps.com/ns/cost-mgmt-dev-tenant/applications/koku/pipelineruns/koku-ci-zzp27/logs?task=run-iqe-cji): 
`= 4172 passed, 3 skipped, 7250 deselected, 132 warnings in 8690.21s (2:24:50) ==`

esebesto:
[smoke-tests](https://konflux-ui.apps.stone-prd-rh01.pg1f.p1.openshiftapps.com/ns/cost-mgmt-dev-tenant/applications/koku/pipelineruns/koku-ci-z6c48/logs?task=run-iqe-cji):
`======= 850 passed, 10575 deselected, 27 warnings in 6461.33s (1:47:41) ========`

esebesto:
[aws-smoke-tests](https://konflux-ui.apps.stone-prd-rh01.pg1f.p1.openshiftapps.com/ns/cost-mgmt-dev-tenant/applications/koku/pipelineruns/koku-ci-qtg5j/logs?task=run-iqe-cji):
`= 3046 passed, 10 skipped, 8369 deselected, 99 warnings in 4657.30s (1:17:37) ==`

koku/masu/processor/report_parquet_processor_base.py
bacciotti, line 1: Nice fix !
one suggestion: please retry only known transient Trino errors, not every TrinoQueryError.

In this repo we usually check error_name (see `HIVE_METASTORE_ERROR` retries in report_db_accessor_base / migrate_trino_tables). For this path, something like ALREADY_EXISTS, HIVE_METASTORE_ERROR, HIVE_FILESYSTEM_ERROR, and JDBC_ERROR would match both the Jira note and the concurrent FULL sync race in the PR description.

That way we avoid 4 attempts on permanent errors, and the retry policy stays consistent with the rest of the code.

esebesto: @bacciotti Appreciate the feedback! 
My main concern with restricting this to a hardcoded list of error names is that we don't know all potential failure modes in advance. ALREADY_EXISTS wasn't originally expected - I only discovered it while testing with the general retry block, so there could be other unexpected errors in edge cases.

Regarding permanent errors, I had a few questions about how we want to treat them here:
1. What do we consider a "permanent error" in this context?
2. Are we concerned about them occurring here? Right now, the existing code logs non-retryable errors and swallows them - is that expected, or is there a different intention for permanent errors?
3. Where do we draw the line? Technically, ALREADY_EXISTS isn't a typical transient failure either - it's a permanent error that happens under specific race conditions.

I want to make sure this fix helps as much as possible and stays resilient without accidentally changing the existing error-handling behavior.

bacciotti: Makes sense — thanks for walking through this.

Agree that a hardcoded allowlist could miss unexpected failure modes (ALREADY_EXISTS is a good example), and that this path already logs + swallows errors today. Keeping the broader TrinoQueryError retry preserves that behavior and stays more resilient.

No change needed from my side. Approving.

bacciotti (dismissed):
LGTM. Thanks for the clarification on the retry scope — keeping the broader TrinoQueryError retry is the right call here.

koku/masu/processor/report_parquet_processor_base.py
bacciotti: as for your questions...

1. What is a "permanent error" here?
Something that will keep failing if we run the same call again right away ( for ex: bad SQL, missing table/schema...) 

2. Are we concerned about them on this path? Is swallow expected?
For `sync_hive_partitions`, yes. the old `_execute_trino_sql` already logs and swallows. 

3. Where do we draw the line? Is ALREADY_EXISTS "permanent"?
By name it looks permanent, but under full sync it is effectively recoverable: after the other sync finishes, a retry can succeed. So for this specific call, "retryable" means "worth trying again soon", not "always transient like a network blip".

esebesto: After thinking this through, I've decided to retry only known errors in the end and improve logging for all other TrinoUserErrors to help us investigate potential issues in future.

esebesto:
[smoke-tests](https://konflux-ui.apps.stone-prd-rh01.pg1f.p1.openshiftapps.com/ns/cost-mgmt-dev-tenant/pipelinerun/koku-ci-9nsx9) after additional code changes:
`======= 868 passed, 10701 deselected, 27 warnings in 7683.67s (2:08:03) ========`

Switching to `hot-fix-smoke-tests` for merging main

bacciotti (approved):
Thanks Eva! Great work

PR #6252, opened 2026-08-09
[COST-8051] Enable readOnlyRootFilesystem by fixing runtime filesystem writes

martinpovolny:
## Motivation

This is needed for the on-prem/operator work: https://github.com/martinpovolny/koku-server-operator/blob/main/docs/design/koku-django-log-handler-problem.md

## Summary
- Conditionally declare Django's `file` log handler only when `"file"` is in `LOGGING_HANDLERS` — prevents `dictConfig()` from instantiating the `FileHandler` at startup and creating `app.log` on a read-only filesystem.
- Use `tempfile.NamedTemporaryFile` in HCS `csv_file_handler.py` instead of writing to CWD (`$APP_HOME`), which would fail with `readOnlyRootFilesystem`.
- Remove `app.log` cleanup from Dockerfile — no longer needed.

## Context
Django's `logging.config.dictConfig()` instantiates ALL declared handlers at startup, even if no logger references them. The `file` handler was always declared but never used in production (`DJANGO_LOG_HANDLERS=console` is the default). Its only effect was creating `app.log` during startup, which blocks `readOnlyRootFilesystem: true`.

The HCS CSV handler wrote temporary files to the current working directory (`$APP_HOME`) before uploading to S3. This also fails on a read-only root filesystem.

## Changes
| File | Change |
|------|--------|
| `koku/koku/settings.py` | Conditionally include `file` handler in `LOGGING` dict |
| `koku/hcs/csv_file_handler.py` | Use `tempfile.NamedTemporaryFile` instead of writing to CWD |
| `Dockerfile` | Remove `app.log` cleanup (no longer created during build) |

## Follow-up
- ClowdApp `securityContext` change to enable `readOnlyRootFilesystem: true` will be done via a separate app-interface MR after this lands.
- Existing `emptyDir` mounts for `$TMP_DIR` and `/var/tmp/masu/` already cover all writable paths.

## Test plan
- [ ] Verify `DJANGO_LOG_HANDLERS=console` (default) — no `app.log` created at startup
- [ ] Verify `DJANGO_LOG_HANDLERS=console,file` — file logging still works
- [ ] Verify HCS CSV upload still works
- [ ] Build Docker image successfully (no `app.log` to remove)
- [ ] Deploy to stage, confirm all pods start correctly

JIRA: https://redhat.atlassian.net/browse/COST-8051

## Summary by Sourcery

Make logging and CSV export compatible with read-only root filesystems by avoiding runtime writes to the application directory.

Bug Fixes:
- Prevent Django's optional file log handler from creating log files on startup when file logging is not configured.
- Ensure HCS CSV uploads use temporary files outside the application directory to avoid write failures on read-only roots.

Enhancements:
- Conditionally register the Django file log handler based on configured logging handlers.
- Simplify Docker image build by removing obsolete cleanup of app.log created during collectstatic.

Build:
- Remove Dockerfile cleanup of app.log now that it is no longer created during image build.


koku/koku/settings.py
```diff
@@ -491,12 +491,18 @@ class CacheEnum(StrEnum):
             "class": "logging.StreamHandler",
             "formatter": LOGGING_FORMATTER,
         },
-        "file": {
-            "level": KOKU_LOGGING_LEVEL,
-            "class": "logging.FileHandler",
-            "filename": LOGGING_FILE,
-            "formatter": LOGGING_FORMATTER,
-        },
+        **(
```
ELK4N4: I think it will be more elegant to add this handler after the declaration. Something like:
```python
if "file" in LOGGING_HANDLERS:
     LOGGING["handlers"].append({
                "file": {
                    "level": KOKU_LOGGING_LEVEL,
                    "class": "logging.FileHandler",
                    "filename": LOGGING_FILE,
                    "formatter": LOGGING_FORMATTER,
                }
            })
```

koku/hcs/csv_file_handler.py
```diff
@@ -65,19 +66,21 @@ def copy_data_to_s3_bucket(
...
         """Copy hcs data to S3 bucket."""
         my_df = pd.DataFrame(data)
-        my_df.to_csv(filename, header=cols, index=False)
         metadata = {"finalized": str(finalize)}
         if finalize and finalize_date:
             metadata["finalized-date"] = finalize_date
         extra_args = {"Metadata": metadata}
-        with open(filename, "rb") as fin:
-            try:
+        with tempfile.NamedTemporaryFile(suffix=".csv", delete=False) as tmp:
```
ELK4N4: You can move the try statement inside the tempfile context manager and then you don't need to add `delete=False` and you can remove the finally statement

martinpovolny:
@ELK4N4 : thx for the comments, addressed

bacciotti (approved):

PR #6253, opened 2026-08-10
fix: daily Slack digest still posts when GH cron is late

bacciotti:
## Summary
- Today's scheduled run succeeded but skipped posting: `Lisbon hour is 8, want 7` (07:02 UTC = 08:00 Lisbon in DST; the 06:00 UTC slot that should hit hour 7 did not fire).
- Accept Lisbon hours **7 and 8** on schedule events so delayed GitHub Actions crons still post during DST.

## Test plan
- [x] Triggered manual `workflow_dispatch` on `main` for today's digest (independent of this PR)
- [ ] After merge, confirm next weekday schedule posts (or skips only outside 7–8 Lisbon)

## Summary by Sourcery

Bug Fixes:
- Ensure the daily Slack digest still posts when the scheduled GitHub Actions cron job is delayed into the next Lisbon hour.

### DIFF
--- a/.github/scripts/daily_slack_digest.py
+++ b/.github/scripts/daily_slack_digest.py
@@ -22,6 +22,9 @@ FORCE_RUN = os.environ.get("FORCE_RUN", "").lower() in {"1", "true", "yes"}
 
 TIMEZONE = ZoneInfo("Europe/Lisbon")
 POST_HOUR = 7
+# GitHub schedule jobs are often delayed; accept the next Lisbon hour too so a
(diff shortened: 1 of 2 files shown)


bacciotti:
Good catch on the duplicate risk — without a guard, summer DST would post twice on a normal day (06 UTC → Lisbon 7 and 07 UTC → Lisbon 8).

Pushed a follow-up: keep the 7–8 hour window for delayed crons, but dedupe with an Actions cache marker keyed by Lisbon calendar date, plus a concurrency group so the second slot waits and then skips.

jvcsizilio (approved):

PR #6263, opened 2026-08-12
Cost 7838 hide cloud apis

jvcsizilio:
## Jira Ticket

[COST-7838](https://redhat.atlassian.net/browse/COST-7838)

## Description

On-prem Cost Management is OCP-only, but AWS/Azure/GCP customer APIs (and OCP-on-cloud variants) were still registered, and Sources create still accepted cloud source types. This change gates those entry points with `settings.ONPREM`.

**What changed**

- Customer API URL registration (`koku/api/urls.py`) now builds patterns via helpers. Pure cloud and OCP-on-cloud routes are registered only when `not settings.ONPREM`.
- On-prem Sources create/update rejects non-OCP source types (including via `source_type_id`).
- On-prem `source_types` lists OpenShift only.
- Sources `aws-s3-regions` action is omitted / 404 on-prem.
- Unit tests cover URL gating, serializer validation, and source_types catalog behavior.

SaaS behavior is unchanged. Cloud view implementations are not deleted.

**Decisions for team review**

1. Hide pure cloud customer routes on-prem (unregister so they 404), not leave empty 200s.
2. Also hide OCP-on-cloud routes (`.../openshift/infrastructures/{aws|azure|gcp}/...`) because on-prem has no cloud providers.
3. Block creating/updating Sources to cloud types on-prem; `source_types` is OpenShift-only on-prem.
4. `docs/specs/openapi.json` / served `openapi.json` is **unchanged** in this PR. Cloud paths may still appear in docs while returning 404 on-prem.
5. Root `resource-types/` list body is **not** filtered here (child cloud resource-type routes are hidden).

**Questions for the team**

1. Should on-prem OpenAPI docs also omit cloud / OCP-on-cloud paths so the contract matches the registered routes?
2. Should the root `resource-types/` response also stop advertising cloud resource types on-prem?

**Branch note**

This branch was cut from `COST-7837-skip-saas-celery-beats` and may include unmerged COST-7837 commits. Merge/rebase strategy should account for that stack.

## Testing

1. Checkout branch `COST-7838-hide-cloud-apis`
2. Run:
   ```bash
   pipenv run python koku/manage.py test koku.api.test.test_onprem_url_gating -v 2
   pipenv run python koku/manage.py test sources.test.api.test_source_type_views -v 2
   pipenv run python koku/manage.py test sources.test.api.test_serializers -v 2
   pipenv run python koku/manage.py test sources.test.api.test_view -v 2
   ```
3. Expect URL gating tests to show cloud / OCP-on-cloud names absent when on-prem and present on SaaS; OCP/shared routes remain on-prem; Sources create rejects AWS/Azure/GCP on-prem; source_types returns OpenShift only on-prem.

## Summary by Sourcery

Gate cloud and OCP-on-cloud customer APIs and SaaS-only Celery beats based on the ONPREM setting, so on-prem deployments expose only OpenShift functionality and skip cloud integrations.

New Features:
- Add helper-based URL construction that cleanly separates shared, SaaS-only, and on-prem-only API routes, including conditional registration of CMMO compatibility endpoints and data-retention settings.
- Introduce on-prem-aware Celery beat registration utilities to control scheduling of currency rates, SaaS-only tasks, and batched report checks via configuration.

Bug Fixes:
- Prevent creation and update of non-OpenShift Sources on on-prem deployments by enforcing provider-type validation in serializers and hiding cloud-only Sources actions and routes.
- Ensure the on-prem source_types catalog lists only OpenShift while SaaS continues to expose the full CMMO mapping, aligning listed types with supported behavior.

Enhancements:
- Refactor API URL configuration into composable helpers that make the ONPREM gating of cloud and infrastructure routes explicit and testable.
- Add targeted tests for on-prem URL gating, Celery beat registration behavior, and SourceTypes/Sources serializers to protect SaaS vs on-prem separation going forward.

Tests:
- Add unit tests validating that cloud and OCP-on-cloud routes are omitted on-prem while shared and OpenShift routes remain, including coverage of the Sources aws-s3-regions action.
- Add Celery beat schedule tests to confirm on-prem deployments skip SaaS-only beats and currency/report-check tasks per configuration.
- Extend SourceTypes and Sources serializer tests to verify OpenShift-only behavior on-prem and full CMMO catalog on SaaS.


koku/api/test/test_onprem_url_gating.py
```diff
@@ -0,0 +1,128 @@
+#
```
myersCody, line 1: There are a few test files under api/

- test_query_params.py
- test_params.py
- test_utils.py

In a follow up can we move these under the test directory you created here. I think that makes more sense then keeping them under api.

myersCody:
This is a bit of a nit pick, but while reviewing `koku/api/urls.py`, noticed the on-prem gating for the Sources `aws-s3-regions` action is implemented twice, via two different sources of truth:

1. [`urls.py`](https://github.com/project-koku/koku/blob/0b0949541ae3f169276141034bda07ce9075d46b/koku/api/urls.py#L550-L557) — `ONPREM_EXCLUDED_SOURCES_ROUTE_NAMES` filters the route out of the router's URLs based on the `onprem` **parameter** passed into `_sources_router_urls()`.
2. [`sources/api/view.py`](https://github.com/project-koku/koku/blob/0e1cc01f4207f5c4b779a1cc3b9e7341e2505ed2/koku/sources/api/view.py#L171-L184) — `SourcesViewSet.get_extra_actions()` already excludes the same `aws_s3_regions` action based on the **global** `settings.ONPREM`, and the action itself also raises `Http404` when `settings.ONPREM` is true.

Functionally this is fine in real deployments (both ultimately read `settings.ONPREM`), and all 67 `sources.test.api.*` tests pass with `ONPREM=True`. But it's redundant, and it makes `build_api_urlpatterns(onprem=...)` a bit misleading as a "pure" parameterized helper — `test_onprem_url_gating.py::test_sources_aws_s3_regions_absent_when_onprem` will actually reflect whatever the real `settings.ONPREM` is (via `get_extra_actions()`), not the `onprem` arg passed in. This only surfaces if a dev's local `.env` sets `ONPREM=True` while running that test with `onprem=False` — it isn't a CI issue since CI's default job runs with `ONPREM=False` — but it's a maintenance trap. Worth relying on just one of these two mechanisms.

myersCody (approved):

PR #6272, opened 2026-08-19
[COST-8112] Lock populate_usage_rates_to_usage against concurrent RTU writers

jordigilh:
## Summary

`populate_usage_rates_to_usage` is the first write into `rates_to_usage` in the RTU pipeline. It performs a DELETE-then-INSERT for a given provider/date-range via raw SQL, structurally identical to `populate_distributed_cost_sql` / `aggregate_rates_to_daily_summary` / `populate_markup_rates_to_usage` / `_populate_cost_breakdown_ui_summary_table` — all of which already acquire `OCPReportDBAccessor._distribution_provider_lock`. This method did not.

`rates_to_usage`'s FKs to `reporting_ocpusagereportperiod` / `reporting_tenant_api_provider` are `ON DELETE CASCADE DEFERRABLE INITIALLY DEFERRED` (migration 0353), so an uncoordinated writer here is exposed to the same concurrent-delete FK-violation crash characterized for `Provider.delete()` in `test_ocp_provider_delete_rtu_race.py` (Finding B, #6102) — but for *any* concurrent event that deletes the report period, not just provider deletion. Blast radius is broader than the cost-breakdown UI: `rates_to_usage` rows feed both `aggregate_rates_to_daily_summary` (the core OCP daily-summary table every OCP cost report reads from) and the cost-breakdown tree.

Fixes [COST-8112](https://redhat.atlassian.net/browse/COST-8112), filed as part of the broader COST-7249 deadlock preflight audit.

## Changes

- Wrap `populate_usage_rates_to_usage`'s body in `_distribution_provider_lock(provider_uuid)`, the same primitive already used by its RTU-pipeline siblings.
- Add a deterministic regression test (`test_populate_usage_rates_to_usage_lock_gap.py`) that races `populate_usage_rates_to_usage` against a concurrent thread already holding `_distribution_provider_lock` for the same provider, and asserts it now blocks for the lock's full hold duration instead of running unserialized. Verified the test fails deterministically (returns in ~0.02s instead of waiting ~1.5s) when the lock is reverted, and passes with the fix applied.

## Testing performed

- New regression test passes with the fix, and was manually confirmed to fail deterministically with the fix reverted.
- Ran the full `masu.test.processor.ocp.test_phase2_rates_to_usage` suite (125 tests) — no new failures introduced. Two pre-existing failures (`test_api_cost_total_matches_daily_summary`, `test_rtu_custom_name_fallback`) reproduce identically with or without this change (confirmed by running that suite alone, without this PR's new test file).
- `black --line-length 119` and `flake8` clean on both changed files.

## Related

- Epic: [COST-7249](https://redhat.atlassian.net/browse/COST-7249)
- Related deadlock-preflight fixes: #6268 (COST-8113), #6269 (COST-8114), #6270 (COST-8115), #6271 (COST-8116)

### DIFF
--- a/koku/masu/database/ocp_report_db_accessor.py
+++ b/koku/masu/database/ocp_report_db_accessor.py
@@ -1397,6 +1397,18 @@ AND (month = {{month_no_zero}} OR month = {{month}})
         cost_model_rate via SQL JOIN.  Distribution is read directly from the
         cost_model table; cte_node_cost computes allocation fractions only;
         rate multiplication happens in Component 6.
+
+        COST-7249 deadlock preflight, COST-8112: this is the very first write
+        into rates_to_usage in the whole RTU pipeline, and does the same
+        DELETE-then-INSERT for the same provider/date-range as every other RTU
+        write step (populate_distributed_cost_sql, aggregate_rates_to_daily_
+        summary, populate_markup_rates_to_usage, _populate_cost_breakdown_ui_
+        summary_table) -- all of which already run under
+        _distribution_provider_lock. This one didn't, leaving it exposed to
+        the same concurrent-Provider.delete() FK-violation crash as Finding B
+        (see test_ocp_provider_delete_rtu_race.py) for an uncoordinated
+        writer. Runs under _distribution_provider_lock (see its docstring) to
+        close that gap.
         """
         sql = pkgutil.get_data(
             "masu.database",
@@ -1413,7 +1425,8 @@ AND (month = {{month_no_zero}} OR month = {{month}})
         }
 
         LOG.info(log_json(msg="populating rates_to_usage (single-pass)", context=sql_params))
-        self._prepare_and_execute_raw_sql_query("rates_to_usage", sql, sql_params, operation="INSERT")
+        with self._distribution_provider_lock(provider_uuid):
+            self._prepare_and_execute_raw_sql_query("rates_to_usage", sql, sql_params, operation="INSERT")
 
     def populate_markup_rates_to_usage(self, start_date, end_date, source_uuid, cluster_id, cost_model_id):
         """Write markup costs as RatesToUsage rows with metric_type='markup'.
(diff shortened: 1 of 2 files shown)


jordigilh:
## Konflux IQE failure investigation (koku-ci-xsxtv, commit f66b886)

Followed up on the CI-triager bot's "possible Koku-side regression" flag for the 4 failing IQE tests by pulling the actual pod logs from S3 (worker + API) and tracing through both failure windows (03:34:26–03:34:37 and 03:53:35–03:53:46 UTC).

**Findings:**
- Zero matches for `deadlock`/`OperationalError`/`IntegrityError`/`traceback` anywhere in `koku-clowder-worker-cost-model`'s full 5.75h log.
- Every `update_cost_model_costs` task succeeded cleanly (4–20s, no exception) during and around both windows — including the RTU pipeline runs for the exact clusters the failing tests exercise.
- Every API request in both windows returned `200` with normal response times/payload sizes. For `test_api_ocp_source_daily_flow` specifically (the case the bot called "particularly telling"), the initial query and the `limit=150` follow-up returned **identical response byte-length (5998 bytes)** — i.e. the same data, contradicting a distribution-value regression.
- The only `5xx`s in the whole run are two `ProgrammingError: relation "api_tenant" does not exist` at 01:12 UTC — 2+ hours before the flagged failures, a known startup-time migration-not-ready race on `/status/`, unrelated to this change.

**Conclusion:** this PR's change (wrapping the existing `populate_usage_rates_to_usage` call in the same `_distribution_provider_lock` primitive already used unmodified by 4 sibling RTU steps) shows no footprint in the logs — no contention, no errors, no timing anomaly. The 4 IQE failures look like ordinary ephemeral-environment flakiness (consistent with the independently-confirmed infra flakiness on #6270/#6271 in this same CI batch), not a regression from this change.

Recommend re-running the smoke tests before merge as a final confirmation, but I don't believe further code changes are needed here.

jordigilh:
Both valid, fixed in f733817bb:

- Moved the `Customer` lookup outside `schema_context` (it's a public-schema model, per the same convention already applied in #6102/#6269).
- Replaced both `skipTest` calls with `assertIsNotNone` so a missing `Customer` or `CostModelMap` fixture fails the regression loudly instead of silently skipping it.

PR #6276, opened 2026-08-27
Parallelize OCP report aggregation by splitting distinct arrays

martinpovolny:
Large-tenant OCP report queries can exceed the 90s timeout because **ARRAY_AGG(DISTINCT ...)** for the clusters/source_uuid response fields forces a serial GroupAggregate on both the main aggregation and the ranked query.

Gate a new path behind the
cost-management.**backend.ocp_report_distinct_arrays_parallel** Unleash flag that computes those two metadata arrays in separate, cheap queries (no currency math / window), letting the heavy aggregations parallelize. Array values are unchanged and the flag defaults off, so the legacy single-query path is preserved.

- provider_map arrays are kept as-is; when enabled they are popped from the main annotation set and backfilled per (date, group) tuple.
- the ranked query **drops its inline DISTINCT arrays and merges them from a separate rank-group-keyed query**; base handlers (AWS/Azure/GCP) are unaffected via an overridable _distinct_arrays_split_enabled.

Add parity tests asserting clusters/source_uuid match the legacy output (as sets) across report types, group-bys, limit, category and CSV.

## Jira Ticket

[COST-8146](https://redhat.atlassian.net/browse/COST-8146)

## Detailed report

Old query plan:
image

New query plan:
image

https://docs.google.com/document/d/1huYOEikUOGFozhcxrS-AJ2SBqLoCpPzW_O1SdeC3fdM/edit?tab=t.0#heading=h.vyg2oznjtvfj


koku/api/report/ocp/query_handler.py
```diff
@@ -231,10 +238,52 @@ def _get_base_currencies_for_conversion(self):
...
+        return merged
+
     @property
     def report_annotations(self):
         """Return annotations with OCP-specific infra_exchange_rate for CSV output."""
         annotations = self._mapper.report_type_map.get("annotations", {})
+        if self._distinct_arrays_split_enabled:
+            # The metadata arrays are computed separately so the main aggregation
+            # can parallelize; drop them from the heavy annotation set.
+            annotations = {k: v for k, v in annotations.items() if k not in DISTINCT_METADATA_FIELDS}
```
jvcsizilio, line 283: When the flag is on this copy means `annotations()` can no longer mutate the mapper dict. `group_by[node]` adds `capacity_count` via `self.report_annotations.update(...)`, so that update is thrown away. `NodeCapacity.update_row` overwrites `capacity` but does not put `capacity_count` on the row. CPU/memory/volume node reports will lose instance counts once the flag is on.

I think we should always copy the mapper annotations, apply `capacity_aggregate["node"]` inside `report_annotations`, and stop mutating the property.

koku/api/report/test/ocp/test_distinct_arrays_parallel.py
```diff
@@ -0,0 +1,131 @@
...
+    Returns a mapping of a stable row identity (the frozenset of the leaf row's
+    string-valued fields -- date plus the group-by labels) to the *sets* of
+    clusters and source_uuid found on that row.
+    """
+    if isinstance(node, dict):
+        if "source_uuid" in node or "clusters" in node:
+            key = frozenset((k, v) for k, v in node.items() if isinstance(v, str))
+            entry = acc.setdefault(key, {"source_uuid": set(), "clusters": set()})
+            entry["source_uuid"] |= {str(v) for v in (node.get("source_uuid") or [])}
+            entry["clusters"] |= {str(v) for v in (node.get("clusters") or [])}
```
jvcsizilio, line 45: `group_by[node]` is in the matrix, but this only compares `clusters`/`source_uuid` as sets. That would still pass with the `capacity_count` drop. Can we also assert `capacity_count` on node values with the flag on and off?

koku/api/report/ocp/query_handler.py
```diff
@@ -231,10 +238,52 @@ def _get_base_currencies_for_conversion(self):
         )
         return (base_currencies | cm_currencies) - {None}
 
+    @cached_property
+    def _distinct_arrays_split_enabled(self):
+        """Whether to compute clusters/source_uuid via separate (parallel-safe) queries."""
+        return is_feature_flag_enabled_by_schema(self.tenant.schema_name, OCP_REPORT_DISTINCT_ARRAYS_PARALLEL_FLAG)
```
jvcsizilio, line 238: This omits `dev_fallback=True`. Local/on-prem will never take the new path unless Unleash is up. Fine if that is an intentional cautious rollout; otherwise I would pass `dev_fallback=True`.

jvcsizilio (changes requested):
I went through this PR against COST-8146. The ticket is the timeout investigation; this is a follow-up query split. Tenant context, cloud handlers, and the default-off flag look fine.

When the flag is on, `report_annotations` returns a new dict, so the existing node `capacity_aggregate` update() is dropped and `group_by[node]` rows lose `capacity_count`. The new tests only compare `clusters`/`source_uuid` as sets, so they would still pass. I think we need that fixed before merge.

koku/api/report/ocp/query_handler.py
martinpovolny: yeah, but we test this via feature flag on a single tenant and then as a follow-up, we can remove the old path and the feature flag and be done with it

jvcsizilio: Thanks, that makes sense. Rolling it out on one tenant and then dropping the flag is a good plan.

jvcsizilio (dismissed):

martinpovolny:
@jvcsizilio : updated, pls, check

jvcsizilio (approved):
I took another look. This looks good to me and I approved it.

I also looked at COST-8146 and I think this covers what that ticket is asking for.

PR #6277, opened 2026-08-28
Capture stacks on Gunicorn worker timeouts

martinpovolny:
## Summary
- emit all Python thread stacks with a faulthandler watchdog before Gunicorn times out a sync request
- send a best-effort GlitchTip event tagged `timeout=hard` when the worker abort hook can run
- expose the watchdog timeout for deployments

## Related
- [COST-8155](https://issues.redhat.com/browse/COST-8155)

## Scope
This PR covers Gunicorn API-worker hard-timeout diagnostics only.

Celery worker/task timeout diagnostics are out of scope. Celery does not use Gunicorn’s `worker_abort` hook or its SIGABRT/SIGKILL lifecycle, so it requires a separate follow-up.

## Testing
- `PROMETHEUS_MULTIPROC_DIR=/tmp /Users/mpovolny/Projects/koku/koku/.venv/bin/python koku/manage.py test koku.test_middleware.RequestTimeoutMiddlewareTest koku.test_gunicorn_conf.GunicornConfigTest --no-input -v 1`
- changed-file `pre-commit` checks


koku/gunicorn_conf.py
```diff
@@ -57,19 +60,77 @@ def on_starting(server):
...
+
+def _capture_worker_timeout(worker, sentry_threads):
+    """Best-effort Sentry capture for a worker abort; never block its termination."""
+    if not ENVIRONMENT.bool("KOKU_ENABLE_SENTRY", default=False):
+        return
+
+    try:
+        import sentry_sdk
+
+        sentry_sdk.capture_event(
```
jvcsizilio, line 109: I think this event still misses COST-8155. `RequestTimeoutError` gives path, method, duration, and tenant. Here we only have the pid and thread stacks.

`worker_abort` does not get the request, so I would store that on the request thread in `RequestTimeoutMiddleware` (including gthread workers) and put it on this event.

I would also send `message` as a string, or raise a dedicated exception through `before_send`, so GlitchTip renders it like the soft-timeout events. `message: {formatted: ...}` plus `threads` may not show the same way.

On gthread, `current` is the abort handler, not the stuck request. I would not mark that thread current.

koku/gunicorn_conf.py
```diff
@@ -57,19 +60,77 @@ def on_starting(server):
...
+        sentry_sdk.flush(timeout=1)
+    except Exception:
+        worker.log.warning("Unable to send worker-timeout stack trace to Sentry.", exc_info=True)
+
+
+def worker_abort(worker):
+    """Log and, when possible, report all thread stacks for a worker timeout."""
+    data, sentry_threads = _get_all_thread_stacks()
 
     worker.log.error(f"Killing worker (pid:{worker.pid})\n{data}")
```
jvcsizilio: This ERROR log can create a second GlitchTip event through Sentry's LoggingIntegration, on top of `capture_event`. I would flush one structured event and keep this log at warning, or filter gunicorn timeout logs in `before_send`.

koku/koku/middleware.py
```diff
@@ -503,10 +518,17 @@ def handler(signum, frame):
 
         signal.signal(signal.SIGALRM, handler)
         signal.alarm(self.SOFT_TIMEOUT)
+        faulthandler.dump_traceback_later(
```
jvcsizilio, line 521: `dump_traceback_later` will not run on our API pods. `GUNICORN_THREADS` defaults to True, so requests are not on the main thread and we return before arming the watchdog. `REQUEST_FAULTHANDLER_TIMEOUT` on koku-reads/writes is unused.

The live production path is `post_fork` `faulthandler.enable` plus `worker_abort`. I would treat that as the product for SaaS, or change how we run Gunicorn if we really want a 95s request watchdog there.

Please do not hook `dump_traceback_later` onto every gthread request. It is process-wide, and one finished request would cancel the dump for another still-running request.

jvcsizilio (commented):
I went through this PR against COST-8155.

What I checked:
- Gunicorn abort vs RequestTimeoutMiddleware
- gthread (`GUNICORN_THREADS=True`) vs sync
- GlitchTip event shape vs RequestTimeoutError
- kustomize/clowdapp wiring
- tests

I think we need a few changes before merge. COST-8155 wants stack plus request context on worker-timeout events. The abort hook still has no path/method/tenant. The new `dump_traceback_later` watchdog also will not run on our API pods, because `GUNICORN_THREADS` defaults to True and the middleware only arms on the main thread.

Please link COST-8155 on the PR. If this is Gunicorn-only, say that Celery is follow-up.

Konflux `koku-ci` failed on the ephemeral environment (triager called it infra). Units is still pending.

martinpovolny:
@jvcsizilio : feedback hopefully addressed. Please, check again.

jvcsizilio (approved):

PR #6278, opened 2026-08-31
[COST-7850] Fail /livez when Celery parent or listener consumer is wedged

jvcsizilio:
## Summary
- Worker `/livez` now fails (503) when the Celery **parent** heartbeat is missing or stale after startup, so Kubernetes can restart a wedged worker whose HTTP probe thread is still answering.
- Listener and sources-listener `/livez` fail when `ready=True` and the Kafka/sources **consumer thread is dead or missing**. v1 does **not** treat a live thread stuck in `listen_for_messages` as dead (long report processing must stay 200).
- Long-running work stays live: heartbeat is the parent hub/timer (re-armed on broker reconnect via `ParentHeartbeatStep`), not "task finished." `/readyz` and gunicorn `BasicProbeServer` `/livez` are unchanged.
- Operators can tune worker liveness via `PROBE_LIVENESS_HEARTBEAT_SECONDS` (default 60) in clowdapp / kustomize worker env.

## Test plan
- [ ] `pipenv run python koku/manage.py test koku.test.test_probe_server --no-input -v 2` (30 tests)
- [ ] `pipenv run python koku/manage.py test koku.test.test_celery --no-input -v 2`
- [ ] Confirm worker `/readyz` is still migrations-only (424 until ready, then 200)
- [ ] Confirm listener `/readyz` still checks Kafka
- [ ] Confirm gunicorn `/livez` is still always 200
- [ ] Confirm a child busy in SQL / ~25h task would not fail `/livez` while the parent still ticks


koku/koku/probe_server.py
```diff
@@ -6,15 +6,91 @@
...
 from masu.prometheus_stats import WORKER_REGISTRY
 
+# Parent-process heartbeat for WorkerProbeServer /livez. Updated by the Celery
+# parent loop (wait_for_migrations + hub/timer), never by a daemon ticker thread.
+_parent_heartbeat_at: float | None = None
+# Kafka/sources consumer thread for ListenerProbeServer / SourcesProbeServer /livez.
+_consumer_thread: threading.Thread | None = None
+
+
+def get_probe_liveness_heartbeat_seconds() -> int:
```
bacciotti, line 27: Could you add a short note in deploy docs or clowdapp for `PROBE_LIVENESS_HEARTBEAT_SECONDS`?
Operators may want to tune this without reading the code.

bacciotti (commented):
Listener `/livez` only checks `thread.is_alive()`, not a wedged consumer stuck in processing.
That matches the tests, but please confirm in the Jira/PR description that dead/missing thread is the target — not a hung message loop.
Worker parent-heartbeat path looks solid, including `ParentHeartbeatStep` after `hub.timer.clear()` on reconnect.

koku/koku/probe_server.py
jvcsizilio: Good call. I added `PROBE_LIVENESS_HEARTBEAT_SECONDS` (default 60) next to `WORKER_PROC_ALIVE_TIMEOUT` on the Celery worker deployments in `deploy/clowdapp.yaml` and the kustomize worker patches, plus the ClowdApp parameter, `.env.example`, docker-compose, and a short note in `docs/architecture/celery-tasks.md`.

jvcsizilio:
@bacciotti thanks for the review.

Confirmed: listener `/livez` v1 is **dead or missing consumer thread** only (`thread.is_alive()`). A live thread blocked in `listen_for_messages` (large report) stays 200 on purpose, so we do not kill long processing. I also spelled that out in the PR description.

On `PROBE_LIVENESS_HEARTBEAT_SECONDS`, I wired it through clowdapp / kustomize worker env (and docs) so it can be tuned without reading the Python. Default stays 60.

bacciotti (dismissed):

bacciotti (approved):

PR #6284, opened 2026-09-01
Add ingress payload Kibana backup scripts and runbook

bacciotti:
## Summary

- Add operational runbook for backing up OCP ingress `.tgz` archives from Kibana `Downloading Payload` logs before presigned URLs expire
- Add reusable incident scripts: parse Kibana export, download payloads, split midnight filter, upload to S3 via AWS CLI
- Document limitations (Kibana snapshot vs Kafka backlog) and link to DLQ work (COST-8164)
- Prod-specific values (Kibana index, buckets) stay in internal service-docs; public docs use placeholders only

## Test plan

- [x] `python3 -m py_compile scripts/incident/*.py`
- [x] Pre-commit hooks pass on commit
- [ ] Review runbook steps against a future incident (no runtime changes to koku service)


scripts/incident/parse_kibana_ingress_payload_logs.py
```diff
@@ -0,0 +1,240 @@
...
+
+SCRIPT_DIR = Path(__file__).resolve().parent
+DEFAULT_INPUT = SCRIPT_DIR / "kibana-downloading-payload-export.json"
+DEFAULT_OUTPUT_DIR = SCRIPT_DIR
+
+HIT_BLOCK_RE = re.compile(
+    r'"@message"\s*:\s*"""(.*?)"""\s*,\s*\n\s*"@timestamp"\s*:\s*"([^"]+)"'
+    r'(?:\s*,\s*\n\s*"@log_stream"\s*:\s*"([^"]*)")?',
+    re.DOTALL,
+)
```
jvcsizilio, line 36: **Blocker:** `HIT_BLOCK_RE` expects `"@message": """..."""` immediately followed by `"@timestamp"`. Kibana Dev Tools `_search` responses are ordinary JSON (`hits.hits[]._source`) with a normal JSON string, and the runbook query field order is `@timestamp`, `@message`, `@log_stream`.

As written, `load_hits_from_export()` will raise `No log hits found` on the file the runbook tells operators to save.

Please parse JSON `hits.hits` first (including paginated responses) and keep this regex only as a fallback. A redacted fixture of a real hit would lock this in. If you already parsed a production export successfully, a redacted snippet in the PR would be enough to drop this.

scripts/incident/download_payloads.py
```diff
@@ -0,0 +1,119 @@
...
+            "status": "skipped",
+            "bytes": out_path.stat().st_size,
+            "error": "",
+        }
+
+    try:
+        req = urllib.request.Request(url, headers={"User-Agent": "koku-incident-backup/1.0"})
+        with urllib.request.urlopen(req, timeout=300) as resp:
+            data = resp.read()
+        out_path.write_bytes(data)
```
jvcsizilio, line 59: **Blocker:** `resp.read()` + `out_path.write_bytes()` writes the final `{request_id}.tgz` in place. An interrupt or OOM leaves a non-empty partial file. The runbook’s command is `download_payloads.py --skip-existing`, which then skips any file with `stat().st_size > 0`. That archive can be uploaded as a “successful” backup.

Stream chunks to a sibling temp file, `os.replace` onto the `.tgz` only after success, and delete the temp on failure. If the manifest has `size`, compare bytes on skip and on download. Do not skip files that are smaller than the logged size.

scripts/incident/upload_payloads.py
```diff
@@ -0,0 +1,255 @@
...
+        if not local_path.is_file():
+            missing_local.append(request_id)
+            continue
+        s3_key = build_s3_key(
+            schema=schema,
+            request_id=request_id,
+            kafka_timestamp=row["kafka_timestamp"],
+            cluster_id=row.get("cluster_id", ""),
+            provider_uuid_map=provider_uuid_map,
+        )
```
jvcsizilio, line 204: **Blocker:** The parser already extracts `org_id` per hit, but `cmd_upload` applies one CLI `--schema` / `--org-id` to every object. A Kibana query that accidentally matches more than one tenant (the sample query is `ORG_ID OR ACCOUNT_ID`) will store customer A’s tarball under customer B’s prefix.

Compute `schema_from_org_id(row["org_id"])` per row. If `--schema` is set, refuse or skip mismatches. If it is omitted, use the row schema. Log skipped request IDs.

jvcsizilio (changes requested):
## Review

Useful ops runbook and scripts, and the security/docs posture is careful (placeholders, gitignore, no prod names). This does not change listener or summarization behavior, and `ok-to-skip-smokes` is appropriate.

The documented happy path is not merge-ready yet. Please address the three inline blockers: parser vs Dev Tools `_search` JSON, truncated downloads treated as complete under `--skip-existing`, and upload keys ignoring per-row `org_id`.

### What looks good

- Honest limitations table (Kafka backlog never logged, export is a snapshot, backup is not reingest)
- Prod identifiers kept out of public git
- Midnight rule labeled as a heuristic, not a guarantee
- S3 prefix adds `ingress-payload`, so raw archives should not collide with daily CSV layout
- Dry-run and `check-access` exist; upload returns non-zero on errors/missing files

### Other notes (non-blocking)

- Empty parse still exits 0 if hits match the wrapper but fail `parse_hit` — return non-zero before writing if `parsed` is empty
- `account` is required in extraction; `handle_message` only requires `org_id`
- `check-access` requires `s3:ListBucket`; PutObject-only incident roles fail before the write probe
- Download URL scheme is not re-checked (`file://` / `http://` in `urls.tsv` would be fetched)
- `request_id` is used as a path without the listener sanitizer
- No tests for parser / midnight heuristic / S3 key layout (`py_compile` does not catch the export-format bug)

### Questions

1. Did you successfully parse a real prod/stage Dev Tools `_search` JSON save with `HIT_BLOCK_RE`? If yes, what does `@message` look like (triple quotes vs escaped JSON string)?
2. Is `data/csv/{schema}/OCP/ingress-payload/` agreed with pipeline owners, or should this stay outside `data/csv` so nothing lists it as report input?
3. Should `cluster_id` come from `b64_identity` only (`identity.system.cluster_id`), or also from the tarball `manifest.json` after download?

After the blockers, I would approve with notes on tests and `check-access`.

martinpovolny:
> See also
[service-docs runbook — Kibana](https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/operations/runbook.md)

points to https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/operations/runbook.md which is not the right URL it seems

jvcsizilio (approved):
I took another look. This looks good to me and I approved it.

PR #6285, opened 2026-09-07
[COST-8179] Decouple on-prem source provider creation from HTTP request

bacciotti:
## Jira Ticket
[COST-8179](https://redhat.atlassian.net/browse/COST-8179)

## Summary
- On `ONPREM=True`, `POST /api/cost-management/v1/sources/` no longer runs `ProviderBuilder.create_provider_from_source()` synchronously in the HTTP request.
- Provider and tenant schema creation are queued via the existing `storage_callback` → Celery `create_provider` path (aligned with SaaS Kafka flow).
- Register `kafka_listener` in `sources.apps.ready()` so koku-server runs the `post_save` handler.
- Add `create_provider` and `create_source_beat` Celery tasks for on-prem recovery.

## Why
First source create for a fresh org blocks on `clone_schema` inside the HTTP handler, causing 504s through Envoy/nginx (~60s) while the source eventually commits in the DB.

## Testing
- `pipenv run tox -e py39 -- sources.test.api.test_serializers sources.test.test_kafka_listener sources.test.test_tasks` (ONPREM-focused tests)
- Local docker `ONPREM=True`: POST `/sources/` returns 201 quickly; `koku_uuid` links within a few seconds via Celery

## Release Notes
```markdown
* [COST-8179](https://redhat.atlassian.net/browse/COST-8179) Queue on-prem provider creation after source create instead of blocking the HTTP request
```

### DIFF
--- a/koku/sources/api/serializers.py
+++ b/koku/sources/api/serializers.py
@@ -207,6 +207,10 @@ class AdminSourcesSerializer(SourcesSerializer):
         auth_header = get_auth_header(self.context.get("request"))
         validated_data["auth_header"] = auth_header
         source = Sources.objects.create(**validated_data)
+        if settings.ONPREM:
+            # Provider/schema creation is queued asynchronously (see storage_callback).
+            LOG.info("Admin created Source on-prem; provider creation queued.")
+            return source
         manager = ProviderBuilder(source.auth_header, source.account_id, source.org_id)
         try:
             provider = manager.create_provider_from_source(source)
--- a/koku/sources/apps.py
+++ b/koku/sources/apps.py
@@ -10,3 +10,7 @@ class SourcesConfig(AppConfig):
     """Sources application configuration."""
 
     name = "sources"
+
+    def ready(self):
+        """Register Sources post_save handlers for provider synchronization."""
+        import sources.kafka_listener  # noqa: F401
--- a/koku/sources/apps.py
+++ b/koku/sources/apps.py
@@ -4,6 +4,7 @@
 #
 """Sources application configuration module."""
 from django.apps import AppConfig
+from django.conf import settings
 
 
 class SourcesConfig(AppConfig):
@@ -12,5 +13,6 @@ class SourcesConfig(AppConfig):
     name = "sources"
 
     def ready(self):
-        """Register Sources post_save handlers for provider synchronization."""
-        import sources.kafka_listener  # noqa: F401
+        """Register Sources post_save handlers for on-prem provider synchronization."""
+        if settings.ONPREM:
+            import sources.kafka_listener  # noqa: F401
--- a/koku/sources/tasks.py
+++ b/koku/sources/tasks.py
@@ -7,6 +7,9 @@ import logging
 
 from django.conf import settings
 from django.core.exceptions import ObjectDoesNotExist
+from django.db import IntegrityError
+from django.db import InterfaceError
+from django.db import OperationalError
 
 from api.common import log_json
 from api.provider.models import Sources
@@ -16,6 +19,8 @@ from common.queues import SummaryQueue
 from koku import celery_app
 from sources.api.source_status import SourceStatus
 from sources.sources_provider_coordinator import SourcesProviderCoordinator
+from sources.sources_provider_coordinator import SourcesProviderCoordinatorError
+from sources.storage import load_providers_to_create
 from sources.storage import load_providers_to_delete
 from sources.storage import mark_provider_as_inactive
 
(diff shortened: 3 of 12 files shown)


koku/sources/apps.py
```diff
@@ -10,3 +10,7 @@ class SourcesConfig(AppConfig):
     """Sources application configuration."""
 
     name = "sources"
+
+    def ready(self):
+        """Register Sources post_save handlers for provider synchronization."""
+        import sources.kafka_listener  # noqa: F401
```
jvcsizilio: **Blocking:** This unconditionally registers `storage_callback` on every deployment (koku-server, Celery workers, etc.). Before this PR, `kafka_listener` was only loaded via `initialize_sources_integration()` in the Sources integration pod.

On **SaaS**, `AdminSourcesSerializer.create()` still synchronously calls `ProviderBuilder.create_provider_from_source()` after `Sources.objects.create()`. That save now also fires `storage_callback`, which can synchronously run `execute_process_queue()` → `create_account()` **before** the serializer's own provider create.

**Risk:** duplicate provider creation, `IntegrityError`, or a failed second create if the in-memory `source` is stale after the callback already linked `koku_uuid`.

**Suggested fix:**
```python
def ready(self):
    if settings.ONPREM:
        import sources.kafka_listener  # noqa: F401
```

The PR goal is koku-server on-prem; this guard matches that scope and avoids SaaS/worker side effects.

koku/sources/test/api/test_serializers.py
```diff
@@ -416,6 +419,33 @@ def test_create_allows_ocp_source_when_onprem(self):
             self.assertTrue(serializer.is_valid(raise_exception=True))
             self.assertEqual(serializer.validated_data["source_type"], Provider.PROVIDER_OCP)
 
+    @override_settings(ONPREM=True)
+    @patch("sources.kafka_listener._dispatch_onprem_provider_create")
```
jvcsizilio: **Non-blocking:** This test patches `_dispatch_onprem_provider_create` and relies on `captureOnCommitCallbacks` inside `@transaction.atomic` `serializer.create()`. This can miss `on_commit` callbacks in nested savepoints (see CI triager discussion). CI passes today, likely due to the manual `post_save.connect`, but patching `sources.tasks.create_provider.delay` would be more reliable and closer to production behavior:

```python
@patch("sources.tasks.create_provider.delay")
def test_create_onprem_returns_before_provider_linked(self, mock_delay, ...):
    ...
    mock_delay.assert_called_once_with(instance.source_id)
```

koku/sources/tasks.py
```diff
@@ -41,6 +46,44 @@ def delete_source(self, source_id, auth_header, koku_uuid, account_number, org_i
     LOG.info(log_json(msg="deleted provider", provider_uuid=koku_uuid, source_id=source_id))
 
 
+@celery_app.task(
+    name="sources.tasks.create_provider",
+    bind=True,
+    autoretry_for=(IntegrityError, InterfaceError, OperationalError, SourcesProviderCoordinatorError),
+    retry_backoff=True,
+    max_retries=settings.MAX_SOURCE_DELETE_RETRIES,
```
jvcsizilio, line 55: **Non-blocking:** `max_retries=settings.MAX_SOURCE_DELETE_RETRIES` reuses a delete-specific setting for create retries. Consider a dedicated setting or a brief comment explaining the reuse — the name is confusing for future maintainers.

koku/sources/tasks.py
```diff
@@ -41,6 +46,44 @@ def delete_source(self, source_id, auth_header, koku_uuid, account_number, org_i
...
+
+@celery_app.task(name="sources.tasks.create_source_beat", queue=SummaryQueue.DEFAULT)
+def create_source_beat():
+    """Re-enqueue on-prem sources that still need a linked provider."""
+    if not settings.ONPREM:
+        return
+
+    for event in load_providers_to_create():
+        provider = event.get("provider")
+        if provider:
```
jvcsizilio, line 84: **Non-blocking:** `storage_callback` and `create_source_beat` can both enqueue `create_provider` for the same source before `koku_uuid` is set. `SourcesProviderCoordinator.create_account()` handles `IntegrityError` for existing providers, but the second task may still burn Celery retries (`IntegrityError` is in `autoretry_for`). Acceptable with recovery, but an atomic claim/lease would be cleaner long-term.

koku/sources/api/serializers.py
```diff
@@ -207,6 +207,10 @@ def create(self, validated_data):
         auth_header = get_auth_header(self.context.get("request"))
         validated_data["auth_header"] = auth_header
         source = Sources.objects.create(**validated_data)
+        if settings.ONPREM:
+            # Provider/schema creation is queued asynchronously (see storage_callback).
+            LOG.info("Admin created Source on-prem; provider creation queued.")
```
jvcsizilio, line 212: **Non-blocking (API compatibility):** On-prem `POST /sources/` now returns **201 with `koku_uuid: null`** initially; linking completes asynchronously (seconds, or up to ~5 minutes if Celery enqueue fails). Clients/UI must poll or wait. Worth noting in release notes or operator docs if not already covered.

jvcsizilio (changes requested):
## Summary

This PR correctly addresses COST-8179: on-prem `POST /sources/` no longer blocks on `clone_schema` during the HTTP request. Provider creation is deferred to Celery via `storage_callback` → `create_provider`, with beat-based recovery. The approach is sound — `transaction.on_commit()` ordering, enqueue failure logging, and test coverage are all solid.

## Blocking

**`apps.ready()` registers `storage_callback` unconditionally.** Before this PR, `kafka_listener` was only loaded in the Sources integration process. Now koku-server and Celery workers also register the `post_save` handler. On SaaS, `AdminSourcesSerializer.create()` still synchronously calls `ProviderBuilder.create_provider_from_source()` after `Sources.objects.create()`, which now also fires `storage_callback` → `execute_process_queue()` → `create_account()`. This is a new double-create race. Please guard the import with `if settings.ONPREM:`.

## Non-blocking

See inline comments for test hardening, duplicate dispatch under concurrency, retry setting naming, and the intentional on-prem API timing change (`koku_uuid: null` on initial 201).

jvcsizilio (approved):
## Re-review

Latest commit `8ff57e8` addresses all prior feedback:

- **Blocking fixed:** `apps.ready()` now guards `kafka_listener` import behind `settings.ONPREM`, so SaaS koku-server/Celery workers no longer register `storage_callback` alongside the synchronous serializer create path.
- **Test hardening:** `test_create_onprem_returns_before_provider_linked` now patches `sources.tasks.create_provider.delay` and asserts the Celery enqueue.
- **Retry setting:** Comment added explaining reuse of `MAX_SOURCE_DELETE_RETRIES`.

CI: `Units - 3.11` and `Units - 3.11 (onprem-sources)` both pass.

Looks good to merge.

PR #6286, opened 2026-09-08
Add a sources capability to the user-access API for on-prem

martinpovolny:
## Jira Ticket

https://redhat.atlassian.net/browse/COST-8189

## Problem

On an on-prem deployment, a non-org-admin user whose only RBAC role is **Sources administrator** (`sources:*:*`) cannot manage integrations in the Cost Management UI. Two backend gaps cause this:

1. `koku/api/user_access/view.py` has no `sources` feature type, so the UI's `/user-access/` response contains nothing it can check for "may this user manage sources".
2. `koku/koku/rbac.py` `get_access_for_user()` only queries RBAC with `?application=cost-management`, so `sources:*:*` grants are never even fetched.

### `rbac.py` — fetch and merge sources permissions (on-prem only)

`get_access_for_user()` now, when `settings.ONPREM` is set, makes a second RBAC call with `?application=sources` and merges the result.

New helper `_normalize_sources_acls()` rewrites each `sources:<resource>:<op>` permission to `sources:sources:<op>`. This is required because `_extract_permission_data()` keys ACLs off the *middle* segment: `sources:*:*` would otherwise land under the `*` wildcard bucket and be expanded across **every** resource type in `_apply_access()`. Rewriting the middle segment maps it onto the single flat `sources` resource type instead.

SaaS behaviour is unchanged — the second call is skipped entirely when `ONPREM` is false.

### `user_access/view.py` — expose a `sources` type

`AccessMapping` gains `sources: ("sources",)`, excluded from the `any` aggregate (same treatment as `cost_model` / `settings`). `/user-access/` now reports `{"type": "sources", "access": …, "write": …}`.

On SaaS this is always `false` for non-admins (no `sources` entry in `RESOURCE_TYPES`); org admins get `true` via the existing `admin_user` short-circuit in `_check_access()`.

## Tests

- `_normalize_sources_acls` unit test (wildcard, normal, non-sources, malformed; input not mutated).
- `get_access_for_user`: SaaS makes 1 RBAC call and does not merge; on-prem makes 2 and merges `sources:*:*` onto `sources` → `{read: ["*"], write: ["*"]}`. Both are hermetic (patch `ONPREM` + `RESOURCE_TYPES`), so they run in the normal CI pass.
- `test_view.py`: `test_sources_view_read` / `_write` / `_query_type`; `build_rbac_permissions` + `matrix_keys` updated for the new type.

No migration.

## Related

Merge and deploy **this PR before** the koku-ui change https://github.com/project-koku/koku-ui/pull/5661. Backend-first is backwards compatible: the extra array entry is ignored by the current UI, and the extra RBAC call is on-prem only and cached (`RBAC_CACHE_TTL`). Shipping the UI first would strip the Integrations tab from org admins until this lands.


dchorvat1 (approved):

lcouzens (approved):

jvcsizilio (approved):
The PR seems good to me and I approved it already.

I also looked at COST-8189 and I think this covers what that ticket is asking for.

PR #6288, opened 2026-09-09
Allow provider viewers to list the built-in sources endpoint (on-prem)

martinpovolny:
## Problem

On-prem, `SourcesViewSet` (`/api/cost-management/v1/sources/`) is gated by
`SourcesAccessPermission`, which for non-admins required `sources:*:read` even
for read-only requests (`SOURCES_PERMISSION_CLASSES = [SourcesAccessPermission]
if settings.ONPREM else [AllowAny]`, added in #6041).

A user whose only role is a cost **viewer** — e.g. the seeded **Cost OpenShift
Viewer** role, which grants only `cost-management:openshift.cluster:*` — has no
`sources` permission and therefore gets **403** on that endpoint.

The cost UI uses the sources list as its "is data available?" oracle
(`filterProviders()` → `hasCurrentMonthData()` / `hasProviders()`). With the
call 403'd, the UI has no provider data, so:

- OCP detail pages show **"Still processing the data"**
- Overview and Cost Explorer show **empty state**

even though the OCP report endpoints return data correctly for that user
(wildcard `openshift.cluster` access → no RBAC filter → same rows as an org admin;
verified against a live on-prem stack).

## Change

`SourcesAccessPermission.has_permission` — for `SAFE_METHODS` only, allow the
request when the user has `sources:*:read` **or org-wide (wildcard) read** on any
provider resource type (`aws.account`, `aws.organizational_unit`,
`azure.subscription_guid`, `gcp.account`, `gcp.project`, `openshift.cluster`,
`openshift.node`, `openshift.project`).

**Resource-scoped readers do not qualify.** The on-prem sources endpoint does no
per-source filtering (`get_excludes()` is disabled for ONPREM), so opening it to
a reader scoped to e.g. `openshift.cluster:["cluster-a"]` would expose every
source's `authentication` / `billing_source` / status. A wildcard read
(`openshift.cluster:*`) is a genuinely org-wide role and is the case the cost
viewer roles actually use.

Write operations (POST / PATCH / DELETE) are unchanged — still require
`sources:*:write`. Org-admin bypass and "no access → 403" are unchanged.

## Tests

- `test_sources_access.py`: provider **wildcard** read → GET allowed (aws/azure/gcp/ocp);
  provider **scoped** read → **denied**; provider empty read → denied;
  provider read → POST/PATCH/DELETE denied. Existing cases (no access, empty
  access, `cost_model`-only, non-wildcard `sources` read, write paths) still
  assert 403.
- `sources/test/api/test_view.py` (`SourcesViewRbacTests`, ONPREM-gated):
  - `{"openshift.cluster": {"read": ["*"]}}` → list `200`, create `403`.
  - Two sources (cluster-a, cluster-b); a reader scoped to `["cluster-a"]` gets
    `403` on list, retrieve, and stats for **both** sources.

Draft: Django test suite not yet run locally; validated with `py_compile` and a
full logic simulation of all old + new scenarios.

https://claude.ai/code/session_01PdbPCKvsAEAev5XommMAJm

### DIFF
--- a/koku/api/common/permissions/sources_access.py
+++ b/koku/api/common/permissions/sources_access.py
@@ -15,9 +15,9 @@ from api.common.permissions.openshift_access import OpenShiftAccessPermission
 from api.common.permissions.openshift_access import OpenShiftNodePermission
 from api.common.permissions.openshift_access import OpenShiftProjectPermission
 
-# Provider resource types. A user who can read cost data for any of these also
-# needs to read the sources list to render the corresponding cost pages -- the
-# UI relies on it to detect whether data is available.
+# Provider resource types. A user with org-wide (wildcard) read on any of these
+# also needs to read the sources list to render the corresponding cost pages --
+# the UI relies on it to detect whether data is available.
 PROVIDER_RESOURCE_TYPES = (
     AwsAccessPermission.resource_type,
     AWSOUAccessPermission.resource_type,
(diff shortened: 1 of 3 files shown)


koku/api/common/permissions/sources_access.py
```diff
@@ -6,12 +6,41 @@
...
+from api.common.permissions.gcp_access import GcpProjectPermission
+from api.common.permissions.openshift_access import OpenShiftAccessPermission
+from api.common.permissions.openshift_access import OpenShiftNodePermission
+from api.common.permissions.openshift_access import OpenShiftProjectPermission
+
+# Provider resource types. A user with org-wide (wildcard) read on any of these
+# also needs to read the sources list to render the corresponding cost pages --
+# the UI relies on it to detect whether data is available.
+PROVIDER_RESOURCE_TYPES = (
+    AwsAccessPermission.resource_type,
```
myersCody: Nit Pick:

Right now on premise only supports OCP sources only.  So, the AWS, Azure, and GCP permissions here feel unnecessary to me.

myersCody (dismissed):

koku/api/common/permissions/sources_access.py
martinpovolny: Good call — done in d04afa66e. `PROVIDER_RESOURCE_TYPES` is now `OCP_RESOURCE_TYPES` (openshift.cluster/node/project only); dropped the aws/azure/gcp imports and updated the tests.

myersCody (approved):

PR #6293, opened 2026-09-13
Optimize ranked OCP project deltas

martinpovolny:
## Summary
- add a schema-scoped, feature-gated bounded prior-period lookup for ranked monthly OCP project cost reports
- retain the full prior-period aggregate for response-level total delta correctness
- preserve legacy behavior for categories, CSV, no-offset, daily, multi-group, and non-cost requests

## Safety
- feature: `cost-management.backend.ocp_report_limited_delta`
- configure default-off in stage/prod; code follows the existing `dev_fallback=True` convention
- includes `namespace IS NULL` handling alongside `namespace IN (...)`

## Validation
- `api.report.test.ocp.test_limited_delta`: 6 passed
- configured pre-commit hooks and flake8 passed
- targeted manual mutants proved predicate/category guards are covered


myersCody (dismissed):

bacciotti:
In the Follow-up section, could we replace the specific customer name with a generic label (e.g. "pilot production schema")? This is a public repo and we try to avoid customer identifiers in PR text.

koku/api/report/test/ocp/test_limited_delta.py
```diff
@@ -0,0 +1,170 @@
...
+            "api.report.ocp.query_handler.is_feature_flag_enabled_by_schema",
+            side_effect=self._flag_enabled,
+        ):
+            handler = self._handler()
+            limited_query = handler._get_previous_rows_query(
+                handler.query_table.objects.all(), [{"project": "ranked"}, {"project": None}]
+            )
+
+        query_sql = str(limited_query.query)
+        self.assertIn('"namespace" IN', query_sql)
```
bacciotti, line 119: Nice SQL-shape check. Could we also evaluate the filtered queryset and assert the namespace values match the selected rows (same as the main test)? That way a regression that still emits `IN`/`IS NULL` but includes extra namespaces would fail.

koku/api/report/ocp/query_handler.py
```diff
@@ -246,6 +248,22 @@ def _distinct_arrays_split_enabled(self):
         """Whether to compute clusters/source_uuid via separate (parallel-safe) queries."""
         return is_feature_flag_enabled_by_schema(self.tenant.schema_name, OCP_REPORT_DISTINCT_ARRAYS_PARALLEL_FLAG)
 
+    @cached_property
+    def _limited_delta_for_ranked_projects_enabled(self):
+        """Whether prior-period deltas can be limited to the returned projects."""
```
bacciotti, line 253: Optional: note that this path requires both `filter[limit]` and `filter[offset]` in the request (not the default offset only), so the limit-without-offset "Others" shape stays on the legacy prior-period scan.

bacciotti (commented):
Non-blocking review notes — no change request. Thanks for the flag gating and the parity test.

koku/api/report/ocp/query_handler.py
martinpovolny: that is intended, because the meaning of "others" can differ between the months --> no pre-selection

bacciotti (approved):

PR #6297, opened 2026-09-18
feat: combine OCP distributed cost delta aggregate

martinpovolny:
## Summary
- add a default-off, schema-targeted Unleash flag for the OCP distributed-cost delta aggregate
- use one SUM only for the monthly, project-only, paginated OCP costs delta shape
- retain legacy behavior for flag-off, daily, CSV, category, multi-group, and non-cost report paths

## Validation
Passed:
- `api.report.test.ocp.test_combined_distributed_cost`, `api.report.test.ocp.test_ocp_provider_map`, `api.report.test.ocp.test_identity_exchange_rate`, and `api.report.test.ocp.test_limited_delta` — 17 tests passed
- `pre-commit run --all-files`
- focused mutation probes for a wrong exchange-rate column, a missing distributed rate type, NULL coalescing, and the disabled-flag path each failed as expected

Not counted as passing validation:
- the local `api.report.test.ocp` package reached 285 tests but is blocked by the absent local RBAC service at `localhost:8111`, which returns expected-environment 424 failures. CI remains required.

## Measurements

```
  Metric                         Legacy (warm)    Combined      Change
  -----------------------------  ---------------  ------------  ----------------
  Execution time                 7.781 s          3.632 s       4.149 s / 53.3% faster
  Rows scanned                   1,308,686        1,308,686    identical
  Final project/month groups     441,919          441,919      identical
  Parallel workers               4 planned / 4    4 planned / 4 identical
  Temporary data read            384 MiB          0 MiB         eliminated
  Temporary data written         658 MiB          0 MiB         eliminated
```


koku/api/report/ocp/query_handler.py
```diff
@@ -250,8 +251,8 @@ def _distinct_arrays_split_enabled(self):
         return is_feature_flag_enabled_by_schema(self.tenant.schema_name, OCP_REPORT_DISTINCT_ARRAYS_PARALLEL_FLAG)
 
     @cached_property
-    def _limited_delta_for_ranked_projects_enabled(self):
-        """Whether prior-period deltas can be limited to the returned projects."""
+    def _is_limited_delta_for_ranked_projects_shape(self):
+        """Whether the Banco-style ranked delta shape is safe for an experiment."""
```
bacciotti, line 255: Please avoid the customer name in this docstring (public repo).
Something like "ranked paginated project-only monthly delta shape" is enough.

bacciotti (commented):
Nice optimization and good gating.
One small clarity note: combined distributed cost keys off the limited-delta *request shape*, but not the limited-delta flag itself. A one-line comment on that property would help future readers.

koku/api/report/ocp/query_handler.py
martinpovolny: Updated in 4ebce6e49: removed the customer references and clarified that the shape is intentionally shared by independently rollable flags.

bacciotti (approved):

PR #6298, opened 2026-09-19
feat: derive OCP report totals outside aggregate

martinpovolny:
## Summary
- add a default-off Unleash flag, `cost-management.backend.ocp_report_derived_totals`, for CPU and memory response totals
- aggregate primitive cost and usage values once, then derive composite costs and efficiency score values in Python
- retain the legacy SQL aggregate for every unflagged tenant and for costs/volume reports

## Verification
- `test_derived_totals_flag_aggregates_only_primitives`
- `test_derived_totals_flag_matches_legacy_cpu_and_memory_totals`
- `test_derived_totals_calculation_handles_score_boundaries`
- `test_derived_totals_flag_ignores_costs_and_volume_reports`
- file-scoped repository pre-commit checks
- staged `gitleaks protect --staged`

The four focused tests pass locally. The full `api.report.test.ocp.test_ocp_query_handler` module requires the local seeded report fixtures and RBAC service; without those it has unrelated fixture/RBAC failures, so CI should run the complete suite.

## Rollout
The flag is default-off. No deployment or tenant enablement is included in this PR.


martinpovolny:
Closing: the bounded RO measurements show no material improvement from moving these derived aggregates to Python (about 0.5% in a warm structural proxy, within noise). The 75-second production shape is more likely dominated by scan/filter and currency-expression work, so this change is not worth carrying as a feature-flagged rollout.

